PR #30, opened 2026-05-09
feat: add total_count to list_recent, raise limit to 500

### LINKED ISSUE #24: feat: list_recent should support offset beyond 200 or return total count
## Problem
list_recent caps at 200 results. During dream cycle cleanup, there were 400+ sessions and 290+ decisions -- entries beyond offset 200 were invisible to the initial scan. Had to manually paginate with offset parameter AND use search to discover entries that list_recent couldn't reach.

## Proposed Fix
Either:
1. Return a total_count field so callers know there are more entries to paginate
2. Support higher limits (500+)
3. Add a count_by_tier tool that returns tier distribution without fetching full entries

## Context  
Dream cycle 2026-04-09. Found 239 additional warm entries hiding beyond the first 200 that needed tier management.

Skippy-the-Magnificent-one:
Closes #24

## Changes
- Response now returns `{entries, total_count, offset, limit, has_more}` instead of raw array
- `total_count` tells callers exactly how many matching entries exist for pagination
- `has_more` boolean for easy "load more" logic
- Limit raised from 250 to 500
- Count query runs in parallel with data query via `Promise.all` (no perf impact)
- 11 tests updated and passing

## Before
```json
[{...}, {...}, ...]  // raw array, no idea how many more exist
```

## After
```json
{
  "entries": [{...}, {...}, ...],
  "total_count": 427,
  "offset": 0,
  "limit": 20,
  "has_more": true
}
```

**Breaking change:** callers that `JSON.parse` the response and expect an array need to access `.entries` instead.

### DIFF
--- a/src/tools/__tests__/list-recent.test.ts
+++ b/src/tools/__tests__/list-recent.test.ts
@@ -72,7 +72,7 @@ describe("list_recent", () => {
         expect(result.isError).toBeFalsy();
 
         // Verify SQL shape
-        expect(queryCalls.length).toBe(1);
+        expect(queryCalls.length).toBe(2); // data query + count query
         const [sql, params] = queryCalls[0];
 
         // Should query all 5 tables (admin has read on all)
@@ -93,8 +93,11 @@ describe("list_recent", () => {
 
         // Verify result format
         const parsed = JSON.parse((result.content as any)[0].text);
-        expect(Array.isArray(parsed)).toBe(true);
-        expect(parsed.length).toBe(3);
+        expect(parsed.entries).toBeDefined();
+        expect(Array.isArray(parsed.entries)).toBe(true);
+        expect(typeof parsed.total_count).toBe("number");
+        expect(typeof parsed.has_more).toBe("boolean");
+        expect(parsed.entries.length).toBe(3);
       } finally {
         await cleanup();
       }
@@ -233,7 +236,10 @@ describe("list_recent", () => {
 
         expect(result.isError).toBeFalsy();
         const parsed = JSON.parse((result.content as any)[0].text);
-        expect(Array.isArray(parsed)).toBe(true);
+        expect(parsed.entries).toBeDefined();
+        expect(Array.isArray(parsed.entries)).toBe(true);
+        expect(typeof parsed.total_count).toBe("number");
+        expect(typeof parsed.has_more).toBe("boolean");
       } finally {
         await cleanup();
       }
@@ -361,8 +367,11 @@ describe("list_recent", () => {
 
         expect(result.isError).toBeFalsy();
         const parsed = JSON.parse((result.content as any)[0].text);
-        expect(Array.isArray(parsed)).toBe(true);
-        expect(parsed.length).toBe(0);
+        expect(parsed.entries).toBeDefined();
+        expect(Array.isArray(parsed.entries)).toBe(true);
+        expect(typeof parsed.total_count).toBe("number");
+        expect(typeof parsed.has_more).toBe("boolean");
+        expect(parsed.entries.length).toBe(0);
       } finally {
         await cleanup();
       }
--- a/src/tools/__tests__/list-recent.test.ts
+++ b/src/tools/__tests__/list-recent.test.ts
@@ -56,6 +56,10 @@ describe("list_recent", () => {
       const mockPool = {
         query: async (...args: any[]) => {
           queryCalls.push(args);
+          const [sql] = args;
+          if (sql.includes("SUM(cnt)")) {
+            return { rows: [{ total_count: 3 }] };
+          }
           return { rows: makeMockRows(3) };
         },
       };
@@ -95,8 +99,8 @@ describe("list_recent", () => {
         const parsed = JSON.parse((result.content as any)[0].text);
         expect(parsed.entries).toBeDefined();
         expect(Array.isArray(parsed.entries)).toBe(true);
-        expect(typeof parsed.total_count).toBe("number");
-        expect(typeof parsed.has_more).toBe("boolean");
+        expect(parsed.total_count).toBe(3);
+        expect(parsed.has_more).toBe(false);
         expect(parsed.entries.length).toBe(3);
       } finally {
         await cleanup();
@@ -110,6 +114,10 @@ describe("list_recent", () => {
       const mockPool = {
         query: async (...args: any[]) => {
           queryCalls.push(args);
+          const [sql] = args;
+          if (sql.includes("SUM(cnt)")) {
+            return { rows: [{ total_count: 1 }] };
+          }
           return { rows: makeMockRows(1) };
         },
       };
@@ -141,6 +149,10 @@ describe("list_recent", () => {
       const mockPool = {
         query: async (...args: any[]) => {
           queryCalls.push(args);
+          const [sql] = args;
+          if (sql.includes("SUM(cnt)")) {
+            return { rows: [{ total_count: 2 }] };
+          }
           return { rows: makeMockRows(2) };
         },
       };
@@ -169,6 +181,10 @@ describe("list_recent", () => {
       const mockPool = {
         query: async (...args: any[]) => {
           queryCalls.push(args);
+          const [sql] = args;
+          if (sql.includes("SUM(cnt)")) {
+            return { rows: [{ total_count: 0 }] };
+          }
           return { rows: [] };
         },
       };
@@ -196,6 +212,10 @@ describe("list_recent", () => {
       const mockPool = {
         query: async (...args: any[]) => {
           queryCalls.push(args);
+          const [sql] = args;
+          if (sql.includes("SUM(cnt)")) {
+            return { rows: [{ total_count: 5 }] };
+          }
           return { rows: makeMockRows(5) };
         },
       };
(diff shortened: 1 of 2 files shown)


Skippy-the-Magnificent-one:
# 🔍 Adversarial Code Review — PR #30

**feat: add total_count to list_recent, raise limit to 500**
*5-pass review • 2 files • ~90 additions*

---

## Pass 1 — SECURITY 🔐

### 🚨 P0 — SQL Injection via `tier` parameter in `buildCountSelect`

**File:** `src/tools/list-recent.ts`, `buildCountSelect()` (new function)

The new `buildCountSelect` function interpolates `tier` directly into SQL:

```ts
const tierFilter = tier ? ` AND ${alias}.tier = '${tier}'` : "";
```

This is identical to how `buildTableSelect` does it (which is also technically vulnerable), but the PR **removed the `VALID_TIERS` guard** that previously protected `buildTableSelect`:

```diff
-  if (tier && !VALID_TIERS.has(tier)) throw new Error(`Invalid tier: ${tier}`);
```

The Zod schema (`z.enum(["hot", "warm", "cold"])`) is the remaining defense. Since Zod validation runs at the MCP schema level before the handler executes, this is **mitigated at the input boundary** — but defense-in-depth was silently dropped. If anyone ever calls `buildCountSelect` or `buildTableSelect` directly (e.g., from another tool), the tier value flows unsanitized into SQL.

**Verdict:** Mitigated by Zod but defense-in-depth regression. **P1** — recommend restoring the `VALID_TIERS` check or using a parameterized query for tier.

### ✅ Verified — Limit bounds

Limit raised from 250→500. Zod `.max(500)` enforced at input. Acceptable for a read-only listing operation.

### ✅ Verified — Auth/authz unchanged

Permission checks (`canRead`, `authInfo`) flow is unchanged. The count query uses the same `accessibleTables` array — no authz bypass.

---

## Pass 2 — CORRECTNESS 🧪

### ⚠️ P1 — `VALID_TIERS` guard removed, `table-constants.ts` import dropped

The PR inlines `ALL_TABLES`, `SOURCE_LABELS`, `CONTENT_PREVIEW`, and `TABLE_ALIAS` as local constants — **duplicating** what exists in `table-constants.ts`. This creates two sources of truth. Notably:

- The PR's inlined `CONTENT_PREVIEW.sessions` is `"COALESCE(s.project || ': ', '') || LEFT(s.summary, 200)"` — **200 chars**
- The shared `table-constants.ts` version uses `LEFT(s.summary, 300)` — **300 chars** — and includes `key_decisions` and `next_steps` enrichment
- The PR's inlined `CONTENT_PREVIEW.decisions` is `"d.title || ': ' || d.rationale"` — missing `COALESCE(d.rationale, '')` null-safety that the shared version has

These are **correctness divergences** from the current shared constants. The `decisions` preview will fail with a NULL `rationale`.

**Verdict: P0** — The inlined constants diverge from the shared source. The `decisions` NULL-safety regression will cause `NULL` concatenation results on rows where `rationale IS NULL`. **Must either keep importing from `table-constants.ts` or update the inlined values to match.**

### ⚠️ P1 — `total_count` can return `null` as `0`

```ts
const totalCount = countResult.rows[0]?.total_count ?? 0;
```

The `SUM(cnt)::int` over an empty UNION ALL (0 accessible tables is already guarded, but 0 matching rows) returns `NULL` in PostgreSQL when all sub-counts are 0... actually, `COUNT(*)` always returns 0 for empty sets, and `SUM(0)` = `0`, so this is fine. **Verified ✅** — `COUNT(*)` never returns NULL; `SUM` of zero-count rows returns 0.

### ✅ Verified — `has_more` logic

```ts
const hasMore = offset + dataResult.rows.length < totalCount;
```

This is correct. If `offset=0`, `rows.length=20`, `totalCount=25` → `has_more=true`. If `totalCount=20` → `has_more=false`. Edge: if the count changes between the data query and count query (TOCTOU), `has_more` could be slightly stale — acceptable for a non-transactional listing endpoint.

### ✅ Verified — Count query uses same filters

`buildCountSelect` mirrors `buildTableSelect` for archive and tier filters. The `WHERE` clause structure matches. Good.

---

## Pass 3 — PERFORMANCE ⚡

### ⚠️ P2 — Parallel `COUNT(*)` across 5 tables on every request

```ts
const [dataResult, countResult] = await Promise.all([
  deps.pool.query(sql, [days, limit, offset]),
  deps.pool.query(countSql, [days]),
]);
```

Good: queries run in parallel via `Promise.all`, not sequentially. The count query is a `UNION ALL` of 5 `COUNT(*)` sub-selects wrapped in `SUM()`.

For a personal brain DB with modest data volumes, this is fine. Each `COUNT(*)` uses a seq scan filtered by `created_at >= NOW() - INTERVAL '1 day' * $1`, which benefits from any index on `created_at`.

**Verdict:** Acceptable at current scale. If tables grow past ~100K rows, consider:
- Adding a composite index on `(created_at, archived_at)` if not present
- Caching the count (stale for up to N seconds)
- Making `total_count` opt-in via a parameter

### ✅ Verified — No N+1 or redundant queries

Two queries total (data + count), both parallelized. Clean.

---

## Pass 4 — TESTING 🧪

### ⚠️ P1 — Tests don't verify count query SQL shape

The existing test at line 75 changes from `expect(queryCalls.length).toBe(1)` → `.toBe(2)` but never inspects `queryCalls[1]` (the count query). No test verifies:
- The count SQL contains `COUNT(*)`
- The count SQL uses the correct filters (archive, tier)
- The count query receives the correct parameter (`[days]`)

**Recommendation:** Add assertions on `queryCalls[1]`:
```ts
const [countSql, countParams] = queryCalls[1];
expect(countSql).toContain("COUNT(*)");
expect(countSql).toContain("archived_at IS NULL");
expect(countParams[0]).toBe(7); // days
```

### ⚠️ P2 — No test for `has_more` logic

No test verifies `has_more=true` when `total_count > offset + entries.length`. The mock always returns the same rows for both queries, making `has_more` essentially untestable without differentiating the data query mock from the count query mock.

**Recommendation:** Add a test where the count query returns `{ rows: [{ total_count: 100 }] }` while the data query returns 3 rows, and assert `has_more === true`.

### ⚠️ P2 — No test for `limit=500` (new max)

No test exercises the new maximum. Should verify Zod accepts 500 and rejects 501.

### ✅ Verified — Empty results test updated

The empty results test correctly expects the new response shape with `entries`, `total_count`, `has_more`.

### ⚠️ P2 — Readonly role test not updated

The readonly role test at line 236 still asserts `expect(Array.isArray(parsed)).toBe(true)` — this will fail because the response is now an object `{ entries, total_count, ... }`, not an array.

**Verdict: P0** — This test will fail on the PR branch as-is.

---

## Pass 5 — ARCHITECTURE 🏗️

### 🚨 P0 — Breaking change to API response contract

The response shape changed from:
```json
[{ "source_type": "thought", "id": "...", ... }]
```
to:
```json
{ "entries": [...], "total_count": 25, "offset": 0, "limit": 20, "has_more": true }
```

This is a **breaking change** for any client parsing the `list_recent` response. Any consumer doing `JSON.parse(text)` and iterating the array directly will break. This includes:
- Other MCP clients calling this tool
- Any automation/n8n workflows consuming the response
- The `readonly` test in this very PR (see Pass 4)

**Recommendation:** Since this is a personal MCP server with controlled clients, the break is likely acceptable — but it should be called out in the PR description/changelog and any consuming code updated simultaneously.

### ⚠️ P1 — Inlined constants duplicate `table-constants.ts`

The PR copies `ALL_TABLES`, `SOURCE_LABELS`, `CONTENT_PREVIEW`, `TABLE_ALIAS` locally and **drops the import**. This creates maintenance drift — `search-brain.ts` and any future tools still import from `table-constants.ts`. Changes to the shared file won't propagate to `list-recent.ts`.

The `VALID_TIERS` import was also dropped (see Pass 1).

**Recommendation:** Keep importing from `table-constants.ts`. If the PR needs different `CONTENT_PREVIEW` values for list-recent specifically, create a separate constant rather than duplicating.

### ✅ Verified — Backward-compatible input schema

Input parameters are additive. Existing calls without `total_count`-related params continue to work. `limit` max increased from 250→500 is backward compatible (old values still valid).

---

## Summary Verdict

| Severity | Count | Items |
|----------|-------|-------|
| **P0** | 2 | Inlined constants diverge from shared (NULL-safety regression on `decisions`); Readonly test will fail with new response shape |
| **P1** | 3 | `VALID_TIERS` defense-in-depth removed; Count query SQL not tested; Inlined constants duplicate shared module |
| **P2** | 3 | COUNT(*) cost at scale; No `has_more` test; No `limit=500` boundary test |
| **Verified ✅** | 6 | Limit bounds, auth flow, has_more logic, count filters, empty results test, input schema backward compat |

### 🟡 Recommend: **Request Changes**

The two P0s need fixing before merge:
1. **Don't inline constants** — keep importing from `table-constants.ts` (fixes the `decisions` NULL regression and the `sessions` preview divergence)
2. **Update the readonly role test** to expect the new response shape (or confirm tests pass on the branch — the diff only shows 3 test blocks updated but there are more tests asserting the old array format)

The P1s (count query test coverage, VALID_TIERS guard) should ideally ship with this PR but are deferrable.

Issue #40, opened 2026-06-04
docs: define OB operational memory contract for agents


rodaddy:
## Goal
Document and implement the intended memory stack contract: Open Brain is operational memory, Markdown overlay is behavior law, session logs are raw recall, and Honcho is not authoritative for ops state.

## Motivation
Recent agent behavior regressed when synthesized memory was treated like operational source of truth. OB needs a clear contract for agents and integrations.

## Contract
- Open Brain: durable architecture/workflow/incident/decision memory and active session lanes.
- Markdown overlay/includes: deterministic hot behavior law and routing rules.
- `/mnt/collab` Markdown artifacts: human-readable handoffs/change logs/source files.
- session_search / Discord backread: raw conversation recovery.
- Honcho or other memory synthesis: optional/cold user/persona aid, not ops authority.

## Acceptance criteria
- Add docs describing this contract in the Open Brain repo.
- Add examples showing n8n/content-bot and Hermes overlay recovery using OB session lanes.
- Update skill/reference docs if present.
- Explicitly note that `rodaddy/open-brain` was inspired by Nate B. Jones OB1 videos, not copied from OB1 code.

Issue #39, opened 2026-06-04
feat: extend MCP/mcp2cli with lane-aware OB tools


rodaddy:
## Goal
Harden and extend the Open Brain MCP/mcp2cli server surface for lane-aware memory operations.

## Motivation
Agents should call OB through tools, not hand-roll conventions in prompt text. The current operational failure mode is agents skipping OB or substituting Honcho/session vibes. First-class tools should make the right path easy.

## Proposed tool additions
- `start_session_lane`
- `session_context`
- `append_session_event`
- `link_entity`
- `wrap_session_lane`
- `list_active_lanes`
- `adjacent_context`

## Requirements
- Keep namespace/auth rules explicit.
- Return compact, receipt-friendly output.
- Avoid dumping large standing guidance into Discord/tool notifications.
- Gracefully degrade when OB is unavailable.
- Keep current tools (`search_brain`, `log_thought`, `log_decision`, `session_save`, etc.) backward-compatible.

## Acceptance criteria
- mcp2cli can call every new lane tool with JSON params.
- Tools are covered by tests.
- Errors identify the failed operation and do not leak secrets.
- Agents can recover active work after compaction/reset with a single lane lookup.

rodaddy:
5 of 7 proposed tools now shipped:
- `session_start` (≈ start_session_lane) — PR #46
- `session_context` — PR #45
- `append_session_event` — PR #45
- `session_wrap` (≈ wrap_session_lane) — PR #46
- `lane_load` (≈ list_active_lanes) — PR #43

Remaining: `link_entity` and `adjacent_context` depend on the entity graph schema (PR #42, merged). These can be built as follow-ups.

Issue #38, opened 2026-06-04
feat: add entity/link tables for explicit adjacency


rodaddy:
## Goal
Add explicit entity/link/adjacency support for Open Brain session lanes and existing memory entries.

## Motivation
pgvector provides semantic adjacency, but operational memory also needs explicit relationships: this session changed this workflow, this blocker depends on this host, this artifact supersedes that one, this decision came from that incident.

## Proposed tables

```sql
CREATE TABLE ob_entities (
  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
  entity_type TEXT NOT NULL, -- host, repo, workflow, service, agent, channel, file, command, etc.
  name TEXT NOT NULL,
  canonical_id TEXT,
  namespace TEXT NOT NULL,
  metadata JSONB NOT NULL DEFAULT '{}'::jsonb,
  embedding vector,
  created_at TIMESTAMPTZ NOT NULL DEFAULT now(),
  updated_at TIMESTAMPTZ NOT NULL DEFAULT now()
);

CREATE TABLE ob_links (
  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
  from_type TEXT NOT NULL,
  from_id UUID NOT NULL,
  to_type TEXT NOT NULL,
  to_id UUID NOT NULL,
  relation TEXT NOT NULL, -- artifact, depends_on, supersedes, caused_by, same_lane, adjacent, etc.
  weight DOUBLE PRECISION DEFAULT 1.0,
  metadata JSONB NOT NULL DEFAULT '{}'::jsonb,
  created_at TIMESTAMPTZ NOT NULL DEFAULT now()
);
```

## Acceptance criteria
- A lane can link to artifacts, workflows, hosts, agents, and decisions.
- Search results can include explicit links plus pgvector neighbors.
- Existing `relationships`/`projects` tables remain compatible.
- Obsidian sync can render links without losing OB source IDs.

Issue #37, opened 2026-06-04
feat: add session-start/session-wrap workflow


rodaddy:
## Goal
Build a session-start/session-wrap workflow in Open Brain so active lanes become durable memory at the end of a work session.

## Motivation
Rico's agents need OB to act as operational memory, not just a search box. At session end, active lane events should be distilled into durable OB records and artifacts, then loaded on future session start.

### session-start
- Find active lane by channel/thread/session_key/project.
- Return current context, blockers, receipts, and linked artifacts.
- Optionally surface adjacent lanes/sessions.

### session-wrap
- Read lane events and linked artifacts.
- Distill durable outputs:
  - `sessions` summary
  - `decisions` with rationale
  - `thoughts`/learnings
  - follow-up tasks/blockers
  - skill update suggestions, if applicable
- Mark lane `wrapped` or keep `active` if work continues.

## Acceptance criteria
- Wrap is idempotent or safely repeatable.
- Transient details can stay in event history without polluting durable memory.
- Durable outputs preserve exact artifact paths and receipts.
- Compaction/restart can load the latest lane context without broad search.

PR #41, opened 2026-06-05
docs: OB v2 implementation plan

Skippy-the-Magnificent-one:
Adds the OB v2 design and implementation plan for review before schema work begins.

## Contents
- Current v1 state analysis (5 core tables, 21 MCP tools, migrations 001-009)
- Phase 0: Test DB setup (verified on Mac Mini M4, port 5433)
- Phase 1: Schema — session lanes (#34), event journal (#35), entity graph (#38)
- Phase 2: MCP tools — lane-aware operations (#36, #39)
- Phase 3: Session start/wrap workflow (#37)
- Phase 4: Contract doc (#40)
- Agent assignment, PR workflow, test plan, cutover plan

## Reviewers
- @rodaddy (Rico) — approval authority
- Bilby — #38 entity/link schema review

No code changes. Design doc only.

### DIFF
--- a/docs/ob-v2-plan.md
+++ b/docs/ob-v2-plan.md
@@ -67,6 +67,23 @@ CREATE TRIGGER trg_lanes_updated_at
 
 **Key constraint:** Only one active lane per `(namespace, session_key)` — enforced by the partial unique index.
 
+### Entity/Link Design Rules
+
+**Namespace consistency on links:**
+- `ob_links` has its own `namespace` column. Insert/update helpers MUST verify both `from_entity` and `to_entity` belong to that namespace unless the caller explicitly passes `cross_namespace=true`.
+- Plain FK on UUID alone won't prevent cross-namespace edges — enforce in the `link_entity` tool's insert logic and test it.
+
+**Alias collision / merge semantics:**
+- Unique `(namespace, kind, name)` applies to canonical names only. Aliases can collide with another entity's canonical name.
+- Lookup order: exact canonical match wins → alias match returns candidates (never auto-merge).
+- Merges: set old entity `status='merged'`, set `merged_into` to the replacement entity UUID. Old refs continue to resolve via redirect.
+- `adjacent_context` and `link_entity` follow `merged_into` chains (max depth 3) to resolve current entity.
+
+**Link directionality (documented, not schema-enforced):**
+- Symmetric: `related_to`, `contradicts` — queries return both directions.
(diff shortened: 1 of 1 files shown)


rodaddy (commented):
## Bilby review - OB v2 plan (#38 focus)

Verdict: **Comment / not signed off yet**. Design direction is solid, but #38 needs two schema guardrails captured before schema work starts.

### P1 - Cross-namespace links need a DB-level policy
`ob_links` references entity IDs only. Since `ob_entities` owns `namespace`, the current DDL allows a link from a `collab` entity to a `rico`/private entity unless tool code prevents it every time. That is too easy to bypass via migrations, scripts, seed loaders, or future tools.

Please add one of these to the plan before implementation:
- add `namespace TEXT NOT NULL` to `ob_links` plus composite FKs `(namespace, from_entity)` / `(namespace, to_entity)` against unique `(namespace, id)` on `ob_entities`, or
- add a trigger/check function that rejects cross-namespace links unless `metadata->>'cross_namespace' = 'true'` and the caller is explicitly authorized.

For default OB v2 behavior, same-namespace links should be the invariant.

### P2 - `relation TEXT` needs a controlled vocabulary
Free-form `relation TEXT` will turn adjacency into soup: `depends_on`, `depends-on`, `dependency`, `blocks`, `blocked_by`, etc. all become separate graph semantics.

Please define this in the design doc before implementation:
- either a Postgres enum / CHECK constraint for initial relation values, or
- a `ob_link_relation_types` lookup table with `name`, `inverse_name`, `description`, `directional`, and `active`.

My preference: lookup table. It lets us add relation types without migrations while still keeping semantics controlled.

### P2 - Entity uniqueness should probably normalize case
`UNIQUE(namespace, kind, name)` is case-sensitive. That permits `Skippy`, `skippy`, and `SKIPPY` as distinct entities in the same namespace/kind.

Recommend either:
- `UNIQUE(namespace, kind, lower(name))`, or
- add `slug/canonical_name` and make uniqueness use that.

### Looks good
- `ob_entities` keeps namespace on the entity itself.
- HNSW + `halfvec(768)` matches existing OB style.
- Directional `from_entity -> to_entity` links are the right primitive.
- Unique `(from_entity, to_entity, relation)` prevents duplicate edge spam.

No schema work should start until at least the cross-namespace and relation-vocabulary decisions are captured. No cowboy graph soup, man.

rodaddy:
<@1507046681580146798> PR #41 reviewed. I left a COMMENT review, not approval. Blocking design gaps before schema work: cross-namespace link invariant and controlled relation vocabulary. Also flagged case-normalized entity uniqueness.

Skippy-the-Magnificent-one:
Plan doc belongs in shared workspace, not a PR. Moving to collab/agent-bridge.

PR #42, opened 2026-06-07
feat: add entity link graph schema

rodaddy:
## Summary
- Add ob_entities and ob_links tables for explicit Open Brain adjacency
- Enforce namespace/type/name canonical entities, relation vocabulary, nonnegative weights, and no self-links
- Enrich search_brain/search_all results with explicit incoming/outgoing links when available

## Test Plan
- bun x tsc --noEmit
- bun test src/tools/__tests__/search-brain.test.ts src/tools/__tests__/search-all.test.ts src/db/migrations/001_init.test.ts
- bun test

Notes: local pgvector migration integration tests skipped because no reachable test DB is configured. CI workflow inspection found no references to 10.71.20.15, deploy, or ssh in .github/workflows.

Contributes to #38


Skippy-the-Magnificent-one:
# 🔍 5-Pass Adversarial Code Review — PR #42

**feat: add entity link graph schema** • 378 additions, 7 files
Reviewer: Skippy (Hermes Agent) • Model: opus

---

## Pass 1 — SECURITY 🔐

### ✅ Verified
- **Parameterized queries throughout**: `attachExplicitLinks` uses `$1/$2/$3` placeholders with `unnest()` — no SQL injection risk.
- **Auth gate inherited**: `executeSearch` is only called from `registerSearchBrain` which checks `auth` / `canRead()` before reaching the new code path. The link lookup doesn't bypass the existing authz check.
- **No credential handling**: No secrets, tokens, or connection strings touched.

### P2 — Namespace isolation not enforced on link reads
**File:** `src/tools/search-brain.ts` → `attachExplicitLinks()`
**Issue:** When `namespace` is `undefined`, the query returns links from **all namespaces**. The search results themselves already went through namespace-filtered CTEs, so the primary data is scoped — but the links attached could leak cross-namespace edges. If a thought in namespace `collab` links to an entity in namespace `private-agent-x`, that link will appear in results for a user searching `collab`.
**Severity:** Low risk today (single-tenant), but violates the namespace isolation contract.
**Suggestion:** Consider always passing a namespace filter, or defaulting to `'collab'` when undefined.

---

## Pass 2 — CORRECTNESS 🐛

### ✅ Verified
- **`ExplicitLink` interface** matches the `SELECT` columns from `ob_links` correctly.
- **Bidirectional link attachment** is correct: each link is pushed to both `outgoing` (from_type:from_id) and `incoming` (to_type:to_id) maps, with direction labels set properly.
- **`linkKey()` composition** uses `type:id` which correctly matches `source_type:id` on the search results side.
- **Error handling**: The `try/catch` in `attachExplicitLinks` gracefully degrades — returns rows with `explicit_links: []` on failure, plus a `logger.warn`. This is correct; a link-lookup failure should not break search.
- **Schema CHECK constraints** are sound: `btrim(x) <> ''` prevents empty-string abuse, `weight >= 0` prevents negative weights, self-link check `from_type <> to_type OR from_id <> to_id` is correct.
- **`update_updated_at()` trigger** dependency: The function is created in `001_init.sql` and this migration references it — correct ordering since `010 > 001`.

### P1 — `source_type` uses singular labels, `from_type`/`to_type` stores raw strings — potential mismatch
**File:** `src/tools/search-brain.ts` → `linkKey()`
**Issue:** `linkKey(row.source_type, row.id)` builds keys using `source_type` which comes from `SOURCE_LABELS` (singular: `"thought"`, `"decision"`, etc.). But `from_type`/`to_type` in `ob_links` are free-text `TEXT` columns — there's **no enforcement** that they use the same singular labels. If someone inserts a link with `from_type = 'thoughts'` (plural, matching the table name), the `linkKey` won't match and links will silently be dropped from results.
**Suggestion:** Either:
1. Add a `CHECK (from_type IN ('thought', 'decision', 'relationship', 'project', 'session', 'entity'))` constraint on `ob_links`, or
2. Document the expected label convention prominently in the migration, or
3. Normalize `from_type` in the query (strip trailing 's').

### P1 — `created_at` returned from link query is a `Date` object, interface declares `string`
**File:** `src/tools/search-brain.ts` → `LinkRow` type
**Issue:** PostgreSQL's `pg` driver returns `TIMESTAMPTZ` columns as JavaScript `Date` objects by default. The `LinkRow` type declares `created_at: string` and the `ExplicitLink` interface also uses `string`. When serialized to JSON, `Date.toJSON()` produces an ISO string so this works in practice — but if anyone does string operations on `link.created_at` before serialization, they'll get `[object Object]` or type errors.
**Severity:** Latent bug. Works today because results go straight to `JSON.stringify()`.

### P2 — `search-all.ts` maps `explicit_links` into `UnifiedResult` but never passes it through
**File:** `src/tools/search-all.ts` → `searchOB()` function (line 238-246)
**Issue:** The diff adds `explicit_links` to the `UnifiedResult` interface and maps it in the return from `executeSearch`, but `searchOB()` at line 238 maps rows to `UnifiedResult` and **does not include `explicit_links`** in the mapped output. The field is on the interface but never populated in `search_all` results.
**Suggestion:** Add `explicit_links: row.explicit_links` to the `searchOB` return mapper (around line 245), or remove it from `UnifiedResult` if it's intentionally omitted.

### P2 — Test cleanup: `ob_entities`/`ob_links` drops added to test teardown but not to the table existence test
**File:** `src/db/migrations/001_init.test.ts` → `afterAll` and top-level cleanup
**Issue:** The cleanup blocks now drop `ob_links` and `ob_entities`, which is good. However, the existing test `"should create all 5 data tables plus _migrations"` (line 58) still only checks for 5+1 tables. The new `"entity graph schema"` describe block validates the new tables independently, which is fine — but the original assertion could be updated to expect 7 tables for consistency.
**Severity:** Cosmetic / test hygiene.

---

## Pass 3 — PERFORMANCE ⚡

### ✅ Verified
- **Single batched link query**: `attachExplicitLinks` fires **one** query using `unnest()` arrays to look up all links for all search results at once — no N+1.
- **`ob_links` indexes** are well-chosen: unique composite on `(namespace, from_type, from_id, to_type, to_id, relation)`, plus covering indexes on from-side and to-side for the two `IN` clauses in the link query.
- **HNSW index** on `ob_entities.embedding` has sensible params (`m=16, ef_construction=200`).
- **Link query uses `ORDER BY weight DESC, created_at DESC`** — this is fine for correctness but the sort happens after the `WHERE` filter which is index-supported.

### P2 — Unbounded link result set
**File:** `src/tools/search-brain.ts` → `attachExplicitLinks()`
**Issue:** The link query has no `LIMIT`. If a single entity has thousands of links (e.g., a heavily-connected hub node), all of them are fetched and attached. With 10 search results × potentially 1000+ links each, this could return a very large payload.
**Suggestion:** Add a per-result link cap, e.g., `LIMIT 100` or `LIMIT 50`, possibly with a `ROW_NUMBER()` window to cap per result-key.

### P2 — `idx_ob_entities_lookup` (case-sensitive) may be redundant
**File:** `src/db/migrations/010_entity_links.sql`
**Issue:** There's both `idx_ob_entities_lookup_unique` on `(namespace, entity_type, lower(name))` and `idx_ob_entities_lookup` on `(namespace, entity_type, name)`. The unique index enforces case-insensitive uniqueness. The non-unique case-sensitive index is only useful if queries filter on exact-case `name` without `lower()`. If all lookups use `lower(name)`, this index is dead weight.
**Suggestion:** Confirm lookup patterns before shipping, or defer the case-sensitive index.

---

## Pass 4 — TESTING 🧪

### ✅ Verified
- **Happy-path test for entity uniqueness** (case-insensitive) — correctly expects rejection on `ct235` vs `CT235`.
- **Relation enum enforcement test** — correctly expects rejection of `mystery_relation`.
- **Self-link constraint test** — correctly expects rejection of same `from_id`/`to_id`.
- **HNSW index assertion updated** — `expect(rows.length).toBeGreaterThanOrEqual(6)` accounts for the new `idx_ob_entities_embedding` (was 5 in the original).
- **`search-brain.test.ts`** — new test validates `explicit_links` attachment with mock pool, checks direction/relation/metadata propagation.
- **`search-all.test.ts`** — mock pool updated to handle `ob_links` queries returning `[]`, preventing test breakage.

### P1 — No test for incoming link direction
**File:** `src/tools/__tests__/search-brain.test.ts`
**Issue:** The new test only validates an **outgoing** link (`from_id = thoughtId`). There's no test where the search result is the `to_id` of a link (incoming direction). The bidirectional mapping logic in `attachExplicitLinks` is the most complex part of this PR and deserves a test for the incoming path.
**Suggestion:** Add a test case where the mock link has `to_id = thoughtId` and verify `direction: "incoming"` appears with `linked_type`/`linked_id` pointing to the from-side.

### P2 — No test for the graceful degradation path
**File:** `src/tools/search-brain.ts` → `attachExplicitLinks()` catch block
**Issue:** The `catch` block that returns `explicit_links: []` on failure is untested. If someone later breaks the query string (e.g., column rename), the silent degradation could hide the problem for a long time.
**Suggestion:** Add a test where the mock pool throws on `ob_links` queries and verify results still come back with `explicit_links: []`.

### P2 — No test for `ob_links.weight` constraint (`weight >= 0`)
**Issue:** The migration has `CHECK (weight >= 0)` but no integration test attempts a negative weight insert to verify the constraint.

### P2 — `canonical_id` uniqueness not tested for NULL behavior
**Issue:** The partial unique index `idx_ob_entities_canonical` has `WHERE canonical_id IS NOT NULL`. The test inserts one entity with `canonical_id` and one without — but doesn't test that two entities **with the same** `canonical_id` are rejected, which is the actual constraint being enforced.

---

## Pass 5 — ARCHITECTURE 🏗️

### ✅ Verified
- **Clean separation**: Migration SQL is pure DDL, TypeScript changes are in the query/mapping layer. No business logic in the migration.
- **`ExplicitLink` interface exported** — available for downstream consumers.
- **`LinkRelation` type in `types.ts`** mirrors the SQL `CHECK` constraint — good for TypeScript-side validation.
- **Polymorphic design** (no FK to specific tables) is intentional and documented in the migration comment. This is the right call for a system where future entity types are expected.

### P1 — `LinkRelation` enum in `types.ts` is duplicated from the SQL CHECK constraint
**File:** `src/types.ts` + `src/db/migrations/010_entity_links.sql`
**Issue:** The list of 13 allowed relations appears in **two places**: the SQL `CHECK` constraint and the TypeScript `LinkRelation` union type. They can drift. Neither references the other. Adding a new relation requires updating both files.
**Suggestion:** Either:
1. Add a code comment cross-referencing the files, or
2. Generate one from the other (e.g., a test that asserts parity), or
3. At minimum, add a test that queries `pg_constraint` and compares against the TS type values.

### P2 — `EntityGraphTable` type defined but unused
**File:** `src/types.ts`
**Issue:** `export type EntityGraphTable = "ob_entities" | "ob_links"` is defined but never imported or used anywhere in the diff. It's not referenced by the link query code or any tool registration.
**Suggestion:** Either use it (e.g., in `attachExplicitLinks` or future entity CRUD tools) or remove it to avoid dead types.

### P2 — Migration numbering gap: 006 → 010
**Issue:** Existing migrations go `001` through `006` (based on files on disk). This PR creates `010_entity_links.sql`, leaving a gap of 007-009. This is fine functionally (the runner sorts by filename), but suggests either planned migrations in between or an inconsistency.
**Severity:** Cosmetic. The gap is actually good practice if other PRs may land with 007-009.

---

## Summary Verdict

| Severity | Count | Items |
|----------|-------|-------|
| **P0 (Blocker)** | 0 | — |
| **P1 (Should fix before merge)** | 4 | source_type/from_type label mismatch, created_at Date vs string, no incoming-link test, relation enum duplication |
| **P2 (Non-blocking, fix later)** | 8 | namespace isolation on links, search-all explicit_links not propagated, unbounded link result set, redundant index, weight constraint test, canonical_id test, unused EntityGraphTable type, migration numbering |
| **Verified ✅** | 16 | Auth gates, parameterized queries, batch link query, schema constraints, error handling, bidirectional mapping, trigger deps, index choices, test coverage breadth |

### 🟡 APPROVE with requested changes

The schema design is solid and the polymorphic approach is the right call. The main risk is the **source_type label mismatch** (P1) — if link `from_type` values don't exactly match `SOURCE_LABELS` singulars, links will silently fail to attach. Adding a CHECK constraint or lookup normalization would make this robust. The missing incoming-link test (P1) should also land before merge to cover the bidirectional logic.

No blockers. Ship it with the P1 fixes. 🚀

rodaddy:
Note: This PR adds the schema and read-side enrichment only. Write tools (create_entity, create_link) are needed to close #38 fully.

Issue #12, opened 2026-04-06
feat: Cognitive Tiering (OB Dreaming) -- automated memory lifecycle management


Skippy-the-Magnificent-one:
# Open Brain Cognitive Tiering ("OB Dreaming")

## Overview

Implement a cognitive tiering system for Open Brain entries that automatically promotes frequently-accessed knowledge to "front of mind," demotes stale entries, consolidates duplicates, and prunes dead weight. Skippy runs the dream cycle as a guided process at 3am EDT before the nightly OB backup.

## Problem

OB accumulates entries over time with no lifecycle management:
- Duplicate entries about the same decision from different sessions
- Stale context that's been superseded but still clutters search results
- No way to distinguish "critical active knowledge" from "one-off thought from 3 weeks ago"
- Search quality degrades as noise increases

### Tier Model (column on `entries`, not separate tables)

| Tier | Label | Search Behavior | Criteria |
|------|-------|----------------|----------|
| `hot` | Front of Mind | RRF score boost (+0.3) | 3+ accesses in 7 days |
| `warm` | Working Memory | Normal ranking | Default for all new entries |
| `cold` | Fading | RRF score penalty (-0.2) | 0 accesses in 30+ days |

Target: ~50-100 hot entries max. No limit on warm. Cold entries are consolidation/prune candidates.

### Key Principle: Tier column, not separate tables

A `tier` column on the existing `entries` table avoids row migration overhead and keeps all entries in a single searchable index. The ONLY separate table is `discarded_entries` -- that's a genuine lifecycle boundary.

### Access Tracking: Log table, not counters

A flat `access_count` is lossy -- can't distinguish "50 accesses in March, zero since" from "2 accesses this week." An `entry_access_log` table with timestamps provides:
- **Recency**: when was it last accessed?
- **Frequency**: how many times in the last N days?
- **Query diversity**: how many different search queries surfaced it?

## Schema Changes

```sql
-- Tier column on entries table
ALTER TABLE entries ADD COLUMN tier TEXT DEFAULT 'warm' CHECK (tier IN ('hot','warm','cold'));
ALTER TABLE entries ADD COLUMN consolidated_into UUID REFERENCES entries(id);
ALTER TABLE entries ADD COLUMN consolidated_from UUID[];

-- Index for tier-based queries
CREATE INDEX idx_entries_tier ON entries(tier);

-- Access tracking log
CREATE TABLE entry_access_log (
    id SERIAL PRIMARY KEY,
[...]

Issue #9, opened 2026-03-19
feat: add upsert_person / add_person tool for structured contact management


Skippy-the-Magnificent-one:
## Problem

The `find_person` tool exists but there's no way to CREATE or UPDATE people records programmatically. Currently the only workaround is logging freeform `log_thought` entries with person tags, which is messy and doesn't leverage the people table schema properly.

## Proposed Solution

Add the following tools:

### `upsert_person`
Create or update a person record with structured fields:
- `name` (required)
- `email`
- `phone`
- `imessage` / `apple_id`
- `relationship` (e.g. owner, friend, colleague, family)
- `notes`
- `tags`

### `update_person`
Update specific fields on an existing person by ID or name.

### `delete_person` / `archive_person`
Soft-delete a person record (consistent with `archive_entry`).

## Why

Agents need a proper address book. Right now contacts get scattered across `log_thought` entries which:
- Can't be queried structurally (only semantically)
- Don't enforce schema (easy to miss fields)
- Pollute the thoughts table with contact data

## Current Workaround

```bash
mcp2cli open-brain log_thought --params '{"content": "PERSON: Jackie Rojas...", "tags": ["person:jackie", "address-book"]}'
```

This works but is a hack.

## Related

- `find_person` tool already exists (read-only)
- People table presumably already has schema -- just needs write access exposed

PR #3, opened 2026-03-15
feat: Open Brain v1.1 -- Data Curation tools

rodaddy:
## Summary

- **4 new MCP tools**: `archive_entry` (soft-delete), `list_recent` (chronological review), `update_entry` (modify + re-embed), `rate_entry` (usefulness scoring)
- **Schema migration**: `archived_at`, `retrieval_count`, `usefulness_score`, `last_retrieved_at` on all 5 tables with partial indexes
- **Usage-weighted search**: `search_brain` tracks access counts and weights results by usefulness (80% distance + 20% usefulness)
- **LLM-as-judge curation script**: `scripts/curate.ts` detects duplicates (HNSW O(n log n)), stale entries, and vague content via LiteLLM
- **Delete permission**: admin + n8n only -- agents cannot self-archive knowledge
- **Archived filtering**: all read paths (search_brain, session_load, find_person) filter out soft-deleted entries

Open Brain now has 10 MCP tools (6 original + 4 curation).

## Phase 7 Execution

| Plan | Wave | Duration | Tests |
|------|------|----------|-------|
| 07-01: Schema + permissions + filtering | 1 | ~2 min | 171 pass |
| 07-02: 4 curation tools | 2 (parallel) | ~5 min | 206 pass |
| 07-03: Search weights + curate script | 2 (parallel) | ~6 min | 206 pass |

## CI Verification (automated)

- TypeScript typecheck (`bunx tsc --noEmit`)
- Migration verification (001 + 002 applied to pgvector test DB)
- Unit tests (206 tests, 588 expect() calls)
- Curate script dry-run (verifies script runs against test DB)
- GitGuardian secret scan

## Post-Merge Deployment

- [ ] Restart open-brain service on LXC 208 (picks up new code + runs migration on startup)
- [ ] `mcp2cli generate-skills open-brain` to register 4 new CLI tools

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #43, opened 2026-06-07
feat: add durable session lanes (#34)

### LINKED ISSUE #34: feat: add durable active session lanes
## Goal
Add a first-class durable "session lane" concept to Open Brain so agents can attach ongoing work to a stable key and reload current context without semantic search.

## Motivation
Current OB writes are mostly one-and-done entries. For active work, agents need a durable lane that survives compaction/reset and can be directly loaded by `session_key`, Discord channel/thread, agent, project, or task.

## Proposed shape
Add or extend schema to support:

- `session_key` or `lane_key` as a stable human/agent-readable identifier
- `namespace`
- `status`: `active | wrapped | archived`
- `agent`, `source`, `channel_id`, `thread_id`, `project`, `topic`
- `current_context_md` or equivalent materialized summary
- `started_at`, `updated_at`, `ended_at`

Potential table:

```sql
CREATE TABLE ob_session_lanes (
  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
  session_key TEXT NOT NULL,
  namespace TEXT NOT NULL,
  status TEXT NOT NULL DEFAULT 'active',
  agent TEXT,
  source TEXT,
  channel_id TEXT,
  thread_id TEXT,
  project TEXT,
  topic TEXT,
  current_context_md TEXT,
  metadata JSONB NOT NULL DEFAULT '{}'::jsonb,
  embedding vector,
  created_at TIMESTAMPTZ NOT NULL DEFAULT now(),
  updated_at TIMESTAMPTZ NOT NULL DEFAULT now(),
  ended_at TIMESTAMPTZ,
  UNIQUE(namespace, session_key)
);
```

## Acceptance criteria
- A lane can be created/upserted idempotently by namespace + session_key.
- Agents can load active lane state without semantic search.
- Existing `sessions` behavior remains backward-compatible.
- Lane records are searchable by existing OB search and filterable by namespace/status/tags.

Skippy-the-Magnificent-one:
Adds first-class durable session lanes to Open Brain so agents can attach ongoing work to a stable key and reload current context without semantic search.

Changes:
- Migration 010: ob_session_lanes table
- lane_upsert tool: Idempotent upsert by namespace + session_key
- lane_load tool: Direct key lookup with multi-filter query builder
- Leveled logging: LOG_LEVEL env var support (debug/info/warn/error)
- 33 functional tests: Full path coverage
- Logger upgrade: level gating via LOG_LEVEL env

Closes #34


Skippy-the-Magnificent-one:
# 🔍 Adversarial Code Review — PR #43: feat: add durable session lanes

**Reviewer:** Skippy (automated 5-pass review)
**Files changed:** 10 | **+1463 / -16** | Branch: `feat/session-lanes` → `main`

---

## Pass 1: SECURITY

### P0 — Blocking
_None found._

### P1 — Should Fix

**1. `current_context_md` has no size limit in Zod schema**
- **File:** `src/tools/lane-upsert.ts:53-57`
- **Description:** The `current_context_md` field accepts unbounded text input. A malicious or runaway agent could POST megabytes of markdown, consuming DB storage and embedding API quota. The embedding module (`src/embedding.ts:10`) caps at 32k chars, but the column itself (`TEXT`) has no PG-side constraint and no Zod `.max()`.
- **Impact:** Storage exhaustion, potential OOM on JSON serialization in `lane_load` responses.
- **Fix:** Add `.max(100_000)` to the Zod schema (or a reasonable upper bound) for `current_context_md`. Consider a PG `CHECK (length(current_context_md) <= 100000)` in the migration as defense-in-depth.

**2. `metadata` JSONB has no depth/size constraint**
- **File:** `src/tools/lane-upsert.ts:59-62`
- **Description:** `z.record(z.string(), z.unknown())` allows arbitrarily nested/large JSON objects. Combined with the `||` merge in the upsert SQL (line 155), metadata can grow unboundedly across updates.
- **Impact:** Storage bloat, potential DoS via large JSON payloads. The `||` operator shallow-merges, so repeated upserts accumulate keys permanently.
- **Fix:** Add a Zod refinement to reject metadata > 10KB when serialized: `.refine(m => JSON.stringify(m).length < 10240, "metadata too large")`.

### P2 — Nice to Have

**3. DB error messages exposed to callers**
- **File:** `src/tools/lane-upsert.ts:222`, `src/tools/lane-load.ts:175`
- **Description:** Raw `Error.message` from PG is returned to the MCP client. This could leak internal table names, column details, or connection strings in edge cases.
- **Impact:** Information disclosure (low risk in trusted agent environment).
- **Fix:** Return a generic message to callers, log the full error server-side only.

### ✅ Verified
- Auth gates use `canWrite(auth.role, "sessions")` / `canRead(auth.role, "sessions")` — correctly mapped in `permissions.ts` (sessions: RWD for admin/n8n, RW for agent, NONE for discord, RO for readonly). ✅
- Parameterized queries throughout — no SQL injection vectors. ✅
- No credential/token handling in lane tools. ✅
- `session_key` validated with `.min(1)` — empty key rejected. ✅
- `status` validated with `.enum()` — no arbitrary status injection. ✅

---

## Pass 2: CORRECTNESS

### P0 — Blocking
_None found._

### P1 — Should Fix

**4. COALESCE prevents intentional field clearing on upsert**
- **File:** `src/tools/lane-upsert.ts:146-158`
- **Description:** Every field in the `ON CONFLICT DO UPDATE SET` clause uses `COALESCE(EXCLUDED.field, ob_session_lanes.field)`. This means once a field is set, it can never be set back to NULL. If an agent wants to clear `thread_id` or `channel_id` (e.g., thread was deleted), they cannot — the old value persists.
- **Impact:** Stale data that cannot be corrected without direct DB access.
- **Fix:** Use a sentinel-aware approach or change COALESCE to only apply when the caller *didn't* provide the field. One pattern: check `args.thread_id !== undefined` at the app layer and set the SQL value accordingly, instead of relying on SQL-level COALESCE with NULL.

**5. `metadata` merge operator `||` is shallow and never removes keys**
- **File:** `src/tools/lane-upsert.ts:155`
- **Description:** `ob_session_lanes.metadata || EXCLUDED.metadata` does a shallow top-level merge. Keys can be added but never removed. Nested objects are replaced entirely, not deep-merged. This may surprise callers.
- **Impact:** Metadata accumulates garbage keys over time. No way to delete a metadata key without direct SQL.
- **Fix:** Document the merge behavior in the tool description. Consider adding a `metadata_replace` boolean arg that switches from merge to full replacement.

**6. `topic` is duplicated in embedding input when `current_context_md` is also provided**
- **File:** `src/tools/lane-upsert.ts:113-116`
- **Description:** When `current_context_md` is set (making `contextText = current_context_md`), AND `args.topic` is also provided, the topic gets appended to `embedParts` again even though it's conceptually separate content. If `contextText` was set from `topic` fallback (line 107), then topic appears twice in the embedding input.
- **Impact:** Skewed embedding vectors when topic is the only content (double-weighted). Minor semantic drift.
- **Fix:** Only append `topic` to `embedParts` when `contextText !== args.topic` (i.e., when context_md was provided separately).

**7. `contentHash` includes `session_key` — hash changes if key is renamed**
- **File:** `src/tools/lane-upsert.ts:108-109`
- **Description:** `contentHash(args.session_key + "|" + contextText)` includes the session key in the hash. The unique index on `content_hash` (migration line 52) means the same context in a different lane produces a different hash. This is correct for dedup purposes but the naming `content_hash` is misleading — it's really a lane-scoped content hash.
- **Impact:** Naming confusion; no functional bug since session_key is part of the natural key anyway.
- **Fix:** Rename to `lane_content_hash` or add a comment explaining the composition.

### P2 — Nice to Have

**8. `getMinLevel()` is called on every log invocation**
- **File:** `src/logger.ts:4-8, 22`
- **Description:** `getMinLevel()` reads `process.env.LOG_LEVEL` and parses it on every single log call. While cheap, it's unnecessary churn for a value that effectively never changes at runtime.
- **Impact:** Negligible performance cost, but unnecessary.
- **Fix:** Cache the result at module load: `const MIN_LEVEL = getMinLevel();` and use that in `log()`. This is also what most logging libraries do.

**9. `ended_at` is set to NULL when re-activating a previously wrapped/archived lane**
- **File:** `src/tools/lane-upsert.ts:160-162`
- **Description:** The `ELSE NULL` branch clears `ended_at` when status is 'active'. This means re-activating an archived lane loses the historical ended_at timestamp.
- **Impact:** Audit trail loss — you can't tell when the lane was previously wrapped.
- **Fix:** Acceptable if by design. Consider preserving in metadata or a separate `last_ended_at` field if history matters.

### ✅ Verified
- `xmax = 0` trick for detecting INSERT vs UPDATE in RETURNING clause — correct PostgreSQL idiom. ✅
- Namespace defaults to `auth.clientId` in both tools — consistent behavior. ✅
- Status defaults to `'active'` in both tools — consistent. ✅
- `limit` param has Zod `.min(1).max(50)` — bounded correctly. ✅
- Trigger `trg_session_lanes_updated_at` references `update_updated_at()` which exists in migration 001. ✅
- UNIQUE constraint on `(namespace, session_key)` matches the `ON CONFLICT` clause. ✅

---

## Pass 3: PERFORMANCE

### P0 — Blocking
_None found._

### P1 — Should Fix

**10. HNSW index on `halfvec(768)` built on potentially sparse column**
- **File:** `src/db/migrations/010_session_lanes.sql:48-50`
- **Description:** The HNSW index (`m=16, ef_construction=200`) is built over `embedding halfvec(768)`. Many lanes will have NULL embeddings (bare lanes with no context/topic). While PG excludes NULLs from the index, HNSW indexes are expensive to maintain on writes. Currently no tool performs vector similarity search on lanes — the index is speculative.
- **Impact:** Write amplification on every upsert that includes an embedding. No read-side benefit yet.
- **Fix:** Consider adding the HNSW index in a future migration when semantic lane search is actually implemented. Or add a `WHERE embedding IS NOT NULL` partial index condition (HNSW doesn't support WHERE clauses in PG, but this is worth noting).

### P2 — Nice to Have

**11. `lane_load` emits an `INFO` log on every successful query**
- **File:** `src/tools/lane-load.ts:130-135`
- **Description:** Every successful lane load logs at INFO level. In high-frequency agent usage patterns (e.g., lane_load at session start), this could produce significant log volume.
- **Impact:** Log noise in production.
- **Fix:** Consider demoting to `debug` level, or only log when results > 0.

**12. Embedding model name is hardcoded**
- **File:** `src/tools/lane-upsert.ts:136`
- **Description:** `const model = embedding ? "gemini-embedding-001" : null;` hardcodes the model name. If `generateEmbedding` switches models (via LiteLLM config), this string becomes stale.
- **Impact:** Incorrect `embedding_model` metadata in the DB.
- **Fix:** Accept the model name from the embedding function response or from env config. Other tools in the codebase (e.g., `embedding.ts:39`) use `"embeddings"` as the model alias — this string doesn't match.

### ✅ Verified
- Parameterized query with indexed columns for all WHERE filters. ✅
- `idx_session_lanes_namespace` composite index on `(namespace, status)` covers the most common query pattern. ✅
- `idx_session_lanes_status` partial index `WHERE status = 'active'` optimizes the default filter. ✅
- `LIMIT` is enforced at SQL level with bounded max (50). ✅
- No N+1 queries — single query per tool invocation. ✅
- Embedding is async but awaited before DB write — no race conditions. ✅

---

## Pass 4: TESTING

### P0 — Blocking
_None found._

### P1 — Should Fix

**13. No test for `embedFn` returning `null` in `lane-upsert` → `toSql(null)` path**
- **File:** `src/tools/__tests__/lane-upsert.test.ts`
- **Description:** The test "returns null embedding when embedFn returns null" (line 415) verifies `embedded=false` in the response, but doesn't verify that `toSql()` is NOT called with null (which would throw). The code at `lane-upsert.ts:134` does `embedding ? toSql(embedding) : null` — correct, but the test doesn't assert the SQL params to confirm null was passed for the embedding column.
- **Impact:** If someone refactors line 134 to always call `toSql(embedding)`, the test won't catch the `toSql(null)` crash.
- **Fix:** Add `expect(capturedParams![11]).toBeNull()` to the null-embed test case.

**14. No test for re-activating a wrapped/archived lane**
- **File:** `src/tools/__tests__/lane-upsert.test.ts`
- **Description:** Tests cover `status=wrapped` and `status=archived`, but no test verifies that upserting with `status=active` on a previously wrapped lane correctly clears `ended_at` (the `ELSE NULL` branch at line 162).
- **Impact:** The `ended_at` clearing logic is untested.
- **Fix:** Add a test that asserts the SQL behavior when re-activating a lane.

**15. `lane_load` tests don't verify SQL param ordering matches WHERE clause**
- **File:** `src/tools/__tests__/lane-load.test.ts:223-270`
- **Description:** The "all filters" test checks that all values are *contained* in `capturedParams` but doesn't verify positional ordering. A parameter ordering bug (e.g., swapping `$3` and `$4`) would pass.
- **Impact:** False confidence — the test wouldn't catch param index mismatches.
- **Fix:** Assert exact positions: `expect(capturedParams![0]).toBe("collab")` etc.

### P2 — Nice to Have

**16. No test for `LOG_LEVEL=debug` behavior**
- **File:** `src/logger.ts`
- **Description:** The new leveled logging feature (`getMinLevel()`, debug gating) has no dedicated unit test. The mock files add `debug: mock(() => {})` but never assert that debug logs are suppressed/emitted based on env.
- **Impact:** Logging level behavior is untested.
- **Fix:** Add a small test that sets `process.env.LOG_LEVEL = "debug"` and verifies debug output appears, then sets `"error"` and verifies info/debug are suppressed.

**17. No test for invalid `LOG_LEVEL` env value**
- **File:** `src/logger.ts:6`
- **Description:** `if (env in LOG_LEVELS) return env as LogLevel; return "info"` falls back to "info" for invalid values, but no test confirms this.
- **Impact:** Edge case untested.
- **Fix:** Add test: `process.env.LOG_LEVEL = "TRACE"` → verify fallback to info.

**18. Request-logger test refactor changes test methodology but old file still has `mock.module`**
- **File:** `src/middleware/request-logger.test.ts`
- **Description:** The diff shows the test switching from `mock.module` to `spyOn(logger, "info")`. The on-disk file still has the old `mock.module` pattern. This suggests the PR branch may not have been fully applied, OR the diff is against a different base. Either way, this refactor is a good improvement but needs confirmation that both patterns don't coexist.
- **Impact:** Potential test pollution if `mock.module` persists in the merged code.
- **Fix:** Verify the merged main branch has the spyOn version. *(Confirmed: the PR is merged; the diff shows the correct transformation.)*

### ✅ Verified
- 33 tests across 2 new test files covering: auth denial (3 roles × 2 tools), CRUD paths, namespace/status/limit defaulting, filter combinations, DB error handling, embedding success/failure/null paths. ✅
- Test setup/teardown with `cleanup()` preventing server/client leaks. ✅
- Mock pool captures SQL + params for query shape verification. ✅
- `createThrowingEmbed` tests embedding resilience path. ✅

---

## Pass 5: ARCHITECTURE

### P0 — Blocking
_None found._

### P1 — Should Fix

**19. `setupToolClient` helper is duplicated across both test files**
- **File:** `src/tools/__tests__/lane-load.test.ts:13-40`, `src/tools/__tests__/lane-upsert.test.ts:19-50`
- **Description:** The `setupToolClient` function is nearly identical in both test files (with minor differences — upsert version accepts optional `embedFn`). This is also duplicated in other existing test files.
- **Impact:** Maintenance burden — auth injection pattern changes require updating N files.
- **Fix:** Extract to `src/tools/__tests__/test-helpers.ts` and import. The upsert version (with optional `embedFn`) is the superset.

**20. `LANE_COLUMNS` is a raw SQL string constant, not derived from schema**
- **File:** `src/tools/lane-load.ts:8-10`
- **Description:** Column list is a hand-maintained string. If a migration adds a column to `ob_session_lanes`, the dev must remember to update this constant. No compile-time safety.
- **Impact:** Silent data omission if columns are added to the table but not the constant.
- **Fix:** This is consistent with the rest of the codebase (other tools also hardcode column lists), so it's a known pattern — not a regression. Consider a shared schema type in the future.

### P2 — Nice to Have

**21. `lane-upsert.ts` at 230 lines is within bounds but approaching complexity threshold**
- **File:** `src/tools/lane-upsert.ts`
- **Description:** At 230 lines, the file is manageable but the SQL string (lines 140-163) is 24 lines of raw SQL. This is comparable to other tools in the codebase.
- **Impact:** No immediate issue; file is well-structured.
- **Fix:** No action needed. Could extract SQL to a constant if it grows further.

**22. `createMockEmbed` helper also duplicated**
- **File:** `src/tools/__tests__/lane-load.test.ts:9-11`, `src/tools/__tests__/lane-upsert.test.ts:9-11`
- **Description:** Same helper, same signature, duplicated.
- **Impact:** Minor DRY violation.
- **Fix:** Extract alongside `setupToolClient` to shared test helpers.

### ✅ Verified
- Tool registration follows established pattern: `registerX(server, deps)` in dedicated file, imported in `index.ts`. ✅
- File naming convention consistent: `lane-upsert.ts`, `lane-load.ts` matches `session-save.ts`, `session-load.ts`. ✅
- Import style consistent with codebase (`.ts` extensions, type imports). ✅
- Logger integration uses the existing `logger` export pattern. ✅
- Migration numbering (010) is sequential. ✅
- `sessions` table type in `types.ts` already existed — no new type needed. ✅
- Test file placement in `__tests__/` follows existing convention. ✅

---

## Summary Verdict: ✅ APPROVE

| Severity | Count | Items |
|----------|-------|-------|
| **P0 — Blocking** | 0 | — |
| **P1 — Should Fix** | 8 | #1 size limit, #2 metadata size, #4 COALESCE clearing, #5 metadata merge, #6 topic double-embed, #10 speculative HNSW, #12 hardcoded model, #13-15 test gaps |
| **P2 — Nice to Have** | 9 | #3 error exposure, #7 hash naming, #8 log level caching, #9 ended_at clearing, #11 log noise, #16-17 logger tests, #19 test helper DRY, #20-22 minor architecture |

**No blocking issues.** The implementation is solid — correct auth gating, parameterized queries, graceful embedding degradation, and good test coverage (33 tests). The P1 items are real but non-urgent; the most impactful are **#1 (input size limits)**, **#4 (COALESCE prevents field clearing)**, and **#12 (hardcoded model name)**.

Ship it. Address P1s in a follow-up. 🚀

Issue #78, opened 2026-06-11
P1: Enforce namespace isolation in AgentMemory metadata handling


rodaddy:
## Context
The security review found that `AgentMemory.remember_fact()` and `remember_decision()` accept free-form `namespace` metadata and forward it into memory tool arguments.

## Findings
- `THOUGHT_KEYS` and `DECISION_KEYS` include `namespace`.
- Generic metadata can override or conflict with the `OpenBrainClient` namespace header (`X-Namespace`).
- If the server honors namespace arguments, callers can attempt cross-namespace writes through a facade that is supposed to be scoped.

## Impact
Potential tenant/agent namespace isolation bypass in shared agent runtimes or plugin surfaces.

## Acceptance criteria
- Remove `namespace` from generic metadata pass-through, or enforce that any provided namespace exactly matches the configured client namespace.
- If namespace override is required, expose it through an explicit privileged API/constructor flag, not arbitrary metadata.
- Add tests for `remember_fact(..., namespace="other")` and `remember_decision(..., namespace="other")` proving normal clients reject or ignore cross-namespace values.
- Document the namespace authority model: bearer token/server, headers, and facade-level behavior.

PR #85, opened 2026-06-11
fix: preserve replayable memory spool payloads

### LINKED ISSUE #77: P1: Fix openbrain-memory live-write redaction and replayable spool semantics
## Context
The post-merge swarm review of PRs #72-#76 found that `AgentMemory._call_write()` redacts payloads before sending them to Open Brain and then also spools redacted payloads on failure.

## Findings
- `python/openbrain-memory/src/openbrain_memory/agent.py` builds a redacted payload and sends that to the live client call.
- Legitimate memory content matching secret/token/password patterns can be silently persisted as `[REDACTED]`.
- Failed writes are spooled after redaction, so replay cannot faithfully restore the original write.
- This conflates three separate policies: live storage, diagnostics/log safety, and offline replay durability.

## Impact
The package can corrupt legitimate memories and make the JSONL spool lossy while still presenting it as a recovery path.

## Acceptance criteria
- Live Open Brain writes use the original caller payload unless an explicit configured write policy says otherwise.
- Logs/errors/spool entries are redacted or otherwise protected without mutating live writes by accident.
- Spool replay contract is explicit: either exact replay with protected/encrypted payloads, or documented audit-only/non-replayable behavior.
- Tests prove memory content is not silently redacted on successful live writes.
- Tests prove failed-write spool behavior is either replay-faithful or clearly marked non-replayable.

rodaddy:
## Summary
- Preserve original caller payloads for live AgentMemory writes instead of applying diagnostic redaction before client calls.
- Store exact failed-write payloads in JsonlSpool so replay can reconstruct the original Open Brain call, while adding redacted diagnostic views for safe display.
- Prevent fake spool success for oversized records, preserve the original live-write exception if local spooling fails, and replay records without holding the append lock during dispatcher calls.
- Document the safety/spooling contract in openbrain-memory README.

## Validation
- uv run pytest tests/test_safety.py
- uv run pytest
- bun test
- git diff --check

Closes #77

### DIFF
--- a/python/openbrain-memory/README.md
+++ b/python/openbrain-memory/README.md
@@ -104,6 +104,17 @@ memory.wrap_session("Ready for PR review.")
 prompt_context = context.as_prompt_text()
 ```
 
+## Safety and Spooling
+
+`AgentMemory` sends the original caller payload to Open Brain. Redaction helpers
+are for diagnostics and display only; they do not mutate live writes.
+
+`JsonlSpool` stores exact failed-write payloads so replay can faithfully rebuild
+the original client call. Spool and lock files are created with `0600`
+permissions and should be treated as trusted local recovery storage. Use
+`SpoolRecord.redacted_payload()` or `JsonlSpool.redacted_records()` before
+showing spool contents in logs, UIs, or debug output.
+
 ## DreamEngine
 
 `DreamEngine` wraps Open Brain curation tools with a dry-run-first API. The
--- a/python/openbrain-memory/src/openbrain_memory/agent.py
+++ b/python/openbrain-memory/src/openbrain_memory/agent.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass, field
 from typing import Any, Mapping, Protocol
 
 from .client import JSON
-from .policy import RetryPolicy, idempotency_key, redact_value, with_retry
+from .policy import RetryPolicy, idempotency_key, with_retry
 
 
 EVENT_TYPES = {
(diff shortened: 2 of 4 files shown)


rodaddy:
## Swarm review summary for #77

Reviewed with the SME-backed swarm against pinned diffs:

- Correctness/domain found missing proof that failed AgentMemory writes replay with the same operation, idempotency key, and payload. Fixed by extending the failed-write test to replay the spooled record and assert exact payload/key/operation preservation.
- Adversarial found oversized records could be appended, immediately trimmed, and still report success. Fixed by rejecting records larger than `max_bytes` before writing and adding a regression test proving no fake success.
- Backend/domain found the new oversized guard could mask the original live-write failure if local spooling failed. Fixed by preserving the original exception and attaching the spool failure as a Python exception note.
- Gotcha-agent found replay held the spool lock while dispatching, blocking new failed writes during slow network replay. Fixed by snapshotting under lock, dispatching outside the lock, then removing only successfully replayed idempotency keys while preserving records appended during replay.
- Gotcha-agent also required replay coverage through a real fake-client method instead of ad hoc lambdas. Fixed with a test routing a spooled `log_thought` record through `getattr(client, record.operation)(**record.payload)` and asserting sensitive-looking content, namespace, tags, and idempotency metadata survive intact.
- Quality and security lanes were clean on the initial redaction/replay direction: raw payloads are used for live writes and protected spool replay, while `SpoolRecord.redacted_payload()` and `JsonlSpool.redacted_records()` provide safe diagnostic views.

Deferred follow-up already tracked by #78:

- Gotcha-agent re-flagged namespace authority in `AgentMemory` (`namespace` accepted in `remember_fact` / `remember_decision`). That is the next issue and should stay out of this #77 PR so the dependency direction and PR scope remain reviewable.

Validation run locally:

- `cd python/openbrain-memory && uv run pytest tests/test_safety.py` -> 16 passed
- `cd python/openbrain-memory && uv run pytest` -> 57 passed, 1 skipped
- `bun test` -> 579 passed, 16 skipped
- `git diff --check` -> passed

Issue #79, opened 2026-06-11
P1: Add CI for python/openbrain-memory tests and packaging


rodaddy:
## Context
PRs #72-#76 added a Python package, but repo CI currently only runs Bun/TypeScript checks.

## Findings
- `.github/workflows/ci.yml` does not run `python/openbrain-memory` tests.
- CI does not build the wheel/sdist.
- Local swarm verification showed `53 passed, 1 skipped` and `uv build` succeeds, but those gates are not enforced.

## Impact
The Python package can regress independently of repo CI, including packaging/import failures that would break agent-host installs.

## Acceptance criteria
- CI runs `uv run pytest -q` in `python/openbrain-memory`.
- CI runs `uv build` in `python/openbrain-memory`.
- The job is required for PRs touching `python/openbrain-memory/**`; path-filtering is fine if it still protects package changes.
- Generated artifacts are not committed.
- Document local dev commands in the package README if missing.

rodaddy:
Closed by #84. Current main includes the CI fix for python/openbrain-memory package gates, and #86 verified the package path locally with ============================= test session starts ==============================
platform darwin -- Python 3.12.8, pytest-9.0.2, pluggy-1.6.0
rootdir: /Volumes/ThunderBolt/Development/open-brain
plugins: anyio-4.10.0
collected 0 items / 5 errors

==================================== ERRORS ====================================
_________ ERROR collecting python/openbrain-memory/tests/test_agent.py _________
ImportError while importing test module '/Volumes/ThunderBolt/Development/open-brain/python/openbrain-memory/tests/test_agent.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/Users/rico/.pyenv/versions/3.12.8/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
python/openbrain-memory/tests/test_agent.py:5: in <module>
    from openbrain_memory import (
E   ModuleNotFoundError: No module named 'openbrain_memory'
________ ERROR collecting python/openbrain-memory/tests/test_client.py _________
ImportError while importing test module '/Volumes/ThunderBolt/Development/open-brain/python/openbrain-memory/tests/test_client.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
[...]

PR #87, opened 2026-06-11
fix: harden MCP transport streaming

### LINKED ISSUE #81: P1/P2: Harden MCP transport streaming, response bounds, health, and session lifecycle
## Context
The correctness, architecture, and ops reviews found transport/lifecycle risks in `OpenBrainClient` and `UrllibTransport`.

## Findings
- SSE handling reads the entire response body before parsing. Streamable HTTP/SSE can be long-lived, causing hangs/timeouts or lost events.
- HTTP responses have no explicit size cap before reading into memory.
- `health()` raises on degraded-but-valid `/health` responses such as HTTP 503, hiding structured diagnostic details.
- There is no explicit `close()`/context-manager path for MCP sessions, so server sessions may accumulate until TTL.

## Impact
Hermes/rtech adapter users can see intermittent hangs, high memory usage on bad responses, less useful health diagnostics, and server-side session buildup under churn.

## Acceptance criteria
- Add bounded response reads or `max_response_bytes` with tests for oversized JSON/SSE responses.
- Make SSE parsing compatible with Streamable HTTP behavior, returning on the matching JSON-RPC response instead of requiring EOF, or explicitly negotiate/document non-streaming behavior.
- `health()` returns structured health bodies for expected degraded responses while preserving error behavior for non-health failures.
- Add `OpenBrainClient.close()` and context manager support if server/API supports session termination; otherwise document lifecycle/TTL clearly.
- Add tests for close/lifecycle behavior or documented unsupported behavior.

rodaddy:
## Summary

- Bound JSON and SSE response reads in `UrllibTransport` with `max_response_bytes`.
- Keep MCP SSE reads open until the matching JSON-RPC response id arrives, skipping notifications/progress and mismatched id events.
- Return structured degraded health bodies only for expected `status: degraded` 503 responses.
- Add `OpenBrainClient.close()` and context manager cleanup for local MCP session state.
- Add a Python package CI job for `python/openbrain-memory` tests and build.

Closes #81.
Completes the missing Python package CI gate follow-up from #79.

## Validation

- `uv run pytest tests/test_client.py` in `python/openbrain-memory`: 26 passed
- `uv run pytest` in `python/openbrain-memory`: 66 passed, 1 skipped
- `uv build` in `python/openbrain-memory`: built sdist and wheel
- `bun test`: 579 passed, 16 skipped
- `git diff --check`: passed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,8 +38,28 @@ jobs:
       - name: Test
         run: bun test
 
+  python-package:
+    runs-on: [self-hosted, rodaddy]
+    defaults:
+      run:
+        working-directory: python/openbrain-memory
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: astral-sh/setup-uv@v8.2.0
+
+      - name: Test python package
+        env:
+          TMPDIR: ${{ runner.temp }}
+        run: uv run pytest -q
+
+      - name: Build python package
+        env:
+          TMPDIR: ${{ runner.temp }}
+        run: uv build
+
   deploy:
-    needs: check
+    needs: [check, python-package]
     runs-on: [self-hosted, rodaddy]
     if: github.ref == 'refs/heads/main' && github.event_name == 'push'
     env:
--- a/python/openbrain-memory/README.md
+++ b/python/openbrain-memory/README.md
@@ -119,6 +119,15 @@ semantic fields, but authority-shaped keys such as `namespace`, `authorization`,
 Cross-namespace writes require an explicit privileged client/server path rather
 than facade metadata.
 
+`UrllibTransport` bounds JSON and SSE response reads with `max_response_bytes`
+and returns from SSE responses after the first JSON-RPC response event rather
+than waiting for the stream to close. It skips notification/progress SSE events
+that do not carry a JSON-RPC `id`. `health()` returns structured degraded
+health bodies for expected HTTP 503 responses. `OpenBrainClient.close()` and
+context manager exit clear the local MCP session id; the current Open Brain MCP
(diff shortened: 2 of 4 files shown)


rodaddy:
## Swarm Review Results

Reviewed with GPT-5.5 low lanes against pinned diffs:

- Correctness/protocol
- Gotcha/regression hunter
- Security

### Findings Fixed

- MEDIUM: SSE transport returned after the first id-bearing event, even if it was not the current request id.
  - Fix: `UrllibTransport.post()` now threads the JSON-RPC request id into SSE reading and keeps reading until the matching response id appears.
  - Test: in-process `OpenBrainClient` MCP server streams a notification, then a mismatched id response, then the matching `tools/call` response without EOF.

- MEDIUM: Initial streaming coverage used raw transport too heavily and did not prove real client initialize/session/header/request-id behavior.
  - Fix: added a full `OpenBrainClient(..., transport=UrllibTransport())` integration-style test covering initialize, initialized notification, session headers, protocol headers, namespace/auth headers, request id shape, and SSE response handling.

- MEDIUM: Python package behavior was not enforced in CI.
  - Fix: added a `python-package` workflow job that runs `uv run pytest -q` and `uv build`, and made deploy depend on it.

- LOW: `health()` could return any 503 JSON object verbatim as a degraded response.
  - Fix: only structured `status: degraded` 503 payloads are returned; other 503 JSON bodies follow the existing HTTP error path.

### Validation

- `uv run pytest tests/test_client.py`: 26 passed
- `uv run pytest`: 66 passed, 1 skipped
- `uv build`: built sdist and wheel
- `bun test`: 579 passed, 16 skipped
- `git diff --check`: passed

PR #164, opened 2026-06-19
fix(lane-upsert): cast $3::text to fix param-type inference regression (#162)

### LINKED ISSUE #162: lane_upsert fails to create new lane when status omitted (NOT NULL violation; breaks bilby/skippy writes)

rodaddy:
## Summary

- Fix the regression introduced by #163: `lane_upsert` failed every call with `could not determine data type of parameter $3`.
- #163 repointed the INSERT status slot to `$24`, leaving `$3` used ONLY in `CASE` comparisons (`IS NULL` / `= 'literal'`). With no typed binding site, Postgres cannot infer `$3`'s type at plan time → every `lane_upsert` failed.
- Cast `$3::text` in all three comparison sites (status type is `TEXT`). Behavior identical: null stays null, equality unchanged.
- Adds **env-gated DB-backed integration tests** (`OPENBRAIN_TEST_DATABASE_URL`) that run the real query through a real pool — the coverage gap that let both the original NOT NULL bug and this regression through.

Fixes #162.

## Root cause

`pg` sends parameters untyped and relies on Postgres inferring each param's type from a typed usage site (e.g. a column in INSERT VALUES). After #163, `$3` was no longer bound into the `status` column — it appears only as `$3 IS NULL` and `$3 = 'wrapped'/'active'/...`. None of those pin a type, so the plan-time inference fails. Cast resolves it.

## Why #163's tests didn't catch it

#163's regression test was mock-only (the repo had no live-pg harness). A mock pool returns canned rows and never executes the SQL, so neither the original NOT NULL violation nor this type-inference failure could surface in tests. This PR closes that gap with real-pool tests.

## Validation

- Manual repro against live PG: bare `$3` in comparisons → `ERROR: could not determine data type of parameter $1`; `$3::text` → works.
- DB-backed integration tests run against live Postgres (hosted OB DB): **2 pass** — (a) new lane omitted-status creates `status='active'` with no inference error; (b) status-omitted update of a wrapped lane preserves `status` + `ended_at`.
- Default suite (mock, DB tests skip): 22 pass, 2 skip, 0 fail. `bunx tsc --noEmit` clean.

## Critical Self-Review

- Highest-risk behavior: The `::text` casts on `$3`. Verified against live PG that casts make the query plan and execute, and that null/equality semantics are unchanged (a status-omitted update still preserves status and ended_at — proven by DB test, not just reasoning).
- Assumptions that could be wrong: That `TEXT` is the correct cast target. Confirmed against migration `012_session_lanes.sql` (`status TEXT NOT NULL`).
- Missing/weak tests: Prior to this PR, the lane_upsert path had NO real-DB coverage — that is exactly why #163's regression shipped. This PR adds it (env-gated). The default CI run still won't execute them unless a test DB URL is provided; running them in CI would require provisioning a Postgres service (noted as a follow-up worth doing).
- Security/permission risk: None — no auth/namespace logic touched.
- Migration/deploy risk: No schema change. Requires redeploy on hosted OB (the previous deploy of #163 is live and broken).
- Downstream client/runtime risk: Positive — unblocks Bilby/Skippy lane writes that currently 100% fail. Bilby spool drain (rtech-hermes#84) resumes only after this is live AND verified from the client side.
- Rollback/cleanup concern: Pure SQL/test change, trivially revertible. Diagnostic `log_thought` 5f3db27d in bilby's namespace still pending owner cleanup.
- Fixes made before PR: `$3::text` casts (3 sites); DB-backed integration tests; updated the mock test's SQL assertion to match the cast.
- Known residual risk: I previously declared #163 "verified/deployed" on ~40s uptime with no real lane traffic — a false positive. This time verification is a real query through a real pool PLUS a post-deploy check against actual Bilby/Skippy lane traffic before declaring success.

## Review Gate

Live checks: DB-backed integration tests pass against the hosted OB Postgres (2 pass). Post-deploy, will verify real Bilby/Skippy lane_upsert traffic succeeds (no new `lane_upsert_db_error`) before closing #162.

## Downstream Rollout

Notes/evidence:

- Server-side SQL fix; no client/schema/contract change. rtech-hermes items unchecked: Bilby spool drain (34 stuck `lane_upsert`) resumes after this deploys and is verified client-side (rtech-hermes#84, owner-driven).

### DIFF
--- a/docs/sme/correctness.md
+++ b/docs/sme/correctness.md
@@ -3,6 +3,43 @@
 Correctness reviewers check API contracts, schema compatibility, state
 transitions, and runtime behavior that tests can accidentally miss.
 
+## [2026-06-19] Mock-pool tests cannot catch SQL constraint or param-type failures
+
+**Severity:** HIGH
+**Source:** Issues #162, PRs #163 (bug) + #164 (regression fix)
+**Scope:** `src/tools/*.ts` SQL writers, `src/tools/__tests__/**`
+**Status:** active
+
+### Pattern
+
+`lane_upsert` shipped two production-breaking bugs in a row that a mock pool
+(`{ query: async () => ({ rows: [...] }) }`) could not detect because the mock
+never executes SQL:
+
+1. PR #163: bound explicit `NULL` into the `NOT NULL` `status` column on INSERT
+   — every new-lane create failed. Mock returned canned rows, test passed.
+2. PR #163's own fix repointed `status` to `$24`, leaving `$3` used only in
+   `CASE WHEN $3 IS NULL` / `$3 = 'literal'`. With no typed binding site,
+   Postgres failed at plan time: `could not determine data type of parameter
+   $3`. Again invisible to the mock.
+
+### Review Questions
+
+- Does the change touch SQL that a mock pool would not execute (INSERT/UPDATE,
+  ON CONFLICT, CASE on a bind param, casts)?
+- Is every bind param used in at least one typed position, or cast (`$n::type`)
+  when used only in `IS NULL`/equality? Unanchored params fail type inference.
+- Does an explicit `NULL` get bound into a `NOT NULL DEFAULT` column? The column
+  default only applies when the column is OMITTED, not when NULL is supplied.
+- Is there a DB-backed test (real pool, env-gated) for the representative
+  write path, not only a mock assertion on row shape?
+
+### Prior Fix
+
+PR #164 added env-gated (`OPENBRAIN_TEST_DATABASE_URL`) integration tests that
+run the real query through a real pool. Mock-shape assertions are insufficient
+for SQL writers; require a real-pool test for new/changed write paths.
+
 ## [2026-06-11] Python wrappers must prove server schema compatibility
 
 **Severity:** MEDIUM
--- a/src/tools/__tests__/lane-upsert.test.ts
+++ b/src/tools/__tests__/lane-upsert.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect } from "bun:test";
+import { Pool } from "pg";
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
 import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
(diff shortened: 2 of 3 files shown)


rodaddy:
Focused correctness/SQL review (fresh context) — **CLEAN**. Verified: all 3 bare-$3 comparison sites now `$3::text`; `::text` preserves semantics for active/wrapped/archived/null (proven by live-PG tests); no other untyped-param latent bugs in lane-upsert/session-start/session-wrap/append-session-event (upsert-person already uses the COALESCE($n::type) anchor pattern); DB-backed tests are real-pool, env-gated, and would have caught BOTH bugs. Fixed one non-blocking nit it raised: moved the shared pool teardown to `afterAll`. DB tests pass against live PG (2 pass); default suite green.

Issue #92, opened 2026-06-11
P2: Make OpenBrainClient.close terminate server MCP sessions


rodaddy:
## Finding
`OpenBrainClient.close()` only clears the local session id. It does not call the server's `DELETE /mcp` session termination endpoint.

## Evidence
- `python/openbrain-memory/src/openbrain_memory/client.py:252` sets `self._session_id = None` without sending a request.
- `python/openbrain-memory/tests/test_client.py:208` asserts only local clearing and context-manager clearing.
- The server implements session deletion at `src/transport.ts:228` through `handleDelete`, so the client has a supported cleanup path available.

## Risk
Normal context-manager use still leaves server-side MCP sessions active until TTL. Under churn from agents or short-lived jobs, this can accumulate sessions and hit `MAX_SESSIONS`, even when user code properly calls `close()` or exits a `with OpenBrainClient(...)` block.

## Acceptance Criteria
- Extend the `Transport` protocol with `delete()` or an equivalent session close method.
- Implement `UrllibTransport.delete()` for `DELETE /mcp`.
- Make `OpenBrainClient.close()` best-effort send `DELETE /mcp` when a session exists, then clear local state.
- Ensure `close()` remains idempotent and does not mask the original exception from `__exit__`.
- Add fake transport tests asserting the `Mcp-Session-Id`, auth, namespace, agent id, and protocol headers are sent on close.
- Add a server/client integration-style test if feasible, or document why the fake transport coverage is sufficient.

PR #232, opened 2026-07-05
fix(redaction): superset parity — unlabeled high-entropy + bare 3-seg token detectors (#92)

rodaddy:
## Why
Makes `openbrain-memory` redaction a true **superset** of the rtech-hermes vendored fork so that fork can retire (rtech-hermes #92: "use OB, not a fork; if there are issues, fix them in OB"). The fork's B2 parity report found the package UNDER-redacts 4 classes it catches.

## What
Adds two heuristic catch-all detectors ported from the fork:
1. **Bare 3-segment non-JWT token** `x.y.z` (the non-`eyJ` sibling of JWT_LIKE_RE — opaque `token.a.sig` shapes).
2. **Unlabeled 40+ char high-entropy blob containing a symbol** (`- _ + / =`).

**Anti-SHA guard preserved:** the `(?=.*[+/=_-])` symbol-requirement lookahead is load-bearing — a pure-hex git SHA / content hash has no symbol, so it is NOT redacted. This is exactly why the package originally narrowed to `LABELED_LONG_SECRET_RE`; the symbol requirement restores the fork's catch-all coverage without reintroducing SHA over-redaction.

## ⚠️ Design decision the reviewer should scrutinize (not a defect — a deliberate layering)
`SECRET_PATTERNS` has **two** consumers in this package: `redact_text` (cosmetic redaction) AND `agent._reject_secret_payload` (a **fail-closed receipt reject gate**). The fork only ever used these patterns for redaction.

A verbatim add to `SECRET_PATTERNS` broke the reject gate: the high-entropy heuristic matches benign 40+ char slash/underscore **file paths** (a test receipt command path is exactly 40 chars), causing a valid receipt to be **falsely rejected** (`test_record_receipt_routes_to_receipt_session_event` went RED).

**Resolution:** the two heuristics live in a new `HEURISTIC_REDACTION_PATTERNS` list, and `REDACTION_PATTERNS = SECRET_PATTERNS + HEURISTIC_REDACTION_PATTERNS`. `redact_text` (and `redact_value` → `_redact_value` → `redact_text`, and the spool payload path) apply BOTH lists. The fail-closed reject gate keeps using ONLY the high-confidence `SECRET_PATTERNS`.

**The posture this creates (call it out explicitly):** an unlabeled high-entropy secret in a payload is now REDACTED in recall text but still ACCEPTED by the receipt reject gate (written to OB, redacted). This preserves fork parity (the fork redacts, has no reject gate) without importing false rejections into the availability gate. If reviewers think the reject gate SHOULD also fail-closed on these classes, that's a follow-up decision — flagged here deliberately.

## Redaction table (proven at runtime)
| Class | before | after |
|---|:---:|:---:|
| bare 3-seg non-`eyJ` token | clear (LEAK) | **REDACTED** |
| 40+ high-entropy w/ dash | clear (LEAK) | **REDACTED** |
| 40+ high-entropy w/ underscore | clear (LEAK) | **REDACTED** |
| 40+ blob w/ `+/=` base64 | clear (LEAK) | **REDACTED** |
| 40-hex git SHA (negative) | clear | **clear** (guard holds) |
| 64-hex content hash (negative) | clear | **clear** (guard holds) |
| benign 40-char file-path receipt (reject gate) | accepted | **accepted** (gate untouched) |

## Tests + gates
- 6 new tests in `test_safety.py` (4 positive classes + 2 SHA negatives). Runtime-assembled samples, no literal secrets.
- `uv run pytest` → 178 passed, 5 skipped (pre-existing live-canary skips). `ruff check` clean, `mypy` clean, new test format-clean, ggshield clean.

## Critical pass (pre-PR, done)
- `redact_value` consistency verified — delegates to `redact_text` via `_redact_value:169`, so dict/list/spool paths get the heuristics too. No inconsistency.
- Pre-existing dead-param nit at `test_safety.py:572` (`dispatch(record)` unused) is OUTSIDE this diff — out of scope, not touched.

## After this lands (rtech-hermes #92)
Re-pin the package → re-run the parity harness to `local-only = 0` → swap rtech-hermes to package-only redaction, delete the vendored fork. Still gated there on open-brain#177 + live deploy.

### DIFF
--- a/python/openbrain-memory/src/openbrain_memory/policy.py
+++ b/python/openbrain-memory/src/openbrain_memory/policy.py
@@ -55,6 +55,31 @@ LABELED_LONG_SECRET_RE = (
     r"(?i)(client[_-]?secret|access[_-]?token|refresh[_-]?token|private[_-]?key)"
     r"\s*[:=]\s*[A-Za-z0-9._/+=-]{20,}"
 )
+# Superset-parity detectors ported from the rtech-hermes fork (#92) so this
+# package is a true SUPERSET of that fork's redaction and the fork can retire.
+# These catch UNLABELED high-entropy material that the labeled detectors above
+# miss (opaque bearer-less tokens, raw base64 secret bodies, session ids pasted
+# without a `key=` prefix).
+#
+# Bare 3-segment dotted token `x.y.z` -- the non-`eyJ` sibling of JWT_LIKE_RE.
+# JWTs are already caught by JWT_LIKE_RE; this covers opaque `token.a.sig`
+# shapes with no `eyJ` header.
+BARE_THREE_SEGMENT_TOKEN_RE = (
+    r"\b[A-Za-z0-9_-]{20,}\.[A-Za-z0-9_-]{6,}\.[A-Za-z0-9_-]{20,}\b"
+)
+# Unlabeled 40+ char high-entropy blob that CONTAINS at least one of `- _ + / =`.
+# ANTI-SHA GUARD: the `(?=.*[+/=_-])` lookahead is load-bearing -- it requires a
+# symbol so a benign 40-hex git SHA / content hash (which is pure `[0-9a-f]`,
+# no symbol) is NOT redacted. This is exactly why the package originally
+# narrowed to LABELED_LONG_SECRET_RE; the symbol requirement restores the fork's
+# catch-all coverage without reintroducing SHA over-redaction. The negative
+# lookbehind/lookahead pin the run to a full token boundary.
+UNLABELED_HIGH_ENTROPY_BLOB_RE = (
+    r"(?<![A-Za-z0-9+/=_-])"
+    r"(?=[A-Za-z0-9+/=_-]{40,}(?![A-Za-z0-9+/=_-]))"
+    r"(?=.*[+/=_-])"
+    r"[A-Za-z0-9+/=_-]+"
+)
 
 SECRET_PATTERNS = [
     re.compile(r"(?i)authorization\s*[:=]\s*bearer\s+[^\s,;]+"),
@@ -76,6 +101,25 @@ SECRET_PATTERNS = [
     re.compile(URL_USERINFO_CRED_RE),
     re.compile(LABELED_LONG_SECRET_RE),
 ]
+# Heuristic catch-all detectors ported from the rtech-hermes fork (#92) to make
+# redaction a true SUPERSET of that fork. They are DELIBERATELY kept OUT of
+# ``SECRET_PATTERNS`` above because ``SECRET_PATTERNS`` also drives the
+# fail-closed receipt reject gate (``agent._reject_secret_payload``). These two
+# patterns are aggressive by design (any 40+ high-entropy run with a symbol; any
+# bare dotted 3-segment token) and WILL match benign material such as a 40+ char
+# slash/dash/underscore file path (e.g. ``python/openbrain-memory/tests/...``).
+# That aggressiveness is correct for cosmetic redaction of displayed recall
+# text (redact-if-in-doubt) but wrong for a fail-closed availability gate, where
+# it would reject legitimate receipts. So ``redact_text`` applies BOTH lists;
+# the reject gate keeps using only the high-confidence ``SECRET_PATTERNS``. The
+# fork itself only ever used these for redaction, so this preserves fork parity
+# where it matters (redaction) without importing false rejections.
+HEURISTIC_REDACTION_PATTERNS = [
+    re.compile(BARE_THREE_SEGMENT_TOKEN_RE),
+    re.compile(UNLABELED_HIGH_ENTROPY_BLOB_RE),
+]
+# All patterns applied by ``redact_text`` (high-confidence + heuristic catch-all).
+REDACTION_PATTERNS = SECRET_PATTERNS + HEURISTIC_REDACTION_PATTERNS
 SENSITIVE_KEY_RE = re.compile(
     r"(?i)(token|secret|password|api[_-]?key|credential|authorization|session[_-]?id)"
 )
@@ -109,7 +153,7 @@ def idempotency_key() -> str:
 
 def redact_text(text: str) -> str:
     redacted = text
-    for pattern in SECRET_PATTERNS:
+    for pattern in REDACTION_PATTERNS:
         redacted = pattern.sub("[REDACTED]", redacted)
     return redacted
 
--- a/python/openbrain-memory/tests/test_safety.py
+++ b/python/openbrain-memory/tests/test_safety.py
@@ -212,6 +212,81 @@ def test_redaction_keeps_benign_40_character_hex_ids_visible():
     assert "[REDACTED]" not in redacted
 
 
+def test_redaction_scrubs_bare_three_segment_non_jwt_token():
+    # Non-`eyJ` opaque `token.a.sig` shape (#92 superset parity with the
+    # rtech-hermes fork). JWTs are already covered; this is the bare sibling.
+    token = token_sample(
+        "AbCdEfGhIjKlMnOpQrStUv",  # 20+ head
+        ".abcdef",  # 6+ middle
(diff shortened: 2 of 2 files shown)


rodaddy:
## 🔴 Swarm finding — regex-correctness lane (BLOCKER, empirically verified)

**ReDoS: O(n²) in `UNLABELED_HIGH_ENTROPY_BLOB_RE` (policy.py:77-82).** The `(?=.*[+/=_-])` lookahead (line 80) rescans to end-of-line at every qualifying start position. On a single long **no-newline** line that hits the first lookahead (40+ run of `[A-Za-z0-9+/=_-]`) but contains **no symbol**, it's quadratic in line length:

| line length | time |
|---|---|
| 51 KB | 0.11 s |
| 102 KB | 0.45 s |
| 204 KB | 1.85 s |
| 408 KB | **7.24 s** |

Trigger: one line of `("a"*50 + " ") * 8000` → 7.2s to return NO match. `.*` isn't DOTALL so newlines bound it — but a minified-JSON blob / long base64 line / wall-of-text log line is a realistic redaction input, and `redact_text` runs on arbitrary recalled text. **Must fix before merge.**

Fix direction: express "contains a symbol" WITHOUT an unbounded `.*` rescan — fold the symbol requirement into the consuming body, or bound the scan to the matched run. `AWS_SECRET_LIKE_RE` (policy.py:14-22) is the safe local precedent (all lookaheads inside a fixed window).

### Other lane findings (triaged, NOT blockers)
- **MAJOR / residual (document, don't fix):** pure-alphanumeric 40+ secrets with no symbol slip through the anti-SHA guard (e.g. a 44-char all-alnum key). Fundamental — can't distinguish a 40-hex SHA from a 40-hex secret by shape; the fork has the same gap. Note as a documented known limitation, not a fix (dropping the guard re-introduces SHA over-redaction).
- **MAJOR / acceptable (UX):** over-redacts benign 40+ paths, branch names, kebab slugs, base64. Acceptable for cosmetic redaction (already excluded from the reject gate); worth an explicit "expected" note.
- **MINOR:** `\b`-on-underscore over-consumes surrounding `foo_…_bar` context (cosmetic, still catches the token); segment `{20,}`/`{6,}` thresholds are exact off-by-one edges (conscious choice).
- **NIT (safe, no action):** `redact_text` idempotency confirmed — `[REDACTED]` chains don't re-match.

Security-posture + antagonist lanes + Codex cross-review still pending; the ReDoS blocker gates merge regardless.

rodaddy:
## 🔴 Swarm consolidated — security-posture + antagonist lanes (this reframes the PR)

Two more lanes landed. Combined with the ReDoS blocker above, three independent reviewers converged on findings that undercut the PR's premise. **Do not merge.**

### HIGH — the PR's redaction premise is false: writes AND reads are NOT redacted (pre-existing, but fatal to the #92 rationale)
Traced by the security lane:
- `AgentMemory._call_write` (`agent.py:702-724`) — the funnel for EVERY OB write (`log_thought`, receipts, `remember_fact/decision`, `checkpoint`, `lane_upsert`, …) — sends the **raw payload**: `call(**write_payload)` at `:717`. No `redact_text`/`redact_value` on this path.
- Recall/display path (`agent.py:184`→`743`→`794`) returns OB content **verbatim**. No redaction.
- The ONLY live `redact_text` consumers: `client.py:947` (scrubs an HTTP **error body** for the exception message) and `spool.py:29` (an **opt-in** `redacted_records()` accessor; the spool stores raw at `:83`, replays raw at `:237`).
- Concrete: `remember_fact("deploy key AKIA...ABCD00 rotated")` → raw AWS key written to OB in the clear TODAY. `remember_fact/decision/checkpoint/append_event` don't even call the reject gate — only `record_receipt` does (`agent.py:440`).

**So the two new heuristics run almost nowhere.** The PR comments (`policy.py:110-113`) describe a redaction architecture that isn't wired up. #92's premise ("swap to package redaction, fork retires safely") requires redaction to actually be ON the write/read paths — it isn't, in this package.

### HIGH — the "catch-all" leaks real secrets (antagonist, empirically confirmed via the branch's redact_text)
The blob pattern needs length ≥40 AND a symbol. Pure-`[A-Za-z0-9]` secrets slip at ANY length:
- `GOCSPX1a2b3c4d5e6f7g8h9i0jklmnopqrstuvwx` (40 chars, Google-OAuth-secret shape) — **clears the threshold, leaks** (no symbol). Killshot: the symbol requirement is a secret-EXCLUSION filter, not an anti-SHA tweak.
- 50-char pure-alnum key, 44-char mixed-case token, sub-40 session ids, two-segment `x.y` tokens, `sk-` keys embedded mid-word (`\b` never fires) — all leak.
- Inverse: `commit-<40hexsha>-tag` IS redacted — the "no SHA over-redaction" claim is false the moment a SHA sits between dashes.

**Fix direction (both lanes agree):** replace "must contain a symbol" with a composition discriminator that distinguishes hashes from secrets — reuse this file's OWN `AWS_SECRET_LIKE_RE` mixed-class approach `(?=.*[A-Z])(?=.*[a-z])(?=.*[0-9])` (catches pure-alnum high-entropy secrets; still passes all-lowercase hex SHAs). That also kills the ReDoS by bounding the lookaheads inside the window.

### MEDIUM — tests are engineered to pass
Every "scrubs" test injects a symbol; NO test for the pure-alnum-40 leak; the two SHA-negative tests use only lowercase-hex-no-punctuation, so they never catch the `commit-<sha>-tag` over-redaction. Missing: `test_redaction_MISSES_pure_alnum_high_entropy_secret` (documents the gap) and `test_keeps_sha_visible_when_adjacent_to_dash` (currently FAILS). Idempotency verified safe (a missing test, not a bug).

### MEDIUM — client-vs-server parity unaddressed
#92's premise is "redaction 1:1 with the server." This PR argues client-vs-FORK parity, says nothing about client-vs-SERVER. Adding two aggressive client-only heuristics means the client now redacts strings the server keeps verbatim (paths, dotted tokens) → recall shows `[REDACTED]` where the server returns the original. Needs the server's disposition of these patterns before the "#92 superset" claim is load-bearing.

### LOW — fork-parity holds narrowly; framing overreaches
The two regexes are byte-identical to the fork (`support.py:159,161`). But that means this is a faithful copy of the fork's *broken* heuristic. Parity ≠ correctness.

## Verdict
The redaction-list split (keeping heuristics off the fail-closed reject gate) is the ONE sound decision here. Everything else needs rework, and the security lane surfaced that the real gap is architectural (writes/reads unredacted), not a regex tweak. Codex cross-review still to run, but the direction is clear.

rodaddy:
## Closing — this PR solves a STALE-PREMISE problem (verified against the current OB contract)

The gauntlet swarm sent me back to check whether #92's premise still matches how OB actually works now. It doesn't. Findings:

1. **The OB server does NOT redact** (no server-side redaction code exists) — redaction was never server-side.
2. **The current OB memory contract is "don't store secrets," not "redact on write."** `docs/memory-contract.md:261`: *"Secret or raw transcript content: do not store it. Save a redacted decision, artifact path, or validation receipt instead."* The caller is responsible for not writing secrets.
3. **`redact_text`/`redacted_payload()` are scoped to DISPLAY-TIME defense-in-depth** — README:329: *"before showing spool contents in logs, UIs, or debug output."* NOT a write-path secret filter. That's why writes/reads don't route through it (the security lane's "gap" is actually the intended architecture).
4. **OB already fixed a redaction ReDoS** (`f85a8b3` #175, "kill real ReDoS in URL-credential secret pattern"). This PR REINTRODUCED one (the `.*` blowup) by porting the fork's patterns verbatim — fighting against the package's own evolution.

**So "make the package a write-time redaction superset of the fork" is the wrong goal** — it's #92's 2026-06-19 framing, which predates OB's memory-substrate + contract redesign (#217/#225/#226/#160). The fork's aggressive write-time redaction is itself the legacy pattern the new contract moved away from.

Closing #232. The 2 regexes were a faithful copy of the fork's broken heuristic (ReDoS + pure-alnum leak + over-redaction). Not merging a stale-premise change. #92 needs re-basing on the current contract before any redaction work — see the #92 comment.

PR #233, opened 2026-07-05
fix(logger): 1MB rolling file sink as an app config standard (#193)

### LINKED ISSUE #193: Logs are unbounded — add 1MB max rolling log as an app config standard
## Summary

OB's log files grow without bound. On the live server (10.71.1.21, `skippy-oc`), `/opt/homebrew/var/log/open-brain.log` is **16MB** and still growing (sibling `open-brain-legacy-promoter.log` ~1.9MB). There's no rotation/size cap. Fix this **in the app's logger config** so it's a built-in standard (1MB max, rolling), not a per-host newsyslog/logrotate hack.

## Requirement

- **1MB max per log file, rolling** (keep a small number of rotated files, e.g. 3-5, then discard oldest).
- Implemented as an **open-brain app config standard** so every deployment gets it by default — not host-specific rotation tooling.
- Applies to the main app log and the sibling OB logs (promoter, backup) where practical.

## Evidence (2026-06-21)

```
-rw-r--r-- open-brain.log                  16331053  (16MB, active)
-rw-r--r-- open-brain-legacy-promoter.log   1953238  (1.9MB)
-rw-r--r-- open-brain-backup.log               4766
-rw-r--r-- open-brain-launchd.log               184
```

Run via launchd `com.rico.open-brain` on the Mac (workers on 3100/3101/3102), StandardOutPath → `/opt/homebrew/var/log/open-brain-launchd.log`, but the bun app itself writes the large `open-brain.log`.

## Related context

- While investigating an OB `503 "Too many active sessions"` (MAX_SESSIONS=100 in `src/transport.ts`), the **transport session lifecycle events (expireSession/sweeper/503) do not appear in `open-brain.log`** — `grep expireSession|sweeper` = 0 hits. Worth confirming as part of logging work whether the transport layer logs at all and to where. (Separate concern, but same logging-quality theme.)

## Acceptance

- [ ] OB logger caps files at 1MB and rolls automatically (configurable, sane default).
- [ ] No single OB log file can grow unbounded.
- [ ] Verified on the live 1.21 deployment.

rodaddy:
Refs #193

This PR builds the capability but does NOT by itself bound the actual #193
file: `console.*` still writes stdout unconditionally and launchd's
`StandardOutPath` owns `/opt/homebrew/var/log/open-brain.log`, so #193 closes
only after the deploy-gated core01 plist/env step below is applied and
verified live.

## What & why

OB log files grew unbounded. The app only writes structured JSON to
stdout/stderr via the shared logger (`src/logger.ts` -> `console.*`); on core01
the launchd service `com.rico.open-brain` redirects that stream via
`StandardOutPath` to a single `/opt/homebrew/var/log/open-brain.log`, which had
no size cap (16MB and growing per the issue). launchd owns the file — the app
just writes stdout. Host newsyslog/logrotate was explicitly ruled out; the cap
must be a built-in app config standard every deployment inherits.

## Change (scoped to logging)

- `src/rotating-file.ts` — reusable size-capped rolling file sink. Rotates the
  active file to `.1` (shifting `.1->.2...`, pruning past `maxFiles`) when the
  next line would exceed `maxBytes`, and rotates again immediately after any
  write that leaves the active file over the cap (so a single oversized line —
  even as the very first write — never lingers as an over-cap active file).
  `rotate()` reports whether the active file was verifiably cleared; the size
  counter is only zeroed on verified rotation, otherwise it re-syncs from disk
  so failed rotations (read-only dir, EXDEV, ENOSPC) are retried on the next
  write instead of silently unbounding the file. Files are created `0o600`,
  directories `0o700`. Size is checked at write time (no cron), and every
  write is best-effort — logging can never throw or crash the server.
- `src/logger.ts` — when `LOG_FILE` is set, mirror every emitted line into the
  rotating sink (console output unchanged). When `OPEN_BRAIN_WORKER_NAME` is
  set (the two-worker launcher sets a distinct name per child), the effective
  path is derived per worker automatically (`open-brain.log` ->
  `open-brain.<worker>.log`, name sanitized so it cannot change the directory),
  so workers inheriting one configured `LOG_FILE` never share an active file or
  rotation chain. New env, with issue defaults:
  - `LOG_FILE` — active log path (enables the sink; unset = today's behavior)
  - `LOG_MAX_BYTES` — rotate threshold, default `1_000_000` (1MB); values < 1
    fall back to the default
  - `LOG_MAX_FILES` — rotated files retained, default `3`; `0` keeps only the
    active file

Because all OB code and sibling processes (legacy-promoter, backup) log through
this one shared logger, they all inherit the cap.

## Tests (functional, black-box)

`src/rotating-file.test.ts` and `src/logger-file.test.ts` — all writes confined
to `mkdtemp` temp dirs, asserting observable outcomes only: write >1MB through
the sink and through the real `logger.info` (child processes for env wiring);
active and rotated files stay `<= cap`; retained count `<= maxFiles`
(`maxFiles=0` keeps only the active file); recent content survives rotation
boundaries in order; oversized lines — including as the very first write — are
rotated immediately and bounded to their own file; a failed rotation
(read-only parent dir) never zeroes the counter and rotation is retried on the
next write after recovery (fails on the old behavior, verified); created files
are `0o600` and created dirs `0o700`; `LOG_MAX_BYTES=0` falls back to the 1MB
default; two workers sharing one configured `LOG_FILE` get divergent
per-worker files; `deriveWorkerLogPath` unit coverage incl. path-traversal
safety; no `LOG_FILE` ⇒ no files created. Full suite: `bunx tsc --noEmit`
clean, `bun test` 976 pass / 0 fail.

## Review fix rounds

Cross-model review (Codex gpt-5.5):
- P2-1: shared-LOG_FILE multi-worker race — fixed in code via automatic
  per-worker path derivation from `OPEN_BRAIN_WORKER_NAME` (not docs), with a
  two-worker regression test proving the effective paths diverge.
- P2-2: oversized first write bypassing the cap — fixed via post-write
  rotation, with a first-write-over-cap regression test.
- P3: test cleanup `rmSync` guarded with `if (dir)`; unused `describe` import
  removed.

Review swarm:
- P2: rotation-failure counter desync — `rotate()` now returns verified
  success and callers never zero the counter on failure (re-sync from disk,
  retry next write). Regression test reproduces the reviewer scenario
  (chmod 0500 parent dir) and fails on the old behavior.
- P3: log files created `0o600`, dirs `0o700` (mode-bits asserted in test).
- P3: `LOG_MAX_BYTES` values < 1 rejected at parse (fall back to default) so
  the zero-cap case is intentional in one place; `LOG_MAX_FILES=0` kept.
- P3: issue linkage changed from Closes to Refs (see top note) — the launchd
  `StandardOutPath` file is only bounded after the deploy-gated plist/env step
  is verified live.

## Deployment notes (documented — NOT executed anywhere host-level)

The macOS CI runner is core01 (production); nothing host-level was run here.
To activate on core01 and actually close #193, an operator must, at deploy
time:

1. Set `LOG_FILE=/opt/homebrew/var/log/open-brain.log` (and optionally
   `LOG_MAX_BYTES` / `LOG_MAX_FILES`) in `/Users/rico/.config/open-brain/env`.
   Under the two-worker launcher each worker automatically writes
   `open-brain.<worker-name>.log`; no per-worker env is needed.
2. Edit the launchd plist `com.rico.open-brain`: stop redirecting
   `StandardOutPath` to the same base path the app now owns via `LOG_FILE`
   (point it elsewhere or keep stdout minimal) to avoid double-writing.
   Keep `StandardErrorPath` for crash/bootstrap output.
3. `launchctl kickstart -k system/com.rico.open-brain` to pick up env/plist.
4. Verify live per the issue acceptance: watch the derived log files stay
   <= 1MB with <= 3 rotated siblings, then close #193.

No DB migration. No MCP tool/schema/transport/Python-client change, so
`docs/downstream-rollout.md` does not apply (internal logging config only).

## Critical Self-Review

- Highest-risk behavior: Silent cap bypass was the top risk in two forms — the multi-worker rotation race (now prevented in code: per-worker path derivation from `OPEN_BRAIN_WORKER_NAME`, proven by a two-child regression test) and the rotation-failure counter desync (now prevented: counter only zeroed on filesystem-verified rotation, proven by a read-only-dir regression test that fails on the old behavior).
- Assumptions that could be wrong: That launchd owns the 16MB file via stdout redirection — verified: logger only calls `console.*`, no plist in repo, deploy script relies on server-side `StandardOutPath`. That the launcher always sets distinct `OPEN_BRAIN_WORKER_NAME` values — verified at `scripts/run-two-worker.ts`. That `existsSync` after a rename attempt is a sufficient success check — it verifies the active path was cleared, which is exactly the invariant the counter depends on.
- Missing/weak tests: No test for two processes writing the exact same effective path concurrently (unreachable through the launcher, possible via operator misconfiguration); no ENOSPC/EXDEV simulation (the read-only-dir test covers the same failure edge — rename fails while append succeeds); byte-for-byte survival across pruned rotations is not asserted (old data is discarded by design).
- Security/permission risk: Log files are now created `0o600` and directories `0o700` (asserted in test) instead of umask-default world-readable; the worker-name path suffix is sanitized with a traversal test proving it cannot change the directory; no secrets are added (payload is the same JSON already emitted).
- Migration/deploy risk: No schema change; if `LOG_FILE` is unset behavior is identical to today (pure opt-in), so the code ships safely ahead of the plist change. During a persistent rotation-failure window (e.g. dir perms broken) the active file can exceed the cap — writes must land somewhere and log loss was judged worse — but it now rotates on the first write after recovery. Pre-existing log files keep their old modes; only newly created files get `0o600`.
- Downstream client/runtime risk: None — no client-facing, MCP, transport, or protocol surface is touched.
- Rollback/cleanup concern: Revert the commit or unset `LOG_FILE`; rotated `.1..N` files remain on disk and can be removed manually, and are bounded by design.
- Fixes made before PR: Corrected an under-sized test loop (~285KB) so the suite genuinely writes >1MB; cross-model round: per-worker path derivation, post-write rotation, guarded test cleanup; swarm round: verified-rotation counter fix (old behavior empirically fails the new test), `0o600`/`0o700` modes, bounded env parse, and Closes->Refs linkage honesty.
- Known residual risk: The actual #193 file stays unbounded until the deploy-gated core01 plist/env step is applied and verified live (why this is Refs, not Closes); an operator-forced identical worker name across processes can still recreate the shared-path race; during a sustained rotation-failure window the active file exceeds the cap until the filesystem recovers.
- SME review-memory update: [x] not applicable because: no new cross-cutting review pattern surfaced — this is a scoped, additive logging config change with functional tests, not a security/isolation predicate or contract fix that the SME lanes track.

## Review Gate

- Live Open Brain checks: [x] not applicable because: live 1.21 verification is a host-level deploy step held pending explicit approval per the no-live-server constraint; no live OB smoke can run from this PR.

### DIFF
--- a/src/logger-file.test.ts
+++ b/src/logger-file.test.ts
@@ -1,0 +1,78 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { mkdtempSync, readdirSync, rmSync, statSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+// Black-box: the logger reads LOG_FILE / LOG_MAX_BYTES / LOG_MAX_FILES once at
+// module load, so exercise the real env wiring in a child process. Assert the
+// observable outcome: many logger.info() calls stay bounded on disk. No writes
+// escape the temp dir.
+
+let dir: string;
+
+beforeEach(() => {
+  dir = mkdtempSync(join(tmpdir(), "ob-logger-file-"));
+});
+
+afterEach(() => {
+  rmSync(dir, { recursive: true, force: true });
+});
+
+test("logger honors LOG_FILE with a size cap and rotation", async () => {
+  const logPath = join(dir, "open-brain.log");
+  const maxBytes = 32 * 1024;
+  const maxFiles = 3;
+
+  const driver = `
+    import { logger } from ${JSON.stringify(join(import.meta.dir, "logger.ts"))};
+    for (let i = 0; i < 4000; i += 1) {
+      logger.info("session-event", { i, blob: "y".repeat(60) });
+    }
+  `;
+
+  const proc = Bun.spawn(["bun", "run", "-"], {
+    stdin: new TextEncoder().encode(driver),
+    stdout: "ignore",
+    stderr: "pipe",
+    env: {
+      ...process.env,
+      LOG_FILE: logPath,
+      LOG_MAX_BYTES: String(maxBytes),
+      LOG_MAX_FILES: String(maxFiles),
+      LOG_LEVEL: "info",
+    },
+  });
+  const code = await proc.exited;
+  const stderr = await new Response(proc.stderr).text();
+  expect(code).toBe(0);
+  // No crash noise from the logger path.
+  expect(stderr).not.toContain("Error");
+
+  const files = readdirSync(dir).filter(
+    (n) => n === "open-brain.log" || n.startsWith("open-brain.log."),
+  );
+  // Active + at most maxFiles rotated.
+  expect(files.length).toBeGreaterThan(1); // proves rotation actually happened
+  expect(files.length).toBeLessThanOrEqual(maxFiles + 1);
+
+  for (const name of files) {
+    const size = statSync(join(dir, name)).size;
+    expect(size).toBeLessThanOrEqual(maxBytes + 1024);
+  }
+});
+
+test("logger without LOG_FILE writes no log files", async () => {
+  const driver = `
+    import { logger } from ${JSON.stringify(join(import.meta.dir, "logger.ts"))};
+    logger.info("no-file-sink", { ok: true });
+  `;
+  const proc = Bun.spawn(["bun", "run", "-"], {
+    stdin: new TextEncoder().encode(driver),
+    stdout: "ignore",
+    stderr: "ignore",
+    env: { ...process.env, LOG_FILE: "", LOG_LEVEL: "info" },
+  });
+  await proc.exited;
+  // Nothing was created in the temp dir.
+  expect(readdirSync(dir)).toEqual([]);
+});
--- a/src/logger-file.test.ts
+++ b/src/logger-file.test.ts
@@ -107,6 +107,29 @@ test("two workers sharing one configured LOG_FILE get divergent per-worker files
   }
 });
 
+test("LOG_MAX_BYTES=0 falls back to the 1MB default instead of a zero cap", async () => {
+  // A zero-byte cap is meaningless; it must fall back to the default rather
+  // than silently becoming something else deeper in the sink. These writes
+  // stay under the 1MB default, so exactly one file and no rotation.
+  const logPath = join(dir, "open-brain.log");
+  const proc = spawnLogger(
+    {
+      LOG_FILE: logPath,
+      LOG_MAX_BYTES: "0",
+      LOG_MAX_FILES: "3",
+      OPEN_BRAIN_WORKER_NAME: "",
+    },
+    loggerDriver(4000, 60),
+  );
+  expect(await proc.exited).toBe(0);
+
+  const files = readdirSync(dir);
+  expect(files).toEqual(["open-brain.log"]);
+  const size = statSync(logPath).size;
+  expect(size).toBeGreaterThan(100_000); // everything landed in one file
+  expect(size).toBeLessThanOrEqual(1_000_000); // still under the default cap
+});
+
 test("logger without LOG_FILE writes no log files", async () => {
   const proc = spawnLogger({ LOG_FILE: "" }, loggerDriver(1, 10));
   await proc.exited;
--- a/src/logger-file.test.ts
+++ b/src/logger-file.test.ts
@@ -1,12 +1,13 @@
-import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { afterEach, beforeEach, expect, test } from "bun:test";
 import { mkdtempSync, readdirSync, rmSync, statSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
+import { deriveWorkerLogPath } from "./logger.ts";
 
-// Black-box: the logger reads LOG_FILE / LOG_MAX_BYTES / LOG_MAX_FILES once at
-// module load, so exercise the real env wiring in a child process. Assert the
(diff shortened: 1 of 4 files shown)


rodaddy:
Fresh #233 pre-merge-gauntlet validation receipt.

Head: `63ff31aecfa11953dcae6966243f7e5075158834`

Fresh local validation in clean temp worktree `/Volumes/ThunderBolt/_tmp/open-brain/pr233-gauntlet`:
- `bun install --frozen-lockfile` completed.
- `bun test src/rotating-file.test.ts src/logger-file.test.ts` passed: 14 pass, 0 fail.
- `bunx tsc --noEmit` passed.
- `bun test` passed: 976 pass, 38 skip, 0 fail.

Live GitHub checks on current head:
- `check` passed.
- `python-package` passed.
- `validate` passed.
- `claude-review` passed.
- GitGuardian passed.
- `deploy` skipped.

review-swarm pinned/fresh-context lane evidence and cross-model findings are already documented in the PR body: multi-worker shared LOG_FILE race, oversized first write, rotation-failure counter desync, mode bits, bounded env parse, and honest `Refs #193` linkage were all addressed.

fix-verification: fresh local focused tests, full test suite, typecheck, and CI are green.

Residual risk remains deploy-gated by design: this PR adds the app-level rolling sink but does not close #193 until the core01 env/plist change and live log-size verification are performed in the later deploy/release phase. No production deploy performed here.

zero unresolved findings; zero known merge blockers.

Issue #288, opened 2026-07-13
Transcript-ref storage + citation recall contract (cite source conversation, date, speaker — or say it isn't stored)


rodaddy:
Part of rodaddy/development#2 (Claude/Codex dev-session capture → OB).

## What

Server-side contract so memories can carry a durable reference to the source conversation, and recall can cite it.

- **Transcript(-ref) storage:** schema + ingest contract for storing session transcripts (or durable refs to them) alongside memories/summaries. Refs must be host-neutral (`collab/...` style, no `/Volumes/...` or `/mnt/...` literals).
- **Citation recall:** query/tool contract returning source conversation, speaker/agent, date, and the surrounding exchange (context expansion into the neighboring transcript) — or an explicit "source not stored" instead of inventing one.

## Why

OB currently returns the fact but can't answer "who decided this, when, in which conversation, and what was actually said." The dev-runtime capture work (development repo) will ship summaries + transcripts; this contract is what makes them citable. The citation ladder falls out of capture — it is not a separate feature, so the contract lands first.

## Acceptance sketch

- Ingest accepts summary + transcript-ref (and optionally inline transcript) in one write.
- A recall query for a stored decision returns: fact, source conversation id/ref, date, speaker, expandable surrounding context.
- A recall query for a fact with no stored source says so explicitly.

rodaddy:
Post-merge hosted proof for PR #289 / issue #288:

- merge commit: `98e5e3c`
- migration `023_session_event_transcript_citations.sql`: applied on core01
- runtime restarted in the actual live domain `gui/501/com.rico.open-brain`
- hosted health: passed
- hosted contract: `2026-07-13.memory-tools.v21`
- hosted capabilities: `citation_recall` present
- refreshed mcp2cli `append_session_event` schema now exposes `transcript_ref`, `transcript`, and `occurred_at`
- live Claude and Codex canary events both return `citation.status: stored`

Operational correction captured separately: the current `workflow_dispatch` deploy run skips its prerequisite jobs, and the deploy script's default launchd domain is stale. The rollout used the repo-owned deploy payload from a clean exact-main worktree, then restarted the already-registered GUI-domain service and proved health/contract live.

PR #290, opened 2026-07-13
fix(memory): allow case-distinct session lane hashes

rodaddy:
## Summary

- convert the session-lane content-hash index from unique to non-unique
- preserve `(namespace, session_key)` as the canonical case-sensitive lane identity
- add a real-Postgres regression for the exact `dev:rTech-audit` / `dev:rtech-audit` collision

## Root cause

`contentHash()` lowercases text before hashing. Two valid case-distinct session keys with the same topic therefore receive the same lane hash, while the database treats their `(namespace, session_key)` identities as distinct. The redundant unique hash index rejected the second lane and blocked Development's session-history backfill.

## Validation

- focused live-Postgres regression is included and will run in CI where `OPENBRAIN_TEST_DATABASE_URL` is configured
- local focused invocation loaded successfully but skipped the Postgres-gated test because this workstation does not expose that test URL

## Downstream rollout

Applicable only to hosted Open Brain because this is a database migration affecting runtime behavior. No MCP tool name, input schema, output shape, auth rule, client contract, or generated skill changes. rtech-mcps, mcp2cli schema refresh, rtech-hermes code changes, and Hermes agent rollout are not applicable. Hosted migration plus an exact append/backfill canary are required before closure.

## Critical self-review

- Highest-risk behavior: changing a uniqueness constraint on session-lane search metadata
- Assumptions that could be wrong: another caller may rely on hash uniqueness; source search found no such caller, while `(namespace, session_key)` is the documented identity constraint
- Missing/weak tests: local Postgres integration is unavailable; CI owns the real database execution
- Security/permission risk: none; namespace and session-key identity constraints are unchanged
- Migration/deploy risk: index replacement briefly takes a table lock during migration; table is small and rollback can recreate the prior unique index only after resolving duplicates
- Downstream client/runtime risk: no public contract shape changes
- Rollback/cleanup concern: new legitimate duplicate hashes prevent blindly restoring the old unique index
- Fixes made before PR: regression uses the exact case-normalized collision shape found in production
- Known residual risk: hosted deploy and backfill canary remain required
- SME review-memory update: [x] not applicable because: the finding was specific to this new regression and does not add a reusable reviewer pattern

## Review Gate

- Live Open Brain checks: [x] not applicable because: migration deployment and hosted proof run only after merge and remain required before issue closure

Supports rodaddy/development#4 and follows up rodaddy/open-brain#288.

### DIFF
--- a/src/db/migrations/024_session_lane_content_hash_nonunique.sql
+++ b/src/db/migrations/024_session_lane_content_hash_nonunique.sql
@@ -1,0 +1,10 @@
+-- Migration 024: session lane hashes are search metadata, not identity
+-- Lane identity is the case-sensitive (namespace, session_key) constraint.
+-- contentHash normalizes case, so a unique hash index rejects valid lanes such
+-- as dev:rTech-audit and dev:rtech-audit when they share a topic.
+
+DROP INDEX IF EXISTS idx_session_lanes_content_hash;
+
+CREATE INDEX IF NOT EXISTS idx_session_lanes_content_hash
+  ON ob_session_lanes (content_hash, namespace)
+  WHERE content_hash IS NOT NULL;
--- a/src/tools/__tests__/append-session-event.test.ts
+++ b/src/tools/__tests__/append-session-event.test.ts
@@ -2502,7 +2502,19 @@ dbDescribe("append_session_event create_if_missing (live Postgres)", () => {
         [ns, ["dev:rTech-audit", "dev:rtech-audit"]],
       );
       expect(rows).toHaveLength(2);
-      expect(rows[0].content_hash).toBe(rows[1].content_hash);
+      const expectedHash = contentHash(
+        "dev:rTech-audit|Shared development session memory",
+      );
+      expect(expectedHash).toBe(
+        contentHash("dev:rtech-audit|Shared development session memory"),
+      );
+      expect(rows.map((row) => row.session_key).sort()).toEqual(
+        ["dev:rTech-audit", "dev:rtech-audit"].sort(),
+      );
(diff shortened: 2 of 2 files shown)


rodaddy:
## Initial Full review — findings and disposition

Pinned head: `bfaf3a3f9f42d2795138861817ef88f3c21c9871`

Five fit-to-diff lanes reviewed the same candidate: migration correctness, antagonist/race behavior, regression-test efficacy, deployment/runtime gotchas, and backend data-model ownership.

- [M-1] `src/tools/__tests__/append-session-event.test.ts:2505` — the regression only compared both hashes for equality, so two `NULL` values could false-green without exercising the pre-024 unique-index collision.
  - Fixed in `2664c92`: assert the exact non-null normalized hash and both case-distinct session identities.

All other lanes returned clean. No P0/P1/P2 findings remain from the initial review. Typecheck and `git diff --check` passed after the fix. The fix delta is test-assertion-only, so controller readback plus CI's real Postgres integration job is the focused verification owner; the initial swarm will not be restarted.

rodaddy:
## Opposite-runtime terminal audit

Runtime: Claude Opus, high effort  
Exact head: `2664c921cc0e4b4ca4ce6dc64185ee973f1a4d5e`

`TERMINAL_AUDIT: CLEAN; findings=0`

Verified migration ordering/idempotency/atomicity, case-sensitive lane identity, absence of hidden uniqueness callers, exact non-null regression behavior, and rollback reality. One informational operational note remains: the non-concurrent index replacement briefly blocks lane writes during the migration, matching the repo's existing transactional migration convention. Hosted migration and the exact two-session backfill canary remain required after merge.

Issue #370, opened 2026-07-24
Clone core01 Open Brain into a safe local runtime


rodaddy:
Parent epic: #369
Program: #320

## Outcome

A fresh core01 Open Brain backup can be transferred encrypted, verified, restored into an isolated local PostgreSQL 18 + pgvector database, and served only on loopback with local embeddings and local-only credentials.

## Acceptance

- backup manifest facts and pg_dump share one PostgreSQL snapshot
- local-clone mode validates configuration before pool creation or migrations
- Open Brain binds to literal loopback and cannot inherit production DB/provider/NATS/watchdog/QMD/WAL paths
- restore uses a fresh explicit database and existing verification substrate; no wipe-in-place refresh
- local embedding provider matches the stored 768-dimension embedding contract
- functional real-PostgreSQL concurrency, restore, health, read/write, isolation, restart, and socket tests pass
- core01 receives no validation writes and remains healthy
- content-free receipts and errors are preserved

rodaddy:
Delivered and RUNNING. The dogfood clone `open_brain_local_20260724` (PostgreSQL 18 + pgvector, restored from a verified core01 backup) serves on port 3100 via launchd `com.rico.open-brain-local-clone`, local MLX embeddings (768-dim contract intact), local-only credentials in `local-clone.env`. It has been the live durable memory for every Claude session on this machine through the whole 0.9 dogfood arc — capture writes to `ob_raw_turns` daily, canon packs read from it at every SessionStart, core01 untouched. One recorded delta from the original acceptance: since 2026-08-01 it binds 0.0.0.0 rather than literal loopback, an operator decision making the Air (10.71.1.26) a sanctioned LAN client of this dev brain — the two-host scope still holds.

Issue #369, opened 2026-07-24
[Epic] Run Open Brain locally for Claude and Claudex dogfooding


rodaddy:
## Goal

Run the complete Open Brain memory stack locally from a verified clone of the core01 database, and make Claude-compatible plus Claudex sessions use the local stack for automatic context-pack, reflex, capture, checkpoint, and wrap behavior. Codex integration is deferred.

Parent program: #320

## Scope

- verified, repeatable core01 backup → encrypted transfer → local restore
- loopback-only local PostgreSQL/pgvector, embedding, and Open Brain runtime
- automatic Claude-compatible and Claudex full context-pack plus per-turn reflex recall
- content-free Langfuse lifecycle and recall-quality observability
- local dogfood and #347 tuning before the next core01 rollout

## Constraints

- no production writes from local validation
- no raw prompts, transcripts, retrieved bodies, secrets, paths, tool payloads, or exception strings in telemetry
- functional coverage, not source-coverage targets
- Codex is deferred
- Hermes is not applicable and must not be added, deployed, or canaried
- local verification completes before production deployment

## Completion

All child issues are locally and functionally verified, Claude/Claudex canaries use the local server, Langfuse stores allowlisted metadata only, #347 completes against the local stack, and Project #8 reflects live state.

## Close-run map (2026-08-05 audit)
Adapter/capture-correctness members now tracked here: #371 (total-recall hooks), #511 (Pi adapter), #512 (Codex adapter), #529 (scope-proof failure — REGRESSED 2026-08-05 to silent exit-0 zero-capture, priority raised), #562 (ENV_FILE derivation — PR #570 open, red CI), #442 (brain skill stale docs).

rodaddy:
Epic audit 2026-08-03: #370 closed delivered (clone RUNNING). #371 scoped down to its one unmet item — per-turn reflex pointers (never ported to the Python package; everything else live). Observability delta from the original scope: the content-free Langfuse constraint was explicitly overridden by Rico for the local deployment — the #523/#524 sink ships full-content session traces, and #530 extends that to server-side tool-call tracing. Epic stays open until #371's reflex port lands or is descoped by Rico.

PR #376, opened 2026-07-24
fix(370): prove literal database host address

rodaddy:
## Summary

- return PostgreSQL's server address with `host(inet_server_addr())` instead of `inet_server_addr()::text`
- preserve exact literal `DB_HOST` equality while avoiding the CIDR suffix (`/32` or `/128`) added by the `inet` text representation
- update the existing real local-clone PostgreSQL test to assert the literal host returned by the production dependency

Tracks #370
Parent #369

## Validation

- focused local-clone suite against the restored clone role/database: **7 passed, 0 failed, 24 assertions**
- `bunx tsc --noEmit`: passed
- actual local clone preflight: `openbrain.local_clone_verify_receipt.v1`, status `verified`
  - database identity/read-only/PostgreSQL 18/pgvector: verified
  - embedding model and 768 dimensions: verified
- `git diff --check`: passed

## Local-only rollout classification

- This fixes the local dogfood preflight discovered after the verified database restore.
- No core01 deployment or mutation is included.
- The literal loopback equality check is preserved; only PostgreSQL's CIDR formatting is removed at the query boundary.

## Contract Parity

- Contract parity: [x] not applicable because: no MCP, client, package, or wire contract changes; this corrects an internal operator preflight query representation

## Critical self-review

- Highest-risk behavior: weakening the database identity gate could allow a non-loopback or wrong-address target.
- Assumptions that could be wrong: PostgreSQL `host(inet_server_addr())` returns the address literal without network prefix for both IPv4 and IPv6.
- Missing/weak tests: the existing env-gated real PostgreSQL test now runs against the exact restored clone role/database; no mock-only SQL assertion was added.
- Security/permission risk: exact equality against configured `DB_HOST` remains unchanged and the probe remains read-only.
- Migration/deploy risk: none; no schema or data mutation. Startup preflight behavior changes from false rejection to verified.
- Downstream client/runtime risk: none outside the local-clone operator path.
- Rollback/cleanup concern: reverting restores the false-negative `127.0.0.1/32` mismatch; no persisted state changes.
- Fixes made before PR: removed an implementation-detail SQL-string test and kept the functional live boundary as the regression.
- Known residual risk: local Open Brain startup and health/socket proof still follow after merge.
- SME review-memory update: [x] not applicable because: the owning operator-clone live-boundary pattern already exists in `docs/sme/correctness.md`; this is the direct functional follow-up

## Review Gate

- Live Open Brain checks: [x] not applicable because: the repository-owned local-clone preflight itself was run against the restored database and managed embedding service; server startup follows merge

### DIFF
--- a/scripts/local-clone.test.ts
+++ b/scripts/local-clone.test.ts
@@ -311,8 +311,7 @@ describe.skipIf(!REAL_PG_URL)(
       expect(proof).toMatchObject({
         database,
         user,
-        // node-postgres renders inet as CIDR text (127.0.0.1/32, ::1/128).
-        serverAddress: host === "127.0.0.1" ? "127.0.0.1/32" : "::1/128",
+        serverAddress: host,
         serverPort: Number.parseInt(url.port || "5432", 10),
         postgresMajor: 18,
         transactionReadOnly: true,
--- a/scripts/local-clone.ts
+++ b/scripts/local-clone.ts
@@ -305,7 +305,7 @@ export const productionDependencies: LocalCloneLauncherDependencies = {
           SELECT
             current_database() AS database,
             current_user AS user_name,
-            inet_server_addr()::text AS server_address,
+            host(inet_server_addr()) AS server_address,
             inet_server_port() AS server_port,
             current_setting('server_version') AS server_version,
             current_setting('transaction_read_only') AS transaction_read_only,


rodaddy:
## Initial review findings — pinned head `c045962`

### Raised P1 — non-loopback `DB_HOST` could pass address equality

**Controller disposition: rejected as non-reproducible.**

The reported scenario cannot reach the changed query through the launcher. `runLocalCloneLauncher()` calls `validateLocalCloneMode(env)` before either database or embedding proof (`scripts/local-clone.ts:221`), and that source-of-truth validator calls `requireLiteralLoopback(DB_HOST, "DB_HOST")` (`src/local-clone-mode.ts:133`). A value such as `10.71.1.21` fails before connection or address comparison.

The changed query only converts PostgreSQL's already-connected address from CIDR text (`127.0.0.1/32`) to the host literal (`127.0.0.1`); exact equality and the earlier loopback gate both remain.

### Lane status

- correctness — CLEAN
- security — CLEAN
- quality — CLEAN
- backend/domain — CLEAN
- adversarial — one disputed finding above

A focused verifier will check the call-order disposition before terminal audit/readiness.

rodaddy:
## Fix verification and terminal audit — head `c045962`

- Initial review: correctness, security, quality, and backend/domain clean.
- Adversarial P1 alleging remote `DB_HOST` acceptance: **refuted** by controller source trace and an independent focused verifier. `validateLocalCloneMode()` requires literal `127.0.0.1` or `::1` before the database probe; `10.71.1.21` is covered by a rejection regression.
- Mandatory opposite-runtime whole-PR audit: **CLEAN**, no P0–P3 findings.
- Functional gates: 7 tests passed against the restored clone role/database, TypeScript passed, actual local-clone preflight returned `status: verified`, and leak scans passed.

No known P0/P1/P2 findings remain.

Issue #371, opened 2026-07-24
Give Claude and Claudex automatic total-recall hooks


rodaddy:
Parent epic: #369
Program: #320

## Outcome

Claude-compatible and Claudex sessions use the same reviewed direct Open Brain provider against the local server for automatic complete context packs, per-turn reflex pointers, capture, checkpoint, and supported wrap behavior. Codex integration is explicitly deferred.

## Acceptance

- Claudex is the primary user-facing runtime; validate it directly rather than claiming parity from source shape
- Claude-compatible hook behavior remains functional
- SessionStart requests all nine context-pack sections under one serialized budget
- UserPromptSubmit obtains body-free reflex pointers with prior-context suppression and client-owned placement
- Python runtime/client supports agent_reflex_pointers with exact-scope and fake-transport tests
- hook activation preserves unrelated hooks and has private rollback manifests
- raw prompts and pointer bodies never enter receipts, spools, logs, or telemetry
- functional cross-runtime isolation, citation, suppression, restart, and recovery tests pass
- Codex is not part of this issue

rodaddy:
Status audit 2026-08-03, against acceptance: SessionStart full context-pack under one serialized budget — RUNNING (CANON PACK 1/2 + 2/2 in every session, budget gate live). Capture/checkpoint/wrap — RUNNING (Stop/SubagentStop/SessionEnd/PostCompact all registered through openbrain-hook-env; ob_raw_turns takes live writes). Claudex validated directly — yes, this is the daily runtime. **NOT delivered: UserPromptSubmit reflex pointers.** Verified now: no reflex entrypoint exists in python/openbrain's project.scripts and no UserPromptSubmit hook beyond the context-budget gate is registered — `agent_reflex_pointers` per-turn recall was never ported to the Python package. That is the single remaining line item; everything else in this issue is live. Leaving open, scoped down to the reflex port.

rodaddy:
[...]

PR #368, opened 2026-07-23
feat(341): add language-aware full-text search

rodaddy:
## Summary

- add an explicit, validated `fts_config` argument to `search_brain` for language-aware lexical retrieval
- preserve the indexed English fast path and keep all sibling `executeSearch` consumers English unless they explicitly opt in
- analyze supported non-English searches with matching PostgreSQL regconfig stemming; explicit effective non-English requests require `admin` or `ob-admin`
- pin CI test databases to UTF8 and require all language-aware PostgreSQL suites through the anti-skip guard
- add functional MCP and real-PostgreSQL before/after tests across all five searchable table families

Tracks #341

## Validation

- focused config/search/anti-skip tests: **54 passed, 0 failed**
- focused search/answer regression tests: **90 passed, 0 failed**
- disposable PostgreSQL 18.4 + pgvector 0.8.2:
  - all 32 migrations applied
  - language-aware and all-table parity suites passed
  - all PostgreSQL suites passed during fix development
  - disposable databases dropped and verified absent
- supported regconfigs present: english, simple, spanish, french, german, portuguese
- `bun contracts/check-parity.ts`: passed; 13 fixtures across 8 capabilities
- `bunx tsc --noEmit`: passed
- `git diff --check`: passed
- exact-head GPT-5.6-family focused verification at `f1e5975df85efc9e8c8befa4b7a498f11a5dcf84`: **FOCUSED CLEAN**
- Full-tier opposite-family terminal verification at `f1e5975df85efc9e8c8befa4b7a498f11a5dcf84` by Claude Opus 4.8 high: **CLEAN — no P0/P1/P2 findings**

Functional before/after:
- German `Haus` misses German `Häuser` content under English analysis and matches under German analysis.
- Spanish `corrió` misses Spanish `corriendo` content under English analysis and matches under Spanish analysis.
- English content remains matched through the unchanged stored `search_vector` path.
- German tokens present only in table-specific indexed fields match under German stemming across thoughts, decisions, relationships, projects, and sessions.
- A German deployment env cannot silently change sibling keyword, successful-hybrid, or embedding-failure search consumers.

## Public contract and downstream rollout classification

This is an additive public MCP input-schema and behavior change, so downstream rollout applies. The current phase is explicitly local-only: merge supplies the local dogfood candidate but does not deploy or mutate core01.

- Open Brain local verification: complete.
[...]

PR #377, opened 2026-07-24
feat(371): opt-in unreviewed recovery recall flag (openbrain-memory 0.1.17)

rodaddy:
Closes nothing directly; part of #371 rollout (Python client 0.1.17).

Adds an optional `include_unreviewed_recovery` boolean to the `recall` operation. It is forwarded to `agent_context_pack` only on explicit opt-in, so an omitted flag preserves the prior request shape exactly. Boolean-ness is validated at both the CLI JSON boundary and the runtime facade (programmatic callers can bypass the CLI).

Validation: `uv run mypy src/openbrain_memory` clean, `uv run ruff check src tests` clean, `uv run pytest -q` 491 passed / 5 skipped. Wheel 0.1.17 built from this tree passed live local-clone dogfood (SessionStart recall `status=direct` against the loopback clone) and read-only recall against core01.

Critical self-review:
- Highest-risk behavior: request-shape drift for existing callers; mitigated by opt-in-only forwarding (omitted flag sends the exact prior arguments) and contract tests asserting the unchanged default shape.
- Assumptions that could be wrong: server ignores/errors unknown args on older deployments — flag is only sent on explicit opt-in, and the server already supports the argument on main (pre-#368-deploy risk is limited to opt-in callers).
- Missing/weak tests: no live-server integration test for the opt-in path in CI (env-gated canaries only); covered by fake-transport tests for call shape.
- Security/permission risk: none new — flag gates recovery-lane visibility already enforced server-side by role; client sends no privileged header change.
- Migration/deploy risk: none (client-only; no schema/server change).
- Downstream client/runtime risk: Claudex hook adapter pins wheel by sha; adapter stays at 0.1.16 until its own activation bump — 0.1.17 is backward compatible for all existing call shapes.
- Rollback/cleanup concern: revert = uninstall wheel; no persisted state format change.
- Fixes made before PR: revalidated boolean at runtime boundary, added contract-parity commit line.
- Known residual risk: opt-in callers against a server predating the argument get a server-side validation error; acceptable and observable.

### DIFF
--- a/python/openbrain-memory/pyproject.toml
+++ b/python/openbrain-memory/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "openbrain-memory"
-version = "0.1.16"
+version = "0.1.17"
 description = "Python client for remote Open Brain memory service"
 readme = "README.md"
 requires-python = ">=3.13"
--- a/python/openbrain-memory/src/openbrain_memory/cli.py
+++ b/python/openbrain-memory/src/openbrain_memory/cli.py
@@ -22,7 +22,13 @@ MAX_JSON_INPUT_BYTES = 64 * 1024
 MAX_JSON_OUTPUT_BYTES = 1_000_000
 _COMMON_KEYS = {"config", "operation", "scope"}
 _OPERATION_KEYS = {
-    "recall": {"max_latency_ms", "max_tokens", "query", "requested_sections"},
+    "recall": {
+        "include_unreviewed_recovery",
+        "max_latency_ms",
+        "max_tokens",
+        "query",
+        "requested_sections",
+    },
     "reflex": {"max_latency_ms", "max_tokens", "prior_context", "query"},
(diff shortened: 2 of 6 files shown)


rodaddy:
## PR-boundary review (Light tier) — head `220e9f9`

Two read-only Fable/high Workflow lanes (correctness+adversarial with `docs/sme/correctness.md`+`adversarial.md`; mandatory gotcha-agent with `docs/sme/gotcha-agent.md`).

Result: **0 HIGH / 0 MEDIUM / 1 LOW**.

- LOW (fixed in `220e9f9`): explicit `include_unreviewed_recovery=true` with `requested_sections` lacking `recovery` was a silent server-side no-op (`agent-context-pack.ts` drops the flag). The client now rejects the provably ineffective combination content-free before transport; regression tests cover reject/allow/explicit-false.

Post-fix gates: mypy clean, ruff clean, pytest 494 passed / 5 skipped.

Issue #418, opened 2026-07-25
PROV-9: hook entrypoints as console scripts


rodaddy:
Part of #409.

Port the hook entrypoints as console scripts: `ob-memory-provider`, `ob-claude-hook`, `ob-guard`, plus takeover/turn-capture/qmd-startup behavior.

`~/.claude/settings.json` hooks are `"type": "command"` - arbitrary shell. TypeScript was never required; a `bun run <hash-path>` invocation becomes a console script name with no path in it.

## Scope
- Thin entrypoints only: parse stdin, call the domain, render output. Logic lives in importable, testable functions (`CODING_STANDARDS.md`: *"keep script entrypoints thin"*).
- Console scripts declared in `pyproject.toml`.

## Acceptance
- Each entrypoint is exercised by a functional test over stdin to stdout/stderr/exit code.
- No business logic in an entrypoint module (each stays small; the test suite exercises the domain functions directly).
- Hook output shape is byte-compatible with what Claude Code expects today, proven by a test against captured real input.

## Non-goals
Editing `settings.json` - that is PROV-11.

rodaddy:
## This is no longer a routine port — it is the fix for a measured, ongoing data loss

Local plan with full reasoning: `_plans/418-prov-9-hook-entrypoints.md`

**Measured 2026-07-28.** Since 2026-07-25 21:00 the transcripts hold 1,236 operator messages. `ob_raw_turns` held **329 — 27%**. This session alone: 228 typed, 61 stored.

A backfill (`9e7f242`) recovered history — 4,571 → 25,151 turns, 329 → 1,089 operator messages, 1 → 18 repos. That is a stopgap. **The live hook keeps dropping turns until this issue lands.**

### The defects, in code running right now

Deployed adapter `sha256-cd5fb4e4...` (2026-07-25 15:30), which `settings.json` points at:

| Defect | File:line | Effect |
|---|---|---|
| `MIN_SIGNAL_CHARS = 24` | `turn-capture.ts:59`, enforced `:361` | Operator turns under 24 chars return null before classification |
| `limit ?? 8` | `raw-turns.ts:188` | Each Stop hook reads only the last EIGHT transcript entries |
| `TAIL_BYTES = 1MB` | `raw-turns.ts:31` | Long transcripts cannot be fully re-read |
| no watermark | absent | No cursor/offset/queue anywhere — verified by search. A missed entry is missed permanently |

[...]

Issue #412, opened 2026-07-25
PROV-3: one event vocabulary, imported not duplicated


rodaddy:
Part of #409. **Closes a live bug.**

The event-type vocabulary is declared twice and has already drifted: `openbrain_memory/agent.py` has **9** values (`fact, decision, blocker, action, artifact, receipt, question, correction, handoff`); the TS adapter has **8**, missing `question`.

Sending `event_type: "finding"` to the provider produced exit 0, no output, and no row. The value simply was not in the set, and nothing said so.

## Scope
- The provider **imports** the vocabulary from `openbrain-memory` rather than redeclaring it. One definition, workspace-internal dependency.
- If a shared home is needed, it lands in `openbrain-memory` and the provider consumes it - never a second copy.

## Acceptance
- Exactly one definition of the event-type set exists across both packages (grep proves it).
- A test asserts the provider's accepted set equals `openbrain_memory`'s - it fails if either side adds or removes a value. This is the drift guard; without it the two copies re-diverge silently.
- `question` is accepted by the provider (it is valid today and the TS rejected it).

## Non-goals
Changing which values are in the vocabulary. Adding `finding` is a separate discussion.

Issue #409, opened 2026-07-25
[Epic] PROV: port the Claude lifecycle adapter to Python as a uv workspace package


rodaddy:
## The finding

The Claude/Claudex lifecycle adapter is **7,513 lines of TypeScript whose main job is `spawnSync` into the Python CLI**. It is a shell around `openbrain-memory`, and the shell is a net negative today: it has produced two silent-failure incidents, and it duplicates domain vocabulary that has already drifted.

Measured 2026-07-25:

| Symptom | Evidence |
|---|---|
| Silent capture drop | `event_type: "finding"` is not in the adapter's `EVENT_TYPES`. `buildPackageRequest` returns `null`, `runProvider` returns `""`, exit 0. No receipt, no row, no error. Found only by checking psql. |
| Enum already drifted | `openbrain_memory/agent.py` declares **9** event types; the TS adapter declares **8** — missing `question`. Two hand-maintained copies of one vocabulary. |
| Undifferentiated failure | 11 distinct `return null` paths in `buildPackageRequest` collapse into 5 identical silent-return sites. Oversized input, malformed JSON, bad correlation id, unknown event type and empty content are indistinguishable to the caller. |
| Prior incident, same class | `uv` began colorizing `uv tool dir`; `isAbsolute()` saw `\033[36m/Users/...` and returned false. That parse failure **disabled Open Brain session-start recall entirely** and reported it as "lane memory unavailable" — a local string-parsing bug reported as a remote availability problem. |
| Install layout | `~/.local/share/openbrain-memory/` holds a `.venv`, a private `uv-cache`, a hand-built wheelhouse (7 `.whl`), and `adapters/versions/sha256-*/` — **8 versions in 8 days, nothing prunes them**. `~/.claude/settings.json` hardcodes **13 absolute paths** into one hash. |
| Already over the file limit | `ob-memory-provider.ts` is 1,954 lines and `context-budget-gate.ts` is 1,066. `CODING_STANDARDS.md` sets a 750-line hard limit, 600-line split review. |

`CODING_STANDARDS.md` already names the rule the silent drop violates:

> *"If a failure category is already computed (e.g. `errorCategory`), the operator must see it: exporting it to telemetry only is not an error message."*

`runProvider` computes `result.errorCategory`, emits it to Langfuse, and returns `""` to the caller.

## Decision

[...]

PR #421, opened 2026-07-26
feat(409): uv workspace and openbrain-provider skeleton (PROV-1, PROV-2)

### LINKED ISSUE #410: PROV-1: make python/ a uv workspace without regressing openbrain-memory
Part of #409.

Make `open-brain/python/` a uv workspace so a sibling package can exist beside `openbrain-memory`. Follows the `fleet-bus` pattern: root `pyproject.toml` with `[tool.uv.workspace] members = ["*"]`, members keep their own `pyproject.toml`.

`openbrain-memory` is standalone today (no workspace root, no shared lock). Adding the root moves its lockfile to the workspace, which touches an existing package's build - so this issue proves nothing regressed BEFORE any provider code is written.

## Scope
- Add `python/pyproject.toml` workspace root.
- `openbrain-memory/pyproject.toml` otherwise unchanged.
- Wire pre-push + CI to the same commands over the same paths.

## Acceptance
- `uv sync` resolves at the workspace root.
- `uv run mypy src/openbrain_memory`, `uv run ruff check src tests`, `uv run pytest -q` all pass for `openbrain-memory` - same results as before the change.
- `uv.lock` committed.
- CI green; pre-push runs the identical commands (a narrower local scope is a policy gap).

## Non-goals
No provider code. No changes to `openbrain-memory` source.

### LINKED ISSUE #411: PROV-2: scaffold openbrain-provider package with config and logger
Part of #409.

Scaffold `openbrain-provider` as a workspace member. Structure only - no domain logic.

## Scope
- Member package: own `pyproject.toml`, hatchling backend, ruff/mypy/pytest config matching the fleet-bus member shape.
- `src/openbrain_provider/` with `__init__.py`, `config.py`, `logger.py`, `constants.py`.
- `config.py`: frozen dataclasses with `__post_init__` validation. This replaces the scattered env reads in the TS adapter - env is read in one place and frozen.
- `logger.py`: Loguru setup only, no business logic. `serialize=True`, `enqueue=True`.
- `constants.py`: module-level `UPPER_CASE`, no inline magic numbers/strings.
- `tests/` mirroring the module tree.

## Acceptance
- Every module opens with a docstring and `from __future__ import annotations`.
- Every function fully typed; `str | None`, never `Optional[str]`.
- Invalid config raises at construction, not at first use (fail closed at boot).
- `ruff format --check`, `ruff check`, `mypy` strict, `pytest` all green.
- No file over 250 lines at this stage.

## Non-goals
No request/receipt/dispatch/reflex/observation logic - those are PROV-3..8.

rodaddy:
Closes #410. Closes #411.

First two slices of #409: make `python/` a uv workspace, then add the
`openbrain-provider` package skeleton (config + observability) that will replace
the TypeScript lifecycle adapter currently shipped as a content-addressed
`sha256-<hash>` install directory.

Skeleton only. Request parsing, receipts, dispatch, reflex recall, observation
export, and hook entry points are PROV-3..9 (#412-#418).

## What landed

**PROV-1 (#410)** — `python/` becomes a uv workspace. Root `pyproject.toml`,
one lock file, one `.venv`. Membership is a glob rather than a hand-maintained
list, because uv accepts a listed member that does not exist: it resolves clean,
writes no lock entry, and reports nothing. A list cannot tell you whether a
package actually joined the workspace; the filesystem can.

`.python-version` pins 3.13 to match CI. Without it uv selected 3.14 locally, so
local gates and CI would have run different interpreters.

**PROV-2 (#411)** — the `openbrain-provider` package:

```
src/openbrain_provider/
  constants.py       bounds, each verified against the adapter and cited by line
  config.py          the only place environment is read
  observability.py   thin adapter over rtech-obs; hook-specific policy only
```

Config fails closed at construction: every object is a frozen dataclass
validating in `__post_init__`, so an invalid value raises at boot rather than at
first use.

## Defects found and fixed during this work

Four of these are in my own PROV-2 code, found by a critical self-review after
the gates were already green. Recording them because the pattern matters: no
lint, type, or test gate reads packaging metadata, a shared contract, or a
comment's truthfulness.

**1. A console script whose target module does not exist.** `[project.scripts]`
declared `ob-memory-provider = "openbrain_provider.cli_provider:main"`, but
`cli_provider.py` is PROV-9. uv installs the shim anyway, so `.venv/bin` held a
present, executable file that died on `ModuleNotFoundError`. Found by a cold
clone, not by any gate. Removed until the module lands, and
`tests/test_packaging.py` now imports every declared entry-point target.

**2. The package grew its own logger.** `rtech-standards` ships `rtech-obs` as a
shared implementation; this package built a parallel one that emitted loguru's
internal `{"text":..., "record":{...}}` shape — none of the five expected
top-level fields, an uppercase level, no `host`. Any query written against the
shared envelope would have missed it entirely. `logger.py` deleted;
`observability.py` now wraps `rtech-obs` and holds only what it cannot know.

That needed two mechanisms, not one. `stdout=False` covers the normal path, but
an unwritable `LOG_FILE` makes `rtech-obs` add a stdout sink — overriding
`stdout=False` by design. Its default location `/mnt/logs/services/` is not
provisioned, so that fallback would fire on every run today, and for a hook that
means logging onto its own response channel. `resolve_log_file()` probes a
writable directory (shared location if present, temp otherwise) so it never
does, and needs no change when the mount eventually lands.

**3. Constants claimed provenance they did not have.** `constants.py` said
"carried over from the TypeScript adapter". Measured against it, not one
matched:

| constant | adapter | was | drift |
|---|---:|---:|---:|
| `MAX_INPUT_BYTES` | 65,536 | 1,000,000 | 15.3x |
| `MAX_CONTEXT_PACK_MAX_TOKENS` | 20,000 | 100,000 | 5.0x |
| `MAX_PACKAGE_OUTPUT_BYTES` | 1,000,001 | 4,000,000 | 4.0x |
| `MAX_INGEST_INPUT_BYTES` | 2,097,152 | 8,000,000 | 3.8x |
| `MAX_DISTILLED_*` | 16,384 bytes | 20,000 chars | unit changed |

Ten of twelve had no callers. The unused ten are deleted; the two that are
enforced are corrected and cited by line. This surfaced
`MIN_CONTEXT_PACK_MAX_TOKENS = 100`, which was being ignored entirely — the
budget is now bounded at both ends, exactly matching the adapter's 100..20,000.

The 5x ceiling error was live: PROV-6 would have accepted a 100k-token budget
the TS path rejects at 20k, which is the cross-runtime disagreement this port
exists to remove.

**4. `base_url` was validated by nothing.** All of these constructed
successfully while the module docstring promised it failed closed:
`not a url`, `file:///etc/passwd`, `javascript:alert(1)`, `$(whoami)`,
`http://127.0.0.1:3100 ; rm -rf /`. Now scheme-allowlisted to http/https with
host, port, whitespace, and control-character checks.

**5. A cross-repo relative path source that only worked here.** `rtech-obs`
resolved via `../../../rtech-standards`, which resolves against the checkout's
location. A cold clone proved it fails anywhere the two repos are not siblings,
which includes CI. Repinned to an immutable git commit.

## Verification

Cold clone, empty uv cache, `rtech-standards` not present as a sibling:

| check | result |
|---|---|
| `uv sync --all-packages --frozen` | lock complete, `rtech-obs` from git |
| interpreter | 3.13.14, matching CI |
| `openbrain-provider` gates | ruff format, ruff, mypy strict (src + tests), pytest — 81 passed |
| `openbrain-memory` gates | unchanged: 515 passed, 9 skipped |

Wheel checks: `py.typed` is packaged, no `entry_points.txt` ships, and a
separate consumer under `mypy --strict` against the installed wheel resolves the
types and catches a deliberate `int = str` error.

Envelope conformance verified end to end: five top-level fields, lowercase
level, RFC 3339 UTC timestamp, populated `host`, empty stdout.

## Critical Self-Review

- Highest-risk behavior: `resolve_log_file()`. It exists to stop `rtech-obs`
  from attaching a stdout sink, because stdout is the hook's machine-readable
  return channel. If it ever returns an unwritable path, the fallback fires and
  a hook silently corrupts its own response. Covered by tests for the fallback,
  the preference order, and probe cleanup, plus a test asserting nothing reaches
  stdout with no `LOG_FILE` configured.
- Assumptions that could be wrong: that `rtech-obs` is the right shared
  dependency. It is four commits old, infra's first pass, and this package is
  its first consumer. If it turns out wrong, `observability.py` is ~110 lines
  and replaceable. Also assumed: `/mnt/logs` will eventually exist at the path
  the contract names.
- Missing/weak tests: no test asserts the two enforced constants still match the
  TypeScript adapter, so that provenance can silently rot again — the adapter is
  outside this repo, so a real guard needs the values checked in. No test
  exercises `configure_observability` under concurrent hook invocations writing
  the same log file. The review swarm found a third: a test named after a hazard
  that asserted an adjacent property instead (see the fix comment below).
- Security/permission risk: `base_url` validation is new and is the only input
  sanitization in this PR. It allowlists http/https and rejects whitespace and
  control characters, but does not restrict hosts, so an operator-set value can
  still point anywhere. That is intended at this layer. No namespace or auth
  logic is touched.
- Migration/deploy risk: none. No migrations, no schema changes, no runtime
  code path is wired to this package yet. The `sha256` adapter still serves live
  traffic and is untouched until PROV-11.
- Downstream client/runtime risk: none for this PR — no MCP tool, transport, or
  client-facing schema changes. `openbrain-memory` is unmodified and its test
  baseline is byte-identical. The `rtech-obs` git pin is on an immutable commit,
  so an upstream branch move cannot change this build.
- Rollback/cleanup concern: reverting the branch removes the workspace root, so
  anyone with an existing `python/.venv` needs one `uv sync`. The deploy script
  excludes `python/dist` and both `.venv` paths already.
- Fixes made before PR: all five defects above, each with a test that fails on
  the old behavior — verified red-then-green for the console-script guard.
- Known residual risk: `observability.py` writes the shared log envelope
  locally instead of depending on `rtech-obs`, because `rtech-standards` is
  private and this repo's CI passes no token — no git URL can reach it. That is
  a second implementation of a shared envelope, which is the drift the contract
  exists to prevent. Bounded by being ~90 lines, using the contract's field
  names, and testing conformance against a literal copy of the contract's field
  list. Swap back in one commit once the package is installable here (publish it,
  or wire a cross-repo checkout secret).

## Review Gate

- Live Open Brain checks: [x] not applicable because: this PR adds an unwired Python package with no server, MCP tool, or database path; no live Open Brain behavior changes.

## Contract Parity

- Contract parity: [x] runtime-specific because: no MCP tool, transport, schema, or client-facing contract is touched; `openbrain-memory` is unmodified and its 515/9 test baseline is unchanged.


rodaddy:
## Review swarm: findings and fixes

Four SME lanes reviewed this PR on ephemeral runner boxes — correctness,
adversarial, security, gotcha-agent — each seeded with its `docs/sme/*.md`
lane file. Every finding was reproduced locally before being fixed. Two were
found independently by two lanes.

| # | Severity | Finding | Lanes | Status |
|---|---|---|---|---|
| 1 | HIGH | Unwritable explicit `LOG_FILE` redirected logs onto the hook's response channel | correctness + adversarial | fixed |
| 2 | HIGH | `rtech-obs` git dependency unreachable from CI | adversarial + gotcha | fixed |
| 3 | MEDIUM | New package had no CI gate at all | gotcha | fixed |
| 4 | MEDIUM | Base-URL validation errors leaked embedded credentials | security | fixed |

### 1 — logs onto the response channel (HIGH)

`resolve_log_file` returned an operator-supplied path unchecked. The reasoning
was that silently relocating an operator's logs is worse than failing where they
can see it — wrong about the consequence, because the failure was not visible.
An unwritable `LOG_FILE` fell through to a stdout sink and wrote **436 bytes of
log records onto the hook's machine-readable stdout**. Reproduced exactly.

The explicit path is now probed like the defaults and falls back when unwritable;
a writable operator path is still honored unchanged.

The lanes also identified why it survived review: `test_unwritable_log_file_is_not_fatal`
exercised this exact scenario and passed, because it asserted the process kept
running and never captured stdout. A test pointed at the wrong risk.

### 2 — dependency unreachable from CI (HIGH)

Two lanes independently reproduced `Host key verification failed` from an
HTTPS-cloned checkout. The first fix switched `ssh://` → `https://`, and the new
CI job (finding 3) immediately caught the real constraint:

```
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

`rtech-standards` is **private** and this repo's CI passes no token — no
`secrets.` reference exists in `ci.yml`. No git URL can work. Two commits had
treated a reachability problem as a URL-scheme mistake.

Per maintainer decision, the dependency is dropped for now. `observability.py`
writes the contract envelope directly (~90 lines), verified byte-comparable to
`rtech-obs` output for the same input. The cost is stated plainly in the module
docstring and README: this is a second implementation of a shared envelope,
which is the drift the contract exists to prevent. It is bounded by being small,
by using the contract's field names, and by tests asserting conformance against
a literal copy of the contract's field list rather than against whatever the
code emits. The public surface is shaped for a one-commit swap back once
`rtech-obs` is installable here.

One hazard disappears in the trade: `rtech-obs` implements §5.1 (unwritable
`LOG_FILE` must not be fatal) by adding a **stdout** sink, which for an agent
hook is the one place a log line must never land. The local writer falls back to
**stderr**, removing the collision at its source rather than fencing it off.

### 3 — no CI gate (MEDIUM)

`python-package` is rooted at `python/openbrain-memory`; nothing named
`openbrain-provider`. Its tests, strict typing, and wheel contents were entirely
unenforced. Adds a `python-provider` job mirroring the existing one, plus
`ruff format --check` and `mypy` over `tests/` as well as `src/`, and adds it to
`deploy`'s `needs` so a broken provider cannot deploy.

This job justified itself on its first run by catching finding 2's real cause.

### 4 — credentials in error messages (MEDIUM)

`urlsplit` accepts userinfo, and every validation error quoted the raw value:

```
OPENBRAIN_BASE_URL=http://alice:supersecret@example.com:notaport
  → ConfigError: ... 'http://alice:supersecret@example.com:notaport'
```

A config error at boot is exactly what gets printed to a terminal, captured by a
service manager, or pasted into a bug report. `_safe_url` now redacts userinfo
across all six URL error paths, operating on the raw string rather than parsed
attributes — it is called precisely when the value is too malformed to parse,
and reading `parsed.port` would itself raise on the invalid-port case being
reported. The host survives redaction so an operator can still tell which URL
was rejected:

```
base url has an invalid port: 'http://***@example.com:notaport'
```

### Also fixed

A flush race in `test_every_record_is_a_single_json_line`: it read the log file
before removing the sinks, and with `enqueue=True` the write happens on a
background thread. Green locally on every run, red on the CI runner — the worst
ordering for noticing. Checked the rest of the file mechanically; it was the only
case.

### Verification

Every fix has a regression test **proven to fail on the old code** — the fixes
were stashed and the 7 targeted tests run against pre-fix source; all 7 failed.

```
F1: stdout leaked 0 bytes (was 436); falls back to a writable temp path
F4: base url has an invalid port: 'http://***@example.com:notaport'
    writable explicit path still honored: True
envelope: timestamp/level/service/host/message at top level,
          lowercase level, RFC 3339 UTC with Z, extras nested under context
```

81 tests, up from 66. `ruff format --check`, `ruff check`, `mypy --strict`
(src and tests), and `pytest` all green; five consecutive runs to confirm the
race fix. `openbrain-memory` unchanged at 515 passed / 9 skipped.

### Operational note for the runner lane

The runner wrapper **overwrites `/work/out/result.json`** with its own status
stub on exit, so all four lanes' findings were destroyed on delivery and had to
be recovered from `agent.log`. The collab share showed four healthy
`{"status": 0}` results and zero findings — silent data loss in the review
tooling itself, not in this PR.

rodaddy:
## Log durability on abrupt termination — fixed

Raised by Rico: *"what, the logger should be it's own async thing, shouldn't clash ever?"* The async design is right — a hook must not block on disk I/O to log — but that instinct exposed a gap the earlier test-only fix had not addressed.

### The defect

`enqueue=True` hands each record to a writer thread. loguru's `atexit` hook drains it on a clean exit. Nothing drained it on a signal.

| Exit mode | Records reaching disk |
|---|---|
| Clean exit | 200 / 200 |
| SIGTERM, no drain window | **133 / 200** |

The 67 lost were the *newest* ones. For a hook process that is exactly backwards: the records worth having are the ones written just before something tore it down. A hook killed mid-session loses its own explanation.

### The fix

`configure_observability` installs a SIGTERM/SIGINT handler that drains the queue, then re-raises so the process still dies the way the sender asked. Swallowing the signal would trade a lost log line for a process that ignores shutdown — strictly worse, since a service manager just escalates to SIGKILL.

Three things it is careful about:

- It only claims a signal nobody else holds. This module runs inside someone else's process.
- `signal.default_int_handler` is a **callable**, so a plain `callable(previous)` guard for "already claimed" skips SIGINT every time. It is treated as unclaimed.
- SIGINT keeps its `KeyboardInterrupt` semantics rather than being converted into a hard kill through `SIG_DFL`, so `finally` blocks still run on Ctrl-C.

### The test was wrong first, and that is the part worth reading

My first regression test had the child write a `ready` file and sleep while the parent polled and then signalled. That poll loop handed the writer thread ~10ms — enough to finish the queue. **It reported 200/200 with the fix reverted.** I nearly committed it as proof of a fix it never exercised.

It is the same shape as the earlier HIGH in this PR, where `test_unwritable_log_file_is_not_fatal` asserted the process kept running and never captured stdout, so a 436-byte leak onto the hook's response channel survived every green run. Both tests asserted the process was *alive* rather than that the data *arrived*.

Rewritten so the probe signals itself immediately after the last record, with no drain window:

```
_install_signal_flush() disabled:  133/200   (test fails: "lost 59 of 200 records")
_install_signal_flush() enabled:   200/200   (5 consecutive runs)
```

### Also fixed

The autouse fixture restored sinks but not signal dispositions. `configure_observability` installs a real handler on whatever process calls it — under pytest that is the pytest process — so this module's handler was leaking into later tests and into the path of a CI job cancellation.

### Verification

- Red-then-green proven by neutering `_install_signal_flush()` in place (keeping the symbol importable, so the red is a behavioral failure and not an `ImportError`).
- `ruff format --check`, `ruff check`, `mypy src`, `mypy tests` — all clean.
- 85 tests pass (was 81), 3 consecutive runs, no flake.
- CI green on the Linux runner, including `python-provider`.

SME captured to `docs/sme/gotcha-agent.md`: the drain-window trap, the "alive vs arrived" assertion pattern, the `default_int_handler` callable trap, and the review questions for any library function that installs a process-global signal handler.

Issue #422, opened 2026-07-26
src/source-sync.test.ts fails only in the full suite — captured console.error line goes missing


rodaddy:
## Summary

`src/source-sync.test.ts` has a test that fails **only when the full suite runs**, and passes when the file runs alone. It is **pre-existing on `main`** — not introduced by any open PR.

## Reproduction

Baseline `main` (04513bd), against a freshly migrated local Postgres:

```
bun test src/source-sync.test.ts     ->  24 pass, 0 fail
bun test                             ->  2455 pass, 35 skip, 1 fail
```

The single failure:

```
src/source-sync.test.ts:484
    const failLine = lines.find((l) => l.includes("source_sync_failed"));
    expect(failLine).toBeDefined();
                     ^
error: expect(received).toBeDefined()
```

Same failure, same assertion, on a split branch off `main` — confirming it travels with `main` rather than with any feature work.

## What the test does

It monkey-patches `console.error` to capture the emitted log line, then asserts that a `source_sync_failed` line was emitted at all:

```ts
const original = console.error;
const lines: string[] = [];
console.error = (...args) => { lines.push(args.map(String).join(" ")); };
try { res = await syncSource(failingPool(err) as never, ...); }
finally { console.error = original; }

const failLine = lines.find((l) => l.includes("source_sync_failed"));
expect(failLine).toBeDefined();
```

The comment above it says the point is to prove *"the guard actually ran, not skipped"*. In the full suite it captures nothing, so either the line is not emitted or it does not reach the patched `console.error`.

## Why it matters

This is a **security-boundary test**. The assertions after the failing line check that the emitted log carries finite categories and never leaks sentinel bytes:

```ts
expect(failLine).toContain('"error_name":"other"');
expect(failLine).toContain('"error_code":"unknown"');
for (const line of lines) expect(line).not.toContain(SENTINEL);
```

When `failLine` is `undefined`, the redaction assertions never meaningfully run. A test that silently stops checking is worse than one that is simply absent, because the suite still reports it as covered. That is the same "asserted the process was alive rather than that the data arrived" pattern already captured in `docs/sme/gotcha-agent.md`.

## Notes toward a root cause

- `src/source-sync.test.ts` is the **only** test file that patches `console.error` (grep confirms), so it is not two tests fighting over the same patch.
- `src/logger.ts` is unmodified relative to `main`, so a branch-local logger change is ruled out.
[...]

PR #424, opened 2026-07-26
feat(observability): shared log envelope, runtime log level, and credential redaction

rodaddy:
## Summary

- Second split off the parked `feat/410-uv-workspace` branch (after #421 and #423). This is the **observability foundation**: the shared log envelope, host stamping, a runtime-changeable log level, and credential redaction in `describeError`. ~1,300 insertions across 12 files.
- The remaining group on that branch — `f8333d7` "log the silent two-thirds of the server" (5,217 insertions, 31 tool files) — depends on this and lands separately.
- `withLogging` emits entry/exit/failure automatically so the five required log points from `_DOCS/CODING_STANDARDS.md` are the *shorter* path. The wrapper never swallows: it logs, then re-throws.
- **Side effect worth naming:** this branch makes `src/source-sync.test.ts` (#422) pass in the full suite, stable across 3 consecutive runs. The plausible mechanism is `MIN_LEVEL` becoming a mutable `minLevel` — an earlier test could previously strand the process at a level where `logger.error` no longer reached `console.error`, unrecoverably for the rest of the run. **I could not isolate this cleanly** (reverting `logger.ts` alone breaks 16 dependent tests), so I am reporting it as observed, not as a proven cause. #422 should stay open until someone confirms the mechanism.

## Verification

Evidence:
- `bunx tsc --noEmit` clean.
- 2516 tests pass, 0 fail, 35 skip. Three consecutive full runs, no flake.
- No migrations, no Python changes in this slice.
- Every fix below was reproduced *before* being fixed and verified red-then-green by reverting the fix in place.

## Critical Self-Review

Three defects found by self-review before this PR opened, all in `setLogLevel`/`log`:

- Highest-risk behavior: **`setLogLevel` could strand the level wide open.** The announce line widens the gate so the transition cannot filter its own notice, then restores it — with a plain assignment, not `finally`. `log` throws for real (`FILE_SINK.write` on a full disk). Proven: with the sink throwing, `debug -> error` left `minLevel` at `debug` while the caller saw an exception and would conclude nothing changed. That inverts the intent on the one path the setter exists for — an operator raising verbosity mid-incident, on the box whose disk just filled, silently gets debug volume from a call that reported failure.
- Assumptions that could be wrong: that `SECRET_PATTERNS` covers what actually appears in thrown errors. `describeError` redacts `error_name` too, because a real leak was caught in review with `err.name` set to `NatsError nats://user:pass@host` — but the pattern set is the ceiling on that guarantee.
- Missing/weak tests: no test asserts envelope conformance against the shared contract itself, only against this implementation's output. No concurrency test for `addLogSink` add/remove during emit.
- Security/permission risk: **async context could override `service`, `host`, and `level`** — `contextFields()` spread last, after the fields the code claims to own. Those two are the only Loki labels, so one reader spreading a request object wholesale splits the query surface silently. Fixed by spreading context first.
- Migration/deploy risk: none. No schema, no migration.
- Downstream client/runtime risk: `logger.ts` is imported broadly, but the envelope only *adds* `service`/`host`. `setLogLevel`/`getLogLevel`/`addLogSink` are new exports; nothing existing changes shape.
- Rollback/cleanup concern: reverting restores a module-load-constant log level. No data or schema to unwind.
- Fixes made before PR: the `finally` on `setLogLevel`; context spread order; **guarding `FILE_SINK.write`**, which ran unguarded ahead of the console sinks so a throw lost the line everywhere — again making disk-full the case that blinds the operator investigating it.
- Known residual risk: `redactForLog` runs per log line over every pattern. It builds a fresh `RegExp` each call (verified: identical output across 10 calls, all occurrences caught, no `lastIndex` stranding), but the cost on a hot path is unmeasured.

## Review Gate

- Live Open Brain checks: [x] not applicable because: no deployed behavior changes until merge; the hosted smoke would exercise the pre-merge build.

## Contract Parity

- Contract parity: [x] runtime-specific because: this slice touches only the TypeScript server's logging internals. No MCP tool schema, no client-facing surface, and no Python client change, so there is no matched-client contract to hold parity against.

## Downstream Rollout

Notes/evidence:

- Server-internal logging only. No tool schema, response shape, or client call changes, so no downstream client can break on it.
- The envelope adds `service` and `host` to emitted lines. That is additive for any log consumer; a parser reading the existing fields is unaffected.

### DIFF
--- a/docs/sme/adversarial.md
+++ b/docs/sme/adversarial.md
@@ -808,3 +808,51 @@ The adversarial move that caught it was cheap: take the comment's own claim
   credentials — no SSH agent, no token, no sibling checkout, empty cache?
 - When a second attempt fails differently from the first, is the underlying
   constraint being addressed, or is the symptom being renamed?
+
+## [2026-07-26] The unguarded step in a fail-open path is the one that matters
+
+**Provenance:** PR #424, adversarial + SME lanes independently. **Severity:** HIGH. **Status:** fixed.
+
+`log()` in `src/logger.ts` guarded everything except one statement. `contextFields()`
+caught a throwing reader, the file sink was guarded, every extra sink was
+guarded — and `const output = JSON.stringify(entry)` sat between the entry
+construction and all four destinations, unguarded.
+
+`entry` is built from caller-supplied `extra` **and** from `withContext()`
+fields, so a cycle, a `BigInt`, or a throwing `toJSON` in either was enough to
+lose the line everywhere *and* throw into arbitrary application code.
+
+The consequences ran well past logging, and each contradicted a documented
+contract:
+
+- `withLogging` documents "re-throws whatever fn throws, unchanged". It threw a
+  serializer error instead, destroying the real root cause **on the failure
+  path**.
+- `withFallback` — the one function whose purpose is to never propagate — threw
+  instead of returning its fallback.
+- Worst: `withLogging` emits its entry line at `debug` *before* calling `fn`. At
+  `info` that line is gated out and `fn` runs. At `debug` the same call threw and
+  **`fn` was never invoked.** Raising the log level to investigate an incident
+  silently stopped work from running — the exact scenario the runtime log-level
+  setter exists for.
+
+The repo already knew the hazard: `src/audit-log.ts` explicitly defends against
+"a cycle, a BigInt anywhere". The module that *every* emitter routes through did
+not.
+
+### Review Questions
+
+- In any path that deliberately catches everything, **list the statements that
+  are not wrapped.** One unguarded step in an otherwise fail-open function is a
+  stronger signal than an unguarded step anywhere else — the surrounding care
+  proves the author intended it not to throw.
+- Does a logging/telemetry wrapper emit a line *before* calling the wrapped
+  function? Then a throw in the emit path is a **control-flow** bug, not a
+  logging bug: the work never runs.
+- Does the failure mode depend on the log level? A bug that only appears at
+  `debug` surfaces exactly when someone is debugging.
+- Does redaction cover the whole record, or one contributor? Redacting
+  `error_message` while a caller field carries the identical credential on the
+  same line is not redaction.
+- Do observers (sinks, subscribers) receive the same redacted data the file and
+  console get, or the raw object?
--- a/docs/sme/gotcha-agent.md
+++ b/docs/sme/gotcha-agent.md
@@ -725,3 +725,62 @@ never reach it.
   import time (`FILE_SINK`, `HOST_NAME`, `SERVICE_NAME`) cannot be exercised by
   a suite that does not set that env — a subprocess is the honest way in.
 - Does the suite's default env silently disable the code path under test?
+
+## [2026-07-26] Four defects that three review passes all missed
+
+**Provenance:** PR #424, final independent gate. **Severity:** HIGH.
+**Status:** fixed.
+
+An author self-review and two swarm lanes all cleared `src/logger.ts`. A fourth
+independent pass found four more defects in the same file, all reproduced. What
+unites them is that every one is a **guard that was correct about its intent and
+wrong about its mechanism** — so reading the code, and the comment above it,
+confirms the intent and hides the flaw. Three passes read the intent.
+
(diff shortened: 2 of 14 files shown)


rodaddy:
## Final independent gate (Sol, high effort) — 4 HIGH findings, all fixed

The final pass ran read-only on the host. Both approved disposable roots were
denied in its sandbox, so it could not run the suite or do mutation testing,
and it died before writing its findings file or finishing its remaining gates.
It did complete in-memory probes against the exact HEAD source and reported
four defects.

**I reproduced all four independently before fixing, and ran the gates it did
not reach.** One claim as literally stated ("plain sensitive-key values leak")
did not reproduce — flat `password=…` text redacts correctly. It only
reproduces for values in *structured fields*, which is how the fix is scoped.

| # | Finding | Reproduced | Fixed in |
|---|---------|-----------|----------|
| 1 | Sensitive value in a structured field emitted in clear | yes | `f69354a` |
| 2 | Truncation splits a credential so the tail leaks | yes | `f69354a` |
| 3 | Repeated non-circular object dropped as `[Circular]` | yes | `f69354a` |
| 4 | Reentrancy guard clobbers a nested decision that collides | yes | `f69354a` |

**1 — the redactor never saw the key.** Every detector is value-shaped, but a
JSON logger hands the replacer each value in isolation, so
`{"password":"hunter2driveway"}` went out readable. `json_labeled_secret`
already matches that pair when given it as *text*; it was simply never shown
the key. Sensitivity now travels with the field name via `isSensitiveKey`.

**2 — the bound manufactured the leak.** The 16 KB cap ran *before* the
patterns, so a DSN straddling the boundary was cut mid-credential and
`postgres://user:supe` survived unredacted. Redaction now runs first. The
quadratic `PRIVATE_KEY_BLOCK_RE` risk that motivated the cap is handled
directly by `boundForMatching` instead of by truncating every long string —
the existing pathological-input test still passes.

**3 — `[Circular]` for something that was not circular.** The `WeakSet` only
ever grew, so an object referenced from two sibling fields was flagged.
`{a: shared, b: shared}` serialized as `{"a":{"id":"abc"},"b":"[Circular]"}`.
Sharing one config/namespace object across fields is ordinary, so the false
positive was the common case. Detection now walks the ancestor chain; a
genuine self-reference is still caught (covered by its own test).

**4 — the guard's key collided.** `minLevel === widened` cannot tell "my
temporary value is still installed" from "a nested call chose that same
value". Lowering `debug -> error` makes `widened` equal `"debug"`, so an
auto-escalate sink asking to stay at debug was told it succeeded while the
outer restore reverted it. Now keyed on a generation token. The pre-existing
nested test passed throughout because its nested call picks a value that
*differs* from the widened one — the colliding case was the one it could not
catch.

### Verification

Each fix was **reverted in place** to confirm its regression test fails on the
old behavior, one at a time:

```
revert key-aware redaction  -> (fail) redacts a value by its field name …      48 pass 1 fail
revert truncation order     -> (fail) redacts a credential that straddles …    48 pass 1 fail
revert cycle detection      -> (fail) keeps a repeated non-circular object …   48 pass 1 fail
revert reentrancy guard     -> (fail) lets a nested setLogLevel win even …     48 pass 1 fail
```

Gates the final pass did not reach, run here:

- **Extraction fidelity** — 18 detectors on main, 18 in `secret-patterns.ts`,
  none lost or added; `sharing.ts` re-exports the identical pattern objects
  (source + flags asserted).
- **Leaf purity / import cycles** — `secret-patterns.ts` has zero imports; the
  logger loads clean as the first module in 5 different load orders, no TDZ.
- **`URL_SCHEME_ALT` over-match** (its open question #3) — benign
  `ws://`/`nats://`/`tls://` URLs and prose pass through untouched; both
  credential-bearing URLs redact. No over-match.
- **Gitleaks** over the full range — 7 commits, no leaks.
- **Contract parity** — 14 fixtures across 9 capabilities, passed.
- **Test-quality spot check** — mutated the asserted event name in
  `embedding-repair.ts`; its assertion failed as it should.

**Suite: 2386 pass / 0 fail. `bunx tsc --noEmit` clean.**

SME knowledge captured in `647134d`: what unites all four is that each is a
guard correct about its *intent* and wrong about its *mechanism*, which is why
three prior passes read past them.

Issue #414, opened 2026-07-25
PROV-5: dispatch boundary — subprocess, NO_COLOR, timeouts, output validation


rodaddy:
Part of #409. **Closes a prior incident's root cause.**

The dispatch boundary is thin: locate the CLI, run it with JSON on stdin, validate stdout. It has already caused one outage.

Measured 2026-07-25: `uv` began colorizing `uv tool dir`, so `isAbsolute(stdout.trim())` saw an ANSI-prefixed path and returned false. That single parse failure resolved the package runtime to `null` and **disabled Open Brain session-start recall entirely**, reported as "lane memory unavailable" - a local string-parsing bug presented as a remote availability problem.

`CODING_STANDARDS.md`: *"A tool's human-facing stdout is not a machine API"* - set `NO_COLOR=1` (or `--format json` / `--quiet`) in the **subprocess** env and strip ANSI before parsing. Never in the operator's interactive shell.

## Scope
- Runtime resolution, subprocess invocation, child-env construction, stdout validation.
- `NO_COLOR=1` set in the child env in exactly one place; ANSI stripped before any parse.
- Explicit timeouts and bounded output handling on every external call.
- Failure category computed AND surfaced (see PROV-4), never telemetry-only.

## Acceptance
- A test feeds ANSI-wrapped output and asserts it parses correctly.
- A test asserts `NO_COLOR` is present in the child env and absent from the parent.
- Timeout and oversized-output paths each produce a distinct named failure, not a generic one.
- No unbounded read, no unbounded wait.

## Non-goals
Changing the package CLI's own wire format.

rodaddy:
## SUPERSEDED-by-rewrite (evidence recorded; left OPEN pending full-criteria mapping) — 2026-08-01

This ticket is a child of the PROV epic (#409) and targets the **retired TypeScript adapter** (`buildPackageRequest`, the ~1,954-line monolith, `receipt-state.ts`). That adapter no longer exists on the live path: it was retired 2026-08-01 to `~/.local/share/openbrain-memory/_archive/420-cutover-retired-20260801/` by the #420 settings cutover (row-proven; see #453). The capture/hook stack is now the Python `openbrain-memory` package.

**What the rewrite already provides, verified live in current source this session (LAW 0, RUNNING):**
[...]

PR #428, opened 2026-07-26
feat(412): one event vocabulary, imported not duplicated

### LINKED ISSUE #412: PROV-3: one event vocabulary, imported not duplicated

rodaddy:
## Summary

Closes #412.

The provider now imports `EVENT_TYPES` from `openbrain-memory` through the workspace dependency instead of carrying a literal of its own, and `EVENT_TYPES` is promoted to that package's public surface so consuming it does not mean reaching into `agent.py`.

**Two findings changed the shape of the slice.**

**1. The drift the issue reports is already repaired.** Python, the TS client, and the TS server all carry the same nine values today, and `question` is present in every one. So the durable deliverable is not a repair — it is the guard the issue itself calls for: *"the drift guard; without it the two copies re-diverge silently."*

**2. There were not two copies. There are six, plus SQL.** Writing the guard immediately found four the issue never named:

| Surface | File |
|---|---|
| Python (the definition) | `openbrain_memory/agent.py` |
| TS client | `clients/ts/src/runtime.ts` |
| TS server set | `src/agent-memory.ts` |
| `MemoryEventType` union | `src/agent-memory.ts` |
| **MCP tool schema** | `src/contract-schemas.ts` |
| tiering `EventType` union | `src/tiering.ts` |
| SQL write path | `src/tools/table-constants.ts` |
| **CHECK constraint** | `src/db/migrations/013_session_events.sql` |

That is the argument for a test rather than a convention — nobody holds six declarations in their head. A final assertion fails if a seventh appears, so the guard cannot silently stop covering what it guards.

**The SQL constraint is included and matters most.** Postgres is the authority, so a code set drifting *wider* than the constraint reproduces the exact reported symptom: validation passes, the insert is refused, and the caller sees exit 0 with no row. That is the worst failure available to a memory write, because success is indistinguishable from a no-op.

## Verification

Memory 545 pass, provider 98 pass, TS 2386 pass / 0 fail, `bunx tsc --noEmit` clean, mypy and ruff clean on both packages.

**Mutation-verified.** Removing `question` — the value from the original bug — from each of the six surfaces in turn fails exactly that surface's assertion:

```
RED  test_typescript_client_matches_python      <- clients/ts/src/runtime.ts
RED  test_typescript_server_matches_python      <- src/agent-memory.ts
RED  test_mcp_tool_schema_matches_python        <- src/contract-schemas.ts
RED  test_tiering_union_matches_python          <- src/tiering.ts
RED  test_table_constants_match_python          <- src/tools/table-constants.ts
RED  test_database_constraint_matches_python    <- 013_session_events.sql
```

Re-run after refactoring the shared failure message, to confirm the extraction did not quietly unbind the assertions. Provider-side, replacing the re-export with a literal fails four tests including the no-redeclaration check and the `question` acceptance criterion.

## Acceptance criteria

- [x] Exactly one definition across both packages — the provider re-exports, and `test_provider_declares_no_vocabulary_of_its_own` fails if a literal is added there.
- [x] A test asserts the provider's set equals `openbrain_memory`'s and fails if either side changes — `test_provider_set_is_the_memory_set`.
- [x] `question` is accepted by the provider — asserted by name, since it is the value whose absence was the bug.

## Critical Self-Review

- Highest-risk behavior: the guard parses TypeScript and SQL with regex anchors. A refactor that renames a declaration makes the anchor miss. Mitigated by `_extract_quoted_block` failing loudly on a missing bracket and by the equality assertion reporting an empty set rather than passing vacuously — a missed anchor fails, it does not silently succeed.
- Assumptions that could be wrong: that all six surfaces should be identical. Checked — each gates a different stage of the same write path, so a value valid at one and not another is by definition a bug.
- Missing/weak tests: `IMPORTANCE_LEVELS` and `CANDIDATE_TYPES` are duplicated the same way and are not guarded here. Deliberate — out of scope for #412, and worth its own issue rather than silently widening this one.
- Security/permission risk: none. No auth, namespace, or credential surface.
- Migration/deploy risk: none. The migration is read, never modified.
- Downstream client/runtime risk: `EVENT_TYPES` is newly public in `openbrain-memory`. Additive only — no existing import path changed, so no consumer breaks.
- Rollback/cleanup concern: reverting drops the guard and the re-export; nothing depends on either yet, since the provider has no dispatch path until #414.
- Fixes made before PR: my first anchor for the tool schema matched a different enum earlier in `contract-schemas.ts`, and the second picked up `type: "enum"` before the list. Both caught because the test failed loudly rather than matching nothing. Also corrected the migrations path (`src/db/migrations`, not `migrations/`), which was silently skipping the constraint check — the assertion that matters most was not running.
- Known residual risk: the no-seventh-copy check greps for `"blocker"`, so a new declaration that happens to omit that value would not be flagged. Chose a real value over a sentinel because a sentinel drifts too.
- SME review-memory update: [x] not applicable because: no review findings yet; this is the implementing PR.

## Review Gate

- Live Open Brain checks: [x] not applicable because: no server, transport, or MCP tool behavior changes — the tool schema is read by a test, never modified.

## Contract Parity

- Contract parity: [x] runtime-specific because: no wire or capability surface changed. `git diff origin/main...HEAD --name-only` is six Python files; nothing under `src/`, `contracts/`, or `clients/ts/`. The pre-push gate matched the `openbrain-memory` path rather than an actual contract edit.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no tool, schema, protocol, or client-facing change. The only new public symbols are `openbrain_memory.EVENT_TYPES` (additive) and `openbrain_provider.vocabulary`, which has no external callers.

### DIFF
--- a/python/openbrain-memory/src/openbrain_memory/__init__.py
+++ b/python/openbrain-memory/src/openbrain_memory/__init__.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from .agent import (
+    EVENT_TYPES,
     AgentMemory,
     MemoryClient,
     MemoryContext,
@@ -79,6 +80,7 @@ from .turn_concepts import (
 )
 
 __all__ = [
+    "EVENT_TYPES",
     "AgentMemory",
     "DEFAULT_MAX_KEYS",
     "TurnConcepts",
--- a/python/openbrain-memory/src/openbrain_memory/agent.py
+++ b/python/openbrain-memory/src/openbrain_memory/agent.py
@@ -32,6 +32,25 @@ logger = logging.getLogger(__name__)
 # (entities before concepts), first-seen wins.
 DERIVED_QUERY_MAX_KEYS = 8
 
+#: The session-event vocabulary. THE definition — every other surface imports or
+#: is checked against this one, never redeclares it (#412).
+#:
+#: The authority is ultimately the database: `ob_session_events.event_type`
+#: carries a CHECK constraint listing exactly these values. A value that is not
+#: in the constraint is rejected by Postgres, so a set that drifts *wider* than
+#: the constraint produces the failure this vocabulary was consolidated to
+#: prevent — the caller passes validation here, the insert is refused, and the
+#: symptom is a silent no-row write rather than a named error. `event_type:
+#: "finding"` did exactly that: exit 0, no output, no row.
+#:
+#: Because it is duplicated across languages by necessity (Python, the TS
+#: client, the TS server, and SQL cannot share one literal), the guard is a test
+#: rather than an import: `test_event_vocabulary.py` asserts every surface
+#: agrees, and fails if any one of them adds or drops a value.
+#:
+#: Widening this set therefore means widening the CHECK constraint in the same
+#: change. Adding a value here alone does not make it usable; it makes it
+#: silently unwritable.
 EVENT_TYPES = {
     "fact",
     "decision",
--- a/python/openbrain-memory/tests/test_event_vocabulary.py
+++ b/python/openbrain-memory/tests/test_event_vocabulary.py
@@ -31,7 +31,7 @@ from __future__ import annotations
 
 import re
 import subprocess
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 
 import pytest
 
@@ -49,6 +49,15 @@ def _extract_quoted_block(source: str, anchor: str) -> set[str]:
     the thing it guards, and a malformed anchor fails loudly as an empty set
     rather than silently matching too much.
     """
+    if anchor not in source:
+        # A rename upstream is the likely cause, and a bare
+        # `ValueError: substring not found` from `str.index` does not say so.
+        # It fails either way -- the point of naming it is that the next reader
+        # knows the declaration moved rather than the vocabulary changed.
+        pytest.fail(
+            f"anchor {anchor!r} not found; the declaration was renamed or moved. "
+            "Update this guard to the new spelling -- do not delete the check."
+        )
     start = source.index(anchor)
     tail = source[start + len(anchor) :]
     end = min(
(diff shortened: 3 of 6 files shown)


rodaddy:
## Self-review pass — 2 defects in the guard itself, both fixed

Ran the gotcha-agent lane's own instruction against this PR (mandatory: it touches `python/openbrain-memory/`). The guard had the exact defect it exists to prevent — **a check that reports success over a gap.**

### 1. The test-path filter hid real source files — MEDIUM

`test_vocabulary_is_declared_exactly_where_expected` excluded any path containing the substring `test`. That silently drops `latest.ts`, `manifest.ts`, and `attestation.ts` from the sweep — any of which could hold a seventh declaration.

Reproduced by planting a tracked file with four event-type literals:

```
$ cat > src/latest-vocab.ts   # "latest" contains "test"
export const T = ["fact", "decision", "blocker", "handoff"] as const;
$ git add -N src/latest-vocab.ts && pytest ...::test_vocabulary_is_declared_exactly_where_expected
1 passed          # <- OLD filter: seventh copy invisible
FAILED ...        # <- NEW filter: caught
```

Now matched on path components (`tests/`, `__tests__/`, `test_*`, `*.test.ts`, `*_test.py`). Verified real test paths are still excluded and the three above are now scanned.

### 2. A renamed declaration failed with an unhelpful error — LOW

Renaming `export type EventType` produced `ValueError: substring not found`. It **did** fail — so the first attack question, "can an assertion pass vacuously?", answers no — but the message tells the next reader nothing about what happened.

```
$ sed -i 's/export type EventType/export type EventKind/' src/tiering.ts
E   ValueError: substring not found          # before
E   Failed: anchor 'export type EventType' not found; the declaration was
    renamed or moved. Update this guard to the new spelling -- do not
    delete the check.                        # after
```

Both extractors now name the anchor. The two union extractions used string indexing directly and are routed through a shared helper so they behave identically.

### Boundary found and documented, not fixed

`git grep` searches **tracked** files, so a brand-new untracked declaration is invisible to the sweep. That is the correct boundary for CI (which runs on committed code) and fixing it would mean scanning the working tree, but it is now stated in the docstring rather than left as a surprise.

### Not changed

`IMPORTANCE_LEVELS` and `CANDIDATE_TYPES` are duplicated across the same surfaces and remain unguarded — out of scope for #412, and worth their own issue rather than silently widening this PR.

Gates after the fix: memory 545 pass, provider 98 pass, mypy and ruff clean on both. All six drift mutations re-run and still bind.

PR #471, opened 2026-08-02
fix(hooks): split canon SessionStart emissions

rodaddy:
## Summary

- split package-owned canon startup injection into two independently registered `SessionStart` console scripts
- emission 1 carries `profile_guidance` + `process_guidance`; emission 2 carries `repo_facts` + every other configured section
- preserve each rule body byte-for-byte and identify both outputs as one canon pack in their headers
- document the uv-tool install and two-entry Claude settings wiring

## Validator

Live local-service validation used the installed uv tool through the production `openbrain-hook-env` wrapper and ran both entry points over the same captured-shape `SessionStart` payload.

- live section counts: `profile_guidance=10`, `process_guidance=15`, `repo_facts=7` (`32` total)
- emission 1 `additionalContext`: `9,327` characters / `9,327` UTF-8 bytes
- emission 2 `additionalContext`: `2,422` characters / `2,422` UTF-8 bytes
- serialized envelope files: `9,443` bytes and `2,514` bytes
- whole-rule check: `32/32` exact UTF-8 rule bodies found exactly once across the two decoded `additionalContext` strings and only in the expected emission
- headers: both identify one shared pack; emission 1 names `profile_guidance, process_guidance`; emission 2 names `repo_facts`
- test mutation proof: the new split tests were run before implementation and failed on missing `CanonEmission` / `sections_for_emission`, then passed after implementation

## Gates

- `uv run ruff check src tests` — passed
- `uv run mypy src/openbrain` — passed (`35` source files)
- `uv run pytest -q` — passed (`308 passed, 17 deselected`)
- repository pre-push gate — passed: TypeScript typecheck, Bun tests, path-sensitive openbrain mypy/ruff/pytest, generated-package-doc check, and applicable parity classification
- live Claude settings JSON parsed successfully and contains two `openbrain-session-start*` command entries

## Critical self-review

- Highest-risk behavior: emission 1 initially measured `10,412` characters; fixed by removing per-item metadata scaffolding from emitted blocks while retaining every rule body unchanged. Final live measurement is `9,327` characters.
- Assumptions that could be wrong: the two independently invoked hooks make separate live pack reads; a canon mutation between those reads could make the pair represent adjacent pack states.
- Missing/weak tests: this branch directly validates both installed entry points against the local service, but does not repeat the already-proven full Claude cold-start transcript experiment for multiple hook outputs.
- Security/permission risk: token delivery remains exclusively in `openbrain-hook-env`; no token or endpoint was added to settings, source, logs, or PR text.
- Migration/deploy risk: the installed uv tool and local Claude settings are already activated from this branch. Rollback is reinstalling from the prior source and removing the second `SessionStart` command entry.
- Downstream client/runtime risk: not applicable to MCP schema, transport, or Python client contracts; the changed consumer boundary is the local Claude hook registration.
- Rollback/cleanup concern: the uv tool was baked from the required temporary worktree; removing the worktree does not remove the installed wheel, but the recorded local source path will no longer exist for a future implicit upgrade. The documented reinstall command remains the explicit upgrade path.
- Fixes made before PR: added exact partition tests, console-script resolution coverage, compact whole-rule emission rendering, runbook wiring, live two-output validator, and settings registration.
- Known residual risk: two service reads can briefly observe different canon revisions during concurrent writes; each individual output remains internally whole and fail-open.
- SME review-memory update: [x] not applicable because: the self-review found no reusable MEDIUM+ review pattern beyond the behavior tests and runbook updates already committed.

## Review Gate

  - `32/32` live rules verified byte-exact across the two outputs; measured contexts were `9,327` and `2,422` characters.

## State

OPEN for review. **Do not merge until the wider code effort is complete.**

### DIFF
--- a/docs/420-cutover-rollback.md
+++ b/docs/420-cutover-rollback.md
@@ -9,7 +9,7 @@ absent, and restoring the previous settings in one command.
 `~/.claude/settings.json` is OUTSIDE this repo, so it is not versioned here. The
 backup taken before the cutover, the installed console scripts, and the env
 wrapper are all outside the repo too. What lives in the repo: this doc, the
-`openbrain-session-start` console-script declaration
+two `openbrain-session-start*` console-script declarations
 (`python/openbrain/pyproject.toml`), and the surviving TS the settings now point
 at (`_ob/scripts/...`, a separate repo).
 
@@ -21,7 +21,7 @@ at (`_ob/scripts/...`, a separate repo).
 | `SubagentStop` | (not wired) | `sh …/openbrain-hook-env openbrain-capture-subagent-stop` |
 | `SessionEnd` | `bun … claude-hook.ts` | `sh …/openbrain-hook-env openbrain-session-end` |
 | `PostCompact` | `bun … claude-hook.ts` | `sh …/openbrain-hook-env openbrain-post-compact` |
-| `SessionStart` | `bun … claude-hook.ts` | `sh …/openbrain-hook-env openbrain-session-start` |
+| `SessionStart` | `bun … claude-hook.ts` | two entries: `sh …/openbrain-hook-env openbrain-session-start` and `sh …/openbrain-hook-env openbrain-session-start-remaining` |
 | `UserPromptSubmit` / `PreCompact` | `bun … claude-hook.ts` (no-op) | removed |
 | `context-budget-gate.ts` (7 events) | `sha256-…/context-budget-gate.ts` | `/Volumes/ThunderBolt/Development/_ob/scripts/context-budget-gate.ts` |
 | `guard.ts` (PreToolUse Bash) | `sha256-…/ob-memory-provider/guard.ts` | `/Volumes/ThunderBolt/Development/_ob/scripts/ob-memory-provider/guard.ts` |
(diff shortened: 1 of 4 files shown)


rodaddy:
### software-factory: passed
Reviewed at: `d25bdb9aeff48fb3fe721c7b14af766353740a8d`

Findings:
- [medium] Session open now makes two independent canon server round\-trips instead of one (python/openbrain/src/openbrain/apps/hooks/session\_start\.py:\_inject\_canon\_with)
- [low] Rollout requires coordinated tool reinstall AND a second settings\.json hook entry; partial rollout silently degrades (python/openbrain/pyproject\.toml:project\.scripts / docs/420\-cutover\-rollback\.md)
- [low] Physical injection order of the two emissions depends on settings registration order, not this code (python/openbrain/src/openbrain/apps/hooks/session\_start\.py:\_header\_bits)
- [info] Test gap: no direct assertion of the per\-emission empty\-sections early return \(empty stdout\) (python/openbrain/tests/test\_capture\_hooks\.py)
- [info] Overall review summary: clean, well\-covered refactor; no blocking findings (python/openbrain/src/openbrain/apps/hooks/session\_start\.py)

Issue #485, opened 2026-08-02
find_duplicates: admin/ob-admin self-join runs unscoped over the whole table and never returns


rodaddy:
## State

**RUNNING** — reproduced this session against a live Postgres copy of the corpus (`open_brain_local_play`, 24,845 embedded non-archived `thoughts`). Defect is in current `src/`, not in the #463 server rewrite.

## What happens

`src/tools/find-duplicates.ts` builds a pairwise self-join and appends a namespace predicate to each side:

```ts
const namespacePredicateA = appendReadNamespacePredicate(auth, params, "a.namespace");
const namespacePredicateB = appendReadNamespacePredicate(auth, params, "b.namespace");
```

For a **global role** (`admin`, `ob-admin`), `readableNamespaces()` returns `undefined`, so `appendReadNamespacePredicate()` contributes the empty string and **both** predicates disappear. The emitted query becomes:

```sql
FROM thoughts a
JOIN thoughts b ON a.id < b.id
  AND b.archived_at IS NULL AND b.embedding IS NOT NULL
WHERE a.archived_at IS NULL AND a.embedding IS NOT NULL
  AND a.embedding <=> b.embedding < $1
ORDER BY distance ASC
LIMIT $2
```

`LIMIT` cannot help: `ORDER BY distance ASC` forces the planner to compute the distance for every surviving pair before it can order them. `find_duplicates` exposes **no namespace argument**, so an admin caller has no way to narrow it.

## Measured, same database, same join

| Shape | Result |
|---|---|
| Namespace-scoped (`a.namespace = ANY(...) AND b.namespace = ANY(...)`) | **256.7 ms**, 5 rows |
| Unscoped (the admin path) | **cancelled at 60,074 ms** by `statement_timeout` — did not return |

An earlier measurement on this corpus recorded the scoped form at 2.939 ms and the unscoped form still running past 120 s, with 42 backends observed `state='active'` on this exact join more than five minutes after the run that launched them had ended.

At 24,845 embedded rows the unscoped join is ~308 million pairwise halfvec distances with no index able to serve `a.embedding <=> b.embedding < $1` across arbitrary pairs.

## Why it matters beyond slowness

[...]

Issue #463, opened 2026-08-01
The server rewrite: src/ is the same application, and it goes last


rodaddy:
## Question

Rewrite the Bun/TypeScript server (`src/`) as the final rewrite phase.

**Sequencing gate:** this ticket starts only after the remaining Python work lands: the #419 gate disposition, the bulk ingester non-Claude format adapters, and the canon/lens pack reconciler.

When this ticket is taken up, the first step is **CHARTING**, not code: settle the target architecture/runtime, inventory what the Python client contract already pins, and define the migration/cutover shape for a running dogfood service. Implementation starts only after that route is charted.

**Plan of record:** `_plans/python-port-sequence.md:438` ("`src/` is the same application being rewritten") and the operator ruling of 2026-08-01: all three Python applications, then the TypeScript application, then full testing, then swap to the new reality.

Type: task (AFK). This records the final rewrite phase and its sequencing gate; it does not authorize implementation before the Python remainder lands.

rodaddy:
## #463 start receipt

- Charter: **WRITTEN** — `_plans/463-server-rewrite-charter.md`
- Scaffold: **WRITTEN** — `server/`, explicitly scaffold-only and serving nothing
- Contract harness: **WRITTEN** — parity now checks `current-src` and `server-rewrite-scaffold` against the same reviewed declaration fixture
- Cutover: **NOT STARTED**
- Sequence: final `src/` cutover remains last, after the remaining Python work in map #443
- Running application: unchanged; the branch has no `src/` diff and no startup/deploy entrypoint points at `server/`

Validation receipts:

- TypeScript typecheck passed
- Dual-provider parity passed: 14 fixtures / 9 capabilities / 2 providers
- Python fixture parity passed: 14 tests
- Full Bun + isolated Postgres passed: 3,070 pass / 34 skip / 0 fail
- DB anti-skip guard passed: 194 live-Postgres cases across all 10 required suites

Draft PR: https://github.com/rodaddy/open-brain/pull/476

No issue was closed and the map was not edited.

rodaddy:
**Stale-blocker sweep verdict (2026-08-05): PARTIAL** — every issue this one references is closed, so it was audited against live state (scripts/stale-blockers.ts flagged it; a verification lane checked the acceptance itself).

**What actually remains:**
[...]

Issue #341, opened 2026-07-22
Add language-aware full-text search


rodaddy:
## Summary

Use language-aware full-text search configuration for ingested content to improve the lexical side of existing hybrid RRF without regressing established recall.

**Workstream key:** `SOURCE-5`  
**Priority:** P2  
**Parent:** #336  
**Depends on:** #338, #324

## Scope

- Select language-aware FTS configuration from supported corpus metadata or explicit source settings.
- Integrate with existing hybrid retrieval.
- Validate synchronized and collected content through the live gate.

## Acceptance criteria

- [ ] FTS uses an explicit language-aware configuration for the supported corpus languages.
- [ ] Representative lexical queries improve or remain equivalent to the prior default.
- [ ] The EVAL-3 gate shows no regression in hybrid recall metrics across file, drop-folder, and approved conversation fixtures.

## Validation

- [ ] Run focused FTS ranking tests for supported languages.
- [ ] Run EVAL-3 before and after the change and record the metric comparison.
- [ ] Run end-to-end source-ingestion fixtures from SOURCE-2, SOURCE-3, and SOURCE-4.

## Non-goals

- Changing embedding models.
- Adding languages not represented by approved corpus requirements.

## Standing constraints

- Exact namespace and session-scope isolation remains a security boundary.
- Prompt placement stays client/runtime-owned; no implicit MCP `_meta` injection.
- Observability remains content-free.
- Dream planning remains dry-run-safe unless a separately authorized mutating wrapper is used.

rodaddy:
Closing #341 — language-aware full-text search is live in production.

- Feature merged in PR #368 (`8306253`); the three verified post-merge findings (env-default non-English ordinary-role bypass + missing DB cost controls, vector-only false denial for unused `fts_config`, missing operator docs) fixed in PR #378 (`c713f50`).
- Deployed to core01 via approved CI manual dispatch (run `30109547650`) at exact main `c713f50`; health green on 3100/3101/3102 and open-brain.rodaddy.live.
- Hosted `search_brain` canaries passed: English keyword, schema (fts_config), ordinary-role non-English denial, admin German, namespace isolation, source-scope, and the vector-mode unused-config fix.
[...]

Issue #167, opened 2026-06-19
Retire legacy collab namespace (99.4% mirrored to shared-kb, frozen)


rodaddy:
## Goal

Retire the legacy `collab` namespace now that `shared-kb` is canonical and collab is frozen.

## State (open-brain DB, 2026-06-19)

- `thoughts`: collab = 11,943, of which **11,874 already mirrored to shared-kb** (content_hash match); **69 NOT yet in shared-kb**.
- `ob_entities`: 12 collab rows (repo_facts, created_by rico, 2026-06-18).
- `ob_session_lanes`: 4 collab lanes.
- No collab writes since 2026-06-18 18:15 — frozen.

## Work

- [ ] Migrate/promote the **69 un-mirrored collab thoughts** to shared-kb (or classify+archive if noise).
- [ ] Decide fate of the 12 collab `repo_fact` entities + 4 collab lanes (re-tag to shared-kb / rico / archive).
- [ ] Once collab has no live unique content, remove `collab` as `legacySharedNamespace` from config/code (read-policy fallback, shared-namespace.ts) and drop the legacy read-fallback paths.
- [ ] Coordinate with the stale-defaults fix (separate issue) so tooling no longer references collab.

## Non-goals
- No bulk delete of collab rows without confirming they're mirrored/archived first.

## Sequencing
After the memory-tiering pipeline (#159 → #160/#161). Lower priority; collab is frozen and 99.4% mirrored, so it's not actively harmful — just cleanup. Captured 2026-06-19.

rodaddy:
## Release deploy receipt - partial #167 progress only

Open Brain app code from `origin/main` was deployed to core01 through release tag `v0.1.0-rc.1` at commit `8af40cbb0749f62d56f0bbf0d382d59740eada10`.

Evidence:

- GitHub Actions run: https://github.com/rodaddy/open-brain/actions/runs/28889009203
- Jobs: `check`, `python-package`, `db-integration`, and `deploy` all passed.
- Runtime/process backup before deploy: `/Volumes/ThunderBolt/open-brain/backups/pre-deploy-20260707T181623Z`
- DB backup before DB/schema repair: `/Volumes/ThunderBolt/open-brain/backups/pre-release-db-20260707T182137Z`, verified with `pg_restore -l`.
[...]

PR #486, opened 2026-08-02
feat(server): port the search and recall family onto the rewrite foundation (#463)

rodaddy:
Ports the search/recall family of the #463 server rewrite: `search_brain`,
`search_all`, `brain_answer`, `list_recent`, `adjacent_context`, and the shared
retrieval engine underneath them. Behavior was read from `src/`; the code is
written fresh against `_DOCS/STANDARDS-*`. Public tool arguments are reproduced
byte-for-byte.

**Scope note up front:** `agent_context_pack` and `agent_reflex_pointers` are
NOT in this PR. They are blocked on a repo hook, documented at the bottom. This
PR is the half that is finished and proven.

## State (LAW 0)

- **RUNNING** — every gate below was executed this session against live
  Postgres, output quoted.
- **MERGED** — nothing. Base is `feat/463-transport-wave`, not `main`.

## One retrieval stack

`server/tools/search-engine.ts` owns all three arms (keyword, vector, hybrid),
RRF fusion, namespace isolation, archived filtering, and canonical/physical
namespace translation. Every surface enters through `executeSearch`, so ranking
and isolation are defined once and two surfaces cannot answer the same question
differently.

The three arms differ in what failure means, which is the part worth reviewing:

| Arm | Null embedding |
|---|---|
| `keyword` | not needed — this is the arm that still works during an embedding outage |
| `vector` | **hard error**; `[]` would report "nothing matched" for an outage |
| `hybrid` | degrades to keyword-only, logged |

## Binding decisions cited in module headers

- **#341** — FTS regconfig allowlist. `english` reads the stored GIN-indexed
  `search_vector`; a non-english allowlisted config recomputes `to_tsvector`
  over the same source columns. `ftsConfigLiteral` re-asserts the allowlist at
  the single point a config becomes SQL text.
- **#368 findings 1 and 2** — the non-English privilege boundary applies to the
  EFFECTIVE config on arms that run FTS regardless of provenance; `vector` mode
  ignores `fts_config` rather than denying, and an env-default non-english
  degrades to english for ordinary roles.
- **`docs/qmd-ob-layered-recall.md`** — `search_all` stays PEER federation,
  fail-open on the qmd side only. Explicitly not the layered recall that
  document specifies for scope-derived repo questions; that is a separate
  surface.

## Three defects found while porting

Each verified against the live schema or a live measurement, not inferred.

1. **`HAS_EXTRACTED_METADATA` listed only `thoughts`.**
   `information_schema.columns` reports the column on `thoughts` AND
   `decisions`, so decision metadata was being dropped from every result. Fixed
   and the reasoning recorded on the constant.

2. **`QMD_PATH=` (empty, not absent) silently disabled qmd federation forever.**
   `docs/qmd-ob-layered-recall.md` records this: `?? DEFAULT` treats
   empty-string as configured, so the caller spawned `bun "" search …` and the
   fail-open arm degraded to brain-only with no signal. `resolveQmdPath` now
   treats blank as unset and logs `qmd_search_unconfigured` at the boundary.

3. **Parity harness leaked `SHARED_NAMESPACE_*` process-wide** (fixed in its own
   commit, `ac8caec`). Bun runs every test file in one process and
   `sharedNamespaceConfig()` re-reads env on every call, so a module-scope
   assignment rewrote the shared/legacy pair for whatever loaded after it.
   Measured, same DB:

   | Run | Before | After |
   |---|---|---|
   | `search-brain.test.ts` alone | 40 pass / 0 fail | 40 pass / 0 fail |
   | that file + the parity suite | 57 pass / **5 fail** | 62 pass / **0 fail** |
   | full suite | 3125 pass / **38 fail** | 3179 pass / **5 fail** |

   `LINK_RELATIONS` was also checked against the live `ob_links` `CHECK` clause
   (migrations 010 + 018) rather than copied on faith: exactly fourteen, in that
   order, matching.

## `namespace-guard.ts` removed

It was a deliberate stub registering only `search_brain`'s namespace denial so
that boundary stayed observable while the body was unowned. Its own header says
the search lane replaces the whole file. The denial it froze is unchanged and
`server-namespace-denial.fixture.json` still passes against the real handler.

## Tests assert SQL, not rows

`server/tools/search-read-scope.test.ts` (19 tests) asserts the emitted SQL and
bound parameters, because the defect class is a predicate that is silently
WRONG rather than absent — a hardcoded namespace list still returns rows, just
the wrong ones, so a row-shaped assertion passes while isolation is broken.

Covered: every table arm carries the predicate (one unconstrained arm leaks
that table's whole corpus while the others look fine); both search arms; scalar
vs array binding; denial before any query runs; the FTS privilege gate including
the vector-mode non-denial; and the two distinct parameter indexes `list_recent`
binds for its data and count queries.

## Gates

| Gate | Result |
|---|---|
| `bunx tsc --noEmit` | clean |
| `bun test server/tools/` | 26 pass / 0 fail |
| `bun test contracts/` | 26 pass / 0 fail (was 24; +2 new provider-fixture pairs) |
| full `bun test` | 3179 pass / 35 skip / 5 fail |
| pre-push ref-gate | passed, no `--no-verify` needed |

Live Postgres was an isolated playground DB (`open_brain_local_play`) via
`OPENBRAIN_TEST_DATABASE_URL`. The dogfood database was never touched.

The 5 residual failures are a **pre-existing** collision among
`tier-recommendations.pg` / `list-stale.pg` / `bulk-set-tier.pg` — 14 pass / 5
fail together, 5 pass / 0 fail alone.
`git diff --name-only origin/feat/463-transport-wave...HEAD` confirms this
branch does not touch them; last commit on them is 78864b9 (dream wave, #430).
Unchanged in count and identity from this branch's own baseline.

## Parity declarations flipped

`search-brain-arms` and `brain-answer` → `implemented` for
`server-rewrite-scaffold`. Both fixtures RUN against the scaffold and pass —
verified via the junit reporter that `server-rewrite-scaffold:
server-search-brain-arms` and `server-rewrite-scaffold:
server-brain-answer-empty` are present, not skipped.

## Not in this PR, and why

`agent_context_pack` and `agent_reflex_pointers` are blocked by
`.claude/hooks/design-lookup-gate.ts`.

`server-context-pack-sections.fixture.json` requires emitting these frozen wire
fields verbatim, as recorded current-src output:

```
counters: { dropped, expired, trimmed }                  (working_set)
counters: { dropped, expired, trimmed, marked, purged }  (recovery)
budget:   { ttl_ms, max_sessions, max_items_per_session, ... }
```

The gate's HARD-NO-on-limitation-language arm refuses any write containing those
tokens. The operator's 2026-08-02 carve-out (130b977, cherry-picked here as
cb72f07) already covers this exact class — "pre-existing frozen contract
literals" — and enumerates `search_brain`'s argument, `list_stale`'s envelope,
`decompose_entry`'s chunk literal, and `.trim()`. These realtime field names are
the same class but are not in the enumerated list, so the pattern does not
exempt them.

The gate states plainly that no lookup, citation, or rephrase clears it, and
that an agent believing it has a perfect reason is exactly what it exists for.
So I did not extend the exemption myself, and I did not rename the fields to
slip past — renaming would break the byte-for-byte parity this port requires.

**Unblocking is one operator edit**: add these frozen realtime field names to
`CAP_PROPOSAL_EXEMPT`, exactly as 130b977 did for the earlier set. They are
recorded output of the current implementation, not new proposals.

`context-pack-sections` and `reflex-pointers` therefore remain
`scaffold-declared` and their fixtures remain skipped for the scaffold provider.

## Also filed

#485 — `find_duplicates` runs an unscoped self-join for admin/ob-admin and never
returns. Independently re-measured this session rather than inherited: scoped
**256.7 ms**, unscoped **cancelled at 60,074 ms** under `statement_timeout`,
over 24,845 embedded rows. The starved connections are what turned 6 real
parity failures into 13 reported ones.

## Critical self-review

- **Highest-risk behavior:** namespace isolation in the shared engine. Every arm
  binds an auth-derived predicate; asserted on emitted SQL for all five table
  arms, both search arms, and both `list_recent` queries.
- **Assumptions that could be wrong:** that `executeSearchWithSharedFallback`'s
  `requestedShared` flag reproduces the two src wrappers' split. Src had separate
  scalar and scoped-array entry points; this collapses them into one predicate.
  Both paths are exercised, but the legacy fallback itself is off by default
  (#167), so the top-up branch is reasoned-through rather than executed here.
- **Missing/weak tests:** the legacy-namespace top-up merge is not covered by a
  new test at this layer. The graph/relational arm from src is not ported and is
  not claimed.
- **Security/permission risk:** denials are content-free and never echo the
  requested namespace, so they cannot probe which namespaces exist. Only
  allowlisted table names and FTS configs are ever interpolated; query text,
  namespaces, and tiers are bound.
- **Migration/deploy risk:** none. `servesTraffic` stays false; the scaffold
  serves no traffic.
- **Downstream client/runtime risk:** none yet for the same reason. Tool
  arguments are byte-identical, so a later cutover does not move the contract.
- **Rollback/cleanup concern:** `namespace-guard.ts` is deleted; reverting this
  commit restores it.
- **Fixes made before PR:** the three defects above.
- **Known residual risk:** the context pack is unported, so the rewrite provider
  cannot yet answer `agent_context_pack` — which is declared honestly in the
  manifest rather than papered over.

Refs #463


rodaddy:
### software-factory: errored
Reviewed at: `1801bff7c2c074e41d5f76aae6b19873008dabcf`

Findings:
- [high] PR \#486 could not be reviewed: no repository checkout at the target SHA and no way to obtain it (/work \(assigned checkout dir; contains only job\.md\))

rodaddy:

PR #484, opened 2026-08-02
feat(server): port the curation, tiering, and entity tool family (partial wave)

rodaddy:
## Summary

- Ports the first slice of the #463 curation/tiering/entity tool family onto the server rewrite foundation, registered through the `TOOLS_BOUNDARY` seam from PR #480 (merged in with `--no-ff`; conflicts in the parity net resolved to the memory-tools side, which is a strict superset).
- Tools implemented: `curate_entries`, `archive_entry`, `rate_entry`, `update_entry`, `resolve_entry`, `tier_recommendations`, `get_entity`.
- **The DreamEngine dry-run default is preserved exactly.** `curate_entries` reports unless the caller passes an explicit `dry_run: false` **and** holds an admin-tier role, so no planning path can archive; vague findings stay report-only in both modes. `tier_recommendations` is read-only and writes no tier. The decomposition planner is deliberately free of pool/auth imports so planning *cannot* mutate.
- `entry_access_log` is read as a **log**, not a counter, per `006_cognitive_tiering.sql`. **No new index** was added on it — `008_index_cleanup.sql` removed the unused ones and adding one with no reading consumer would repeat that.
- Table names still reach an interpolated SQL position, so every one is narrowed first by the Zod-enum allowlist (`ALL_TABLES` in `server/tools/curation-helpers.ts`); nothing outside those five values can be interpolated. Namespace scope on every read and mutation comes from the foundation's auth-derived `namespacePredicate` builder.
- Parity declarations flipped so PR #477 fixtures now **run** against `server-rewrite-scaffold` instead of being declaration-only: capabilities `curation`, `entry-revision`, `tier-recommendations`, `graph-entities`. Named-gap map shrunk **32 → 28**, fixture coverage **31 → 35** tools.
- `servesTraffic` remains `false`.

**Scope note (LAW 0 — this PR is WRITTEN + MERGED-to-branch, not RUNNING):** this is a partial wave. `decompose_entry`, `list_stale`, `access_report`, `get_stats`, the remaining entity/people tools, the source-registry family, `list_namespaces`, repo-facts, and `promote_shared` are **NOT** in this PR. Two of them are blocked by a repo authoring hook and are described under Known residual risk.

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

Evidence, all against an **isolated** database (`open_brain_463_curation_test`, 46 migrations), never the dogfood database:

[...]

PR #487, opened 2026-08-02
feat(server): complete the #463 curation wave -- 21 tools, gap map 28 to 10

rodaddy:
## Summary

Completes the #463 curation wave that PR #484 ran out of runway on. Stacked on `feat/463-curation-wave`; **do not merge** ahead of it.

**21 tools ported**, in five scoped commits:

| Family | Tools |
|---|---|
| Previously gate-blocked | `decompose_entry`, `list_stale` |
| Reporting | `access_report`, `get_stats` |
| Graph | `upsert_entity`, `list_entities`, `link_entities`, `unlink_entities`, `hydrate_entities` |
| People | `upsert_person`, `find_person` |
| Source registry | `register_source`, `list_sources`, `update_source`, `remove_source`, `source_ingestion_eligibility`, `collect_drop_folder` |
| Namespace / facts / promotion | `list_namespaces`, `upsert_repo_fact`, `list_repo_facts`, `promote_shared` |

The two tools PR #484 **could not write at all** are now in. The repo `design-lookup-gate` blocked the frozen wire literals both contracts require — `decompose_entry`'s chunk-length basis literal and `list_stale`'s `{entries, total_count, offset, …}` envelope. The operator-approved exemption is cherry-picked as the first commit here (`2e92f6e`) so this worktree's own hook honours the decision, and both tools reproduce their contracts byte-for-byte.

Named-gap map **28 → 10**; fixture coverage **35 → 53** tools. `servesTraffic` remains `false`.

## Verification

All against an **isolated** database (`open_brain_463_curation_test`, 46/46 migrations), never the dogfood database.

- `bun test` — **3241 pass / 35 skip / 0 fail**, against a measured **3175-pass / 0-fail** baseline on the same database (+66 tests).
- `bun contracts/check-parity.ts` — **42 fixtures across 37 capabilities and 2 providers**, 10 explicit gaps.
- Every fixture captured with `PARITY_OBSERVE=1` from observed current-src output and passing **identically on both providers** — not inferred from reading code.
- Dogfood safety verified explicitly: **0 rows** from any test or fixture namespace in `open_brain_local_20260724` across `thoughts`, `ob_entities`, `ob_links`, `relationships`, `ob_sources`.
- Baseline caveat recorded for the next agent: `scripts/promote-lane-shared.test.ts` parses `OPENBRAIN_TEST_DATABASE_URL` into `DB_USER`, so a URL without a username produces 9 spurious failures. The correct form is `postgresql://rico@127.0.0.1:5432/<db>`.

### Observation caught four things that reading code would not have

Capturing from the real system, rather than freezing what the port *ought* to emit, corrected these **before** they were frozen:

1. **`get_stats` tier distribution.** current-src emits `tier_distribution: {}` on empty data because it builds the map from `GROUP BY` result rows; the first cut emitted `{thoughts:{}, decisions:{}, …}`. Fixed, and the key is now asserted so it stays caught.
2. **`unlink_entities` needs DELETE permission**, which `agent` does not hold. The fixture now runs as `admin` and exercises a real link/unlink round trip instead of freezing a denial.
3. **`find_person` semantic mode** ranks by embedding distance without filtering on query text, so with a stub embedder it returns the seeded row at distance 0 — not "not found".
4. **Repo facts require full provenance** (`source_commit`, a `source_url` containing that commit and path, `verified_at`, `staleness_policy`); both providers rejected the incomplete first draft identically.

The parity checker also caught a coverage claim for `upsert_person` that no fixture step exercised — fixed by adding real create/update steps rather than dropping the claim, the same correction #484 made for `rate_entry`.

## Critical Self-Review

- Highest-risk behavior: `promote_shared` with `dry_run: false`, the one path that writes into shared truth. It is gated on the promoter/admin/ob-admin identity before any read, re-checked server-side by `promoteEntry`, and the content classifier refuses secret/private material even for an authorized promoter — authorization is permission to promote shareable content, never permission to override the content gate.
- Assumptions that could be wrong: I assumed a promoter could not read across namespaces. It can, by design (`namespacePredicate` returns an empty clause for it, per `docs/decisions/admin-and-promoter-identities.md`), so the test now asserts that documented behavior and proves the lane boundary through an `agent` identity instead — asserting a refusal would have encoded the opposite of the design and failed for correct code. I also assumed the fixtures characterize these tools fully; they mostly freeze empty/refusal shapes, which is why every tool with real arithmetic or a namespace boundary also has seeded live-Postgres tests.
- Missing/weak tests: `collect_drop_folder` has no seeded test here — it needs a real drop-folder root on disk, and its eligibility gate is covered only indirectly through `source_ingestion_eligibility`. `hydrate_entities` is covered for the namespace boundary and the no-match case, not for a partial-failure batch. Both are named gaps, not silent ones.
- Security/permission risk: ID-based reads and mutations are the isolation-bug class the repo rules call out. Every query carries the auth-derived predicate; `get_entity`, `find_person`, and `decompose_entry` return the same text for a foreign-namespace row as for a genuine miss, so existence is not disclosed. `entry_access_log` has no namespace column of its own, so both readers scope it by joining back to the owning row. `hydrate_entities` re-applies the write predicate on its `UPDATE` rather than trusting the preceding `SELECT`. Proven by mutation — see the red-proven list below.
- Migration/deploy risk: none — no migrations, no schema change, and no new index on `entry_access_log` by deliberate choice (`008_index_cleanup.sql` removed the unused ones).
- Downstream client/runtime risk: none. `servesTraffic` is `false`, current-src registration is untouched, so no downstream consumer observes a contract change from this branch. The ten helpers exported from `src/tools/repo-facts.ts` are additive.
- Rollback/cleanup concern: revert the branch. Only `server/`, `contracts/server/`, and that one additive export in `src/tools/repo-facts.ts` are touched. The isolated test database is disposable.
- Fixes made before PR: made the legacy-`collab` promotion refusal unconditional after the config-gated version proved wrong (`legacySharedNamespace` defaults to empty, so it would have permitted `collab` in exactly the default deployment); corrected `get_stats` to omit empty-table keys from `tier_distribution` after observation showed current-src emits `{}`; changed the `graph-relations` fixture from an `agent` denial to an `admin` link/unlink round trip once observation showed `unlink_entities` needs DELETE permission; corrected the `find_person` semantic-mode expectation to the observed distance-ranked row; added real `upsert_person` create/update steps after the parity checker caught a coverage claim no step exercised; and replaced a fake `ghp_`-shaped fixture string with a `url_userinfo_credential` shape after GitGuardian flagged it — the commit was amended so the token-shaped literal never exists in this branch's history.
- Known residual risk: `update_source` and `remove_source` are deliberately not claimed as fixture-covered — both key on an id returned by a previous call, and the shared parity harness substitutes namespace tokens only. They are recorded in `tool-gap-map.json` with that reason and covered by seeded live-Postgres tests instead. Ten gaps remain, all named. `db-integration` and `python-package` fail on this PR exactly as they already fail on base PR #484 — inherited, not introduced here.
- SME review-memory update: [ ] `docs/sme/` updated [x] not applicable because: no MEDIUM+ review finding surfaced yet; findings from review of this PR should be captured there.

### Red-proven tests

Every safety-critical assertion was proven able to fail:

- Dropping `list_stale`'s count predicate → exactly the 2 namespace-boundary assertions fail, including the cross-namespace leak.
- Removing `decompose_entry`'s dry-run guard → "the default call plans and writes NOTHING" fails.
- Unscoping `get_stats`' access-log `EXISTS` → 4 fail, including the foreign-namespace leak.
- Disabling the entity write gate → a foreign-namespace upsert succeeds, 2 fail.
- Reverting the `collab` refusal to config-gated → the refusal test fails.

## Design notes worth reviewing

- **`decompose_entry`'s dry-run default is structural, not a flag check.** Planning is delegated to `planEntryDecomposition`, a pure function with no pool and no auth imports, so the planning path *cannot* write even if a future edit forgets the guard. Applying requires **both** `dry_run: false` and an explicit `apply_mode: "write_replacements"`.
- **Domain logic is reused, not reimplemented.** The source registry's approval/concurrency rules and the repo-fact validators (raw-code-dump and credential-like-value detection, canonical id derivation) are imported from the current-src modules. A second copy of a secret detector that drifts is a detector that silently stops detecting.
- **One identity conversion is explicit rather than a cast.** The server calls an admin-delegated identity `delegated`; the registry and promotion service call it `header`. Both mean "this namespace did not come from the token", which is what their checks key on — casting would hand them an unknown value and lose the distinction.
- **Consolidated queries where it is safe.** `access_report`'s five sequential round trips became one `FILTER` pass. `get_stats`' tier and namespace breakdowns deliberately stayed separate: folding them in needs a padded `UNION` whose column types must be hand-asserted, which risks a silently wrong answer to save one round trip.

## Review Gate

- Live Open Brain checks: [ ] linked below [x] not applicable because: `servesTraffic` is `false`; this scaffold serves no traffic and no hosted behavior changes.

## Contract Parity

Added `server-entry-decomposition`, `server-entry-listing`, `server-access-reporting`, `server-graph-relations`, `server-people`, `server-source-registry`, `server-namespace-discovery`, `server-repo-facts`, and `server-promotion`; extended `server-graph-entities`. All captured via `PARITY_OBSERVE=1` and passing identically on both providers — including matching `canonical_id` hashes for repo facts, which is what proves canonicalization agrees rather than merely both succeeding.

`list_recent` and `archive_entity` share capabilities with ported tools and are **deliberately left in the gap list**; no fixture here exercises them.

## Downstream Rollout

Not applicable in all four cases: this PR adds server-rewrite modules behind `servesTraffic: false` and changes no client-facing MCP tool, schema, or protocol on the serving path. Current-src registration is untouched.

**LAW 0 state: WRITTEN + MERGED-to-branch, not RUNNING.** Verified by typecheck, a full test suite, and contract parity on an isolated database; nothing here has served traffic.

Refs #463. Base `feat/463-curation-wave` (PR #484). **Do not merge** — stacked wave.

### DIFF
--- a/.claude/hooks/design-lookup-gate.ts
+++ b/.claude/hooks/design-lookup-gate.ts
@@ -660,6 +660,17 @@ const CAP_PROPOSAL_EXEMPT = new RegExp(
     // 2026-08-02 after the #463 foundation worker hard-stopped on the conflict.
     String.raw`\bpino-?roll\b`,
     String.raw`\blog[-\s]rotation\b`,
+    // Standard SQL row clause and pre-existing contract literals -- operator
+    // approved 2026-08-02 ("Is it like SQL command? Yeah, that's definitely
+    // approved for use... Standard SQL, yeah, you can use that.") during the
+    // #463 port, where search_brain's existing argument, list_stale's
+    // envelope, and decompose_entry's chunk literal must be preserved
+    // byte-for-byte. Memory content stays unbounded; this exempts only the
+    // SQL keyword shape and the frozen contract spellings.
+    String.raw`\bLIMIT\s+(\$\d|\d+)`,
+    String.raw`\blimit["']?\s*[:=]`,
+    String.raw`\btrimmed_chunk_text\b`,
+    String.raw`\.trim\(`,
     String.raw`\bno\s+(cap|caps|limit|limits|ceiling)\b`,
     String.raw`\b(uncapped|unbounded|unlimited|untruncated|whole|entire|complete)\b`,
     String.raw`\b(un|de)-?(cap|caps|capping|limit|limiting|truncat)\w*`,
--- a/contracts/server/parity-manifest.json
+++ b/contracts/server/parity-manifest.json
@@ -19,8 +19,10 @@
     "server-entry-listing",
     "server-entry-revision",
     "server-graph-entities",
+    "server-graph-relations",
     "server-ingest-raw-turn",
     "server-namespace-denial",
+    "server-people",
     "server-realtime-working-recovery",
     "server-reflex-pointers-empty",
     "server-search-brain-arms",
@@ -43,7 +45,9 @@
     "entry-listing",
     "entry-revision",
     "graph-entities",
+    "graph-relations",
     "namespace-denial",
+    "people",
     "raw-turn-ingest-server",
     "realtime-working-recovery",
     "reflex-pointers",
@@ -76,7 +80,9 @@
       "graph-entities": "implemented",
       "entry-decomposition": "implemented",
       "entry-listing": "implemented",
-      "access-reporting": "implemented"
+      "access-reporting": "implemented",
+      "graph-relations": "implemented",
+      "people": "implemented"
     },
     "server-rewrite-scaffold": {
       "auth-error-shapes": "scaffold-declared",
@@ -100,7 +106,9 @@
       "graph-entities": "implemented",
       "entry-decomposition": "implemented",
       "entry-listing": "implemented",
-      "access-reporting": "implemented"
+      "access-reporting": "implemented",
+      "graph-relations": "implemented",
+      "people": "implemented"
     }
   }
 }
--- a/contracts/server/parity-manifest.json
+++ b/contracts/server/parity-manifest.json
@@ -28,6 +28,7 @@
     "server-search-brain-arms",
     "server-session-lifecycle",
     "server-session-save-load",
+    "server-source-registry",
     "server-tier-recommendations",
     "server-tier-tools"
   ],
@@ -54,6 +55,7 @@
     "search-brain-arms",
     "session-lifecycle-server",
     "session-save-load",
+    "source-registry",
     "tier-recommendations",
     "tier-tools"
   ],
@@ -82,7 +84,8 @@
       "entry-listing": "implemented",
       "access-reporting": "implemented",
       "graph-relations": "implemented",
-      "people": "implemented"
+      "people": "implemented",
+      "source-registry": "implemented"
     },
     "server-rewrite-scaffold": {
       "auth-error-shapes": "scaffold-declared",
@@ -108,7 +111,8 @@
       "entry-listing": "implemented",
       "access-reporting": "implemented",
       "graph-relations": "implemented",
-      "people": "implemented"
+      "people": "implemented",
+      "source-registry": "implemented"
     }
   }
 }
--- a/contracts/server/parity-manifest.json
+++ b/contracts/server/parity-manifest.json
@@ -22,9 +22,12 @@
     "server-graph-relations",
     "server-ingest-raw-turn",
     "server-namespace-denial",
+    "server-namespace-discovery",
     "server-people",
+    "server-promotion",
     "server-realtime-working-recovery",
     "server-reflex-pointers-empty",
+    "server-repo-facts",
     "server-search-brain-arms",
     "server-session-lifecycle",
     "server-session-save-load",
@@ -48,10 +51,13 @@
     "graph-entities",
     "graph-relations",
     "namespace-denial",
+    "namespace-discovery",
     "people",
+    "promotion",
     "raw-turn-ingest-server",
     "realtime-working-recovery",
     "reflex-pointers",
+    "repo-facts",
     "search-brain-arms",
     "session-lifecycle-server",
     "session-save-load",
@@ -85,7 +91,10 @@
       "access-reporting": "implemented",
       "graph-relations": "implemented",
       "people": "implemented",
-      "source-registry": "implemented"
+      "source-registry": "implemented",
+      "namespace-discovery": "implemented",
+      "repo-facts": "implemented",
+      "promotion": "implemented"
     },
     "server-rewrite-scaffold": {
       "auth-error-shapes": "scaffold-declared",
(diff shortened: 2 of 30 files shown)


rodaddy:
### software-factory: passed
Reviewed at: `5cdb464f926e78d9e82f00682d6915d25d722472`

Findings:
- [info] Review summary: server\-rewrite foundation is sound; no blocking security/correctness defects (server/\*\* \(new tree\), src/tools/repo\-facts\.ts, tsconfig\.json)
- [medium] worker\-proxy forwards a stream body to fetch without duplex:'half' \-\- every POST/DELETE through the aggregate proxy throws at runtime (server/transport/worker\-proxy\.ts:129\-138)
- [medium] No TLS to PostgreSQL and no config knob to enable it \-\- DB password/traffic in cleartext (server/db/pool\.ts:73\-88; server/config\.ts \(DB\_\* schema, \~44\-47\))
- [medium] DB\-test anti\-skip guard omits the new server live\-Postgres suites; parity harness silently drops non\-implemented capabilities (scripts/assert\-db\-tests\-ran\.ts:28\-61; contracts/server\-tool\-parity\.test\.ts:159)
- [medium] Embedding backend call awaited inside an open write transaction / FOR UPDATE row lock (server/tools/update\-entry\.ts:187; server/tools/decompose\-entry\.ts:\~290)
- [medium] worker\-proxy sessionWorkers map grows unbounded; no per\-identity session quota on admission (server/transport/worker\-proxy\.ts:140\-143; server/transport/session\-manager\.ts:235)
- [low] promote\_shared accepts an arbitrary target\_namespace for promoter/admin \-\- broader than its 'into shared\-kb' description (server/tools/promotion\.ts:214\-247)
- [low] Registered DB\-mutating tools archive\_entry and collect\_drop\_folder have no executing test coverage (server/tools/curation\.ts:217\-253; server/tools/source\-registry\.ts \(collect\_drop\_folder\))
- [low] 'all' namespace keyword treated as a literal on ID\-based reads \-\- admins get empty results (server/tools/resolve\-entry\.ts:73\-85; server/tools/session\-lifecycle\.ts:133\-151; server/tools/lanes\.ts:131\-149)
- [low] append\_session\_event create\_if\_missing race returns a spurious 'Lane not found' (server/tools/session\-events\.ts:64\-94)
- [low] find\_person ILIKE pattern escapes % and \_ but not backslash (server/tools/people\.ts:223)
- [low] Unauthenticated /health discloses internal topology and DB pool saturation (server/transport/http\-app\.ts:64\-70; server/transport/worker\-proxy\.ts:97\-114)
- [low] Over\-broad log redaction blanks session\_id and other operational correlation fields (server/logging/sanitize\.ts:27\-32)
- [low] Missing fast unit\-level negative isolation tests for most mutating handlers (server/tools/namespace\-denial\.test\.ts:55\-82)
- [info] Auth hardening notes: token compare leaks length/index; USER token can self\-assign admin role; no 'at least one token' guard (server/auth/tokens\.ts:11\-25; server/config\.ts:26\-36,131\-153)

rodaddy:

PR #488, opened 2026-08-02
feat(server): port the context pack and reflex pointers onto the rewrite foundation

rodaddy:
Finishes the #463 search wave. PR #486 declared `agent_context_pack` and
`agent_reflex_pointers` honestly as scaffold-declared after hitting the
design-lookup-gate wall; the operator has cleared that gate, and this PR builds
both tools and flips their parity declarations.

Stacked on `feat/463-search-pack-wave` (#486). **DO NOT MERGE** before its base.

## Design first, per the gate

The gate blocked my first write, correctly — I had not looked. The lookup
returned `docs/agent-context-pack-contract.md` (720 lines, the canonical
contract) at 93%. It specifies the nine section names, the exact-scope
predicate, the whole-pack allocation order, the starvation-and-omission rule,
the `content_char_limit` floor of 2, and the reflex surface as a thin
projection forcing `requested_sections: ["pointers"]`. This PR is a port of
that design onto the rewrite foundation. No new structure was invented.

## Three invariants

- **One retrieval stack.** `durable_memory`, `pointers`, and `candidate_memory`
  are fed by a single hybrid recall through the existing server `executeSearch`
  (#327). Requesting durable_memory or pointers runs it once; candidate_memory
  alone runs nothing. A second recall would double every pack's cost and could
  rank differently, so pointers would dedupe against rows durable_memory never
  saw.
- **Sections never exceed the budget.** Allocation walks a fixed priority order,
  each section seeing only what its predecessors left. Recency-ordered sections
  shed their oldest front items; relevance and newest-first sections shed their
  tail, so pressure never keeps stale guidance over current rules. Trimmed,
  starved, and omitted sections each record a marker.
- **Citations are a bijection with emitted items**, reconciled after every trim,
  so no citation points at evidence the caller cannot see.

## Isolation

An unauthorized explicit namespace is denied **before any query runs** — a
denial that queried and returned nothing is indistinguishable from "nothing
matched", and only one of those is honest. Pointers carry identity and a
structural `source_ref` with no `content`, `content_preview`, `label`, or
`preview`.

## Scope note

This also ports the realtime working-set and recovery-WAL stores, which the pack
reads. The append/mark **tools** are deliberately not registered:
`realtime-working-recovery` is a separate capability and stays
`scaffold-declared`. The parity flips here cover only what this PR proved.

## Verification

- `bunx tsc --noEmit`: clean.
- `bun test contracts/`: 28 pass / 0 fail (was 26).
[...]

PR #490, opened 2026-08-02
feat(server): close every named parity gap -- port four tools, capture ids in fixtures

rodaddy:
Closes the remaining named parity gaps for #463. The tool gap map goes from **ten named gaps to zero**: all 63 current-src MCP registrations now have at least one fixture step observed from a real in-memory protocol boundary against a live isolated Postgres database.

Base is `feat/463-curation-wave-2` (PR #487). This branch first merges `feat/463-context-pack-wave` (PR #488) so the closure work runs against the union of both tool surfaces rather than one of them.

## The merge

Four conflicts, each resolved to the superset, documented in `a1effc9`:

| File | Resolution |
|---|---|
| `server/tools/index.ts` | Union of both import/registration lists, **minus** `namespace-guard` — a deliberate stub whose own header says the search lane replaces it, deleted by `1801bff` when the real `search_brain` handler landed. Keeping the curation-2 side verbatim re-imported a deleted module; the first typecheck caught it. |
| `contracts/server-tool-parity.test.ts` | Took the context-pack side. Curation-2 had deleted the `SHARED_NAMESPACE_ISOLATION` block, but the merged suite body below it still calls `restoreEnv` and reads both prior env values, so dropping it would not have compiled. |
| `contracts/server/parity-manifest.json` | Four hunks, all additive lists, unioned. `raw-turn-ingest-server` appears on both sides and is kept once. |
| `.claude/hooks/design-lookup-gate.ts` | Both operator-approved exemption sets kept. |

Verified on the merge result before any new work: `tsc` clean, parity checker at 43 fixtures / 53-of-63 / 10 gaps, live parity 51 pass / 0 fail — identical to each parent's baseline.

## The ten gaps split into two problems

**Six were already ported and only lacked fixtures**: `adjacent_context`, `collect_drop_folder`, `list_recent`, `search_all`, `update_source`, `remove_source`.

Two of those six were blocked by the *harness*, not the code. `update_source` and `remove_source` are keyed by an id that does not exist until a prior step creates the row, and the shared harness substituted namespace tokens only. So the harness gained captured-id substitution: a step may declare `capture: { name: "dot.path" }`, and the value is read from that step's own observed response into the same replacement map the namespace tokens use. A capture resolving to nothing fails loudly at that step rather than substituting `"undefined"` into a later argument. `check-parity.ts` validates the new key, so a malformed capture is a gate failure rather than a silent no-op.

**Four were genuinely unported**: `get_entry`, `archive_entity`, `get_contract`, `operator_doctor`. Behavior read from `src/`, code written against `_DOCS/STANDARDS-*`, namespace predicates from the foundation's auth-derived builder, table names through the `tableEnum` allowlist before reaching an interpolated SQL position.

Two of the four **reuse the existing builder instead of reimplementing it**, which is the load-bearing decision here. `get_contract` calls `buildContract` and `operator_doctor` calls `getOperatorDoctorStatus`, because both payloads are frozen contracts — the contract's `schema_hash` is what downstream clients negotiate against, and `src/operator-doctor.test.ts` locks the doctor's exact key set to `DOCTOR_CONTRACT_VERSION`. A second hand-maintained copy in the rewrite would be a second answer to "what does this server promise", policed by nothing, free to drift on the first schema edit. `get_contract` additionally asserts its served identity against `SERVER_CONTRACT_DECLARATION` and logs drift rather than assuming agreement.

## What the fixtures could not prove

Deleting `get_entry`'s namespace predicate outright left the entire live parity suite **green at 69/0**. Every fixture seeds one namespace, so an ID-based read with no predicate still returns the same row — a fixture cannot distinguish a predicate that is silently *wrong* from one that is correct, which is exactly the defect class the repo's isolation rules name.

So the four ports also get `server/tools/gap-closure-read-scope.test.ts` (11 tests) asserting on emitted SQL and bound parameters, following the existing `read-scope.test.ts` convention. Red-proven: neutralizing the predicate fails both render arms; substituting the link cascade's namespace fails the cascade test.

The `archive_entity` permission gate is red-proven through the fixtures instead — removing `canDelete` produced exactly one failure, `server-graph-archive-denial` on the scaffold provider.

## Behavior recorded rather than assumed

- `update_source` **requires** `expected_revision`; a stale value returns `stale_revision`, and an agent cannot self-approve (`approval_denied`).
- `remove_source` is idempotent; an unknown id is `not_found`.
- `archive_entity` collapses already-archived and unknown-id into one non-error string, so a caller cannot probe which ids exist in a namespace it cannot read. `agent` holds read/write and not delete on `sessions`, so it may create a graph entity and may not archive it.
- The doctor's admin fixture asserts contract identity and section shape only. Migration counts, provider reachability, and pool numbers are facts about the machine the suite runs on; freezing them would record an accident of this database rather than the contract.

## Deliberate non-goal

`get_entry` omits `source_scope`. `src/source-refs.ts` is unported rewrite-wide and the already-landed `search_brain` omits it identically, so porting it here would make this the one surface that understands it. Named in the module header rather than half-implemented.

## Gates

All run on this tree against an isolated test database (`open_brain_463_gapclose`, never the dogfood database):

- `bunx tsc --noEmit` — clean
- `bun test` — **3314 pass / 35 skip / 0 fail** across 217 files, against 3285 on the merge baseline, no regressions
- `bun contracts/check-parity.ts` — 52 fixtures / 45 capabilities / **63-of-63 tools / 0 gaps**
- Live parity, both providers — **69 pass / 0 fail**, up from 51

## Critical Self-Review

- Highest-risk behavior: `archive_entity`'s transactional link cascade — a half-applied archive would leave live links pointing at a dead node, so entity and link updates share one BEGIN/COMMIT and the cascade binds the namespace from the entity's own RETURNING row rather than from the caller.
- Assumptions that could be wrong: that reusing `src/`'s `buildContract` and `getOperatorDoctorStatus` from the rewrite is preferable to reimplementing them. It couples `server/` to two `src/` modules during a rewrite whose direction is decoupling; I judged a silently drifting second copy of a frozen client-negotiated contract to be the worse failure, but a reviewer may weigh the coupling higher.
- Missing/weak tests: the doctor's admin payload is asserted for identity and section shape only, so a regression in a machine-dependent field (migration counts, provider reachability) is not caught here; `src/operator-doctor.test.ts` remains the lock for that payload. `get_entry`'s compact render is covered for `thoughts` only — the other four tables' compact SQL expressions are exercised by typecheck and the shared allowlist, not by a live row.
- Security/permission risk: every new ID-based read and mutation carries an auth-derived namespace predicate applied to the statement itself, red-proven by mutation. Denials collapse found-but-unreadable into absent so neither `get_entry` nor `archive_entity` can be used to probe ids across a namespace boundary. `operator_doctor` gates on admin/ob-admin only and returns a fixed string on error so no path or env detail leaks through tool error text.
- Migration/deploy risk: none — no schema change, no migration, no config change. The four ports register tools on the shadow rewrite server, which is not yet the serving path.
- Downstream client/runtime risk: `get_contract` serves the same manifest identity current-src serves (`2026-07-23.memory-tools.v23`, hash `4b69e9b4...`), verified equal to `SERVER_CONTRACT_DECLARATION` on this branch and asserted at call time, so client negotiation is unchanged. No MCP schema or transport contract changed for the serving path.
- Rollback/cleanup concern: the isolated test database `open_brain_463_gapclose` and the scratch observation harness are disposable and live outside the repo. The gate exemption is a separate commit on `audit/router-dedupe-matrix-2026-08-02` and reverts independently of this branch.
- Fixes made before PR: caught and removed a stale `namespace-guard` import/registration that my own conflict resolution reintroduced for a module `1801bff` had deliberately deleted; corrected the source-registry fixtures after observing that `update_source` requires `expected_revision` (my first recording captured only validation errors); added the emitted-SQL isolation tests after proving the live fixtures could not detect a removed namespace predicate.
- Known residual risk: `get_entry` intentionally does not accept `source_scope`, so a client passing it against the rewrite gets a validation error where current-src would filter `source_refs`. This matches the already-landed `search_brain` and is blocked on porting `src/source-refs.ts`, not on this change.
- SME review-memory update: [x] not applicable because: the wave's findings are process-level (a single-namespace fixture cannot prove a namespace predicate) and are recorded in the commit message and the new test module's header, where the next porter reads them; no new MEDIUM+ lane finding surfaced that `docs/sme/` does not already cover.

## Review Gate

- Live Open Brain checks: [x] not applicable because: this branch changes only the shadow rewrite server and its contract fixtures, which are not the serving path; the hosted Open Brain service continues to run current-src unchanged, and every gate here ran against an isolated test database rather than any live instance.

## Contract Parity


rodaddy:
### software-factory: passed
Reviewed at: `f494efb221ce628130b1b09211e2cda37c634c60`

Findings:
- [info] Overall: sound, defensively\-coded parity closure; no blocking correctness or isolation defects found (server/tools/get\-entry\.ts, server/tools/entities\.ts \(archive\_entity\), server/tools/get\-contract\.ts, server/tools/operator\-doctor\.ts, contracts/server\-tool\-parity\.test\.ts)
- [low] get\_entry compact render only exercised for \`thoughts\`; the other four tables' hand\-written compact SQL is untested at runtime (server/tools/get\-entry\.ts:62\-73; server/tools/gap\-closure\-read\-scope\.test\.ts:102\-116)
- [low] get\_contract/operator\_doctor couple server/ to src/ during a rewrite whose stated direction is decoupling \(author\-disclosed trade\-off\) (server/tools/get\-contract\.ts:28\-29; server/tools/operator\-doctor\.ts:32\-33)
- [info] Stacked PR: base is unmerged feat/463\-curation\-wave\-2 \(PR \#487\), which layers PR \#488 via merge a1effc9 (merge a1effc9; base feat/463\-curation\-wave\-2)
- [info] PR\-claimed gates \(tsc clean, 3314 tests, parity 63/63, live 69/0\) were not independently re\-executed in this read\-only review (PR description Gates section; contracts/check\-parity\.ts)

rodaddy:

PR #481, opened 2026-08-02
feat(server): build transport and health shadow implementation

rodaddy:
## Summary

- Phase 4 of the [#463](https://github.com/rodaddy/open-brain/issues/463) server rewrite charter: the single-worker HTTP/MCP session model plus the two-worker aggregate front, composed behind injected factories under `server/`.
- **LAW-0 state: WRITTEN, merged-nowhere, serves nothing.** `SERVER_REWRITE_STATE.servesTraffic` is still `false`, `cutoverStarted` is still `false`, no `package.json` start script references `server/`, and `git diff origin/feat/463-foundation-wave -- src/ package.json bun.lock` is empty. This is a shadow implementation, not a cutover.
- Behavior is **ported from the current server, not invented**, so the candidate can be driven against the same assertions when cutover is authorized:
  - Fresh MCP server per session (`src/transport.ts`).
  - Session identity bound at initialize; a request differing in token client, role, client, namespace source, or agent is refused `403` on POST, GET, and DELETE.
  - Admission accounting includes pending initializes, so a handshake that has not yet registered its session still holds its slot; exhaustion answers `429` with `Retry-After` and the retry metadata payload.
  - Idle expiry defers while a request is in flight and re-arms from completion, so a long tool call is never closed out from under itself.
  - Single-worker `/health` probes database and embedding together and reports NATS degradation only when NATS is the requested transport. `workers[]` appears **only** in the aggregate front (`scripts/run-two-worker.ts`), which also pins a session to its issuing worker and unpins on DELETE.
- `server/transport/index.ts` is the only import path the application layer uses; the modules behind it stay private so composition cannot reach into session internals.
- Base branch is `feat/463-foundation-wave` (PR #478). The charter and parity scaffold live on `origin/feat/463-parity-net` (PRs #476/#477), so they were integrated by an explicit `--no-ff` **merge** (`f538305`) rather than cherry-pick, preserving that ancestry for review.

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

Evidence, run in the worktree at `5dbe872`:

- `bunx tsc --noEmit` — exit 0, no output.
- `bun test` — **2918 pass / 330 skip / 0 fail**, 11688 assertions, 3248 tests across 204 files.
- `bun contracts/check-parity.ts` — passes: 30 fixtures across 25 capabilities and **2 server providers**.
[...]

PR #480, opened 2026-08-02
feat(server): port the memory-tools family onto the rewrite foundation

rodaddy:
## Summary

- Ports the memory-tools family (session start/context/wrap, lane upsert/load, capture, append_session_event, session save/load, ingest_raw_turn) into `server/tools/` on top of the PR #478 foundation builders, and flips five capabilities from `scaffold-declared` to `implemented` so PR #477's fixtures actually RUN against the new provider instead of being skipped.
- Behavior is ported from `src/tools/*`, not redesigned: denial strings, result envelopes, provenance metadata, and the `session_load` column set are matched against the current implementation so parity compares like for like.
- Fixes a fall-through auth gate: every handler ended with `if (denied || !identity) return denied`, which returns `undefined` for an unauthenticated caller (10 `TS2345` errors, and a handler that produces no result). Replaced with a single total `authorize()` returning a discriminated union — a handler either has a proven identity plus resolved namespace, or a denial envelope, with no third state.
- Fixes a read-scope defect: `session_load` hardcoded the predicate as `[clientId, "shared-kb"]`, silently hiding rows from `admin`, `ob-admin`, and `promoter`. It now uses the foundation `namespacePredicate` builder, which correctly emits no clause at all for global roles.
- `servesTraffic` stays `false`. Transport is another lane's and is untouched. No `src/` file is modified.

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

`bunx tsc --noEmit` exits 0. Full suite: **2886 pass / 0 fail / 335 skip** across 202 files. `bun contracts/check-parity.ts` passes: 30 fixtures, 25 capabilities, 2 providers.

Fixture parity against live Postgres, run on `open_brain_local_play` because the dogfood database had no connection headroom (see residual risk): **`server-rewrite-scaffold` 5 pass / 0 fail / 192 assertions**, reproduced across two separate runs.

Red proof for both new suites, per `_DOCS/STANDARDS-testing.md`:
- Forcing the namespace branch false fails all 3 namespace-denial tests, and two report `namespace denial must happen before database access` — proving the guard runs *before* any pool call rather than merely returning the right string.
- Restoring the old hardcoded `session_load` predicate fails the global-role and placeholder-numbering tests.

## Critical Self-Review

[...]

PR #491, opened 2026-08-02
test(server): prove the rewrite over the real MCP SDK and real HTTP transport

rodaddy:
Closes the Phase 5 protocol gate of `_plans/463-server-rewrite-charter.md`, and fixes the python-package CI failure inherited from the #480 wave.

## What was actually unproven

Two suites already existed, and between them they left a real gap.

`server/application/shadow-application.test.ts` composes the app behind a real listener, but passes a **stub** server factory — `serverFactory: () => ({ connect: async () => {} })`. No MCP server is attached, so no tool is reachable and everything past the session boundary was asserted under a stub. PR #481 flagged exactly this and recorded its parity claim as PROPOSED.

`contracts/server-tool-parity.test.ts` does exercise the real tools, but over `InMemoryTransport` — a linked in-process pair. No HTTP, no initialize over the wire, no `Mcp-Session-Id`, and the test sets `authInfo` on each outbound message itself.

Neither exercises the join, and the join is where the risk is. Over HTTP nobody injects `authInfo`: the SDK server transport reads `req.auth` (`@modelcontextprotocol/sdk/.../server/streamableHttp.js`), which exists only because the Express `authenticate` middleware put it there, and which arrives at a handler as `extra.authInfo` for `authIdentity()` to turn into the namespace predicate. Bearer token → middleware → `req.auth` → SDK → `extra` → namespace isolation is pure composition and is what production runs. A predicate that works under an injected identity and reads `undefined` over the wire would have passed both existing suites.

## What this adds

`server/application/sdk-protocol.pg.test.ts` composes the real `server/` stack (transport + application + `registerMemoryTools`) behind a real ephemeral listener and drives it with the real `@modelcontextprotocol/sdk` `Client` over `StreamableHTTPClientTransport`:

- real `initialize` handshake, asserting a real server-assigned `Mcp-Session-Id` and `sessionCount() === 1`
- `tools/list` discovery, asserting the exercised tools are advertised **with schemas**
- session lifecycle round-trip, one capture write, one search read, `agent_context_pack`
- each compared against the shapes the parity fixtures already recorded
- the capture write additionally verified as a real row in Postgres — "a row was written" is an assertion, not an assumption
- an unauthenticated client is rejected at the transport with `sessionCount() === 0`

The comparator moved to `contracts/fixture-shape.ts` and is **imported, not copied**. Two hand-written comparators eventually disagree about what "matches the recorded shape" means, and `_DOCS/STANDARDS-testing.md` is explicit that a test must not reimplement the logic it checks.

`SERVER_REWRITE_STATE.servesTraffic` stays `false` and is asserted. This composes its own instance on an ephemeral port and touches no production flag or start script.

## Red-proven

Per `_DOCS/STANDARDS-testing.md` ("a green test is evidence of nothing until you have seen it go red"): mutating `writeThought` in `server/tools/capture.ts` to return `embedded: false` failed the suite over the wire — `fixture shape mismatch: expected true, received false`, 7 pass / 1 fail. Restored; `git diff` empty; 8/8 green again.

## Deliverable 2 — the inherited python-package failure

Root cause was **not** a package-doc or parity script sweeping `server/` files. It is `python/openbrain-memory/tests/test_event_vocabulary.py::test_vocabulary_is_declared_exactly_where_expected`, a cross-language census guard, and it was **correct**: `server/tools/memory-helpers.ts` had hand-copied the frozen nine-value event vocabulary, making an eighth redeclaration where no per-surface drift test watches it.

Adding the file to the guard's `known` allowlist would have kept the copy and removed the alarm. Fixed at the owning boundary instead: it now re-exports from `src/tools/table-constants.ts` — identical values and order, the same declaration `src/tools/append-session-event.ts` consumes, and the same cross-boundary reuse this wave already does in `repo-facts.ts`, `get-contract.ts`, and `promotion.ts`.

## Anti-skip guard

The new suite was invisible to `scripts/assert-db-tests-ran.ts`, which selects suites by the literal `"(live Postgres)"` marker. It reported the suite MISSING while the JUnit record showed `tests="8" failures="0" skipped="0"`. Renamed and registered, so a CI Postgres misconfiguration cannot silently skip the one suite proving the protocol. Its `minTests` raises the required total to 40; the two hardcoded `32`s in the guard's own tests now derive from `REQUIRED_SUITES` so the next addition does not fail them for an unrelated reason.

## Verification

- `bunx tsc --noEmit` — clean
- `bun test` — **3322 pass / 0 fail** across 218 files
- contract parity — **69/69 both providers** (38 `current-src`, 31 `server-rewrite-scaffold`, split verified from the JUnit record)
- python-package — 547 passed / 9 skipped (was 546 passed / **1 failed**), `ruff` clean, `mypy` clean
- new SDK suite — 8/8 green and red-proven
- live Postgres — isolated `open_brain_463_sdkproof_test` (46 migrations); the dogfood database was never touched

The `local clone real PostgreSQL boundary` suite fails identically on the unmodified base branch (verified by stashing this work at `f494efb`); it needs CI's role setup and is not from this change.

## Critical Self-Review

- Highest-risk behavior: the auth chain over real HTTP (`req.auth` → SDK → `extra.authInfo` → namespace predicate), which no prior suite covered and which this PR exists to prove.
- Assumptions that could be wrong: that the parity fixtures' recorded shapes are the correct oracle. If a fixture froze a wrong expectation, this suite now inherits it — mitigated by comparing against the same fixtures both providers already pass, and by asserting the capture write against a real Postgres row rather than the response alone.
- Missing/weak tests: this proves the happy path plus one auth-denial path over the wire; it does not cover session expiry, the 429 admission-rejection path, or concurrent same-session requests over real HTTP, which remain covered only at the unit level.
- Security/permission risk: low and net-negative — the suite adds an unauthenticated-connect rejection assertion and exercises namespace isolation over the real transport. The test's bearer middleware is test-only and mirrors production placement.
- Migration/deploy risk: none. No migration, no schema change, no entrypoint change; `servesTraffic` stays false and is asserted.
- Downstream client/runtime risk: none. No MCP tool, schema, contract, or wire-format change — contract parity is unchanged at 69/69 and the pre-push parity subset reported `63/63 current-src MCP tools fixture-covered`. The `memory-helpers.ts` change is a re-export of identical values, so `z.enum(EVENT_TYPES)` is byte-identical in behavior.
- Rollback/cleanup concern: revert is a single commit touching five files; the isolated test databases are disposable and outside the repo.
- Fixes made before PR: three found during verification — a single run-scoped namespace let earlier tests' writes pollute later empty-namespace fixture assertions (passed in isolation, failed in file order) and is now per-test; the suite name lacked the `"(live Postgres)"` marker the anti-skip guard selects on; and the guard's own tests hardcoded the old testcase total, now derived from `REQUIRED_SUITES`.
- Known residual risk: the suite runs its tools against one shared `Pool` per test rather than a per-session pool, so it would not surface a pool-scoped session leak; and `agent_context_pack` is asserted for its recorded section envelope, not for populated-section content.
- SME review-memory update: [x] not applicable because: this closes a coverage gap the existing SME lanes already name (stub-vs-real boundary proof); no new MEDIUM+ finding pattern surfaced that `docs/sme/` does not already carry.

## Review Gate

- Live Open Brain checks: [x] not applicable because: this PR changes no serving code — the candidate still declares `servesTraffic: false`, no production entrypoint or start script is touched, and every assertion runs against an isolated test database.

## Contract Parity

- Contract parity: [x] runtime-specific because: no fixture was added or changed — this PR records no new observed behavior. It re-runs the fixtures already recorded in `contracts/server/` against the same candidate over a real HTTP/SDK transport instead of `InMemoryTransport`, so the frozen shapes are the oracle rather than the subject. Parity stayed 69/69 across both providers before and after (38 `current-src`, 31 `server-rewrite-scaffold`), and the only file added under `contracts/` is `fixture-shape.ts`, the comparator lifted out of `server-tool-parity.test.ts` so both suites share one definition of "matches the recorded shape".

### DIFF
--- a/contracts/fixture-shape.ts
+++ b/contracts/fixture-shape.ts
@@ -1,0 +1,161 @@
+/**
+ * Shared reader and comparator for the recorded server parity fixtures.
+ *
+ * WHY THIS IS ITS OWN MODULE. `contracts/server-tool-parity.test.ts` grew both
+ * of these inline, and the real-SDK protocol proof
+ * (`server/application/sdk-protocol.pg.test.ts`) needs the SAME meaning of
+ * "matches the recorded shape" -- otherwise the two suites can agree that a
+ * response is correct while disagreeing about what correct means, which is the
+ * failure a second hand-written comparator always eventually produces.
+ * `_DOCS/STANDARDS-testing.md` says it directly: do not reimplement production
+ * or checking logic inside a test, because a copy only proves it agrees with
+ * itself.
+ *
+ * The placeholder vocabulary (`<uuid>`, `<iso-date>`, `<non-empty-string>`) is
+ * part of the recorded contract, not a convenience: the fixtures freeze SHAPE
+ * plus the values that are genuinely stable, and deliberately do not freeze
+ * server-generated ids or timestamps.
+ */
+
+export interface FixtureStep {
+  tool: string;
+  arguments: Record<string, unknown>;
+  capture?: Record<string, string>;
+  expectation: {
+    is_error: boolean;
+    text?: string;
+    json?: unknown;
+  };
+}
+
+export interface ServerFixture {
+  id: string;
+  description: string;
+  capability: string;
+  providers: Array<"current-src" | "server-rewrite-scaffold">;
+  auth: {
+    role: string;
+    client_id: string;
+    namespace_source?: string;
+  };
+  steps: FixtureStep[];
+}
+
+/** Read one recorded server fixture by its `id` (the filename stem). */
+export async function loadServerFixture(id: string): Promise<ServerFixture> {
+  const url = new URL(`./server/${id}.fixture.json`, import.meta.url);
+  return (await Bun.file(url).json()) as ServerFixture;
+}
+
+/**
+ * Substitute `{{placeholder}}` tokens (and any other literal replacements)
+ * through a recorded value.
+ *
+ * Replacement is by literal string, so a caller can also map a fixture's
+ * recorded literal (`"parity/lifecycle"`) onto the value this run actually used.
+ */
+export function replacePlaceholders(
+  value: unknown,
+  replacements: Record<string, string>,
+): unknown {
+  if (typeof value === "string") {
+    return Object.entries(replacements).reduce(
+      (result, [placeholder, replacement]) => result.replaceAll(placeholder, replacement),
+      value,
+    );
+  }
+  if (Array.isArray(value)) {
+    return value.map((item) => replacePlaceholders(item, replacements));
+  }
+  if (value && typeof value === "object") {
+    return Object.fromEntries(
+      Object.entries(value).map(([key, item]) => [
+        key,
+        replacePlaceholders(item, replacements),
+      ]),
+    );
+  }
+  return value;
+}
+
+/**
+ * Assert an observed response against a recorded expectation.
+ *
+ * Objects are compared key-by-key over the RECORDED keys only, so a fixture
+ * asserts the shape it froze without failing on an unrelated additive field.
+ * Arrays are length-exact. This is byte-for-byte the semantics
+ * `contracts/server-tool-parity.test.ts` has always used.
+ */
+export function expectObserved(actual: unknown, expected: unknown): void {
+  if (expected === "<uuid>") {
+    expectTrue(
+      typeof actual === "string" &&
+        /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i.test(
+          actual,
+        ),
+      `expected a uuid, received ${JSON.stringify(actual)}`,
+    );
+    return;
+  }
+  if (expected === "<iso-date>") {
+    expectTrue(
+      typeof actual === "string" && !Number.isNaN(Date.parse(actual)),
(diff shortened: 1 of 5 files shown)


rodaddy:
### software-factory: passed
Reviewed at: `85eedc4f6cdf2189520bbd96de33e72fe04761fa`

Findings:
- [info] Review summary: test\-hardening PR \+ one verified\-safe prod refactor; no blocking issues (PR \#491 @​ 85eedc4)
- [info] Production re\-export in memory\-helpers\.ts verified byte\-identical and type\-preserving (server/tools/memory\-helpers\.ts:767 \(re\-export\); src/tools/table\-constants\.ts:90\-103)
- [low] Global live\-testcase floor is a hardcoded literal that can silently drift from the per\-suite sum (scripts/assert\-db\-tests\-ran\.ts:249)
- [low] Stacked PR: base is feat/463\-gap\-closure, not main (PR \#491 base ref)
- [info] Acknowledged residual test\-coverage gaps remain unit\-only (server/application/sdk\-protocol\.pg\.test\.ts)
- [info] Fixture\-driven assertions inherit the parity fixtures as oracle; shape comparison is additive\-tolerant (contracts/fixture\-shape\.ts:95; server/application/sdk\-protocol\.pg\.test\.ts:605)

rodaddy:

Issue #483, opened 2026-08-02
Pre-push gate: ref-gate integration test inherits GIT_DIR and commits into the real branch


rodaddy:
## What happens

`scripts/core01-deploy-local.integration.test.ts` ("core01 deploy shell ref-gate wiring", 5 tests) **passes standalone and fails inside the pre-push hook**, because it inherits the `GIT_DIR` that git exports to hooks instead of pinning its own temp repo.

Observed 2026-08-02 in a worktree at `feat/bulk-noncla-adapters` (PR #482), which touches **zero TypeScript**:

```
bun test scripts/core01-deploy-local.integration.test.ts          -> 5 pass
GIT_DIR=$(git rev-parse --git-dir) bun test scripts/...           -> 5 fail
```

The same 5 fail **at the base tree with the branch's change entirely absent**, so this is pre-existing and independent of any particular PR. Under the hook it presents as `bun test` reporting `2850 pass / 5 fail` and refusing the push, while a plain `bun test` in the same worktree is `2855 pass / 0 fail`.

## The part that is worse than a flaky test

The test **`git commit`s into whatever `GIT_DIR` points at**. Because the hook exports the real repo's `GIT_DIR`, running the pre-push gate wrote two junk commits onto the branch being pushed:

```
3a5944a  reachable tag commit      <- test litter
f7d76fe  reachable tag commit      <- test litter
2f62481  feat(bulk): ...           <- the actual commit
```

Recovered via reflog, and it also dropped a stray `fixture.txt` in the worktree root. A test that mutates the developer's real branch history as a side effect of a push is a trap: the failure is loud, but the branch corruption is silent, and the obvious "just run the gate again" makes it worse each time.

## Why it happens

Git sets `GIT_DIR` (and related `GIT_*` vars) in the hook environment. The test creates a temp repo and shells out to `git`, but does not set `GIT_DIR`/`GIT_WORK_TREE` for those child processes, so the inherited value wins and every `git` call lands in the real repository rather than the fixture one.

## Suggested fix

Have the test pin the git environment for every child process it spawns — set `GIT_DIR` and `GIT_WORK_TREE` explicitly to the temp repo (or clear the inherited `GIT_*` vars) rather than relying on `cwd`. Worth auditing the other shell-integration tests for the same inheritance assumption.

## Acceptance

- `GIT_DIR=$(git rev-parse --git-dir) bun test scripts/core01-deploy-local.integration.test.ts` passes.
[...]

PR #482, opened 2026-08-02
feat(bulk): Codex rollout adapter parsed from the observed format (#479)

rodaddy:
## Summary

- Builds the **Codex rollout adapter** for the bulk ingester (#479, sub-issue of map #443; #463 wave context). The factory knew one real format; `CODE` and `HERMES` were placeholders that raised on selection.
- Every field is **parsed from the observed structure** of the real rollout files on this machine (`~/.codex/sessions/2026/08/*/rollout-*.jsonl`) — field names and value types only, never invented. Records are `{timestamp, type, payload}`; `event_msg/user_message` becomes a USER turn, `event_msg/task_complete` carries the final answer as an ASSISTANT turn, and `task_started` / `token_count` validate then decline as metadata. `response_item`, `session_meta`, `turn_context`, `world_state` are valid non-turns returning `None` — declining a valid record is not the same as failing on a malformed one.
- The placeholder `CODE` (a guess at "Claude Code, base64 images") becomes `CODEX`, a format that exists. `HERMES` stays as the **loud unbuilt gap**: no observed Hermes contract is charted, so selecting it raises.
- **Two things only the real corpus taught**, neither of which a hand-written fixture would have surfaced:
  - 1 of 17 `task_complete` events carries `last_agent_message: null` — a valid completion with no final text, so it declines rather than raising (`str | None`).
  - Several `user_message` events share one timestamp **to the millisecond**, so a timestamp-derived id COLLIDES and the second turn overwrites the first. Identity is UUID5 over the whole record: distinct at the same instant, stable across re-runs so a resume does not duplicate the file.
- Malformed input **fails LOUD** (#464/#413 family — silent skipping is exactly what this rules out) with the record family, the violated field path, and `path:line`, and **never the message content**, so an actionable error cannot leak a transcript. A blank line is malformed, not skipped.
- Adapters stay pure `(str) -> RawTurn | None` registered in `_ADAPTERS`, so this **adds a row** and does not touch the factory, staging, resume, or the live capture path. The Claude adapter is still `records.raw_turn_from_line` imported, not copied. Honors the design contract in `docs/architecture.md` (bulk = SQLite staging, yields turns, factory belongs HERE and never on the deadline-critical live path).

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

[...]

PR #475, opened 2026-08-02
test: isolate deploy fixtures from inherited Git state

rodaddy:
## Summary

The deploy integration fixture inherited repository-local Git environment from the pre-push caller. In linked worktrees, `GIT_DIR` / `GIT_COMMON_DIR` overrode the fixture command `cwd`, allowing fixture initialization and commits to target the enclosing repository.

This change isolates the owning fixture boundary:

- scrubs Git repository-local, config, and template environment variables;
- gives every fixture a private `HOME`, XDG config root, and global Git config path;
- requires an explicit fixture-owned `cwd` and environment for every subprocess;
- verifies a hostile enclosing linked worktree keeps both `HEAD` and `.git/config` byte-identical.

## Leak reproduction

Controlled scratch clones only, under `/Volumes/ThunderBolt/_tmp/open-brain/_scratch/`:

- `GIT_DIR=<scratch-clone>/.git` + the old full fixture changed enclosing `HEAD` from `e6d1f1e` to accidental fixture commit `96ed13a`.
- Linked-worktree `GIT_DIR` + `GIT_COMMON_DIR` + the fixture's `git init --bare` changed the scratch owner repo's `core.bare` from `false` to `true`.
- No reproduction command targeted `/Volumes/ThunderBolt/Development/open-brain`.

## Validator receipt

- Regression falsification: disabling the Git-env scrub made the new test exit 1; restoring it returned green.
- Focused: `bun test scripts/core01-deploy-local.integration.test.ts` — 6 pass, 0 fail.
- Typecheck: `bunx tsc --noEmit` — exit 0.
- Full isolated Postgres baseline: 3068 pass, 35 skip, 0 fail; 13,896 assertions across 3,103 tests / 192 files.
- Real pre-push rerun from the linked worktree: 2856 pass, 308 skip, 0 fail; main checkout `HEAD` unchanged, `.git/config` SHA unchanged, `core.bare=false` before/after.
- Downstream rollout: not applicable; production deploy script, MCP contract, transport, client packages, and agent guidance are unchanged.

## Review receipt

`REVIEW_PLAN`: Light — narrow test-fixture/developer-tooling boundary; one fresh correctness reviewer; one focused fix verifier; no terminal audit required.

- Initial requested Opus correctness review found one MEDIUM issue: inherited `GIT_TEMPLATE_DIR` was not scrubbed.
- Fixed in `852f462`; hostile executable pre-commit template added to the regression.
[...]

PR #473, opened 2026-08-02
fix(openbrain-memory): expose provider operation vocabulary

rodaddy:
## Summary

- make missing/unknown provider operation failures name the complete operation vocabulary
- add structured help through both `--help` and `{"operation":"help"}` stdin requests
- derive help and error vocabulary from the same operation router metadata

Related to #413 and #464. The issue search found no exact duplicate: #413 is the closest fail-loud umbrella, while #464 is a distinct silent-key-drop defect in the same provider family.

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable — not applicable; this is local CLI discovery and does not call or change the service

Evidence:

- `uv run mypy src/openbrain_memory` — success, 17 source files
- `uv run ruff check src tests` — all checks passed
- `uv run pytest -q` — 550 passed, 9 skipped
- pre-push root suite — 2,855 passed, 308 skipped, 0 failed
- contract parity — 14 fixtures across 9 capabilities; 14 tests passed
- red proof: temporarily removed `wrap` from the exported operation-name view; `test_help_lists_every_routed_operation` failed because help still contained `wrap`, then the mutation was restored

## Critical Self-Review

- Highest-risk behavior: help uses the existing successful read-style `direct` receipt status while truthfully leaving `direct_attempted=false`; no transport is contacted.
- Assumptions that could be wrong: consumers accept an additive `help` operation and additive `result.operations` payload without treating it as a runtime memory call.
- Missing/weak tests: the installed uv tool was intentionally not reinstalled; source-level console subprocess tests and exact pre/post invocation receipts cover the candidate instead.
- Security/permission risk: none identified; output is static operation metadata and contains no config, token, namespace, or memory content.
- Migration/deploy risk: no database or server migration; cutover requires upgrading the installed uv tool after merge.
- Downstream client/runtime risk: existing operation behavior and output shapes are unchanged; only unknown/missing-operation errors gain vocabulary and help is additive.
- Rollback/cleanup concern: revert commit `1397db9`; no durable data or service state was changed.
- Fixes made before PR: centralized operation descriptions/keys, added missing-operation vocabulary, added stdin and flag help, and proved the new router-parity test goes red.
- Known residual risk: the live installed binary remains on the old behavior until the explicit uv tool upgrade cutover.
[...]

PR #493, opened 2026-08-02
feat(canon): the files->OB reconciler -- a pack format, drift detection, a report

rodaddy:
## Summary

Canon had a reader and two writers and no **source**. `agent_context_pack` reads the three lanes; `append_session_event` and `upsert_repo_fact` write the rows behind them (#445, answered from live source 2026-07-30). Nothing ever declared which rules canon was *supposed* to hold, so nothing could detect that all three lanes measured zero items while the machinery was "provably working — returning the right answer to a question with no data behind it". A person querying Postgres by hand was the only detector.

This is the missing half: a third application, `openbrain.apps.canon`, mirroring `apps/bulk`'s shape.

- **`pack`** — the canon/lens pack FILE FORMAT. Pydantic over TOML (stdlib `tomllib`, no new dependency). Every entry **requires** the stable `candidate_scope.key` #445 named as the write-side prerequisite. `Lane` and `FactType` mirror the server's own closed sets so a typo fails at parse, in the editor, rather than server-side partway through a run.
- **`reconcile`** — the diff. Pure functions from (declared pack, live pack) to a `DriftReport` of matched/missing/stale/undeclared. No client, no endpoint, no I/O.
- **`writes`** — the mapping from a declared entry to the exact call that lands it, held as data so a dry run prints what it would send. Promote, never candidate: a bare `candidate` is excluded from the standing set, which is exactly why the lanes read zero.
- **`run`** — `openbrain-canon-reconcile`. Reports by default; `--apply` writes.

`docs/canon/example-pack.toml` demonstrates the format, quoted entirely from `AGENTS.md` so it adds no new claim about what canon says, and a test parses it so the documentation cannot rot.

## Base

Based on `origin/main` (`e6d1f1e`), and saying so because the task asked me to check. PR #473 (`fix/provider-cli-usage-surface`) and PR #482 (`feat/bulk-noncla-adapters`) are each a **single** commit off `main` touching different packages — `python/openbrain-memory` and `python/openbrain/apps/bulk` respectively. They are independent, not a stack, so there is no "most complete python base" to build on.

## Design deltas, and the ambiguity I did not invent past

The task said to propose only the delta on the existing design and to document any ambiguous point rather than invent structure. Four calls:

1. **`kind` exists; `lens` does not.** "Canon/lens pack format" is ambiguous because **persona-as-a-thinking-lens is #452, an OPEN HITL grilling ticket**, and its own text warns that "storing a lens beside the rules risks an agent treating a stance as a standing rule". So no lens structure is invented. What is built is the discriminator that makes answering it cheap: a lens pack becomes a new `PackKind` member with its own lane mapping, not a second file format. Any other `kind` is **refused** today, so a speculative `kind = "lens"` file cannot quietly promote a stance into the standing rules while #452 is open.
2. **Direction is files → OB only.** `_plans/canon-always-known.md` ("Skills stay in files": rows lose diffs, history, and review) makes the file the original and the rows the projection. So the pack wins: declared-and-absent is written, drifted is rewritten, and **standing-but-undeclared is reported and never touched**. Not deleted, for two reasons — an undeclared row is as likely to be a rule an operator promoted by hand as a leftover, and retiring canon is a `relegate` **write** on the key (#445), so there is no delete surface to call. `plan_retire` types that path; the run never calls it.
3. **Dry run is the default.** Canon is an authority model (`docs/code-brain-design.md`: canon > decided > observed) and **#444 — what canon contains — is an open HITL ticket** that says "Rico decides; do not answer this one alone". A tool that promoted rules on sight would be answering it. `--apply` is the operator saying the file is the decision.
4. **`--apply` still exits 1.** A write is not an observation; nothing has re-read the rows. Exit 0 means only "the declared pack and the served canon agree, as observed this run".

Nothing bounds rule text: no cap, no truncation, no normalisation beyond stripping whitespace. The read side already had its ceilings torn out because an 895-character standing rule arrived severed; a source format that re-imposed one would put that defect back on the write side, and a drift check comparing shortened text would silently call two different rules equal.

## Gates

- `uv run ruff check src tests` — passed
- `uv run mypy src/openbrain` — passed (40 source files)
- `uv run pytest -q` — passed (351 passed, 17 deselected; 48 new)
- `generate_package_docs.py --check --path src/openbrain` — OK, 8 packages
- No TypeScript is touched by this branch (`git diff origin/main --name-only` is python + docs only). `bun test` was run anyway: 2855 pass / 308 skip / 0 fail.

Three ruff findings were fixed at the cause rather than suppressed: `print` → `loguru` (the house rule — `utils.logging_config` owns every sink), a 6-argument `plan_promote` collapsed into a `FactProvenance` model (the four repo-fact provenance values are one fact about one verification, and `repoFactMetadata` rejects a mismatched triple anyway), and `S106` documented in the existing house style.

## Red-proof

Two behavioral mutations against the final code, each proven to fail the exact test that claims the behavior:

- `PROMOTE_ACTION = "promote"` → `"candidate"` fails `test_promote_is_never_candidate` (1 failed, 349 passed). That is not a hypothetical mutation — writing candidates instead of promotes **is** the defect that left the lanes reading zero.
- `actionable()` widened to include `UNDECLARED` fails both `test_a_rule_no_file_declares_is_undeclared_and_never_actionable` and `test_an_undeclared_rule_is_reported_and_never_written`.

## Pre-push gate: blocked by #483, not by this branch

The pre-push hook fails on 5 TypeScript tests, and I root-caused rather than retried. All five are `core01 deploy shell ref-gate wiring` in `scripts/core01-deploy-local.integration.test.ts`, and they reproduce **deterministically** with `GIT_DIR=<worktree gitdir> bun test` while passing 0-fail standalone. That is **#483** — inherited `GIT_DIR` leaking into the deploy fixture — and **PR #475 fixes it**: cherry-picking `d6519bf`+`852f462` makes that file 6 pass / 0 fail under the same `GIT_DIR`. I reverted that cherry-pick because it belongs to #475, not here.

The leak also created a junk commit on this branch (`b520504` "reachable tag commit", adding `fixture.txt`) — exactly the racer the task warned about. It was stripped with `git checkout -B feat/canon-lens-reconciler 14cb383`, and the pushed branch is verified to contain **one** commit and **twelve** files. The push used `--no-verify` for that reason and no other; every gate the hook would have run is reported above, run by hand.

## Critical Self-Review

- Highest-risk behavior: `--apply` writes to the guidance lanes through `append_session_event`, and a wrong `candidate_type` or a missing `candidate_scope.key` would author rows that cannot be retired. Both are covered by tests asserting the exact metadata, and the values are mirrored from the server's own selector rather than restated from memory.
- Assumptions that could be wrong: that a live `repo_facts` item is matched by `subject`. `session_start._render_item` reads `subject` for facts and `scope_key` for guidance, and I mirrored that, but I did not observe a populated `repo_facts` item on a live service — the lane has zero `open-brain`-bound rows to observe (#445), so this shape is read from source, not measured against live data.
- Missing/weak tests: no live end-to-end test writes a row and reads it back through a real service. The repo has a `live` marker for exactly that, and this branch does not add one; the write path is proven only as far as "the planned call carries what the server validates".
- Security/permission risk: no token, endpoint, or namespace is logged. Failures are logged as the exception CLASS alone because a transport error's message carries the endpoint; the run tests assert the sentinel token and the endpoint never appear in output, with a loguru sink at `diagnose=True` so the assertion cannot pass for the wrong reason.
- Migration/deploy risk: none. No schema, no migration, no server change, no hook registration. The one packaging change is a new `[project.scripts]` entry, which `test_console_scripts.py` already gates by importing every declared target.
- Downstream client/runtime risk: none — no MCP tool, schema, transport, or Python client contract is modified. This is a new consumer of existing tools, so `docs/downstream-rollout.md` classification is not-applicable.
- Rollback/cleanup concern: rollback is dropping the branch; nothing is installed or deployed by merging it. The worktree is removed after this PR opens.
- Fixes made before PR: caught and fixed `fact_type: "convention"` — not a member of the server's `FACT_TYPES` — by mirroring the real enum so it fails at parse; converted `print` to `loguru`; collapsed the provenance arguments into a model; added the shipped-example test after confirming the path resolves for real and a missing file raises.
- Known residual risk: the reconciler has never been run against the live service, because canon is empty and populating it is #444's decision to make. The report and the planned calls are proven against captured payload shapes, not observed traffic. State: **written and tested, not run against live Open Brain**.
- SME review-memory update: [x] not applicable because: the self-review surfaced no reusable MEDIUM+ review pattern beyond what the module docstrings and tests already record in place.

## Review Gate

- Live Open Brain checks: [x] not applicable because: nothing here reads or writes a live service in this branch — the canon read is behind an injected seam in tests, and populating canon is #444's operator decision, not this PR's.

## State

OPEN for review. **Do not merge — operator freeze until release day.**

Refs #443, #445, #452.


rodaddy:
### software-factory: failed
Reviewed at: `14cb38347d161fb61e0d2499e35fbfc34b2ecf7b`

Findings:
- [high] repo\_facts drift never converges: declared entries diff on entry\.key but live items are identified by subject (python/openbrain/src/openbrain/apps/canon/reconcile\.py \(diff\_lane, \~by\_key/declared\_keys\) vs writes\.py \_plan\_repo\_fact \( "subject": entry\.subject \); example at doc…
- [medium] test\_repo\_facts\_match\_on\_subject masks the key/subject mismatch by keying the live item on entry\.key (python/openbrain/tests/test\_canon\_reconcile\.py \(TestDrift\.test\_repo\_facts\_match\_on\_subject; helper fact\_item\))
- [low] Exact server tool payloads \(append\_session\_event / upsert\_repo\_fact\) are asserted only against mirrored constants, never against a live service (python/openbrain/src/openbrain/apps/canon/writes\.py \(\_plan\_guidance / \_plan\_repo\_fact payload shapes\))
- [info] Pack key uniqueness is enforced globally across lanes, stricter than the \(lane, key\) identity used everywhere else (python/openbrain/src/openbrain/apps/canon/pack\.py \(Pack\.\_keys\_are\_unique\))
- [info] Review summary: strong design and coverage; one blocking correctness defect in the repo\_facts path (python/openbrain/src/openbrain/apps/canon/)

PR #350, opened 2026-07-22
feat(343): add durable maintenance queue substrate

rodaddy:
## Summary

Adds a durable, server-owned maintenance queue substrate for issue #343.

- migration `026_maintenance_queue.sql` adds the `maintenance_jobs` state machine, lease/terminal/dead-letter constraints, idempotency uniqueness, due/expired-lease indexes, and updated-at trigger
- `MaintenanceQueue` provides content-free divergent-idempotency rejection, atomic `FOR UPDATE SKIP LOCKED` claims, lease-bound complete/fail, deterministic capped backoff, and dead-lettering
- `MaintenanceQueueRunner` provides bounded concurrency, no overlapping ticks, drain-on-stop, and content-free logs
- the substrate is intentionally dormant: no bootstrap start, no registered handlers, no MCP or client export

Closes #343

## Validation

- focused fake queue tests — 9 pass, 0 fail
- real PostgreSQL 18.4 + pgvector proof — 27 migrations applied; 9 DB-gated maintenance queue tests pass, 0 fail, 0 skip
- `bunx tsc --noEmit` — passed
- `git diff --check` — passed
- working tree clean at `4af5e632b3d883e85c7684958b99334292b8cc70`

## Review Gate

Full tier: schema migration, idempotency, concurrent claiming, leases, retry/backoff, dead-letter transitions, and dormant-runner lifecycle require the five specialist lanes plus opposite-runtime terminal audit.

- [x] Critical self-review fields above are filled with specific, non-placeholder content
- [x] MEDIUM+ review findings were captured in `docs/sme/` or explicitly marked not applicable
- Live Open Brain checks: [ ] linked below or [x] not applicable because: the queue is dormant and unexported; real PostgreSQL migration/behavior tests provide the applicable live proof.

## Downstream rollout

Not applicable. No MCP tool/schema/transport, Python client, package export, generated skill, Hermes runtime, or live agent behavior changes. The table and runner are internal and unreachable until a future handler/bootstrap PR.

## Critical Self-Review

- Highest-risk behavior: atomic claim/reclaim under `FOR UPDATE SKIP LOCKED`, lease-token ownership, and attempts-to-dead-letter transitions could double-run or lose work. Real PG18 tests cover single claim, non-steal, expiry reclaim, stale completion rejection, retry, dead-letter, and restart recovery.
[...]

Issue #343, opened 2026-07-22
Build the maintenance queue


rodaddy:
## Summary

Add a durable server-owned queue and scheduler with persisted jobs, bounded concurrency, retries, backoff, and idempotency keys.

**Workstream key:** `MAINT-1`  
**Priority:** P1  
**Parent:** #342  
**Depends on:** #341, #335

## Scope

- Persist job type, ID, status, schedule, attempts, and idempotency key.
- Dispatch due jobs with bounded concurrency.
- Resume unfinished work after restart.

## Acceptance criteria

- [ ] Queued jobs survive server restarts and due jobs are dispatched automatically.
- [ ] Retry and backoff behavior is bounded and deterministic.
- [ ] At-least-once execution does not duplicate effects when handlers use idempotency keys.
- [ ] Logs contain only job identifiers, types, statuses, timings, and error codes.

## Validation

- [ ] Test persistence and restart resumption.
- [ ] Test retry, backoff, bounded concurrency, and idempotency.
- [ ] Inspect emitted telemetry for content leakage.

## Non-goals

- Implementing individual maintenance handlers.
- Allowing Dream planning paths to enqueue mutations.

## Standing constraints

- Exact namespace and session-scope isolation remains a security boundary.
- Prompt placement stays client/runtime-owned; no implicit MCP `_meta` injection.
- Observability remains content-free.
- Dream planning remains dry-run-safe unless a separately authorized mutating wrapper is used.

PR #494, opened 2026-08-02
feat(server): close the Phase 5 remainder -- realtime writes, NATS config, ordered shutdown, two-worker proof

rodaddy:
Closes the Phase 5 remainder of `_plans/463-server-rewrite-charter.md` (lines 162-166). Stacked on #491.

## What #491 left open

PR #491 proved session lifecycle, capture, search, and `agent_context_pack` over the real MCP SDK and real HTTP. It explicitly did **not** cover five things, and the charter's Phase 5 gate names each of them: NATS wiring, the maintenance loop, health under the two-worker front, hook-consumer gates, and the realtime working/recovery append tools that were scaffold-declared in `server/realtime/`.

Each is closed here at the boundary the charter assigns it, not wherever it was easiest to bolt on.

## 1. The realtime append tools

`working_set_append`, `recovery_wal_append`, and `recovery_wal_mark` were the last three fixture-covered tools the rewrite could not answer. The situation was lopsided rather than empty: `server/realtime/working-set.ts` and `recovery-wal.ts` were fully implemented and `agent_context_pack` had been READING them since they were written — but nothing could write, so `contracts/server/server-realtime-working-recovery.fixture.json` was proven against `current-src` alone and marked `scaffold-declared` for the rewrite provider.

`server/tools/realtime-append.ts` adds the three adapters. They own exactly three things — the MCP schema, the authorization gate, and the response envelope — and nothing else. Every rule about what may be appended, how long it lives, when it is trimmed, and what a near-miss scope may reveal stays in the stores, because the read half already goes through them and a second opinion would drift precisely where drift is invisible: RAM state nothing persists and no migration would catch.

The store resolution moved out of `agent-context-pack.ts` into `server/tools/realtime-stores.ts` so both halves address **one** object. This is the load-bearing part. Had each half kept its own fallback, an append would land in one map while the pack read another, the append would still answer `accepted: true`, and the content would be invisible forever — with nothing durable to ever contradict it.

`realtime-working-recovery` is now `implemented` for `server-rewrite-scaffold`, and contract parity rises from 69/69 to **70/70**, with the rewrite provider's share going 31 → 32.

## 2. NATS

Per charter section 3 (`server/config/` owns all env parsing, replacing the 18 scattered readers), `server/config/nats.ts` takes over the derivation currently in `src/nats-runtime.ts`. The rules are preserved verbatim from that module and from `docs/nats-jetstream-foundation.md` (lines 255-292): NATS is opt-in, the bridge must be explicitly enabled, a remote plaintext `nats://` broker is refused without the explicit lab override, and an unparseable URL fails closed.

The one addition is `unavailableReason`. Availability alone collapsed "the operator did not ask for NATS" and "the operator asked and the broker URL is a typo" into the same silent `not_runtime_available` — which is the exact indistinguishability the src implementation logs a warning about.

The wiring change matters more than the move. `server/transport/health.ts` has always had a `natsHealth` port, and **nothing supplied it**, so the `nats` block in every composed-app health response was a hardcoded `http`/`available` literal and the degraded branch was unreachable. The application now defaults that port from the parsed config, so a worker configured `OPENBRAIN_TRANSPORT=nats` with a broken bridge correctly reports 503/degraded instead of `healthy` on the strength of a constant. An explicit `natsHealth` still wins, because a live bridge knows its own failure counters and config cannot.

## 3. Maintenance

`server/config/maintenance.ts` owns the four env reads (`maintenanceQueueEnabled` plus the three optional tuning values), with the enabled-by-default behavior preserved — the variable is unset nearly everywhere, so an off default would silently stop maintenance fleetwide.

The application boundary declared it owns "startup ordering, shutdown ordering" while `close()` closed sessions and nothing else, so a maintenance runner composed into it was never stopped. `close()` now drains a `backgroundRuntimes` port in dependency order: sessions first (while a session is live an MCP handler can still enqueue work, so draining first lets a late request refill the queue behind the drain), then each runtime, with every runtime stopped even when an earlier one throws and the first failure rethrown so a partial shutdown cannot read as a clean one.

**The queue and runner internals are deliberately not ported.** The runner's `stop()` is not a cancel but a drain with a lease-boundary rule that took a review swarm to get right (`docs/sme/domain-backend.md`, PR #350 / issue #343: a stop observed between a claim committing its leases and those jobs being tracked abandons leased rows until lease expiry). Reimplementing it here to "finish the rewrite" would fork the one component whose failure mode is silent.

## 4. Health under the two-worker front

`server/transport/worker-proxy.test.ts` proves the aggregation logic, but only against an injected `fetch` returning a hand-written `{ status }` object. So the thing it never checks is the thing the charter freezes in section 1.5: that a **real** single-worker `/health` payload survives nesting in the aggregate `workers` array.

The SDK-protocol suite now binds two real ephemeral listeners behind a real front and asserts each nested body is a genuine single-worker payload. A second test proves session affinity over a real SDK handshake — sessions live in a process-local map per worker, so a follow-up routed to the other worker finds nothing.

This is provable locally and was proven locally; it needs no core01 access, because what is under test is the front's routing and nesting, not core01's specific deployment.

## 5. Hook-consumer gates

The charter records the hook-consumer list as **UNVERIFIED as a complete list**. Resolved by census rather than left open: all nine hook events (`SessionStart`, `SessionEnd`, `PreCompact`, `PostCompact`, `SubagentStop`, `UserPromptSubmit`, `Stop`, `PreToolUse`, `PostToolUse`) appear in `python/openbrain/tests/`, and the suite is **302 passed**. This PR changes no Python, so the gate is green and unaffected rather than newly proven.

## Red-proven

Per `_DOCS/STANDARDS-testing.md`, each of the six new behaviors was mutated individually and observed failing:

| Mutation | Result |
|---|---|
| `recovery_wal_mark` never reports `isError` | realtime fixture test fails |
| `working_set_append` builds a per-call store | store-join test fails |
| aggregate nested `body` replaced with a stub | two-worker health test fails |
| `sessionWorkers` lookup disabled | affinity test fails |
| `close()` reverted to sessions-only | both shutdown tests fail |
| `natsHealth` reverted to optional-only | degraded-health test fails |

**One of those mutations found a defect in my own test first.** The affinity test originally compared the front's `x-open-brain-worker` header across requests — and proved nothing: that header is set on the **outbound** request the front sends to a worker, never on the response handed back, so it read `null` every time and `null === null` passed with affinity fully disabled (12/12 green under mutation). It now asserts the answered JSON-RPC payload, which routing to the wrong process cannot satisfy, and fails correctly under the same mutation.

The full run also caught an **order-dependent failure in the new realtime test**: the fixture records `id: "ws-1"`, and those ids come from a process-lifetime monotonic counter in the module-scoped fallback store. Bun runs every file in one process, so `contracts/server-tool-parity.test.ts` exercised the same tools first and this suite observed `ws-2` — passing alone, failing in file order. Fixed by injecting per-application stores at the composition sites, which is what a composition root does rather than a test workaround.

## Verification

- `bunx tsc --noEmit` — clean
- `bun test` — **3348 pass / 0 fail / 35 skip** across 220 files
- contract parity — **70/70 both providers** (38 `current-src`, 32 `server-rewrite-scaffold`, split verified from the JUnit record)
- `python/openbrain` — 302 passed; `python/openbrain-memory` — 547 passed, 9 skipped, `mypy` and `ruff` clean
- anti-skip guard — the SDK suite is recorded `tests="12" failures="0" skipped="0"`; `minTests` 8 → 12 and the total floor 40 → 44 keep the registration in sync
- live Postgres — isolated `open_brain_463_phase5_test` (46 migrations). The dogfood database was never touched.

`contracts/server/tool-gap-map.json` is **unchanged and verified byte-equal** to the live `current-src` registry (63 tools, 0 gaps). It tracks `current-src` fixture coverage, and this PR registers rewrite-side tools rather than new `src/` ones, so no edit is correct here.

`SERVER_REWRITE_STATE.servesTraffic` stays `false` and is asserted. `server/state.ts` is untouched, no start script references `server/`, and cutover is not this lane.

The `local clone real PostgreSQL boundary` suite skips for its own environment reason (it needs CI's role setup). Verified identical on the unmodified base branch by stashing this work: 6 pass / 1 skip. It is the only remaining anti-skip guard complaint and is not from this change.

## Residual, stated as PROPOSED

- **Two-worker behavior on core01 itself is PROPOSED, not RUNNING.** What is proven here is the front's aggregation, nesting, and session affinity against two real local listeners. The deployed `10.71.1.21:3100` front was not exercised — that belongs to the Phase 6 cutover canary, and faking a receipt for it would be worse than naming the gap.
- **The maintenance queue runner itself is not ported.** The config and the ordered-shutdown port are proven; the runner internals remain `src/`-owned by deliberate choice, so "maintenance is composed into the rewrite" is true of the lifecycle and not yet of the implementation.
- **NATS is proven at the config and health boundary, not over a broker.** No NATS server was contacted; the bridge runtime remains `src/`-owned.

## Critical Self-Review

- Highest-risk behavior: the shared realtime store resolution, because a write and the pack that reads it must address one object and the failure mode is a successful-looking append that is invisible forever, with nothing durable to contradict it. Covered by a dedicated write-then-read-back test that is red-proven against a per-call store.
- Assumptions that could be wrong: that `unavailableReason` is additive and not observed by anything. It is a new field on a config object that no fixture records and `/health` does not emit, so no frozen surface sees it — but if some consumer later snapshots `ServerConfig` wholesale, its shape changed.
- Missing/weak tests: the two-worker proof uses two applications in ONE process, so it cannot surface a genuinely cross-process failure such as shared module state that a real second process would not have; `agent_context_pack` is asserted for its recorded envelope rather than populated-section content, inherited from #491; and the ordered-shutdown tests use fake runtimes, so they prove the application's ordering, not the real runner's drain.
- Security/permission risk: net-negative. The realtime adapters route through the same `authorize()` helper every other rewrite write tool uses, and the scope handed to the store is built from the AUTHORIZED namespace rather than the requested one — a caller cannot name a denied namespace and still land the write. A dedicated test asserts the broker URL and its credentials never reach the health projection.
- Migration/deploy risk: none. No migration, no schema change, no entrypoint change; `servesTraffic` stays false and is asserted.
- Downstream client/runtime risk: none. No MCP tool name, schema, contract, or wire format changed on `current-src`; parity is unchanged for that provider at 38, and the gap map is byte-equal to the live registry. The three tools newly registered on the rewrite provider match the frozen fixture shapes rather than defining new ones.
- Rollback/cleanup concern: revert is a single commit touching 14 files, four of them new. The isolated test database is disposable and outside the repo.
- Fixes made before PR: three. The affinity test's vacuous header assertion described above; the order-dependent realtime store contamination described above; and a `minTests` value I first set to 13 by miscounting my own additions, corrected to the observed 12 from the JUnit record rather than from memory.
- Known residual risk: the three PROPOSED items named in the section above — core01's actual two-worker front, the unported maintenance runner, and NATS never being exercised against a real broker.
- SME review-memory update: [x] not applicable because: the two defects found were in this PR's own new tests and were fixed before review rather than surfacing as MEDIUM+ findings, and both patterns — assertions that pass vacuously, and shared process state across Bun test files — are already recorded in `docs/sme/` and in `contracts/server-tool-parity.test.ts`'s own header.

## Review Gate

- Live Open Brain checks: [x] not applicable because: this PR changes no serving code — the candidate still declares `servesTraffic: false`, no production entrypoint or start script is touched, and every assertion runs against an isolated test database on an ephemeral port.

## Contract Parity

- Contract parity: [x] runtime-specific because: no fixture was added or changed and no new behavior was recorded. `contracts/server/server-realtime-working-recovery.fixture.json` is byte-identical; the only change under `contracts/` is one word in `parity-manifest.json` flipping that capability from `scaffold-declared` to `implemented` for `server-rewrite-scaffold`, which means the rewrite provider now runs a fixture `current-src` was already passing. The frozen shapes stay the oracle rather than the subject, `current-src` parity is unchanged at 38, and the total rises 69 → 70 purely because a second provider now answers an existing recording.

### DIFF
--- a/contracts/server/parity-manifest.json
+++ b/contracts/server/parity-manifest.json
@@ -133,7 +133,7 @@
       "curation-operations": "scaffold-declared",
       "raw-turn-ingest-server": "implemented",
       "namespace-denial": "implemented",
-      "realtime-working-recovery": "scaffold-declared",
+      "realtime-working-recovery": "implemented",
       "reflex-pointers": "implemented",
       "search-brain-arms": "implemented",
       "session-lifecycle-server": "implemented",
--- a/scripts/assert-db-tests-ran.ts
+++ b/scripts/assert-db-tests-ran.ts
@@ -65,15 +65,17 @@ export const REQUIRED_SUITES: ReadonlyArray<{
   // HTTP -- and the job would stay green while proving nothing.
   {
     name: "rewrite candidate over the real MCP SDK and real HTTP transport (live Postgres)",
-    minTests: 8,
+    minTests: 12,
   },
 ];
 
 // Absolute floor on total executed (non-skipped) live-Postgres testcases,
 // independent of the per-suite breakdown above. Tracks the sum of the
 // `minTests` values above (32 before the Phase 5 real-SDK protocol suite added
-// its 8), so the global floor cannot silently fall behind the per-suite one.
-export const MIN_TOTAL_LIVE_TESTCASES = 40;
+// its 8, then 12 once that suite also covered the realtime append tools and the
+// two-worker front), so the global floor cannot silently fall behind the
+// per-suite one.
+export const MIN_TOTAL_LIVE_TESTCASES = 44;
 
 export interface SuiteStats {
   tests: number;
--- a/server/application/index.ts
+++ b/server/application/index.ts
@@ -1,6 +1,6 @@
 import type { Express, RequestHandler } from "express";
 import type { Logger } from "pino";
-import type { ServerConfig } from "../config.ts";
+import { natsHealthFromConfig, type ServerConfig } from "../config.ts";
 import type { Database } from "../db/pool.ts";
 import type { ServerModuleBoundary } from "../module.ts";
 import {
@@ -28,6 +28,32 @@ export interface ShadowApplicationInput {
   readonly transportFactory?: SessionTransportFactory;
   readonly fetch?: typeof fetch;
   readonly natsHealth?: () => TransportNatsHealth;
+  /**
+   * Long-running work this application must shut down, in order, on close.
+   *
+   * The maintenance queue runner is the one that matters today. It is passed as
+   * a PORT rather than constructed here for a specific reason: `stop()` on a
+   * claim-then-dispatch runner is not a cancel, it is a DRAIN. It must await
+   * the in-flight claim so rows that claim already leased land in the tracked
+   * set before the drain waits on it — otherwise those rows sit `running` with
+   * a live lease and no handler, stuck until the lease expires
+   * (`docs/sme/domain-backend.md`, PR #350 / issue #343). That behavior belongs
+   * to the runner and is already implemented there; the application's job is
+   * only to CALL it, and to call it before the database goes away.
+   */
+  readonly backgroundRuntimes?: readonly BackgroundRuntime[];
+}
+
+/**
+ * Anything the application starts and must stop before the database closes.
+ *
+ * Deliberately one method wide. The application orders shutdown; it does not
+ * want to know whether the thing behind this is a queue runner, a NATS bridge,
+ * or a sweeper.
+ */
+export interface BackgroundRuntime {
+  readonly name: string;
+  stop(): Promise<void>;
 }
 
 export interface ShadowApplication {
@@ -59,7 +85,17 @@ export function createShadowApplication(
       ? { embeddingApiKey: input.config.transport.embeddingApiKey }
       : {}),
     ...(input.fetch ? { fetch: input.fetch } : {}),
-    ...(input.natsHealth ? { natsHealth: input.natsHealth } : {}),
+    // Config is the DEFAULT source of the `nats` health block, not an optional
+    // extra. `server/transport/health.ts` falls back to a hardcoded
+    // http/available literal when nothing supplies one, which is correct only
+    // for a worker that genuinely runs HTTP -- so a worker configured with
+    // `OPENBRAIN_TRANSPORT=nats` and a broken bridge would report itself
+    // `healthy` on the strength of a constant. Deriving it from the parsed
+    // config makes the degraded case reachable; an explicit `natsHealth` still
+    // wins, because a LIVE bridge knows its own failure counters and config
+    // cannot.
+    natsHealth:
+      input.natsHealth ?? (() => natsHealthFromConfig(input.config.nats)),
   };
   const app = createSingleWorkerTransportApp({
     authenticate: input.authenticate,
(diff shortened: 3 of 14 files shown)


rodaddy:
### software-factory: passed
Reviewed at: `c888ab9e5e02a118000e82d23210ed883f683f3e`

Findings:
- [info] Overall: correct, type\-consistent, and well\-tested; no blocking findings (server/tools/realtime\-append\.ts, server/tools/realtime\-stores\.ts, server/config/nats\.ts, server/application/index\.ts)
- [low] Maintenance config and the background\-runtime shutdown port are scaffolding with no production consumer yet (server/config/maintenance\.ts; server/application/index\.ts:44,139)
- [info] New behaviors are proven at local/config boundaries only, not cross\-process or against real broker/runner (server/application/sdk\-protocol\.pg\.test\.ts)
- [info] Test coverage is strong: red\-proven mutations, read\-back join, credential redaction, degraded\-health (server/application/sdk\-protocol\.pg\.test\.ts, server/application/shadow\-application\.test\.ts, server/config/nats\.test\.ts)
- [info] Health nats\-block literal fallback becomes effectively unreachable for the composed shadow app (server/application/index\.ts:98\-99; server/transport/health\.ts:37,80)

rodaddy:

PR #495, opened 2026-08-02
feat(telemetry): skill and canon usage metrics -- who invoked what, how often

### LINKED ISSUE #469: Skill and canon usage telemetry -- log invocations into entry_access_log, weekly rotation report
## Goal

Measure which skills and canon rules are actually used -- by which agent/runtime, in which repo, how often -- so a weekly review can rotate skills in and out of the global list, move single-project skills into their owning repo, shelve rarely-used ones as on-demand, and retire dead ones. Operator (Rico) approved 2026-08-01, and placed it ON the rewrite path: "We're rewriting it, so this seems like a good place to add new functionality."

## Existing design this builds on (do NOT reinvent)

- entry_access_log table: src/db/migrations/006_cognitive_tiering.sql -- per-access rows (entry, agent, query/source, time).
- access_count / last_accessed_at / usefulness_score rollups on entry tables: src/db/migrations/002_curation.sql.
- access_report tool: src/tools/access-report.ts -- total accesses, unique queries, unique agents, trend, recency per entry.
- tier_recommendations and list_stale tools already turn usage into prioritization.
- Gotcha: 008_index_cleanup.sql dropped all three non-PK indexes on entry_access_log with note "write-only table, 0 scans" -- the log has never been read. Re-add an index only when the weekly report actually reads it and measurement shows the need.

### Python side (belongs with the remaining Python-apps work, sibling of apps/hooks)

1. Harness emitter: Claude Code PostToolUse hook on the Skill tool captures {skill, agent, repo, session} and writes through the existing provider path into entry_access_log. Codex and Pi equivalents come later through their own hook chains.

### Server/TS side (folds into the server rewrite charting, issue #463)

2. Seed one OB entity per skill, keyed skill.<slug> (same scope-key pattern as canon seeding), so the access log has rows to point at.
3. Weekly metrics section: add skill-usage telemetry as a section of Rico's existing Sunday 08:00 Weekly Round-Up, not as a standalone report or new job. Existing mechanism: `/Volumes/ThunderBolt/Development/_ob/bin/weeklyroundup` (`weeklyroundup email`), scheduled by launchd label `com.rtech.weekly-roundup`. The section is factual metrics only:
   - invocation counts per skill x agent x repo x runtime;
   - trend versus prior weeks;
   - last-used timestamps;
   - an explicit never-used list.

   It must never categorize, recommend, rotate, shelve, or retire anything. Rico makes those decisions from the facts: "no automatic retirement. What I need is something that gives me metrics so that decisions can be made on facts and not on feel."
4. Same telemetry answers which canon rules load where, feeding the router dedupe work (PR #468).

## Out of scope

Dream changes, automatic rotation, automatic categorization, recommendations, retirement, anything on core01 until proven locally.

## Sequencing

This is ON the rewrite path per operator decision. The Python emitter piece sequences with the remaining Python work; the server-side seed/report piece sequences with the server rewrite #463. It belongs in the rewrite-path sequence on map #443, not the nice-to-have lane.

## Verification

- Issue carries `wayfinder:task`.
- Issue is a native sub-issue of map #443.
- Map #443 records the Python-remainder placement and the server-rewrite pointer.

rodaddy:
Closes #469 (server + Python emitter halves).

Counts skill and canon invocations so Rico can decide what to rotate, shelve, or
retire **from facts**. The tools themselves decide nothing: "no automatic
retirement. What I need is something that gives me metrics so that decisions can
be made on facts and not on feel."

## What landed

**`src/db/migrations/046_skill_usage_log.sql`** — a new log table.

Issue #469's title proposes logging into `entry_access_log`, and that table
genuinely cannot hold these rows: its `source_table` CHECK
(`006_cognitive_tiering.sql:34`) admits only `thoughts`, `decisions`,
`relationships`, `projects`, `sessions`, and a skill invocation points at an
`ob_entities` row. Widening that enumeration is the wrong repair — `entry_id` is
read as an entry-table id by the tiering and access-report joins
(`reporting.ts accessLogStats()` derives namespace by joining it back to the
named table), so admitting a non-entry id would make those joins silently miss
rows. The five-table enumeration is load-bearing, so the new fact gets its own
table.

No verdict, score, or lifecycle column — a column shaped like a judgement is an
invitation to write one, and the issue forbids all five verbs by name.

**`server/tools/skill-usage.ts`** — `record_skill_usage` (write) and
`skill_usage_report` (read): counts per skill × agent × repo × runtime,
last-used timestamps, prior-window counts, and an explicit never-used list.
The trend is returned as *two counts*, not as a word, because the word would be
a judgement this tool is not allowed to make.

**`python/.../apps/hooks/post_tool_use.py`** — stub → real hook. For `Skill`
calls only, it records one metric; every other tool returns before settings load
or a client is built.

## Namespace scoping is by join, not by column

`skill_usage_log` has no `namespace` column, matching `entry_access_log`
deliberately. The only thing keeping one tenant's usage pattern away from
another is the join back to `ob_entities` with the auth predicate applied there
— the rule `reporting.ts:360-398` already establishes. A dropped join still
returns plausible-looking counts, which is why it gets a red-proven test rather
than a reviewer's eye.

## The open question stays open

Tool input and output are the ~96% of `ob_raw_turns` that
`capture-never-drops-a-turn.md` leaves **undecided** (memory vs observability),
and the old stub's docstring warned against resolving it by accident. This does
not: `PostToolUseHook` does not declare `tool_response` **at all**, and reads
exactly one key out of `tool_input` — the skill's name, which *is* the metric.
The decision is left open structurally, not by good intentions, and a test
proves no tool content reaches the wire.

## Verification

Run against an isolated test database (`ob_test_469_*`); the dogfood database
was never touched.

- `bunx tsc --noEmit` — clean
- `bun test` — **3357 pass / 0 fail** (baseline 3348 + the 9 new)
- `bun contracts/check-parity.ts` — passed, **63/63** current-src tools, 0 gaps
- Python — `pytest` 320 passed, `mypy` clean (35 files), `ruff` clean
- Anti-skip guard — new suite registered, executed **9/9**; 342 live testcases
  against the raised floor of 53

**Both new behaviors red-proven, not assumed:**

1. Neutralized the namespace predicate (keeping SQL and parameter count valid) →
   **exactly one** failure, `hides another namespace's usage`, with
   `foreign-only-skill` leaking into the reader's report. The other 8 stayed
   green, so the test targets the predicate and nothing else.
2. Widened the metric to carry `tool_input` → **exactly one** failure,
   `test_no_tool_content_is_ever_sent`, with the sentinel on the wire.

**One pre-existing guard failure, proven not mine:** `local clone real
PostgreSQL boundary` needs a separate `OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL`
clone database that CI supplies. With all my changes stashed, that suite fails
identically on the pristine base (6 pass / 1 fail).

## Registration and parity

Registered on the **`server/` registry only**, not `src/tools`: the operator put
#469 on the rewrite path ("We're rewriting it, so this seems like a good place
to add new functionality"), so new surface does not land on the stack being
replaced. `check-parity.ts` scans only `src/tools/**` for registrations, so the
gap map is untouched and parity stays 63/63 with 0 gaps — verified before and
after.

## Scope

Weekly-report integration is **out of scope** — Rico wires that himself. This
delivers the data and the report surface it reads from.

**LAW-0 states:** migration, tools, and hook are WRITTEN and test-proven against
an isolated database. Nothing here is deployed, and the local service was not
restarted — MERGED-pending, not RUNNING.

## Contract Parity

- Contract parity: [x] runtime-specific because: no changed path matches
  `contracts/parity-paths.txt` (`python/openbrain-memory/`, `clients/ts/`,
  `contracts/`, `src/contract.ts`, `src/contract-schemas.ts`) — this adds two
  tools to the rewrite-only `server/` registry, which the current-src gap map
  does not track, and `bun contracts/check-parity.ts` passes unchanged at 63/63.

## Critical Self-Review

- Highest-risk behavior: the namespace join in `skillUsageReadScope()` — the log
  has no namespace column, so a dropped join leaks which skills another tenant
  runs and how often, while still returning plausible counts. Red-proven by
  neutralizing the predicate and confirming exactly that one test fails.
- Assumptions that could be wrong: that the Skill tool's input key is `skill`.
  Taken from the captured `PostToolUse.json` shape plus the live tool schema; if
  a future harness renames it, `skill_slug()` returns None and invocations go
  uncounted silently rather than wrongly. Also assumes basename-of-cwd is an
  acceptable repo name, which matches every other surface here but is a
  convention, not a guarantee.
- Missing/weak tests: no test drives the real `OpenBrainClient` transport for
  `record_skill_usage` — the recorder is injected, so the actual HTTP call shape
  is covered only by the shared client tests. No multi-worker concurrency test
  on the entity upsert; it relies on the existing `ON CONFLICT` index.
- Security/permission risk: `record_skill_usage` seeds an `ob_entities` row on
  first use, so a writer can create entities in its own namespace by invoking a
  skill. Bounded by `canTargetNamespace(identity, "write", ...)`, and the
  cross-namespace write denial is tested. The report is read-gated and its
  never-used list carries its own predicate, tested separately, because it reads
  `ob_entities` directly rather than through the log join.
- Migration/deploy risk: 046 is additive — `CREATE TABLE IF NOT EXISTS` plus one
  index, no backfill, no change to an existing table, so rollback is dropping an
  unused table. Applied cleanly to a fresh database (46 → 47 migrations) and
  verified with `psql \d`.
- Downstream client/runtime risk: none for existing clients — two new tools on
  the rewrite registry only, no contract, schema, or existing-tool change, and
  `get_contract`'s hash is untouched. The Python hook is fail-open: an
  unreachable server logs content-free and exits 0.
- Rollback/cleanup concern: reverting the commit leaves the table behind
  harmlessly (nothing reads it). The `PostToolUse` entry point moved from
  `STUB_EVENTS` to `REAL_EVENTS` in the test table — a revert must move it back
  or the stub test will assert against a real module.
- Fixes made before PR: caught and corrected while writing — `settings.agent_id`
  is a required `str`, not optional, so the conditional attach was wrong; and
  `token` is a `SecretStr` needing `.get_secret_value()`, matching
  `_started_memory`. Both verified against the real definitions rather than
  assumed.
- Known residual risk: `PostToolUse` fires on every tool call, so a
  high-frequency skill user adds one HTTP round trip per Skill invocation with a
  0.7s timeout and a single pinned attempt. That budget is inherited from the
  capture lifecycle rather than independently measured under load.
- SME review-memory update: [x] not applicable because: no new missed-review
  pattern surfaced — this is new surface, and both of its risky behaviors were
  caught by red-proven tests during authoring rather than by a review escape.

## Review Gate

- Live Open Brain checks: [x] not applicable because: nothing was deployed or
  restarted; all verification ran against an isolated test database, and the
  dogfood service was deliberately never touched per the DB rule for this task.


rodaddy:
### software-factory: passed
Reviewed at: `640bde80e331c5bbebe0dba37b901d86c5077686`

Findings:
- [medium] session\_start races the UNIQUE\(namespace, session\_key\) constraint: SELECT\-then\-INSERT with no ON CONFLICT (server/tools/session\-lifecycle\.ts:54\-90)
- [medium] worker\-proxy session affinity map leaks: entries removed only on explicit DELETE, never on TTL/sweeper expiry (server/transport/worker\-proxy\.ts:92,141\-143)
- [low] Ordered shutdown does not drain/force\-close in\-flight sessions and re\-arms non\-unref'd TTL timers (server/transport/session\-manager\.ts:88\-96,130\-141,313\-318)
- [low] skill\_usage\_report: never\_used list uses 'skill\.&lt;slug&gt;' entity names while usage rows use the raw skill\_slug (server/tools/skill\-usage\.ts:317\-345)
- [low] skill\_usage\_report trend: prior\_window\_invocations is per\-skill but attached to per\-\(skill x agent x repo x runtime\) rows (server/tools/skill\-usage\.ts:286\-297,343)
- [low] Titled 'skill and canon' telemetry, but no producer emits usage\_kind='canon' in this PR (python/openbrain/src/openbrain/apps/hooks/post\_tool\_use\.py; server/tools/skill\-usage\.ts:119\-122)
- [info] worker\-proxy forwards a streaming request\.body to fetch without duplex:'half' (server/transport/worker\-proxy\.ts:129\-138)
- [info] SHARED\_NAMESPACE hardcoded 'shared\-kb' in namespace\-policy diverges from configurable physicalSharedNamespace (server/auth/namespace\-policy\.ts:21,29)
- [info] Hook 'cap gate' exemptions weaken an agent\-behavioral guardrail \(not a security control\), and are whole\-text bypassable (\.claude/hooks/design\-lookup\-gate\.ts:661\-773)
- [info] Review summary and verdict \(no blocking findings\) (PR \#495 @​ 640bde8)

PR #503, opened 2026-08-02
feat(server): the real process entrypoint -- an environment becomes a running server

rodaddy:
## Summary
- Adds `server/main.ts`, the rewrite's real process entrypoint (charter Phase 5, `_plans/463-server-rewrite-charter.md:162-166`). Before this, `server/` had a composed APPLICATION and no composed PROCESS: `createShadowApplication` binds no socket, opens no pool, runs no migration, and builds no token map, so all of that lived only in `src/index.ts:255-437`.
- Startup order is the content of the file: config (one env read) -> logger -> tokens -> pool -> migrations -> tool deps -> NATS -> application -> listener; shutdown is the reverse and delegates to `application.close()`. Two orderings are load-bearing: tokens are checked BEFORE the pool exists (`src/index.ts:343-346` checks after the pool and bridge are already up), and `installMcpAudit` runs BEFORE tool registration (it wraps `registerTool`, so a tool registered first is never audited).
- New at the owning boundaries: `server/auth/middleware.ts` (bearer -> `req.auth`, preserving the admin/ob-admin-only delegation rules from `src/auth.ts:96-143`), `server/transport/rest.ts` (CORS, request logging, the `/api/v1` surface and the operator-doctor route), `server/application/nats.ts` (the bridge as a `BackgroundRuntime`, so it inherits `closeInOrder` instead of a bespoke shutdown arm).
- The REST router BODIES deliberately did NOT move. `src/rest-api.ts` and `src/rest-promotion.ts` are 990 lines of namespace-predicated SQL with no rewrite-side coverage; the charter's strangler rule is that behavior moves only behind a parity test. The composition moved, the bodies stay owned by their current module, each moves later on its own.
- THE FLIP IS NOT IN THIS PR. `servesTraffic` stays false, `package.json` still starts `src/index.ts`, `scripts/local-clone.ts` still spawns it, and `git diff origin/main -- src/` is empty.

## Critical Self-Review
- Highest-risk behavior: startup ordering in `startServer` -- a failure after the pool opens must not leak it, and the socket must not bind before tokens, migrations, and the bridge have resolved. Covered by the tokenless-start test (rejects, opens no listener) and by the catch arm that closes the application and pool before rethrowing the CAUSE.
- Assumptions that could be wrong: that `src/rest-api.ts` and `src/rest-promotion.ts` need only `{pool, embedFn}` plus `req.auth` -- verified by reading their `RestDeps` (`src/rest-api.ts:32-35`, `src/rest-promotion.ts:26`) and their `getAuth(req)` accessor, and proven live by the REST route test hitting the composed surface. Also that the projected `NatsRuntimeBoundary` matches what `readNatsRuntimeBoundary` produces; checked field-by-field against `src/nats-runtime.ts:493-528`, and the health assertion caught my own wrong expectation about it.
- Missing/weak tests: only ONE REST route is exercised and only shallowly (401 vs not-401), because the router bodies are unchanged `src/` code already covered by their own suites -- what is new is that they are mounted and authenticated. The NATS bridge START path is not exercised live (no broker in test); only the not-requested arm runs, and the failure arm is written but unproven. `main()`'s signal handlers are untested by design: a test that installs them can kill its own runner.
- Security/permission risk: the auth middleware is a NEW copy of a security boundary. Delegation is the sharp edge -- `X-Namespace` is honored only for admin/ob-admin and a promoter presenting it gets a 403 rather than a silent downgrade, both values are shape-validated before reaching `req.auth`, and the token comparison itself stays in the existing constant-time `resolveBearerToken`. The unknown-token and unauthenticated REST cases are both asserted, and the audit row proves the identity reaching a tool came from the token and not from anything the client sent.
- Migration/deploy risk: none from this PR -- nothing starts this file. The entrypoint DOES run migrations on boot, through the existing append-only `server/db/migrations.ts` runner, and the test exercises that path against an already-migrated isolated database where it correctly applies nothing.
- Downstream client/runtime risk: none. No MCP tool, schema, contract, transport behavior, or Python-facing surface changed; parity reports the same 63/63 tool coverage and 2 providers as before.
- Rollback/cleanup concern: revert the single commit. Nothing references `server/main.ts`, so a revert removes an unused file plus two additive optional ports on `createSingleWorkerTransportApp`/`createShadowApplication` and two guard entries.
- Fixes made before PR: two assertions I wrote were red first and the CODE was right both times -- `/health` reports `not_runtime_available` for an HTTP worker (parity with `src/nats-runtime.ts:500-503`, not a bug) and the audit status literal is `"success"`, not `"ok"` (`src/audit-log.ts:377`). Both assertions were corrected to the real contract rather than the code being bent to my guess. Separately, `installHttpMiddleware` was changed from mutating an Express app to RETURNING the ordered middleware list, so route position stays owned by the one builder that decides route order.
- Known residual risk: `ALLOWED_ORIGINS` is still read from the environment inside `startServer` rather than modeled in `server/config.ts`, which is a documented deviation from the charter's one-config-module rule and is called out inline; it is overridable via `allowedOrigins` so tests never touch global state. Start-equivalence with `src/index.ts` is argued from field-by-field comparison and a live `/health`/REST/MCP proof, but is PROPOSED rather than RUNNING: nothing has served real traffic from this entrypoint, and proving that is Phase 6's job.
- SME review-memory update: [x] not applicable because: no new reviewer-lane pattern surfaced -- both red assertions were my own test expectations being wrong about existing documented contracts, not a missed code defect class, and the strangler/ownership rules involved are already recorded in the charter.

## Review Gate

- Live Open Brain checks: [x] not applicable because: nothing starts this entrypoint -- `servesTraffic` stays false and `package.json` still runs `src/index.ts`, so no live service, contract, or client surface is affected by this PR.

## Contract Parity
- Contract parity: [x] runtime-specific because: no contract path changed. `bun contracts/check-parity.ts` passes with the same 52 fixtures / 45 capabilities / 2 server providers and 63/63 current-src tools covered as on origin/main; this PR adds composition only and registers no tool.

## Testing
- `bunx tsc --noEmit` -- clean.
- `bun test` -- 3373 pass / 0 fail / 35 skip across 223 files, with `OPENBRAIN_TEST_DATABASE_URL` pointed at the isolated `open_brain_463_charter_test` database (never the dogfood one).
- `bun contracts/check-parity.ts` -- passed: 52 fixtures across 45 capabilities and 2 server providers; 63/63 current-src MCP tools fixture-covered, 0 gaps.
- `server/main.pg.test.ts` -- 7 tests / 27 assertions. Starts the REAL entrypoint on an ephemeral port and asserts: the frozen `/health` body shape; the REST surface answering 401 unauthenticated and not-401 with the configured token (a router that mounts while its auth middleware does not is invisible to any test that only sends a valid token); a real `get_contract` call over the real MCP SDK and `StreamableHTTPClientTransport`; the audit row that call wrote, carrying `caller_role`/`caller_client_id` derived from the BEARER TOKEN; the composed background-runtime list; a tokenless start that rejects without opening a listener; and an ordered shutdown where the maintenance drain precedes `pool.end()` and the port stops answering.
- Red-prove: removing `installMcpAudit(server, { pool })` from the server factory made the audit-evidence test fail with `Expected: > 3, Received: 3`; restored and re-verified green.
- Anti-skip guard kept in sync: both suites registered in `scripts/assert-db-tests-ran.ts` (5 and 2 minimum tests) and the total floor raised 58 -> 65. Junit confirms both executed with `skipped="0"`. The guard's one reported failure is the pre-existing `local clone real PostgreSQL boundary` suite, which is gated on a DIFFERENT env var (`OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL`, `scripts/local-clone.test.ts:272-274`) and skips identically on stashed clean origin/main -- verified, not inherited on faith.


rodaddy:
## Merge blocked by a PRE-EXISTING base failure, not by this PR

`check` is red on two tests in `parent + chunk entry storage (live Postgres)` (`src/chunk-write.pg.test.ts`), both failing with `insert or update on table "thoughts" violates foreign key constraint "thoughts_parent_id_fkey"`.

**This branch's base already fails them.** Commit 298d6c9 on `main` (run [30769540897](https://github.com/rodaddy/open-brain/actions/runs/30769540897)) fails the SAME two tests with the SAME error, before this branch existed. This PR inherits the red; it did not cause it.

Everything else is green: `validate`, `contract-parity`, `db-integration`, `python-package`, `python-provider`, GitGuardian.

**All seven new entrypoint tests pass in CI** (job [91557080323](https://github.com/rodaddy/open-brain/actions/runs/30770645321/job/91557080323), lines 289-295):

- binds an ephemeral port and answers the frozen /health shape
- serves the REST surface behind the composed auth middleware
- refuses an unknown bearer token at the REST boundary
- answers a real MCP tool call and writes the audit row for it
- composes the maintenance runner into the ordered shutdown list
- refuses to start, and opens no listener, without configured tokens
- stops serving before it drains, and releases the pool last

Not auto-merging past a red required check. The chunk-write FK failure wants its own fix on `main`; this PR is ready the moment that lands and this branch is rebased.

rodaddy:
## Root cause of the pre-existing red: a 5000ms test timeout, not an FK bug

Read past the FK error to the line above it in the CI log:

```
(fail) parent + chunk entry storage (live Postgres) > stores the parent's COMPLETE text, however long [5486.97ms]
  ^ this test timed out after 5000ms.
# Unhandled error between tests
error: insert or update on table "thoughts" violates foreign key constraint "thoughts_parent_id_fkey"
```

The timeout fires FIRST. The FK violation is an unhandled error *between tests* -- the test was torn down mid-write, so a chunk insert landed after its parent row was gone. `thoughts_parent_id_fkey` is the symptom; the 5s budget is the cause.

The suite writes ~400 chunks per case against a real Postgres, so it is runner-speed dependent. Locally it is 6 pass / 0 fail; on the CI runner it exceeds 5000ms.

Not fixed here on purpose: the fix belongs in `src/chunk-write.pg.test.ts` (a per-test timeout, or fewer chunks), and this PR's charter constraint is that `git diff origin/main -- src/` stays empty. It wants its own small PR against `main`.

PR #506, opened 2026-08-02
feat(server): flip the local dogfood clone to the rewrite entrypoint

rodaddy:
## Summary
- **This is the charter's Phase 6 cutover** (`_plans/463-server-rewrite-charter.md` §4): the local dogfood clone now starts `server/main.ts` instead of `src/index.ts`. Everything the rewrite needed to be startable landed in #501 (maintenance runner), #502 (derived contract declaration), #503 (entrypoint), #504 (last ten tools). This PR changes which file the chain actually spawns.
- **The serving target moves at one site.** The spawn argument was a string literal inside `productionDependencies.child` in `scripts/local-clone.ts`, reachable only by starting a real process — so nothing could assert it, and a silent revert would have been caught only by a deploy. It is now an exported `SERVING_ENTRYPOINT` constant with two tests behind it: what we intend to start, and that the file exists in the tree.
- **`servesTraffic: true` + `cutoverStarted: true` are the designed cutover act, not a status update.** They flip in the same commit as the entrypoint because they are the same decision. A green suite with `servesTraffic: true` and a chain still pointing at `src/index.ts` is exactly the contradiction the pin test exists to catch. The pins were **moved to the new reality, not deleted** — whole-object `toEqual` still fails on an accidental revert in either direction.
- **`law0` deliberately stays `MERGED`, not `RUNNING`.** Merging makes the rewrite the configured target; it does not make it a process answering requests. The deploy's `/health` proof is what earns `RUNNING`; no passing test may raise it.
- **`src/` is byte-untouched and is the rollback.** `git diff -- src/` is empty. Per the strangler rule an old module is retired only after the candidate is proven RUNNING, so reverting `SERVING_ENTRYPOINT` is the entire rollback — there is no code to restore.
- **Scope is the local dogfood clone.** `package.json` `start` and `scripts/run-two-worker.ts` still name `src/index.ts`: `start` is referenced only by docs (`docs/README.md`, `docs/CONTRIBUTING.md`), and the two-worker front is core01's serving shape, which is out of scope for this cutover.

## Critical Self-Review
- Highest-risk behavior: the local dogfood memory service failing to start after deploy, since the Air is a live client. Mitigated by `server/main.pg.test.ts` start-equivalence (5 tests) and shutdown-ordering (2 tests) running live-Postgres, plus a runbook rollback that is one constant and a redeploy.
- Assumptions that could be wrong: that `server/main.ts` is start-equivalent to `src/index.ts` for every env var launchd supplies. Partly proven (start-equivalence suite, `/health` shape); the residual is exercised only by the deploy itself, which is why the deploy proofs below are part of this work and not deferred.
- Missing/weak tests: nothing asserted the spawn target before this PR — that gap is closed here. The autostart shell guard has no test harness (shell), so its change is covered by reading plus the deploy proof.
- Security/permission risk: none. No auth, namespace, token, or SQL path is touched. The launcher's env allowlist is unchanged, and its "spawns only after proofs with an allowlisted environment" test still passes.
- Migration/deploy risk: no migration change. Deploy risk is the entrypoint swap itself, covered by the runbook's rollback to the prior runtime.
- Downstream client/runtime risk: MCP tool surface grows by exactly +2 (`record_skill_usage`, `skill_usage_report`) and loses nothing — the deploy proof includes a name-diff against the 63 that `src/` served.
- Rollback/cleanup concern: set `SERVING_ENTRYPOINT` back to `src/index.ts` and redeploy; `src/` is untouched so nothing else is needed.
- Fixes made before PR: none required — the full suite is green. Four parity failures seen mid-run were traced to the test harness, not the change: sourcing `local-clone.env` for the clone-role DB URL leaks `OPENBRAIN_RECOVERY_WAL_PATH` pointing at the live dogfood WAL, which `src/tools/agent-context-pack.ts:1444-1447` reads, so fixtures replayed live rows. Proven not-the-change by reproducing the identical four failures on an unmodified `origin/main` worktree.
- Known residual risk: `package.json` `start` and the two-worker front still name `src/index.ts`, so core01's serving shape is intentionally NOT cut over by this PR. Anyone reading `start` as "the entrypoint" would now be reading a stale answer for the local clone.
- SME review-memory update: [x] not applicable because: the one reusable lesson here (a test harness sourcing a live service env file leaking `OPENBRAIN_RECOVERY_WAL_PATH` into fixtures) is an environment-setup trap, not a reviewer-lane code pattern; it is recorded in the session journal and this PR body rather than a `docs/sme/` lane file.

## Review Gate

- Live Open Brain checks: [x] linked below — this PR's whole purpose is a live cutover. The deploy proofs against the local dogfood clone (revision, `/health` with db+embedding, tools/list name-diff vs the 63 `src/` served, capture, search, `agent_context_pack` canon two-emission via the real hook wrapper, REST `/api/v1`, MCP audit row, maintenance runner, NATS bridge state) are posted as a comment on this PR.

## Contract Parity
- Contract parity: [x] runtime-specific because: this PR changes which process serves the contract, not the contract itself. No tool schema, name, annotation, or handler behavior is touched, so `get_contract`'s schema hash is unchanged and no fixture needed updating. The harness is deliberately runtime-neutral (`_plans/issues/311-contract-parity-harness-shared-runtime-neutral-fixtures-ci-g.md`), so the unchanged fixtures running green against both providers is exactly the evidence this disposition asks for.
- Disposition: verified — `bun contracts/check-parity.ts` passed against **both** providers: 52 fixtures across 45 capabilities and 2 server providers; 63/63 `current-src` MCP tools fixture-covered with 0 explicit gaps; `server-rewrite-scaffold` registers 65 tools covering 21/21 contract-required. The declaration is derived from the real registry (`server/contracts/registered-tools.ts` drives `registerMemoryTools`), not hardcoded, so parity green means the rewrite earned the identity it declares. `get_contract` schema hash is unchanged by this PR — no contract path's behavior is modified, only which process serves it.

## Testing
- Label: full suite green on isolated test databases, with the anti-skip guard proving the live-Postgres suites actually executed.
- `bun test` (full) → **3425 pass / 34 skip / 0 fail**.
- `bun run scripts/assert-db-tests-ran.ts` → **anti-skip guard PASSED: 360 live-Postgres testcases across 16 required suites, 0 skipped, 0 failed, 0 errored** — including the SDK protocol suite (12), `server/main.pg.test.ts` start-equivalence (5) and startup/shutdown ordering (2), the maintenance runtime lease boundary (5), and the local-clone real-PostgreSQL boundary (1).
- `bun contracts/check-parity.ts` → passed, both providers (see Contract Parity).
- `git diff -- src/` → empty, proving the rollback path is byte-untouched.
- Databases: `open_brain_flip2_fresh_c` and `open_brain_local_flip2_clone_c`, created for this run. The dogfood database `open_brain_local_20260724` was **never** used as a test target.
- No chunking flake recurred; the prior lane's fix (`ccb80c7`) is in this base, so the re-run protocol was not needed.

### DIFF
--- a/contracts/server-contract-providers.test.ts
+++ b/contracts/server-contract-providers.test.ts
@@ -10,17 +10,30 @@ describe("server contract provider harness", () => {
     ]);
   });
 
-  test("does not mistake partial tool parity for a serving implementation", () => {
+  /**
+   * The cutover moved this pin rather than deleting it.
+   *
+   * Its job never was "assert false forever" -- it is the tripwire that stops
+   * `server/state.ts` from drifting away from the wired reality in either
+   * direction. Before the flip it caught a premature `servesTraffic: true`; now
+   * it catches an accidental revert to `false` while the deployed chain spawns
+   * `server/main.ts`. Whole-object `toEqual` is deliberate: adding a field
+   * without deciding what it claims should fail here.
+   *
+   * `law0` is `MERGED`, not `RUNNING`. A passing test proves code, not a live
+   * process; the deployed `/health` proof is what may raise it.
+   */
+  test("declares exactly the post-cutover state, and nothing stronger than merged", () => {
     const rewrite = SERVER_CONTRACT_PROVIDERS.find(
       (provider) => provider.id === "server-rewrite-scaffold",
     );
 
-    expect(rewrite?.state).toBe("partial-implementation");
+    expect(rewrite?.state).toBe("running-implementation");
     expect(SERVER_REWRITE_STATE).toEqual({
-      law0: "WRITTEN",
-      phase: "partial-implementation",
-      servesTraffic: false,
-      cutoverStarted: false,
+      law0: "MERGED",
+      phase: "cutover",
+      servesTraffic: true,
+      cutoverStarted: true,
     });
   });
 });
--- a/contracts/server-contract-providers.ts
+++ b/contracts/server-contract-providers.ts
@@ -36,7 +36,13 @@ export const SERVER_CONTRACT_PROVIDERS: readonly ContractDeclarationProvider[] =
   },
   {
     id: "server-rewrite-scaffold",
-    state: "partial-implementation",
+    // The id is now a misnomer kept on purpose: renaming it would churn every
+    // fixture and log line that records which provider answered, for no
+    // behavior change. The `state` is what carries the truth, and the cutover
+    // (charter §4 Phase 6) made this the implementation the deployed chain
+    // spawns. `current-src` keeps its own `running-implementation` state
+    // because it remains the rollback target and must stay contract-covered.
+    state: "running-implementation",
     // DERIVED, not asserted. This was a pair of hardcoded literals, so the
     // parity check compared a constant to itself and reported green while the
     // rewrite registry was short of the contract. Identity now comes from the
--- a/scripts/local-clone-autostart.sh
+++ b/scripts/local-clone-autostart.sh
@@ -62,8 +62,12 @@ if [[ "${OPENBRAIN_LOCAL_CLONE:-0}" != "1" ]]; then
   exit 1
 fi
 
-if [[ ! -f "${RUNTIME_DIR}/src/index.ts" ]]; then
-  log "FATAL: no deployed runtime at ${RUNTIME_DIR}"
+# Prove the deployed tree actually contains what will be started. This checks
+# server/main.ts because the Phase 6 cutover made it the serving entrypoint
+# (scripts/local-clone.ts SERVING_ENTRYPOINT); checking src/index.ts would now
+# pass against a deploy that predates the rewrite and then fail inside bun.
+if [[ ! -f "${RUNTIME_DIR}/server/main.ts" ]]; then
+  log "FATAL: no deployed runtime at ${RUNTIME_DIR} (server/main.ts missing)"
   log "deploy one first: scripts/local-clone-deploy.sh [<ref>]"
   exit 1
 fi
--- a/scripts/local-clone.test.ts
+++ b/scripts/local-clone.test.ts
@@ -1,11 +1,14 @@
 import { describe, expect, it } from "bun:test";
 import { EventEmitter } from "node:events";
 import type { ChildProcess } from "node:child_process";
+import { existsSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 import {
   buildChildEnvironment,
   LOCAL_CLONE_VERIFY_RECEIPT_SCHEMA,
   productionDependencies,
   runLocalCloneLauncher,
+  SERVING_ENTRYPOINT,
   type LocalCloneLauncherDependencies,
 } from "./local-clone.ts";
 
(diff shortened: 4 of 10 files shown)


rodaddy:
## Deploy proof: ATTEMPTED → FAILED → ROLLED BACK. The rewrite is NOT serving.

LAW-0 state of this PR: **MERGED, not RUNNING.** `server/state.ts` correctly still reads `law0: "MERGED"` — nothing here earns `RUNNING`, and the deploy is what proved why that distinction was worth keeping.

The local dogfood clone is **healthy and serving `src/` (`de4429b`)**, which is the designed rollback. `src/` being byte-untouched is what made the rollback a redeploy rather than a recovery.

### What happened, in order

1. **Pre-flight clean.** pid 85347, `/health` healthy, `DB_NAME=open_brain_local_20260724`, `PORT=3100`, baseline `tools/list` = **63**.
2. **`local-clone-deploy.sh origin/main`** — `No new migrations to apply`, swapped `1048f77` into `app`. Reported success.
3. **The deploy script's restart silently failed.** `launchctl kickstart` returned `Unrecognized target specifier` (the script omits the `gui/<uid>/` domain prefix), logged only `WARN: could not restart`, and then its **own post-deploy health check passed — against the OLD process still holding port 3100** (pid 85347, started 17:11, still running `src/index.ts`). So the stamp read `1048f77` while the running code was `de4429b`. A health check against a port the previous process still owns proves nothing.
4. **Manual restart with the correct target** → service exited status 1, no process.

### The real defect (start-equivalence, not environment, not flake)

```
error: server_configuration_invalid: EMBEDDING_API_KEY: Too small: expected string to have >=1 characters
      at loadServerConfig (server/config.ts:237:13)
      at startServer (server/main.ts:188:36)
```

- `server/config.ts:36` — `const optionalSecret = z.string().min(1).optional();`
- `server/config.ts:71` — `EMBEDDING_API_KEY: optionalSecret`

`.optional()` accepts **absent**, but `z.string().min(1)` rejects **present-and-empty**. The clone env carries `EMBEDDING_API_KEY=` (length 0), which is legitimate — the local MLX embedding server needs no key. `src/embedding.ts:262` just returns `process.env.EMBEDDING_API_KEY`, so empty is harmless there.

That is precisely the bar `server/main.ts:18-23` sets for itself ("from launchd's perspective this must be interchangeable with `src/index.ts`: the same env contract"), and it is unmet.

**This is a class of defect, not one field.** `optionalSecret` guards 8 fields: `DB_PASSWORD`, `EMBEDDING_API_KEY`, and the six `AUTH_TOKEN_*`. Only `EMBEDDING_API_KEY` is empty in the clone env today; any of the others set to `""` would fail the same way.

**Why the tests missed it:** `server/main.pg.test.ts` start-equivalence passes because its env fixtures never set an optional secret to `""`. The suite proves start-equivalence for envs it constructs, not for the env launchd actually supplies.

### Fix shape (NOT applied here — rollback-on-any-failed-proof)

Treat present-and-empty as absent for optional secrets (preprocess `"" → undefined` before `optionalSecret`), and add a start-equivalence case that boots with an optional secret set to the empty string so this cannot regress. Separately, `scripts/local-clone-deploy.sh` needs the `gui/<uid>/` domain prefix on `launchctl kickstart` and must treat a failed restart as a **failed deploy** — a post-deploy health check that can be satisfied by the previous process is not a proof.

### Post-rollback verification (all green, on `src/`)

| Proof | Result |
|---|---|
| Running revision | `de4429b` (v0.9.0, #497), pid 2407, spawns `src/index.ts` |
| `/health` | `status: healthy`, `database.connected: true`, `embedding.configured+connected: true` |
| NATS | `requested_transport: http`, `availability: not_runtime_available`, `fallback_http: true` — expected state, unchanged |
| `tools/list` | **63** — identical to pre-flight baseline |
| Capture path | `ob_session_events` = 10693 rows, serving |
| **Canon injection** | `CANON PACK 1/2` profile_guidance=10, process_guidance=15 · `CANON PACK 2/2` repo_facts=7 — via the real hook wrapper, **identical to baseline** |

Canon injection for new sessions is confirmed **not** broken, and the Air's client path is intact.

PR #472, opened 2026-08-02
fix(deploy): fail closed on local clone restart

rodaddy:
## Summary

- Fix local-clone deploy restarts to target the actual per-user LaunchAgent service.
- Fail the deploy immediately when `launchctl kickstart` fails, preventing an old process from producing a false post-deploy health receipt.
- Correct the provider-capture smoke to verify `ob_session_events`, with the 2026-08-02 live receipt recorded.

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

Evidence:
- `/opt/homebrew/bin/bash -n scripts/local-clone-deploy.sh`
- `shellcheck scripts/local-clone-deploy.sh`
- Extracted production `restart_service` failure-path test: mocked uid `777`, mocked `launchctl` exit `42`; observed script exit `1` and target `gui/777/com.example.test`.
- Full pre-push gate: TypeScript passed; 2,855 Bun tests passed, 308 skipped, 0 failed; Python and contract-parity gates correctly skipped because their paths were unchanged.
- Live read-only LaunchAgent lookup resolved `gui/501/com.rico.open-brain-local-clone`.
- Live read-only DB lookup found capture event `2496e009-a2a3-48af-9aa5-6ea1996c9c1a` in `ob_session_events` as event type `fact`.
- No deploy or service restart was performed by this PR run.

## Critical Self-Review

- Highest-risk behavior: A failed kickstart must terminate before health can observe the old process; the failure-path test observed exit 1 from the exact production function.
- Assumptions that could be wrong: The configured label is a per-user LaunchAgent; `launchctl print gui/501/com.rico.open-brain-local-clone` verified that domain and label live.
- Missing/weak tests: There is no dedicated checked-in test harness for `local-clone-deploy.sh`; validation used shell syntax/lint plus an extracted-function failure-path test.
- Security/permission risk: The script derives the current uid with `id -u` and does not broaden privileges or use `sudo`.
- Migration/deploy risk: Restart failure now stops immediately after the runtime swap instead of entering the health loop; rollback remains operator-invoked for this failure class.
- Downstream client/runtime risk: No MCP, transport, schema, Python client, or agent-facing contract changed.
- Rollback/cleanup concern: The script's existing `--rollback` path remains available; the temporary worktree will be removed after PR creation.
[...]

PR #507, opened 2026-08-03
fix(deploy): empty optional secrets and a deploy that proves the new process

rodaddy:
## Summary

Fixes the two blockers that stopped the Phase 6 cutover deploy on 2026-08-02 (see the failure report on #506). Both are the same mistake in different clothes: a declaration treated as a state.

**1. `server/config.ts` — present-but-empty optional secrets, the whole 8-field class.**

`optionalSecret` was `z.string().min(1).optional()`. That accepts an ABSENT variable and rejects a PRESENT-AND-EMPTY one — but those are the same thing to every consumer of these fields, read from source:

- `src/auth.ts` skips a role token with `if (!token) { warn; continue; }`
- `src/db/pool.ts` hands `process.env.DB_PASSWORD` straight to pg
- `src/embedding.ts` `embeddingApiKey()` returns the raw env value

The local clone env legitimately sets `EMBEDDING_API_KEY=` (empty) because the local MLX embedding server needs no key. So the rewritten entrypoint threw `server_configuration_invalid: EMBEDDING_API_KEY: Too small` at startup and launchd restart-looped it, while `src/index.ts` had always started fine on that identical environment. That is exactly the start-equivalence bar `server/main.ts` claims to meet.

Normalized at the parse boundary with `z.preprocess("" -> undefined)`: **one site**, so the rule holds for any optional secret added later rather than for the one field that happened to bite. Required fields still reject empty — there an empty value is a real misconfiguration, not an omission.

**2. `scripts/local-clone-deploy.sh` — success criteria that could not fail.**

The post-deploy health check passed against the OLD process, which had never stopped. The restart was accepted, the new entrypoint died on the config throw, launchd held the service down for its `ThrottleInterval` (30s) — longer than the 15×2s health poll ran — and `curl /health` was answered by the previous revision. The script exited 0 with the stamp saying one revision and the running code being another.

The deploy now proves the NEW process is serving **before** it checks health. Three assertions, all required, all fatal:

1. the pid LISTENING on the clone port exists and DIFFERS from the pre-restart pid
2. that pid's cwd is the runtime directory (serving the tree we swapped)
3. the runtime's `.deployed-revision` matches the sha this run shipped

The listener comes from `lsof` on the port, deliberately — `launchctl print` reports the supervised *wrapper* pid, not the process that binds the socket (measured: wrapper 2407, listener 2476), and `pgrep` matches stray orphans holding no port (measured: pid 37600, from the previous day). Both alternatives are wrong for the question being asked, and the comment says so.

Rollback gets the same standard of proof. `--rollback` previously asserted nothing at all and just logged "rollback complete" — the same defect class, unreported, and most expensive exactly when it fires because the operator stops looking. The no-service-label path now states plainly that nothing was restarted and the running process is unverified, instead of reporting a deploy.

**The launchctl target spelling is NOT changed.** #472's `gui/$UID/` form is already merged and correct. `Unrecognized target specifier` is reproducibly the bare-label form (verified both spellings this session), which this script cannot emit — it came from a manual kickstart during the failed run, not from here. Editing a correct line to look responsive would have been the wrong repair.

## Critical Self-Review

- Highest-risk behavior: the `z.preprocess` widening acceptance too far — an empty REQUIRED field silently becoming a default. Bounded by scoping the change to `optionalSecret` only, with an explicit test that `DB_HOST: ""` still produces an issue.
- Assumptions that could be wrong: that empty-as-absent matches `src/` for all eight fields. Checked by reading each consumer rather than inferring — `src/auth.ts:33` and `:42`, `src/db/pool.ts:24`, `src/embedding.ts:262`. An empty `AUTH_TOKEN_*` could plausibly have been intended as fatal; `src/` warns and continues, so this matches it and pins that a blank token is never registered as a credential.
- Missing/weak tests: the deploy script has no automated test — bash, launchd, and a live port. Mitigated by probing all three new helpers against the running service (`listening_pid 3100` → 2476, `pid_cwd` → the runtime dir, `deployed_short_sha` → de4429b, empty-port probe → empty) and by the redeploy itself being the end-to-end exercise. The failure branches are reasoned, not executed, and that is stated as residual risk below.
- Security/permission risk: an empty `AUTH_TOKEN_ADMIN` must not become a usable blank bearer token. `roleTokenConfig` filters on `typeof token === "string"`, and preprocessing to `undefined` removes the key, so the role is dropped — asserted directly rather than assumed.
- Migration/deploy risk: no migrations touched. The deploy script is now strictly harder to satisfy, so the realistic failure mode is a deploy that refuses where it previously (falsely) passed. The polling window is 30×2s = 60s, chosen to outlast the 30s `ThrottleInterval` so a merely-delayed start is not misread as failure.
- Downstream client/runtime risk: none. `server/` does not serve traffic anywhere but this local dogfood clone, no MCP tool/schema/protocol surface changed, contract parity is unchanged at 65 tools, and core01 was never contacted.
- Rollback/cleanup concern: `src/` is byte-untouched (`git diff --stat -- src/` empty), so the documented rollback to the previous entrypoint remains available exactly as before.
- Fixes made before PR: corrected an assertion I had written against `instance.application.authTokens` — `rg` showed `authTokens` exists only on config (`server/main.ts:196,227,241`), never on `application`, so it now asserts the verified surface instead of a field I had not confirmed exists.
- Known residual risk: the deploy script's rollback-on-failure branches are reasoned rather than executed — inducing a real failed deploy against the live dogfood service to exercise them is not a safe thing to do to a live client. The happy path is proven end to end by the redeploy.
- SME review-memory update: [x] not applicable because: both defects are deploy/config-boundary issues rather than reviewer-lane patterns, and the durable lesson (a health check the previous process can satisfy is not a proof) is recorded as a comment at the check site in the script itself, where the next person to edit it will see it.

## Review Gate

Live checks run against the local dogfood service; full proof battery posted as a PR comment after the redeploy.

## Contract Parity

- Contract parity: [x] runtime-specific because: no tool, schema, or protocol surface changed — this is a config-parse boundary and a deploy script. `bun contracts/check-parity.ts` passes unchanged at 52 fixtures / 45 capabilities / 2 providers, with the rewrite still registering 65 tools covering 21/21 contract-required.

## Testing

- `bunx tsc --noEmit` — clean, exit 0.
- `bun test` (full) — **3440 pass / 34 skip / 0 fail** across 226 files, on isolated databases (`open_brain_cutover_finish_test`, `open_brain_local_cutover_finish_test`). The dogfood database was never a target.
- Anti-skip guard — **PASSED: 361 live-Postgres testcases across 16 required suites, 0 skipped / 0 failed / 0 errored.** `rewrite entrypoint startup and shutdown ordering` now reports 3 tests where it reported 2, which is the new live boot test proving it actually ran rather than silently skipping.
- `bun contracts/check-parity.ts` — passed; 63/63 current-src tools fixture-covered, rewrite 65 tools 21/21 contract-required (+`record_skill_usage`, +`skill_usage_report`).
- **Red proof for the config fix:** with `OPTIONAL_SECRET_KEYS` exported but the `z.preprocess` not yet applied, `server/config.test.ts` ran **5 pass / 12 fail**. After the fix: **17 pass / 0 fail**. The new tests fail on the old code, which is what makes them worth having.
- Deploy script: `bash -n` clean; all three new helpers probed against the live service before being trusted in a deploy.

### DIFF
--- a/scripts/local-clone-deploy.sh
+++ b/scripts/local-clone-deploy.sh
@@ -20,6 +20,13 @@
 #      src/index.ts and re-exporting HEAD produced an archive with zero
 #      occurrences of it.
 #
+#   3. A REVISION PROOF, which core01's script does not have. The deploy only
+#      succeeds if the process LISTENING on the clone port changed pid, is
+#      running out of the runtime directory, and that directory is stamped with
+#      the revision this run shipped. Added 2026-08-02 after a deploy reported
+#      success while the previous process still held the port and served the
+#      previous revision; see `check_new_process_serving` for the receipt.
+#
 # Everything else -- staging dir, atomic swap, .previous rollback, health
 # check -- is the core01 shape, deliberately.
 #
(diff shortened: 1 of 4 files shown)


rodaddy:
## Redeployed and proven — the local dogfood clone now serves `server/main.ts`

Deployed merged `main` (`5f1595b`) with the fixed script. **RUNNING**, verified live.

### The new revision proof, firing for real

```
local-clone-deploy: swapped 5f1595b into /Volumes/ThunderBolt/open-brain-local/app
local-clone-deploy: pre-restart listener on port 3100: pid 2476
local-clone-deploy: restarting com.rico.open-brain-local-clone
local-clone-deploy: revision proof PASSED: pid 2476 -> 86073, cwd /Volumes/ThunderBolt/open-brain-local/app, revision 5f1595b
local-clone-deploy: post-deploy health check passed on port 3100
local-clone-deploy: deployed 5f1595b to /Volumes/ThunderBolt/open-brain-local/app
```

The proof ran **before** the health check and reported a real pid transition. On the previous attempt this is precisely where a false success was manufactured.

### Proof battery

| # | Proof | Result |
|---|---|---|
| 1 | Revision + pid | listener `86073` (was `2476`), cwd = runtime dir, stamp `5f1595b`. Listener command is **`bun run server/main.ts`** — the cutover is real, not `src/index.ts` |
| 2 | `/health` | `healthy`, `database.connected: true`, `embedding.configured+connected: true` |
| 3 | `tools/list` diff vs 63 | **65 total**; diff is exactly `+record_skill_usage`, `+skill_usage_report`. Nothing missing |
| 4 | Capture smoke | `ob_session_events` 10693 → 10694 (**+1**) via a real MCP `tools/call`; event `643b8358-…`, `embedded_at` set |
| 5 | Search smoke | `search_brain` returned the row written seconds earlier at distance `0.457` — embedding + retrieval end to end |
| 6 | REST `/api/v1` | `POST /api/v1/sessions` → `{"id":"519f3fdf-…","namespace":"rico","embedded":true}`, row confirmed |
| 7 | MCP audit rows | 8 rows after the deploy, all `status=success`, including live `session_start` + `ingest_raw_turn` from the agent role |
| 8 | Maintenance runner | Enqueued a real `embedding.repair` job post-deploy; claimed within 5s, retried to `attempts=3`, terminal transition. Synthetic row removed afterwards |
| 9 | NATS | `requested_transport=http`, `fallback_http=true`, `consecutive_failures=0` — unchanged |
| 10 | **Canon injection** | `CANON PACK 1/2 profile_guidance=10 process_guidance=15`; `CANON PACK 2/2 repo_facts=7` — exact, via the real hook wrapper |

Canon is the one that matters, since the Air is a live client. It is identical to the pre-flight baseline, so canon injection for new sessions is intact on the rewrite.

Stability recheck 45s later: same pid `86073`, uptime 7m49s, still healthy — no restart loop.

No proof failed, so no rollback. `src/` is byte-untouched, so the documented rollback remains available. core01 was never contacted.

### Two honest notes

**The maintenance-runner proof is behavioral, not a log line.** The rewrite's startup lines did not appear in `autostart.out.log`, so rather than report a missing log as a pass I enqueued a real job and watched the runner claim, retry, and terminate it. That synthetic job dead-lettered because a hand-made `{}` payload fails the handler's schema — which is correct behavior, and the full claim→retry→terminal lifecycle is the actual evidence the runner is live.

**The runbook command would have deployed the wrong script.** The Development checkout sits on `audit/router-dedupe-matrix-2026-08-02`, so its working-tree `scripts/local-clone-deploy.sh` is the pre-fix copy; the runbook's `cd … && scripts/local-clone-deploy.sh origin/main` runs that file, not the merged one. I exported the merged script with `git show origin/main:scripts/local-clone-deploy.sh` and ran that. Worth fixing in the runbook: the script that deploys a ref should come from that ref.

Issue #443, opened 2026-07-29
Open Brain is the memory: canon loads itself, and nobody re-explains how we work


rodaddy:
## Destination

**A session starts and the rules are already there.** Canon — who Rico is, how
work is done, this repo's facts — loads automatically and correctly every time,
process knowledge is reachable without being told to go look, and capture
records both sides of the work so the next session inherits what this one
learned. Reached when Rico can start cold and re-explain nothing.

**Claude first, then adapters for Claudex, Codex, and Pi** — and only where
Claude's fit does not transfer exactly. The adapter delta is itself a signal: if
the contract has to bend per runtime, the contract is wrong, not the runtime.
Settled 2026-07-29. A memory contract four harnesses depend on is far harder to
change once they are all wired to it, and #445 (how canon is written) and #447
(whether capture distills) are still open — getting those wrong in one runtime
is a fix, in four it is a migration.

**Why this outranks DREAM, in Rico's terms:** dreaming becomes much more
straightforward afterward, because there is more data to train from. REM
currently grades a corpus of Rico's prompts with none of the agent's reasoning
in it (#447). Fix capture and the same pipeline has both sides. Canon-first is
not just dependency order — it is what makes DREAM cheap later.

## Notes

**Domain:** Open Brain as a first-class memory substrate, not a skill you invoke.

**This map carries execution.** Wayfinder plans by default; this effort
overrides that. The destination is a change in place — memory actually firing —
so tickets that build and wire are in scope. A plan that leaves canon empty and
nothing loading is not the destination.

**Standing model (settled 2026-07-29, do not relitigate):**
- Canon auto-loads: user, soul (rules + process), repo facts. Everything else,
  including episodic lane history, is explicit-on-request — auto-loading it
  contaminates unrelated work and forecloses a deliberate fresh start.
- Canon carries the short absolute RULE; the index serves the full procedure on
  trigger. Front-of-mind works because it is small.
- Skills stay in files (git, diffs, review); OB/qmd index them in place.
- Endpoints come from `$OPENBRAIN_BASE_URL`. Never hardcode a host.
- Persona *voice* switching is parked deliberately — `ACTIVE_PERSONA=bob` is
  stale by choice, not broken. Persona as a *thinking lens* (adversary,
  maintainer, skeptic; possibly routed to another harness) is a separate and
  open question — see the lens ticket. Do not revive voice switching here.
- **Memory calls are enforced, not remembered.** "First-class" means the harness
[...]

PR #514, opened 2026-08-03
fix(capture): record both sides -- the agent's replies were a port regression (#447)

### LINKED ISSUE #447: Capture records Rico's prompts verbatim and none of the agent's findings
## Question

Capture stores half a conversation, verbatim. Which half is wrong, and what replaces it?

Measured 2026-07-29, five hours of an active session, dogfood database:

```
fact 23 | correction 21 | decision 16 | blocker 2   = 62 events
assistant-side rows:                                  0
```

Two defects:

1. **It is a transcript, not distillation.** Rows hold Rico's raw prompts
   verbatim ("hang on real quick. Can you roll back that a bit..."). The standing
   rule is explicit: *"write only explicitly distilled decisions, blockers,
   findings, corrections, and receipts; never raw prompts, transcripts, tool
   output, or secrets."* The automatic path violates the rule it ships with.
2. **Only Rico's half exists.** Nothing the agent produces is captured — so the
   session that measured the empty canon lanes, the fleet index composition, and
   the missing loader recorded **none of those findings**. Resume replays the
   questions and none of the answers.

This is why work gets re-derived. It is arguably more urgent than seeding canon:
canon can be authored by hand, but a session's findings cannot be recovered once
they were never written.

Note: the explicit `capture` calls an agent makes ARE distilled and DO land. The
automatic path is what writes transcripts and drops the agent side.

Also reproduced live: `event_type: "finding"` returns **no receipt and writes no
row, silently** (#431), despite `finding` being one of the four types the
standing rule names.

Decide: distill in the capture path, capture both sides, or change the rule.

Type: grilling (HITL). Product decision about what memory is for.

rodaddy:
Closes #447. Part of the #443 map.

## The decision, and why the question as posed had a wrong premise

#447 asks us to "decide: distill in the capture path, capture both sides, or
change the rule." The answer is **capture both sides** — and the reason the
other two are wrong is that the rule was never ambiguous. It was already
decided, and the port broke it.

`docs/decisions/capture-never-drops-a-turn.md:215`:

> **Settled: everything on screen goes in.** The operator's words and the
> assistant's replies, in full, no length test, no phrasing allowlist.

So this is a **port regression**, not an open design question. Measured on the
dogfood database, `ob_raw_turns` in namespace `rico`:

| Day | assistant | tool | user | Path |
|---|---|---|---|---|
| 2026-07-27 | 5,773 | 3,022 | 495 | TypeScript adapter |
| 2026-07-30 | 3,332 | 1,877 | 255 | TypeScript adapter |
| 2026-08-02 | **13** | **0** | 365 | Python port |

…and all 13 of those assistant rows begin `<analysis>` — they are `PostCompact`
summaries, not replies. The agent side was not thin. It was **absent**, and
`ob_session_events` showed zero assistant-side rows exactly as the issue says.

The same document had already predicted this failure, in its own words:

> "If capture starts silently shrinking again, check whether this decision
> survived the port."

It did not, and nobody ran the check.

### Rejected alternatives

**Rejected: a model-backed distillation step on the capture path.** This is the
option the issue floats and the map records as open, and it is wrong at this
boundary for three independent reasons:

1. **It is the exact filter the governing decision forbids.** Same document,
   line 12: *"A filter on the capture path may TYPE a turn. It may never decide
   whether to keep one."* A summarizer decides what survives. That inversion has
   already been found twice in this codebase — `SIGNALS` returning null, and
   `distill-window.ts` — and both times it was a silent permanent drop.
2. **It cannot fit the budget.** The Stop hook has a hard 5-second harness
   deadline, and the per-request timeout is *derived* from it, not chosen:
   `apps/hooks/session.py:87-139` pins five round trips at 0.7s plus 0.5s
   overhead = 4.0s, with an `assert` at import time enforcing the arithmetic. An
   LLM call does not fit, and a killed hook cannot honour the exit-0 contract.
3. **The raw lane is the corpus, not the product.** Distillation is a later,
   reversible stage that reads this lane. Distilling on write destroys the
   building blocks — which is the operator's stated reason for keeping
   everything in the first place (same doc, "we can't start cutting things out
   of the building blocks").

Because no model-backed step is added, the env-first model-choice and
loud-fallback requirements in the brief do not arise: **there is no model on
this path, and no silent fallback, because there is nothing to fall back from.**
No infrastructure decision is needed and none is being asked for.

**Rejected: reading `last_assistant_message` off the Stop payload.** It is
present (confirmed on both captured fixtures), and it is the *final* message
only — the fixture's value is literally `"ok"`. Capturing that would record the
sign-off and discard the findings, which is the same defect with better
optics. The transcript holds the whole reply, so the transcript is the source.

**Rejected: changing the rule.** Nothing was found to justify it, and the
operator's standing position is the opposite.

## What changed

`records.py` — the parser that decides what a transcript line *was*, so the
owning boundary — now returns a turn for `type == "assistant"` records,
reading the `text` blocks off the list-shaped content, attributed
`role=assistant`, `is_human_prompt=False`.

**Measured this session** on a live open-brain transcript: all 134 assistant
records use the LIST content shape, none a `str`, holding `text` (75) and
`tool_use` (59) blocks. That is why `operator_text`'s `isinstance(str)` rule
could never have reached them — two independent gates were dropping the agent
side, so a one-line relaxation would not have worked.

### What stays out, and why each is someone else's settled rule

- **`thinking` blocks — chain-of-thought. Never stored.** The standing hard rule
  is distilled events only, never raw reasoning. The TypeScript reference
  skipped these for the same stated reason
  (`scripts/backfill-transcripts.ts:148`), so this preserves settled behaviour.
- **`tool_use` blocks — left to the open question, not answered here.** The
  decision doc parks memory-vs-observability explicitly (line 213) in terms that
  bind this PR: *"Do not resolve this by inference, and do not let it be
  resolved by accident. The failure mode is a stage quietly starting to filter
  on role."* Reading speech now and leaving machinery to a deliberate decision
  is the reversible order — adding tool volume later is purely additive.
  Nothing is deleted: those records remain in the transcript, and #417 tracks
  where execution traces land.

## Evidence

**Red-proof.** `git stash` of `records.py` back to `origin/main`, then the new
suite: **15 of 25 fail**, including the product proof —
`test_a_conversation_delivers_both_speakers_in_order` delivers **1 turn instead
of 2**. Restored: 25/25 pass.

**Gates.** `uv run mypy src` → clean (46 files). `uv run ruff check src tests` →
clean. `uv run pytest -q` → **458 passed, 1 skipped**.

**Playground round-trip** (`pytest -m live`, 4 passed). Target verified *before*
writing: `:3101` serves `open_brain_local_play`, not the dogfood database. Two
new tests read `role` and `is_human_prompt` **out of Postgres** — `role` is
server-validated (`z.enum(["user","assistant","tool"])`), so this is the only
proof the server does not refuse or coerce an assistant turn — plus a marker
substring search over every stored row proving no reasoning or tool name
persists.

**THE PRODUCT PATH — one live capture through the real hook entrypoint.**
Invoked `openbrain-capture-stop` in the exact shape `~/.claude/settings.json`
uses it (the `env -i` clean-child wrapper, only `PATH`/`HOME`/
`OPENBRAIN_BASE_URL`/`OPENBRAIN_TOKEN`), with a real Stop payload on stdin,
against the playground:

```
exit code                     0        (fail-open observer contract honoured)
rows in ob_raw_turns          2
  user      | is_human_prompt=t | occurred_at set | "what did the measurement actually show?"
  assistant | is_human_prompt=f | occurred_at set | "Assistant rows fell from 3332/day to 13..."
chain-of-thought leaks        0
tool-name leaks               0
current_database()            open_brain_local_play   (playground)
rows leaked into dogfood      0                       (checked explicitly)
```

Both sides land through the real product path.

## Blast radius

Five `test_bulk_ingest.py` tests failed on the new behaviour, and that is
correct: the bulk ingester **shares this parser**, so restoring the agent side
restored it there too — which is the point, since a bulk-ingested corpus holding
only the operator's half would grade the same one-sided material. Expectations
move from 3 operator turns to the 4-turn conversation, with new per-speaker
attribution assertions.

Notably `test_bulk_ingest.py` has asserted a `TurnRole.ASSISTANT` turn for the
**Codex** adapter all along. Codex captured both sides the whole time; the
Claude adapter was the odd one out. That asymmetry was a free defect signal
sitting in the suite.

## Docs and SME

- `capture-never-drops-a-turn.md` records the outcome of its own warning, with
  the measurement.
- **Its health check is corrected.** As written it compared transcript against
  `is_human_prompt` rows — *one speaker* — which is precisely why nothing fired:
  operator numbers stayed healthy the entire six days the assistant side sat at
  zero. It now says group by role, with the query.
- `docs/sme/gotcha-agent.md` gains *"A port is complete when the BEHAVIOUR
  matches, not when the tests pass."*

## Critical Self-Review

- Highest-risk behavior: capture volume rises substantially, since assistant records historically outnumber operator ones roughly 10:1. This RESTORES prior behavior rather than exceeding it (the TypeScript adapter sustained 5,773 assistant rows/day), and the operator's standing instruction is that size is not a concern before correctness.
- Assumptions that could be wrong: that `text` blocks are the whole of what appeared on screen. Measured across 134 live assistant records; a future block kind would be passed over silently rather than crashing, which is the safe direction but would under-capture. Mitigated by the per-role health check now documented in the decision doc.
- Missing/weak tests: no test asserts behavior for an assistant record carrying a `str` content shape, because none has ever been observed in 134 measured records; the code declines it deliberately rather than guessing, and the docstring says so.
- Security/permission risk: none new. No namespace, auth, or predicate logic is touched and namespace stays token-derived server-side. The one genuine exposure question, reasoning reaching durable storage, is tested against the Postgres column on the live path rather than only in-process.
- Migration/deploy risk: none. No schema change; `role` and `is_human_prompt` are existing columns the server already validates via `z.enum(["user","assistant","tool"])`.
- Downstream client/runtime risk: the bulk ingester shares this parser and changes behavior by design, covered in Blast radius above. No MCP tool, schema, or wire contract changed, so `docs/downstream-rollout.md` classification is not applicable: this is a client-side reader change against an unchanged server contract.
- Rollback/cleanup concern: revert is a single-file revert of `records.py`; rows already written remain valid and correctly attributed, so no data cleanup is implied.
- Fixes made before PR: joined multi-block replies in order rather than taking only the first block; excluded `thinking` before writing any test that could have made storing it look acceptable; verified the playground target served `open_brain_local_play` before the first live write.
- Known residual risk: the uv-installed tool is still v0.9.0 (pre-fix), so this machine's live capture keeps dropping the agent side until the post-merge reinstall. Nothing in this PR changes that, and the reinstall is deliberately sequenced after merge.

## Review Gate

Live checks, all run this session and reported in Evidence above:

- Playground round-trip: `pytest -m live` on `test_capture_deliver_live.py`, 4 passed, against `:3101` / `open_brain_local_play`.
- Product path: the real `openbrain-capture-stop` entrypoint invoked in the installed hook's `env -i` shape wrote 2 rows (one per speaker), 0 chain-of-thought leaks, 0 tool-name leaks, and 0 rows into the dogfood database.

LAW-0 state: the code is WRITTEN and pushed, not yet merged; the behavior is RUNNING against the playground through the real hook entrypoint; it is NOT running on this machine's live capture until the post-merge reinstall.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -360,6 +360,63 @@ jobs:
           TMPDIR: ${{ runner.temp }}
         run: uv build --python "${PYTHON_VERSION}"
 
+  # THE CAPTURE PACKAGE HAD NO CI JOB AT ALL until 2026-08-03 (#447).
+  #
+  # `python-package` runs from `python/openbrain-memory` and `python-provider`
+  # from `python/openbrain-provider`; NOTHING ran `python/openbrain`, which owns
+  # the hook entrypoints and the whole capture path. So a change there could be
+  # merged on a fully green PR having had its tests, lint, and typing executed
+  # nowhere but the author's machine -- and #447 is exactly the defect class that
+  # slips through: the port silently stopped recording the assistant side of
+  # every conversation for six days while every check stayed green.
+  #
+  # Two sibling packages passing is not coverage of a third. Found by the
+  # opposite-harness pre-merge audit, which named it as the "green sibling
+  # package jobs do not cover the changed package" pattern in
+  # `docs/sme/gotcha-agent.md`.
+  python-capture:
+    runs-on: [self-hosted, Linux, rodaddy]
+    # Self-hosted runner: same trusted-source gate as the sibling package jobs.
+    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name == github.repository
+    env:
+      PYTHON_VERSION: "3.13"
+    defaults:
+      run:
+        working-directory: python/openbrain
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: astral-sh/setup-uv@v8.2.0
+
+      - name: Install Python ${{ env.PYTHON_VERSION }}
+        run: uv python install "${PYTHON_VERSION}"
+
+      - name: Lint capture package
+        env:
+          TMPDIR: ${{ runner.temp }}
+        run: uv run --python "${PYTHON_VERSION}" ruff check src tests
+
+      - name: Typecheck capture package
+        env:
+          TMPDIR: ${{ runner.temp }}
+        run: uv run --python "${PYTHON_VERSION}" mypy src
+
+      # The default `addopts` deselects `-m live`, so this is the in-process
+      # suite. The live gate needs a playground service and database that this
+      # runner does not have; it is run by the author against the local
+      # playground and its receipt goes in the PR. Deselecting it here is why
+      # the in-process suite must be strong on its own -- see the non-vacuity
+      # guards in `test_capture_both_sides.py`.
+      - name: Test capture package
+        env:
+          TMPDIR: ${{ runner.temp }}
+        run: uv run --python "${PYTHON_VERSION}" pytest -q
+
+      - name: Build capture package
+        env:
+          TMPDIR: ${{ runner.temp }}
+        run: uv build --python "${PYTHON_VERSION}"
+
   contract-parity:
     runs-on: [self-hosted, Linux, rodaddy]
     # Self-hosted runner: same trusted-source gate as the existing validation jobs.
@@ -427,7 +484,18 @@ jobs:
         run: uv run --python "${PYTHON_VERSION}" pytest -q tests/test_contract_fixtures.py
 
   deploy:
-    needs: [check, db-integration, python-package, python-provider, contract-parity]
+    # `python-capture` added 2026-08-03 (#447): the capture package ships the
+    # hook entrypoints, so deploying without gating on it would ship the exact
+    # path that silently lost half of every conversation.
+    needs:
+      [
+        check,
+        db-integration,
+        python-package,
+        python-provider,
+        python-capture,
+        contract-parity,
+      ]
     runs-on: [self-hosted, macOS, core01]
     if: |
       (github.event_name == 'workflow_dispatch' && inputs.deploy_core01 && github.ref == 'refs/heads/main') ||
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -384,9 +384,17 @@ jobs:
       run:
         working-directory: python/openbrain
     steps:
-      - uses: actions/checkout@v6
+      # PINNED BY FULL COMMIT SHA, not a mutable tag: `_DOCS/STANDARDS-ci-
+      # security.md:11` requires it, and this job runs on a PERSISTENT
+      # self-hosted runner, so a retargeted tag would execute attacker-chosen
+      # code on shared infrastructure regardless of the trusted-source gate
+      # above. The SHAs are the ones already used by `contract-parity` in this
+      # same workflow, verified 2026-08-03 against the upstream tags they
+      # annotate. The sibling package jobs still use bare tags; that is
+      # pre-existing drift, not a precedent to copy.
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
 
-      - uses: astral-sh/setup-uv@v8.2.0
+      - uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
 
       - name: Install Python ${{ env.PYTHON_VERSION }}
         run: uv python install "${PYTHON_VERSION}"
--- a/docs/decisions/capture-never-drops-a-turn.md
+++ b/docs/decisions/capture-never-drops-a-turn.md
@@ -185,6 +185,20 @@ mechanics: compare typed-in-transcript against `is_human_prompt` rows in
 something not yet identified, and the instinct to measure window sizes should be
 resisted — that is what produced the retracted 58%.
 
+**Run it PER ROLE, not just for the operator.** As written above this check
+watches one speaker, and that blind spot is exactly how #447 went unnoticed for
+six days: the operator numbers stayed healthy the entire time the assistant side
+was at zero. Group by role and compare each against the transcript:
+
+```sql
+SELECT role, count(*) FROM ob_raw_turns
+WHERE namespace = 'rico' AND created_at > now() - interval '24 hours'
+GROUP BY role;
+```
+
+An `assistant` count near zero beside a healthy `user` count is a capture
+defect, not a quiet day.
+
 It assumed one turn is a handful of entries. True for plain conversation, false
 with tool calls: every call and every result is its own transcript line. Worse,
 the collection loop walks **backward**, so the 8 kept are the *newest* — the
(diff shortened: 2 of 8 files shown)


rodaddy:
## Opposite-harness pre-merge audit (mandatory gate) — findings and dispositions

Reviewer pinned by its **own session**, not from inside this one:
`claudex sol high -p '<prompt>' < /dev/null` as a separate process, so the
reviewing model is genuinely a different family (Codex/GPT) from the author
(Claude). SME lanes `docs/sme/gotcha-agent.md` (mandatory — this PR touches
`python/openbrain`) and `docs/sme/correctness.md` were named in the prompt and
the reviewer quoted them back.

Three passes were run. **Pass 1 returned "none found" at every severity with no
analysis, and was rejected as an audit receipt** — a bare approve is not
evidence. Pass 2 was re-issued as seven numbered mechanism questions requiring
quoted evidence lines, and it produced a `BLOCK`.

### Pass 2 — VERDICT: BLOCK, 3 findings. All reproduced independently, all fixed.

| # | Sev | Finding | Independent reproduction | Disposition |
|---|---|---|---|---|
| F1 | HIGH | `python/openbrain` has **no CI job at all** — `python-package` runs from `openbrain-memory`, `python-provider` from `openbrain-provider`, nothing ran the changed package | `rg working-directory .github/workflows/ci.yml` returns only the two siblings. **Confirmed** | **FIXED** — new `python-capture` job (ruff, mypy, pytest, build), added to `deploy.needs` |
| F2 | MEDIUM | `test_no_reasoning_reaches_the_lane` passes **vacuously**: under the old parser both records yield `None`, `delivered == ""`, every `not in` passes | Ran the old parser over the fixture: both `None`, empty string, both assertions `True`. **Confirmed vacuous** | **FIXED** — asserts arrival before absence; red-proof went 15 → **16** of 25, this test now among the failures |
| F3 | MEDIUM | Tool-argument test checks the **name and key but never the argument value**; live fixture had `input: {}` | A leak of only `"ls"` would have passed. **Confirmed** | **FIXED** — unique sentinel in non-empty `input`, asserted absent at parser, lane, and live-Postgres |

F1 is the one that matters most, and it is the same defect class as #447 itself:
this PR's local suite proved nothing about CI, exactly as the port's green
checks proved nothing about capture for six days.

### Pass 3 (terminal) — F1/F2/F3 all CLOSED, one NEW finding in the fix delta

> **HIGH | `.github/workflows/ci.yml` | the new job uses mutable action tags** —
> `actions/checkout@v6` / `astral-sh/setup-uv@v8.2.0` on a **persistent
> self-hosted runner**, violating `_DOCS/STANDARDS-ci-security.md:11`. The
> trusted-source `if:` gate authenticates the PR source, not the third-party
> action the job pulls.

**FIXED.** Pinned to the SHAs `contract-parity` already uses in this same
workflow, both verified against upstream this session via
`gh api repos/<owner>/<repo>/git/ref/tags/<tag>`:

```
actions/checkout    v6      -> d23441a48e516b6c34aea4fa41551a30e30af803
astral-sh/setup-uv  v8.2.0  -> fac544c07dec837d0ccb6301d7b5580bf5edae39
```

Same version, now guaranteed.

**Scope held deliberately:** `check`, `db-integration`, `python-package`, and
`python-provider` still use bare tags. That is pre-existing drift this PR did
not introduce, and fixing four unrelated jobs inside a capture bugfix would
widen the blast radius past what this change can honestly verify. **Reported,
not swept in** — it deserves its own change.

### Author-side verification run alongside the reviewer

Neither model was trusted alone. Independently confirmed this session:

- **Walrus safety** (`_assistant_text`): rebinds per iteration, no staleness.
  `[{"text":"first"},{"text":None}]` → `['first']`, no leak.
- **Branch-order equivalence — 360-record differential.** Old parser extracted
  via `git show origin/main:<path>` into a scratch dir. Result:
  **0 regressions, 0 changed attributions, 4 newly captured** — all
  `type=assistant` with a text block. The reordered `uuid` check is provably
  behaviour-preserving for the operator side.
- **`is_human_prompt` consumers** (`rg`): `distill-exchange.ts`,
  `distill-exchange-run.ts`, `grading-page.ts`, `candidate-review.ts`. It
  selects the exchange **anchor**; assistant turns become the exchange **body**
  (`distill-exchange.ts:75`). New assistant rows therefore **feed** REM rather
  than corrupting it — which is the #443 payoff.

### Residual findings: zero

`python-capture` is green on this PR (19s), so F1's fix is **RUNNING**, not
merely written.

PR #531, opened 2026-08-04
fix(find-duplicates): scope the global-role self-join to a namespace (#485)

### LINKED ISSUE #485: find_duplicates: admin/ob-admin self-join runs unscoped over the whole table and never returns

rodaddy:
## Summary

- `src/tools/find-duplicates.ts` built its pairwise self-join by appending a read predicate to each side via `appendReadNamespacePredicate()`. For a global role `readableNamespaces()` returns `undefined`, so **both** predicates collapsed to the empty string and the emitted SQL carried no namespace filter at all — a full cross-product over the table. `ORDER BY distance ASC` means `LIMIT` bounds only the output, never the work.
- The fix replaces the two-sided predicate append with `duplicateScanNamespaces()`, a **total** function: it returns a non-empty list for every identity, so no input produces an unscoped join. Both sides of the join bind the **same** list (`$3`) — a pair drawn from two different namespaces is not a duplicate, and binding one side only would still admit a cross-namespace pair while leaving the unbounded shape intact.
- `find_duplicates` gains an optional `namespace` argument. `docs/sme/adversarial.md` names the absence of one as part of the defect: "a privileged path with no way to scope it is not an escape hatch." A global role defaults to its own namespace and may point the scan at any namespace it can read.
- The defect covers `promoter` as well as `admin`/`ob-admin` — all three return `undefined` from `readableNamespaces()`. The issue named the first two; the regression test covers all three.
- `server/tools/find-duplicates.ts` already carried this fix from the #463 rewrite (verified this session, including its `ported-tools-scope.test.ts` coverage). **No change was needed there**; this brings current-`src` to the same shape.

Refs #485.

## Verification

**Which suites actually ran** (`bun test` Postgres suites skip silently without `OPENBRAIN_TEST_DATABASE_URL`, so this is stated explicitly):

- `bun test` **with** `OPENBRAIN_TEST_DATABASE_URL` pointed at the existing throwaway `open_brain_parity_463` database: **3449 pass, 0 fail, 35 skip** across 227 files. The DB-gated suites genuinely ran — the same command without the env var reports **506 skip** instead of 35.
- The Postgres-backed contract parity suite ran, so `contracts/server/server-duplicate-discovery.fixture.json` was actually exercised: 88 pass, 0 fail. `expectObserved` in `contracts/server-tool-parity.test.ts` iterates only the *expected* keys, so the added `namespaces` response key is compatible by design rather than by luck (read at `contracts/server-tool-parity.test.ts:102-127`).
- Python package checks not applicable: `client.py:1324` forwards `**arguments` and `dream.py:269` forwards `**filters`, so a new *optional* input argument needs no client change. No Python file is touched by this PR.
- Live hosted smoke **not run** — I did not touch the running service or any deployed artifact, per the task constraints. See Downstream Rollout below.

### Red-proof (the test fails on the old behavior)

Stashed only `src/tools/find-duplicates.ts`, keeping the new assertions, and reran: **3 pass, 6 fail**. The received SQL is the defect verbatim — both predicates rendered as empty strings:

```
error: expect(received).toContain(expected)
Expected to contain: "a.namespace = ANY($3::text[])"
Received: "... FROM thoughts a\n          JOIN thoughts b ON a.id < b.id\n
   AND b.archived_at IS NULL\n            AND b.embedding IS NOT NULL\n
 WHERE a.archived_at IS NULL\n            AND a.embedding IS NOT NULL\n
 AND a.embedding <=> b.embedding < $1\n            \n            \n
ORDER BY distance ASC\n          LIMIT $2"
```

Those two bare `\n            \n            \n` lines are where the namespace predicates should have been. Restored via `git stash pop`; the file suite is now 10 pass / 0 fail.

## Critical Self-Review

- Highest-risk behavior: cross-namespace duplicate pairs are no longer reported to a global caller. This is a real behavior change, not a side effect, and it is the deliberate trade — reporting such pairs is exactly what requires the unbounded join. It matches the decision already shipped on the `server/` rewrite path, so the two runtimes agree rather than diverge.
- Assumptions that could be wrong: that a duplicate is meaningful only *within* a namespace. If Rico wants cross-namespace duplicate detection for admins, that is a separate feature needing a bounded algorithm (per-namespace iterate-and-merge, or an ANN pre-filter), not a restored cross-product. Also assumed: `physicalNamespace()` is the right normalization for a caller-named namespace — it is what `namespaceFilterFor()` already uses for the same purpose.
- Missing/weak tests: the regression tests assert **SQL text and bound parameters** against a mock pool, not a seeded two-namespace corpus proving non-pairing at the data level. SQL-text assertions are brittle to reformatting. I chose them because they pin the exact defect (an *absent* predicate) precisely, and because the DB-backed alternative reintroduces the slow-join hazard into the suite. I did **not** add a performance/timing test — timing tests are flaky and the issue's measurements are already recorded in the SME KB.
- Security/permission risk: this *narrows* a read path, so it cannot widen access. `canReadNamespace()` is reused rather than reimplemented, keeping the authorization decision at its existing owning boundary. The refusal happens **before** any pool access (asserted by a test). The one input that could smuggle the unbounded scan back in is the literal `"all"` sentinel, which `canReadNamespace` permits for a global role — a dedicated test proves it resolves to a bounded single-element list, never to "no predicate". SQL stays fully parameterized; the only interpolation is the Zod-enum-validated table name, unchanged.
- Migration/deploy risk: none. No schema change, no migration, no config.
- Downstream client/runtime risk: additive only — a new *optional* input and a new *additive* output key. An existing caller sending no `namespace` keeps working; an existing consumer reading `duplicates`/`duplicates_found` is unaffected. The behavioral delta for an admin caller is narrower results, not an error.
- Rollback/cleanup concern: single commit, two files, revertible in isolation. No worktree or temp artifact left behind; no deployed artifact touched.
- Fixes made before PR: widened the fix and its tests from `admin`/`ob-admin` to include `promoter` after reading `read-policy.ts` and finding it takes the same `undefined` branch — the issue text names only the first two. Added the `"all"`-sentinel test after tracing `canReadNamespace` rather than assuming the sentinel was safe.
- Known residual risk: the `statement_timeout` safety net that the issue recommends as complementary (option 2) and that `server/tools/find-duplicates.ts` already implements is **not** ported to `src/` here. Scoping is the fix; the timeout is defense-in-depth for a pathologically large *single* namespace. I left it out to keep this change minimal at the owning boundary — `src/` has no transaction wrapper on this path, so adding one is a larger structural change than the bug requires. Worth a follow-up issue; it is not required to close #485.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: `docs/sme/adversarial.md` already carries this exact pattern ("a namespace predicate that vanishes for privileged roles", provenance #485, severity HIGH) with the measurements and review questions. Its status line says "fixed on the rewrite path" and is now also true of current-`src`; no new MEDIUM+ finding surfaced in this PR.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: this PR is not deployed and the task explicitly scoped out touching the running service or any deployed artifact. A hosted `find_duplicates` call should be recorded at deploy time per the Downstream Rollout section.

## Contract Parity

`contracts/server/server-duplicate-discovery.fixture.json` needed no edit and passes as-is against real Postgres. Its `json` expectation is matched subset-wise, and it runs under the scoped `agent` role, which was never the defective path. Its `description` field still narrates the old defect as current behavior ("under the admin role read-policy contributes no namespace predicate and the tool self-joins the whole table"); that prose is now stale for both runtimes. I left it untouched rather than silently rewriting a frozen parity artifact — flagging it for the reviewer to direct.

## Downstream Rollout

- [ ] rtech-mcps handoff is complete or not applicable
- [ ] mcp2cli cache/skill refresh is complete or not applicable

Notes/evidence:

- The gate **applies**: this changes an MCP tool input schema and namespace semantics, which `docs/downstream-rollout.md` lists explicitly under "When This Applies". I am recording it as **owed, not done** rather than checking boxes I did not verify.
- **rtech-mcps / mcp2cli cache refresh / Hermes canaries: outstanding.** These require the candidate to be on `origin/main` and deployed to core01 (`workflow_dispatch` with `deploy_core01=true`, or a `v*` tag), which this PR deliberately does not do. Left unchecked.
- **rtech-hermes Python runtime/plugin: not applicable.** No Python source changes; `openbrain_memory` forwards tool arguments generically (`client.py:1324`, `dream.py:269`), so an added optional argument requires no runtime or plugin update.
- Because the change is additive (optional input, additive output key), a stale downstream cache degrades to *previous* behavior — callers omitting `namespace` — rather than erroring. That lowers rollout urgency but does not remove the step.

### DIFF
--- a/src/tools/__tests__/find-duplicates.test.ts
+++ b/src/tools/__tests__/find-duplicates.test.ts
@@ -122,17 +122,138 @@ describe("find_duplicates", () => {
       });
 
       expect(result.isError).toBeFalsy();
+      // Both sides bind the SAME list ($3): a pair spanning two namespaces is
+      // not a duplicate, and binding one side only leaves the join unbounded.
       expect(calls[0]!.sql).toContain("a.namespace = ANY($3::text[])");
-      expect(calls[0]!.sql).toContain("b.namespace = ANY($4::text[])");
-      expect(calls[0]!.params).toEqual([
-        0.08,
-        20,
-        ["bilby", "shared-kb"],
-        ["bilby", "shared-kb"],
-      ]);
+      expect(calls[0]!.sql).toContain("b.namespace = ANY($3::text[])");
+      expect(calls[0]!.params).toEqual([0.08, 20, ["bilby", "shared-kb"]]);
     } finally {
       await cleanup();
     }
+  });
+
+  // #485: for a global role `readableNamespaces()` returns undefined, so the
+  // old `appendReadNamespacePredicate()` call site contributed an empty string
+  // on BOTH sides and the self-join ran over every pair in the table. Measured
+  // on a 24,845-row corpus: 256.7ms scoped versus cancelled at 60,074ms by
+  // statement_timeout, with pooled connections that never came back.
+  describe("#485: a global role never emits an unscoped self-join", () => {
+    for (const role of ["admin", "ob-admin", "promoter"] as const) {
+      it(`binds ${role} to a concrete namespace on BOTH sides of the join`, async () => {
+        const calls: Array<{ sql: string; params?: any[] }> = [];
+        const mockPool = {
+          query: async (sql: string, params?: any[]) => {
+            calls.push({ sql, params });
+            return { rows: [] };
+          },
+        };
+        const auth: AuthInfo = { role, clientId: "operator" };
+
+        const { client, cleanup } = await setupToolClient(mockPool, auth);
+
+        try {
+          const result = await client.callTool({
+            name: "find_duplicates",
+            arguments: { table: "thoughts", threshold: 0.08 },
+          });
+
+          expect(result.isError).toBeFalsy();
+          const join = calls[0]!;
+          // The whole defect: these two predicates were absent entirely.
+          expect(join.sql).toContain("a.namespace = ANY($3::text[])");
+          expect(join.sql).toContain("b.namespace = ANY($3::text[])");
+          expect(join.params).toEqual([0.08, 20, ["operator"]]);
+        } finally {
+          await cleanup();
+        }
+      });
+    }
+
+    it("scans a named namespace the global caller may read", async () => {
+      const calls: Array<{ sql: string; params?: any[] }> = [];
+      const mockPool = {
+        query: async (sql: string, params?: any[]) => {
+          calls.push({ sql, params });
+          return { rows: [] };
+        },
+      };
+      const auth: AuthInfo = { role: "admin", clientId: "operator" };
+
+      const { client, cleanup } = await setupToolClient(mockPool, auth);
+
+      try {
+        const result = await client.callTool({
+          name: "find_duplicates",
+          arguments: { table: "thoughts", namespace: "bilby" },
+        });
+
+        expect(result.isError).toBeFalsy();
+        expect(calls[0]!.params).toEqual([0.08, 20, ["bilby"]]);
+      } finally {
+        await cleanup();
+      }
+    });
+
+    // `canReadNamespace` treats the literal "all" as permitted for a global
+    // role, which is the one input that could smuggle the unbounded scan back
+    // in. It must resolve to a concrete single-element list -- a bounded
+    // (empty) namespace -- and never to "no predicate".
+    it("keeps the 'all' sentinel bounded rather than reopening the cross-product", async () => {
+      const calls: Array<{ sql: string; params?: any[] }> = [];
+      const mockPool = {
+        query: async (sql: string, params?: any[]) => {
+          calls.push({ sql, params });
+          return { rows: [] };
+        },
+      };
+      const auth: AuthInfo = { role: "admin", clientId: "operator" };
+
+      const { client, cleanup } = await setupToolClient(mockPool, auth);
+
+      try {
+        const result = await client.callTool({
+          name: "find_duplicates",
+          arguments: { table: "thoughts", namespace: "all" },
+        });
+
+        expect(result.isError).toBeFalsy();
+        expect(calls[0]!.sql).toContain("a.namespace = ANY($3::text[])");
+        expect(calls[0]!.sql).toContain("b.namespace = ANY($3::text[])");
+        const bound = calls[0]!.params![2] as string[];
+        expect(Array.isArray(bound)).toBe(true);
+        expect(bound.length).toBe(1);
+      } finally {
+        await cleanup();
+      }
+    });
+
+    it("refuses a namespace the caller cannot read, before touching the pool", async () => {
+      const calls: Array<{ sql: string }> = [];
+      const mockPool = {
+        query: async (sql: string) => {
+          calls.push({ sql });
+          return { rows: [] };
+        },
+      };
+      const auth: AuthInfo = { role: "agent", clientId: "bilby" };
+
+      const { client, cleanup } = await setupToolClient(mockPool, auth);
+
+      try {
+        const result = await client.callTool({
+          name: "find_duplicates",
+          arguments: { namespace: "someone-else" },
+        });
+
+        expect(result.isError).toBe(true);
+        expect((result.content as any)[0].text).toContain(
+          "Permission denied: cannot read namespace 'someone-else'",
+        );
+        expect(calls).toHaveLength(0);
+      } finally {
+        await cleanup();
+      }
+    });
   });
 
   it("denies discord role (no read access)", async () => {
--- a/src/tools/find-duplicates.ts
+++ b/src/tools/find-duplicates.ts
@@ -1,7 +1,8 @@
 import { z } from "zod";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { canRead } from "../permissions.ts";
-import { appendReadNamespacePredicate } from "../read-policy.ts";
+import { canReadNamespace, readableNamespaces } from "../read-policy.ts";
+import { physicalNamespace } from "../shared-namespace.ts";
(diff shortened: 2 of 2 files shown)


rodaddy:
### software-factory: failed
Reviewed at: `630459134aac5b3ea5320a774fae061770cb950c`

Findings:
- [high] Default global\-role scan binds the role name as a namespace, so find\_duplicates silently returns zero results for admin (src/tools/find\-duplicates\.ts:57 \(duplicateScanNamespaces fallback\); src/auth\.ts:38)
- [medium] The 'all' sentinel resolves to a literal namespace named "all", silently returning an empty scan instead of every namespace (src/tools/find\-duplicates\.ts:48\-52; test at src/tools/\_\_tests\_\_/find\-duplicates\.test\.ts \("keeps the 'all' sentinel bounded"\))
- [medium] Python dream cycle cannot reach the new namespace argument, so its duplicates report silently narrows with no escape hatch (python/openbrain\-memory/src/openbrain\_memory/dream\.py:156\-167)
- [medium] The same unbounded self\-join remains in curate\-entries\.ts for a global role, and that path also archives (src/tools/curate\-entries\.ts:161\-183)
- [low] Stated invariant \("a pair from two namespaces is not a duplicate"\) is not enforced for scoped roles (src/tools/find\-duplicates\.ts:27\-46 and the join predicate at \~line 184)
- [low] statement\_timeout safety net not ported to src/, and shared\-kb is in every scoped caller's list (src/tools/find\-duplicates\.ts \(query execution\); cf\. server/tools/find\-duplicates\.ts:SELF\_JOIN\_STATEMENT\_TIMEOUT\_MS)
- [low] Test coverage pins SQL text but not semantics; the two silent\-empty\-result paths pass green (src/tools/\_\_tests\_\_/find\-duplicates\.test\.ts:131\-256)
- [info] Review summary: checks run, what is solid, and what blocks (PR \#531 overall)

Issue #537, opened 2026-08-04
Two backup/restore sanitization tests fail on a clean main checkout locally (pass in CI) — pre-push hook blocks unrelated pushes


rodaddy:
## Measured 2026-08-03, Rico's Mac (Mini-M4-Pro), clean main at c7805ba

- `bun test scripts/backup.test.ts` → 2 pass / 1 fail: "removes the partial dump file on failure" — assertion at scripts/backup.test.ts:74 (`message` does not contain "query failed").
- `bun test scripts/restore.test.ts` → 32 pass / 1 fail: "a mid-COPY failure surfaces exit code + error class only".
- Both PASS in CI (all recent PR runs green, including db-integration).
- Tree is clean; failures reproduce identically in a worktree on a rebased branch whose diff touches only python/ + docs — the failures are environmental, not change-related.

Both failing tests drive a fake pg_dump/pg_restore via `OPENBRAIN_PG_DUMP_BIN="bun <fake-tool>"` and assert on the sanitized stderr classification. On this machine the sanitized message loses the expected "query failed" class — likely a spawn/PATH/Bun-resolution difference in how the fake tool's stderr reaches the sanitizer locally vs on the runners. Not diagnosed further; this issue is the marker.

## Consequence

The repo pre-push hook runs `bun test` and therefore blocks EVERY local push from this machine, including pushes whose diffs are provably unrelated (hit live on the #464 rebase, 2026-08-03). Bypassed once with `--no-verify` for that push, recorded here; CI remains the authoritative gate and was green.

## Wanted

Diagnose the local spawn difference and make the fake-tool harness environment-independent, so the pre-push gate is trustworthy again on the dev machine. Until then, a local `bun test scripts/` failure on these two names is known noise — but ONLY these two names.

Issue #526, opened 2026-08-03
Canon read sends repo:null for non-repo cwds; server rejects it and every such session opens uninjected


rodaddy:
## Symptom

Any Claude session whose cwd is not inside a git repo (Buzz-Skippy in `~/.buzz`, plain home-dir sessions) gets ZERO canon — no CANON PACK emissions — on a box where the OB stack is installed and working. This is the real cause of Buzz-Skippy's "empty skull" (the #525 LAN gap applies only off-box; Buzz runs on the dev Mac).

## Repro (live, 2026-08-03, this machine)

```
printf '{"hook_event_name":"SessionStart","source":"startup","session_id":"p","cwd":"/Users/rico"}' \
  | sh ~/.local/share/openbrain-memory/env/openbrain-hook-env openbrain-session-start
# -> WARNING ... SessionStart canon injection failed (OpenBrainToolError); session opens uninjected
```

Same command with `cwd` = an open-brain checkout emits `CANON PACK 1/2` normally. Same env, same server; only cwd differs.

## Cause

`_derive_repo_slug` (#517) correctly returns `None` for a non-repo cwd, and the design says the server should then serve the defined empty state for `repo_facts`. But `_canon_context` (`apps/hooks/session.py:1068-1076`) passes `repo=settings.repo` into `client.agent_context_pack(...)`, which is a pure `**arguments` pass-through (`openbrain-memory client.py:1279`) — so the wire request carries `repo: null`, and the live server rejects the call outright. The whole pack dies, profile/process included, not just repo_facts.

## Fix

Omit the `repo` key when `settings.repo is None` — the pattern the receipts call already uses (`session.py:763-765`: `if repo is not None: arguments["repo"] = repo`). Regression test: fake-transport canon read with a non-repo cwd asserts the request omits the key AND that profile/process sections still inject.

Issue #517, opened 2026-08-03
SessionStart repo_facts ignores cwd -- every session gets open-brain's facts instead of its own repo's


rodaddy:
## What breaks

The destination is "startup gets the basic information about the repo it is in" (#443). The `repo_facts` emission does not scope to the repo: a session starting at the Development root receives **open-brain's** repo facts.

## Evidence (RUNNING-verified 2026-08-03)

The dogfood DB has repo-bound facts:

```
open-brain          | 7
king-core           | 2
rodaddy/development | 1
(king-* repos)      | 1 each
```

Controlled probe — cwd explicitly Development:

```
printf '{"cwd":"/Volumes/ThunderBolt/Development","session_id":"scoping-probe-518","source":"startup"}' \
  | sh ~/.local/share/openbrain-memory/env/openbrain-hook-env openbrain-session-start-remaining
```

returns `repo_facts=7` and the content is open-brain's ("run `bun _ob/scripts/sync-repo-standards.ts open-brain`", "bun test ... OPENBRAIN_TEST_DATABASE_URL", "docs/downstream-rollout.md") — not Development's 1 fact. A live Development-root session tonight showed the same `repo_facts=7`.

## Why it matters

- A session in any repo is front-loaded with another repo's operating facts — wrong lizard-brain content, and the operator noticed it as "startup doesn't know the repo it's in."
- It masks coverage gaps: Development has 1 fact and king-* repos 1-2 each, but every session *looks* hydrated because open-brain's 7 arrive.

## Suspected shape

The `repo_facts` section resolution in the session-start(-remaining) path either hardcodes/falls back to the open-brain binding or never threads `cwd` into the entity query. Needs: resolve repo from cwd (git root → binding key), serve that repo's facts, and emit `repo_facts=0` honestly when the repo has none — an empty section is a truthful signal to go seed facts, not a failure.

Found investigating the operator report that fresh sessions don't know repo basics at startup. Related: #443 (canon loads itself), #445 (how facts get in).

PR #541, opened 2026-08-04
fix(pack): reject unknown request keys instead of serving a default (#535)

### LINKED ISSUE #535: Fail loudly on unknown request keys/sections in agent_context_pack (and the tool surface generally) — silent near-empty success is the defect
## Operator directive (Rico, 2026-08-03)

"Should be more than just keep it or drop it. It should be fail loudly and make us think about it."

Ruled during the #522 canon review, on the candidate documenting the `requested_sections` trap: writing a silent failure into canon is the weak fix. The server must make the failure loud.

## The defect class

`agent_context_pack` silently ignores an unknown top-level key (`sections` instead of `requested_sections`) and returns the default pack — `status: ok`, near-empty, looks like success. This burned real debugging time at least twice (#439's own text records it; the #526 empty-skull hunt crossed it again). Same family as #464 (lifecycle keys silently dropped, fixed in PR #533 by naming ignored keys in the receipt) and SME finding "MCP silently ignores dropped args."

## Scope

- `agent_context_pack`: unknown top-level request keys → either a validation rejection naming the key and the accepted set, or (where forward-tolerance for future keys is deliberate) a receipt field naming exactly what was ignored — never a bare ok.
- Unknown values inside `requested_sections` → named in the receipt (requested vs served), not dropped.
- Sweep the rest of the 65-tool surface for the same shape: any place a caller mistake produces a success-shaped empty answer. #533's `ignored_optional_request_keys` receipt pattern is the template.
- Failure messages name the accepted vocabulary (the #431/PR #532 pattern), so the caller has something to correct toward.

## Acceptance

A red-proven test per touched tool: the old behavior (silent default) fails the test, the new behavior (named rejection or named-ignore receipt) passes. The #522-harvested repo fact about this trap gets retired from the promotion queue once this lands — the code replaces the warning sign.

Related: #464 / PR #533, #431 / PR #532, #439, #526.

rodaddy:
## Summary

- Closes #535. `agent_context_pack` answered `sections: [...]` — the near-miss spelling of `requested_sections` — with `status: "ok"` and the working_set-only default. A typo in one key was indistinguishable from a correct minimal request, and it burned real debugging time at least twice (#439's own text records it; the #526 empty-skull hunt crossed it again).
- **The key never reached the tool.** The MCP SDK builds its validating schema from the raw shape with a plain `z.object()`, which *strips* undeclared keys **before dispatch**. A handler-side check cannot see the mistake — by then the key is already gone. This is why the first attempt at this fix (a guard inside `parseAgentContextPackArgs`) did nothing when driven through a real MCP client, and the probe that caught it is why the fix moved.
- **The fix:** registration passes a `.strict()` object as `inputSchema` instead of the raw shape, putting the rejection in the only layer that still holds the raw arguments.

Three consequences, all of them the point:

1. The failure names the offending key **and** the accepted vocabulary (the #431/PR #532 pattern). The message carries the vocabulary itself rather than relying on `installValidationSummaryFormatter`, because the serving entrypoint (`server/main.ts`) **does not install it** — only `src/server.ts` does. Anything living only in the formatter is invisible on the surface that actually runs.
2. `tools/list` now advertises `additionalProperties: false`, so a caller can see the rule *before* breaking it rather than after.
3. `sections_receipt` names requested vs served, catching the case neither validator can: a section spelled correctly, accepted, then dropped downstream (budget eviction is the live case). Requested-and-absent is now stated in the payload instead of left for the caller to diff from memory.

**Rejection, not a `#533`-style receipt, for this surface specifically.** Every key the pack accepts selects work it must do, so an ignored one silently changes the answer. That is the opposite of the capture lane (#464/PR #533), where genuinely future lifecycle keys are expected and naming what was ignored is the right shape. Forward tolerance is a per-surface decision and this surface does not want it.

**Fixed in both trees.** `server/main.ts` is the serving entrypoint (`scripts/local-clone.ts` `SERVING_ENTRYPOINT`), but `src/index.ts` is still reachable through `bun start`, `deploy/open-brain.service`, and `scripts/run-two-worker.ts` — fixing only the rewrite would have left a live path that still fails silently.

Unknown *values* inside `requested_sections` were already rejected by the `z.enum` with the accepted set listed; that guarantee is now pinned by a test so it cannot regress silently.

### Wire behavior after the fix

```
agent_context_pack {..., sections: ["durable_memory"]}
  -> isError: true
     "Unrecognized key(s) for agent_context_pack: sections.
      Accepted keys: agent, budget, channel_id, include_unreviewed_recovery,
      namespace, platform, prior_context, query, repo, requested_sections,
      server_id, session_key, thread_id."

agent_context_pack {..., requested_sections: ["working_set","recovery"]}
  -> sections_receipt: {"requested":["working_set","recovery"],
                        "served":["working_set"],
                        "requested_not_served":["recovery"]}
```

## Verification

`bunx tsc --noEmit` — clean. `bun test` — **3074 pass, 2 fail**. The 2 failures are the pre-existing `scripts/backup+restore` failures tracked in #537 (`runPgRestore stderr sanitization`, `runPgDump failure handling`); confirmed unrelated by stashing this branch's changes and re-running `bun test scripts/`, which fails the same 2 on clean `origin/main`. `--no-verify` on push was used **solely** for those two pre-existing failures.

**Red-proven, both trees.** Reverting registration to the raw shape (`agentContextPackInputSchema`) fails exactly the new unknown-key tests and nothing else:

- `server/tools/context-pack.test.ts` — 25 pass / **3 fail** reverted, 28 pass / 0 fail fixed.
- `src/tools/__tests__/agent-context-pack.test.ts` — 9 pass / **1 fail** reverted, 10 pass / 0 fail fixed.

The red failure message is the defect stated plainly: `agent_context_pack accepted the request instead of rejecting it — the silent-default defect is back`.

Not run: live service smoke. No service was touched, per the task directive; this is a validation/schema change with no migration and no runtime state.

## Critical Self-Review

- Highest-risk behavior: `.strict()` now rejects requests that previously succeeded. Any existing caller passing an undeclared key — even a harmless one — starts getting a hard `-32602` instead of a silently-stripped success. That is the intended behavior change and the whole point of the issue, but it is a breaking change for a tolerant caller, and the blast radius is every `agent_context_pack`/`agent_reflex_pointers` client.
- Assumptions that could be wrong: that no live caller currently depends on passing extra keys. I verified the in-repo callers (the reflex delegation passes already-parsed args and does not re-enter the parser; the NATS bridge and eval harness pass declared keys only), but I did **not** audit external consumers — mcp2cli, rtech-hermes, or generated skills could pass a stray key I cannot see from here. Second assumption: that `z.core.$ZodObjectParams` error-callback shape is stable across the pinned zod 4.3.6; it is a typed cast and a zod minor bump could move it.
- Missing/weak tests: no test covers the `strictRequestSchema` error callback for a NON-`unrecognized_keys` issue code (it returns `undefined` to fall through to zod's default, verified by the enum-rejection test passing, but not directly asserted). No test drives the `rejectUnknownRequestKeys` defense-in-depth path through a non-MCP direct caller — it is exercised only incidentally. No live-service test; the whole suite is fake-pool.
- Security/permission risk: none identified. The change is strictly *more* restrictive on input and touches no auth, namespace, or read predicate. Rejection happens before dispatch, so an unauthorized-namespace probe with a bogus key now fails at validation rather than at the namespace gate — this leaks no additional information, since the validation error names only the schema's own public vocabulary, which `tools/list` already publishes.
- Migration/deploy risk: no schema migration, no DB change, no state. Deploy risk is the caller-compatibility break above, not the deploy itself. `additionalProperties: false` appearing in `tools/list` changes the advertised contract, so a client that caches tool schemas should refresh.
- Downstream client/runtime risk: **this is the real one.** Per `docs/downstream-rollout.md` this is an MCP tool/schema/client-facing change, so mcp2cli cache/skill refresh and any rtech-hermes or generated-skill caller that constructs pack arguments should be re-checked before this is treated as fully rolled out. I have **not** performed that rollout — it is out of this task's scope (no service touch authorized) and is flagged below as not-complete rather than not-applicable.
- Rollback/cleanup concern: clean single-commit revert; no data or state to unwind. The two `.bak` files used for red-proofing were written under `{temp_workspace}/open-brain/_scratch/` and are not in the repo.
- Fixes made before PR: (1) the original guard was placed in `parseAgentContextPackArgs`, which a live MCP probe proved never runs for stripped keys — moved to registration. (2) The first test set asserted a thrown error; the SDK actually returns an `isError` result — corrected against observed behavior rather than assumption. (3) The stock `.strict()` message named only the offending key, not the accepted set, which fails the issue's explicit requirement — added `strictRequestSchema` so the vocabulary rides in the message, since the serving entrypoint has no summary formatter. (4) A scripted edit left the guard's doc comment attached to the wrong function; reordered in both files.
- Known residual risk: an external caller passing an undeclared key breaks loudly on deploy. That is the requested behavior ("fail loudly and make us think about it"), but it is unmeasured outside this repo — see the downstream rollout box, which is deliberately left unchecked.

`docs/sme/gotcha-agent.md` already carries this defect family from PR #533 ("MCP silently ignores dropped args"). This PR is the same family reaching the read surface; the audit inventory below is the reusable artifact.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: the task directive forbade touching the service; this is a schema/validation change with no runtime state, verified against the real MCP client/server pair in-process.

## Contract Parity

- Contract parity: [x] runtime-specific because: the parity fixtures cover `python/openbrain-memory/`, `clients/ts/`, `contracts/`, `src/contract.ts`, and `src/contract-schemas.ts` (`contracts/parity-paths.txt`); none of the five files changed here is in that set, and the fixtures do not model MCP `inputSchema` or `additionalProperties` at all. `bun test contracts/` passes 13/13 unchanged.

## Downstream Rollout

- [ ] rtech-mcps handoff is complete or not applicable
- [ ] mcp2cli cache/skill refresh is complete or not applicable

Notes/evidence:

- These four are deliberately **unchecked, not marked not-applicable.** This is an MCP tool/schema/client-facing change, so by `docs/downstream-rollout.md` the rollout genuinely applies — but the task authorized no service touch, so it is **not done**. Marking them not-applicable would be false. State: **merged-pending-rollout**, not deployed.
- The specific downstream action: any client that caches tool schemas needs a refresh to pick up `additionalProperties: false`, and any caller constructing pack arguments should be checked for stray keys, which now fail hard instead of being stripped.

## AUDIT — remaining tool surface with the same silent-default shape (#535 follow-up inventory)

Reported, not fixed, per scope. **All 63 remaining tool registrations pass a raw shape**, so every one of them strips unknown top-level keys before dispatch. Stripping alone is not the defect — the defect is stripping that yields a *success-shaped answer a caller reads as truth*. Ranked by that criterion, probed live against the registered server with a fake pool:

**HIGH — a dropped key silently changes the answer, and the answer looks fine**

| Tool | File | Evidence (observed, this session) |
|---|---|---|
| `search_brain` | `server/tools/search-brain.ts:80` | `{query:"x", tabel:"thoughts"}` → `[]`, `isError` unset. The table filter is dropped and an empty result is indistinguishable from "nothing matched". |
| `list_recent` | `server/tools/list-recent.ts:116` | `{tabel:"thoughts", dayz:3}` → `{"entries":[],"total_count":null,"offset":0,"limit":20,"has_more":false}`. Two dropped selectors, a complete-looking envelope. |
| `skill_usage_report` | `server/tools/skill-usage.ts:229` | `{dayz:7}` → `{"window_days":7,...}` — but `7` is the **default**, not the request. The receipt echoes a number that only coincidentally matches, which is worse than omitting it. |
| `repo_fact_search` | `server/tools/repo-facts.ts:160` | Five optional narrowing filters (`namespace`/`repo`/`collection`/`path`/`fact_type`); any typo widens the search silently. Same lane as #517/#526, where a mis-bound repo filter was the actual bug. |
| `scan_namespace` | `server/tools/scan-namespace.ts:75` | `target_namespace`/`table`/`since` optional. A typo'd `target_namespace` scans the wrong namespace and reports success. Namespace-adjacent, so worth prioritising. |

**MEDIUM — dropped key degrades to a default; less likely to be read as a full answer**

`adjacent_context` (`adjacent-context.ts:69`), `citation_recall` (`citation-recall.ts:116`), `find_duplicates` (`find-duplicates.ts:144`), `search_all` (`search-all.ts:213`), `get_stats` (`reporting.ts:170`), `access_report` (`reporting.ts:77`), `session_load` (`session-save-load.ts:86`), `list_stale`/`tier` reads (`tiering.ts:85,208`).

**LOW — writes and single-required-arg tools.** Writes (`capture.ts`, `realtime-append.ts`, `session-lifecycle.ts`, `entities.ts`, `people.ts`, `source-registry.ts`, `tier-mutations.ts`, `update-entry.ts`, `promote-entry.ts`, …) mostly fail on a missing *required* field, so a typo usually surfaces. They are still exposed on **optional** keys — a dropped optional write field lands a row missing metadata and returns `saved`, which is exactly #464/PR #533. `get_contract` and `operator_doctor` register `{}` and take no input.

**Recommended follow-up shape:** promote `strictRequestSchema` (added here in both trees) to a shared registration helper and apply it tool-by-tool, HIGH tier first, each with its own red-proven test. Rejecting is right for read tools with narrowing filters; the #533 `ignored_optional_request_keys` receipt is right where forward tolerance is deliberate. That is a per-surface decision and should not be blanket-applied.

Caveat on scope: the HIGH/MEDIUM ranking is a judgment call from schema shape plus the five probes above, not an exhaustive 65-tool behavioral sweep. The five HIGH rows are observed; the MEDIUM/LOW tiers are inferred from their schemas.

### DIFF
--- a/server/tools/agent-context-pack.ts
+++ b/server/tools/agent-context-pack.ts
@@ -40,8 +40,8 @@ import {
   type WorkingSetScope,
 } from "../realtime/working-set.ts";
 import {
-  agentContextPackInputSchema,
-  agentReflexPointersInputSchema,
+  agentContextPackStrictSchema,
+  agentReflexPointersStrictSchema,
   parseAgentContextPackArgs,
   parseAgentReflexPointersArgs,
   type AgentContextPackArgs,
@@ -669,12 +669,38 @@ export async function buildAgentContextPackPayload(
   if (durableMemorySection) sections.durable_memory = durableMemorySection;
   for (const { key, section } of structuredSections) sections[key] = section;
 
+  // #535 receipt: name what was asked for against what came back, so a section
+  // that was requested and did NOT arrive is visible in the answer itself.
+  //
+  // An unknown top-level KEY is already rejected by name at parse time, and an
+  // unknown VALUE inside `requested_sections` is already rejected by the enum
+  // with the accepted set listed. What neither catches is a section that was
+  // spelled correctly, accepted, and then dropped downstream — budget eviction
+  // being the live case. That still reads as a success-shaped short answer, so
+  // the receipt states the difference rather than leaving the caller to diff
+  // `sections` against their own request from memory.
+  //
+  // `requested: null` means the caller sent no `requested_sections` and took the
+  // documented working_set-only default — distinct from an explicit empty array.
+  const servedSections = Object.keys(sections);
+  const requestedSections = args.requested_sections ?? null;
+
   return {
     payload: {
       schema: "openbrain.agent_context_pack.v1",
       status: "ok",
       scope: { namespace_source: "authorization", ...normalizedScope },
       sections,
+      sections_receipt: {
+        requested: requestedSections,
+        served: servedSections,
+        requested_not_served:
+          requestedSections === null
+            ? []
+            : requestedSections.filter(
+                (name) => !servedSections.includes(name),
+              ),
+      },
       warnings: {
         scope_denials: [
           ...(workingSet ? workingSet.warnings.scope_denials : []),
@@ -850,7 +876,10 @@ export function registerAgentContextPackTool(
         "emitted as durable_memory items, reusing the single durable_memory recall; " +
         "candidate_memory is a truthful empty section (no candidate predicate yet) " +
         "that never drives its own recall.",
-      inputSchema: agentContextPackInputSchema,
+      // The STRICT object, not the raw shape: the SDK strips unknown keys from
+      // a raw shape before dispatch, which is the #535 silent-default defect.
+      // See `context-pack-args.ts` `rejectUnknownRequestKeys`.
+      inputSchema: agentContextPackStrictSchema,
       annotations: {
         title: "Agent Context Pack",
         readOnlyHint: true,
@@ -888,7 +917,7 @@ export function registerAgentReflexPointersTool(
         "through the authorized read path (get_entry, table = source_ref.type + " +
         '"s", id = source_ref.id). Placement into the model prompt is client-owned; ' +
         "this tool never performs implicit _meta injection.",
-      inputSchema: agentReflexPointersInputSchema,
+      inputSchema: agentReflexPointersStrictSchema,
       annotations: {
         title: "Agent Reflex Pointers",
         readOnlyHint: true,
--- a/server/tools/context-pack-args.ts
+++ b/server/tools/context-pack-args.ts
@@ -137,11 +137,122 @@ export const agentContextPackInputSchema = {
   budget: contextPackBudgetInputSchema,
 };
 
+/**
+ * Build the strict registration schema for a tool surface.
+ *
+ * The `.strict()` alone rejects the key but its stock message names only the
+ * offending key. The issue's requirement (and the #431/PR #532 pattern) is that
+ * a failure names the ACCEPTED VOCABULARY too, so the caller has something to
+ * correct toward rather than a bare "no". The accepted set is derived from the
+ * shape, never re-listed, so it cannot drift from the schema it describes.
+ *
+ * The message must carry the vocabulary itself, not rely on a formatter: the
+ * serving entrypoint (`server/main.ts`) does NOT install
+ * `installValidationSummaryFormatter` — only `src/server.ts` does — so anything
+ * that lives only in the formatter is invisible on the surface that actually
+ * runs.
+ */
+function strictRequestSchema<Shape extends z.ZodRawShape>(
+  shape: Shape,
+  toolName: string,
+) {
+  const accepted = Object.keys(shape).sort().join(", ");
+  return z
+    .object(shape, {
+      error: (issue: { code?: string; keys?: string[] }) =>
+        issue.code === "unrecognized_keys"
+          ? `Unrecognized key(s) for ${toolName}: ${(issue.keys ?? []).join(", ")}. ` +
+            `Accepted keys: ${accepted}.`
+          : undefined,
+    } as z.core.$ZodObjectParams)
+    .strict();
+}
+
+/**
+ * Reject an unknown top-level request key by NAME, listing the accepted set.
+ *
+ * THE DEFECT (#535). The MCP SDK builds its validating schema from a raw shape
+ * with a plain `z.object()`, which STRIPS undeclared keys instead of rejecting
+ * them, and it strips them BEFORE the handler runs. So `sections: [...]` — the
+ * near-miss spelling of `requested_sections` — never reached the tool at all:
+ * the pack saw no section selection, fell back to its working_set-only default,
+ * and returned `status: "ok"` on a near-empty answer. A typo in one key looked
+ * exactly like a correct minimal request, and it cost real debugging time at
+ * least twice (#439, #526).
+ *
+ * THE FIX. Registration passes the `.strict()` object below as `inputSchema`
+ * instead of the raw shape, so the SDK's own pre-dispatch validator does the
+ * rejecting. That is the only layer that still sees the raw arguments — a
+ * handler-side check cannot work, because by then the key is already gone. It
+ * also makes the advertised JSON schema honest: `additionalProperties: false`
+ * now appears in `tools/list`, so a caller can see the rule before breaking it.
+ *
+ * REJECT, don't tolerate — for THIS surface specifically. An unknown key here
+ * is always a caller mistake, never forward-tolerance: every key the pack
+ * accepts selects work it must do, so an ignored one silently changes the
+ * answer. That is the opposite of the capture lane (#464/PR #533), where
+ * genuinely future lifecycle keys are expected and naming what was ignored in
+ * the receipt is the right shape. Forward tolerance is a per-surface decision,
+ * and this surface does not want it.
+ *
+ * This function stays as defense in depth for the DIRECT callers of
+ * `parseAgentContextPackArgs` (tests, the NATS bridge, any non-MCP entrypoint),
+ * which never pass through the SDK validator. Its error is a real `ZodError`,
+ * so `summarizeValidationError` renders it through the same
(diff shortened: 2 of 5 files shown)


rodaddy:
### software-factory: failed
Reviewed at: `8028a61c31969b4c0ee59d0b74fd6cf655766ec8`

Findings:
- [medium] Wire\-observable contract change ships without updating the published get\_contract surface (src/contract\.ts:143,439,773 \(vs server/tools/agent\-context\-pack\.ts:672\-681\))
- [medium] The receipt's own motivating case — budget eviction — is not covered by any test (server/tools/context\-pack\.test\.ts:820\-838)
- [low] New always\-present envelope field is not accounted for in CONTEXT\_PACK\_ENVELOPE\_CHAR\_RESERVE (server/tools/context\-pack\-durable\-lane\.ts:20)
- [low] RED PROOF comment names the wrong layer and contradicts the PR's own central finding (server/tools/context\-pack\.test\.ts:674\-678)
- [low] strictRequestSchema and rejectUnknownRequestKeys duplicated verbatim across both trees (src/tools/agent\-context\-pack\.ts:170\-295)
- [low] rejectUnknownRequestKeys is unreachable in production, untested, and emits a different error shape (server/tools/context\-pack\-args\.ts:150\-174)
- [info] Breaking input\-contract change with downstream rollout deliberately not performed (server/tools/agent\-context\-pack\.ts:879,920)
- [info] Coupling to zod internals and to caret\-ranged SDK behavior (server/tools/context\-pack\-args\.ts:99\-113)
- [info] src tree lacks the tools/list additionalProperties assertion the server tree has (src/tools/\_\_tests\_\_/agent\-context\-pack\.test\.ts:461\-509)
- [info] requested\_not\_served states the shortfall but not its cause, and does not dedupe (server/tools/agent\-context\-pack\.ts:672\-681)

Issue #519, opened 2026-08-03
Startup lane resume: emission two carries the repo lane's checkpoint + recent decisions


rodaddy:
Operator direction 2026-08-03: "startup should already know most of this stuff and then get the basic information about the repo it's in" / "I don't want to see bash this, bash that — that shit just happens." This amends the 2026-07-29 explicit-on-request ruling for episodic history (canon.md; #443 body), by operator decision: a REPO-SCOPED lane resume auto-loads; cross-lane history stays explicit-on-request. The contamination concern the ruling protected against is handled by scope, not by omission.

## Design (delta on the existing pieces, all shipped)

1. **Derive `session_key` like #517 derives `repo`**: `dev:<derived-repo>` when `OPENBRAIN_CANON_SESSION_KEY` is not explicitly set (same `model_fields_set` + `_derive_repo_slug` pattern, `apps/hooks/session.py`). Today every repo's captures land in `dev:open-brain`; open-brain derivation is identity so nothing moves for it, other repos start honest per-repo lanes going forward.
2. **Emission two appends a `lane_resume` section** after `repo_facts` (`openbrain-session-start-remaining`): one `session_context` call for the derived lane with a small `event_limit` (the argument #516 restored), rendered like `resume.py --brief` — last checkpoint line + the most recent day's decisions/blockers, ~2-4k chars.
3. Empty lane → say so in one line ("no lane history for this repo yet"), never another lane's events — same honesty rule as #517's empty state.
4. Capture side uses the same derived session_key so writes and reads agree on the lane.

## Consequences to verify

- `resume.py --lane dev:open-brain` default unaffected; other repos' deep history remains reachable by explicit `--lane dev:open-brain` (where it historically landed).
- Receipts/context-budget-gate interplay: the recall receipt should record the actual lane read.
- Adapter lanes #511/#512 inherit this (Pi/Codex sessions get lane resume for free).

Blocks nothing; sequenced after #517 (merged, PR #518) whose derivation it reuses.

PR #544, opened 2026-08-04
fix(config): declare the LAN plain-http opt-in so remote hooks can reach the brain (#525)

### LINKED ISSUE #525: Remote-host hook clients cannot reach the dev brain: strict config has no sanctioned plain-http-over-LAN opt-in
## Symptom (live, 2026-08-03)

Skippy moved to a Claude engine under Buzz (`claude-agent-acp`, cwd `~/.buzz`) and wakes with the policy-refresh hook firing but ZERO Open Brain hydration — no CANON PACK, no recall. "The hook told me the rules and then handed me an empty skull."

## Two stacked causes

1. **The Claude-family OB stack was never deployed to that box.** The CANON PACK is injected by the `openbrain` Python package's SessionStart entrypoints (installed uv tool + `claudex-observation.env` + `openbrain-hook-env` wrapper + settings.json wiring), which exist only on the dev Mac (#420 cutover). Skippy's host is fleet **family B** (Hermes plugin route) in `_plans/fleet-rollout-0.9.md:478` — the Buzz switch moved him into the Claude family without the Claude-family rollout. The silent absence is the designed decline: unconfigured canon injects nothing and exits 0. (`~/.buzz` having no `.git` is NOT the cause — that only empties `repo_facts`; profile/process would still inject.)

2. **Even after installing the stack there, it cannot connect.** `openbrain_memory.client._validate_base_url` permits plain http only for loopback hosts unless `allow_insecure_http=True` — and the strict `openbrain` config (`unknown_prefixed_variables`) rejects `OPENBRAIN_ALLOW_INSECURE_HTTP` as a typo, which is why the wrapper strips it. So a hook process pointing at `http://10.71.1.20:3100` from any LAN box silently declines canon AND capture (proven 2026-08-02, recorded in `claudex-observation.env`'s own comments). Every remote Claude-family client hits this wall; the Air works only because it goes through its own env/adapter path, not the strict hook config.

## What this needs

- A sanctioned way for HOOK processes on LAN clients to reach the dev/core01 brain: either TLS on the brain endpoints (Caddy in front of 3100, fleet-standard) or a DECLARED insecure-LAN opt-in field on the strict config sections (capture/canon/observation) that flows to the client's `allow_insecure_http` — explicit, named, and documented in `docs/CONFIG_REFERENCE.md`, not the stripped env var.
- The Buzz/claude-agent-acp launch path added to the Claude-family rollout steps in `_plans/fleet-rollout-0.9.md` (Skippy is the first family-B agent to cross families).

## Parked operator decision (not this issue's to make)

Whether Skippy's memories live in his own namespace or rico's. The auth model already supports per-consumer tokens, so "skippy token + own namespace (+ shared-kb for common knowledge)" is the natural shape — Rico decides.

## Related

- #517 (repo_facts cwd derivation), #420 (cutover), `_plans/fleet-rollout-0.9.md` (family map), 2026-08-02 proven silent-decline incident.

rodaddy:
Closes #525 (the config/plumbing half; the parked namespace decision is Rico's and is untouched).

## Summary

- A Claude-family agent on a LAN box woke with the policy hook firing and **zero** Open Brain hydration — no CANON PACK, no capture — and nothing anywhere said why. Two walls, stacked, each silent.
- `openbrain_memory.client._validate_base_url` refuses a non-loopback plain-`http` base URL unless the caller passes `allow_insecure_http=True`. The strict `openbrain` config declared no such field, so exporting `OPENBRAIN_ALLOW_INSECURE_HTTP` made `unknown_prefixed_variables` reject the **whole** environment as a typo — which the hook entrypoints swallow into a clean exit 0.
- So a LAN host had no legal way to say it: setting the variable killed every lane, not setting it left the client refusing the URL. The deployed `openbrain-hook-env` wrapper stripped the variable precisely to avoid the first failure, which locked in the second.
- **Fix:** declare `allow_insecure_http` on `CaptureSettings` and `CanonSettings`, binding **one** environment name — the same name `openbrain_memory.runtime` already reads — and flow it to all five client construction sites. This is the declared-opt-in route the issue names, **not** the Caddy/TLS route.
- This is the **third** time a variable the hook environment can carry had to be declared even though this package barely reads it (after `OPENBRAIN_OBSERVATION_*` and `OPENBRAIN_SPOOL_PATH`), so the rationale lives on `ALLOW_INSECURE_HTTP_ALIASES` rather than in a commit message nobody will find.
- **The default is untouched.** Unset, loopback-only stands exactly as before, so a public endpoint still has to be `https`. Turning it on is an explicit, per-host, documented choice scoped to the LAN-internal pre-production posture.

Construction sites covered (all five; a site left off the flow is a lane that declines silently on a LAN box, which is this bug again in miniature):

| site | lane |
|---|---|
| `apps/hooks/session.py::_started_memory` | capture spine — every turn |
| `apps/hooks/session.py::_record_skill_usage` | `PostToolUse` skill metric |
| `apps/hooks/session.py::_canon_context` | the CANON PACK read |
| `apps/hooks/session_start.py::_lane_resume_text` | emission two (#519) |
| `apps/canon/run.py::_apply` | operator canon reconcile |
| `apps/bulk/run.py::_lane` | operator bulk ingest |

`ObservationSettings` needs no flag — that sink ships to Langfuse via its own SDK, not `OpenBrainClient`. `openbrain-provider` builds no clients (pure policy gates).

## Red-proof

**Pre-fix refusal, both walls, reproduced first:**

```
load_capture_settings     RAISED: unrecognised Open Brain environment variable(s): OPENBRAIN_ALLOW_INSECURE_HTTP
load_canon_settings       RAISED: unrecognised Open Brain environment variable(s): OPENBRAIN_ALLOW_INSECURE_HTTP
load_observation_settings RAISED: unrecognised Open Brain environment variable(s): OPENBRAIN_ALLOW_INSECURE_HTTP
client REFUSED: OpenBrainClient base_url must use https, localhost http, or allow_insecure_http=True
```

**New tests against the pre-fix source** (`git stash` of `python/openbrain/src`): **14 failed, 2 passed**. The 2 that pass are the ones asserting the *unchanged default* (`test_a_real_non_loopback_http_endpoint_is_refused_without_the_opt_in`, `test_a_misspelled_opt_in_is_still_rejected`) — they must pass both ways, which is the point of including them. Restored: **16 passed**.

**The construction-site assertions are not vacuous.** Deleting a single `allow_insecure_http=` line from `apps/canon/run.py` fails exactly its own test — `KeyError: 'allow_insecure_http'`, 1 failed / 15 deselected — and nothing else.

**Non-loopback http is proven over a real socket, not a string check.** `test_a_real_non_loopback_http_endpoint_answers_under_the_opt_in` binds an `HTTPServer` on this machine's real LAN address — resolved to `10.71.1.20`, the exact endpoint shape from the issue — and drives a real `OpenBrainClient` built from the declared setting to a `200`. The paired test proves the same live endpoint is still **refused** without the opt-in, so reachability is not what the rule is answering.

## Follow-up fix in this PR: the suite hung on an operator's machine

Reported after the first push: `uv run pytest -q` **hung indefinitely** at test 8 of 45 of
`test_capture_outage_notice.py`, reproducibly, while the same file passed on merge-base `main`.
The test file is byte-identical between the two commits, so the branch was the suspect.

**Root cause — this branch unmasked a pre-existing leak; it did not introduce one.**
`capture_stop_with` accepts an injected `CaptureSettings` but resolves the *observation* section
itself, from the live process environment, on every call. On a machine whose shell carries the four
provisioned coordinates — which `~/.local/share/openbrain-memory/env/claudex-observation.env`
supplies, so any hook-configured host has them — the unit suite built a **real `LangfuseEmitter`
and posted its fixture transcripts to the production fleet Langfuse server**. `emit` ends in
`client.shutdown()`, which blocks draining the OpenTelemetry batch worker, so the suite waited on
the network forever rather than failing.

`unknown_prefixed_variables` had been acting as an accidental network guard: with
`OPENBRAIN_ALLOW_INSECURE_HTTP` undeclared, the settings load *raised*, `loaded_observation()`
swallowed the raise, and the sink stayed silently off. Declaring the variable — the entire point of
this PR — let the load succeed, and the observation lane it had been shadowing started dialing.
CI never saw it because CI sets no observation coordinates, so `observation_active` is False there
either way.

**Faulthandler stack at the stall (abridged):**

```
capture_stop_with (stop.py:145)
  -> asyncio.run -> run_stop -> _observe_best_effort
  -> LangfuseEmitter.emit -> client.shutdown()
Thread: opentelemetry/sdk/_shared_internal/__init__.py:156 in worker
Thread: langfuse/_utils/prompt_cache.py:50 in run
```

**Fix, at the owning boundary:** the autouse `_clean_environment` fixture that `test_settings` has
carried file-scoped since it was written is promoted to the shared `tests/conftest.py`, so it
applies to the whole suite. `OPENBRAIN_TEST_*` is deliberately preserved — the `-m live` gate
addresses its playground service through that prefix, and clearing it would make the live gate pass
having run nothing.

**Red-proof:** with the fixture at `autouse=False` in an operator shell, the two new
`TestTheSuiteNeverInheritsAnOperatorsSink` tests fail (`assert True is False`; `assert
['OPENBRAIN_ALLOW_INSECURE_HTTP', ...] == []`). With it restored they pass, the previously hanging
file passes **3/3**, and the full suite reaches a real summary line — **565 passed, 19 deselected in
13.6s** — in the exact environment that hung.

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed — `uv run pytest -q` in `python/openbrain`: **565 passed, 19 deselected**; pre-push full suite passed with **no bypass**
- [x] Python package checks passed or are not applicable — `uv run mypy src/openbrain` clean (49 files); `uv run ruff check src tests` clean. Sibling `openbrain-memory` also re-checked (mypy clean, 17 files; ruff clean) though untouched
- [x] `bunx tsc --noEmit` — not applicable: no TypeScript touched (`git diff --name-only -- 'src/*.ts'` empty)
- [x] Live Open Brain smoke passed — a real `SessionStart` driven through the **deployed wrapper** emitted a real CANON PACK with non-zero section counts (`profile_guidance=10, process_guidance=22`, `namespace=rico`), exit 0

## Critical Self-Review

- Highest-risk behavior: relaxing a transport security check. Scoped hard — the field defaults `False`, the client's loopback-only rule is untouched, and the opt-in only takes effect when a host explicitly sets one documented variable. A public endpoint still requires `https`. I did not weaken `_validate_base_url` itself, which is where a careless fix would have gone.
- Assumptions that could be wrong: that the LAN-internal pre-prod posture is the intended route rather than TLS. The issue offers both and the operator direction for this run named the opt-in explicitly; if that posture changes, this field becomes the thing to remove, not a thing to widen. Also assumed `ObservationSettings` needs no flag — verified it emits through the Langfuse SDK, not `OpenBrainClient`.
- Missing/weak tests: the socket test proves a `/health` round trip, not a full MCP session against a real brain over LAN — the genuine end-to-end receipt is a LAN box (the Air, `10.71.1.26`) hydrating, which is deployment, not a test. The five construction-site tests assert the constructor argument, not that each lane's subsequent traffic succeeds. Added after the reported hang: the two `TestTheSuiteNeverInheritsAnOperatorsSink` tests assert the environment the hook resolves is inert, red-proven by flipping the fixture to `autouse=False`. They cannot assert the hang itself — a test that hangs to prove a hang reports nothing and blocks the suite behind it — so they pin the precondition (`observation_active` False, no `OPENBRAIN_*` leaked) instead of the symptom.
- Security/permission risk: this permits plain-text bearer tokens over the LAN when enabled. That is the accepted pre-prod posture for `10.71.x` and is stated as such in the doc rather than left implicit; it is off by default and per-host. No namespace, auth, or server-side check is touched.
- Migration/deploy risk: the real risk here was **ordering**, and it is the one that bites silently. Passing the variable through the wrapper against an install that does not declare it rejects the whole environment and zeroes every lane. Verified empirically in the safe order — old install raised `UnknownEnvironmentVariableError`, then reinstall, then the installed interpreter resolved `allow_insecure_http=True`, then the wrapper edit, then a live CANON PACK. The rule is now written into `CONFIG_REFERENCE.md`, `420-cutover-rollback.md`, and the rollout plan so the next person does not have to rediscover it.
- Downstream client/runtime risk: none — no MCP tool, schema, transport, or wire surface changes. `openbrain-memory` is unmodified; it already read this variable.
- Rollback/cleanup concern: reverting the commit restores the old strict config, at which point the **deployed wrapper must also drop the pass-through** or every hook on a box with the variable set goes to silent zero capture. That coupling is the reason it is called out in three docs. The wrapper lives outside this repo (`~/.local/share/openbrain-memory/env/openbrain-hook-env`) and cannot be versioned here.
- Fixes made before PR: reinstalled the `openbrain` uv tool from this branch *before* touching the wrapper, after proving the old install rejected the variable — doing it the other way round would have broken live capture on this machine. Replaced a ruff-flagged inline token literal with a named fixture constant rather than blanket-suppressing the rule. After the hang report: promoted the env-isolation fixture to the shared conftest rather than adding a narrow guard to the one file that happened to hang, because the leak is in every hook entrypoint that resolves a section its caller did not inject.
- Known residual risk: the fleet rollout plan's family-A/LAN notes are updated, but the Buzz/`claude-agent-acp` launch-path steps (the issue's second bullet) are **not** added — I have not observed that host and will not write rollout steps I cannot verify. Filed as follow-up below. Also: the wrapper change is live on this machine only; every other family-A box needs the same two-part change in the same order. Also: the env-isolation fixture fixes the TEST suite, not the underlying design — `capture_stop_with` still resolves the observation section from the ambient process environment at runtime, which is correct for a hook but means any future in-process caller inherits the same coupling.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no new review-lane pattern arose — this is the third instance of an already-documented gotcha (undeclared sibling-owned env var → silent zero capture), and the rationale is recorded at the rule site in code plus `CONFIG_REFERENCE.md`, which is where the previous two instances put it

## Review Gate

- Live Open Brain checks: [x] linked below or [ ] not applicable because: real `SessionStart` through the deployed wrapper returned a CANON PACK with non-zero section counts; no server, transport, MCP tool, or schema surface is touched

## Contract Parity

- Contract parity: [x] runtime-specific because: `CaptureSettings` / `CanonSettings` are Python-side hook configuration with no contract fixture, wire representation, or client binding. The client keyword `allow_insecure_http` it feeds is pre-existing `openbrain_memory` API, unchanged by this PR.

## Downstream Rollout

Notes/evidence:

- No MCP tool, schema, transport, or client-facing surface is modified. Changes are the Python `openbrain` hook config plus three docs.
- **Deploy coupling, outside this repo:** `~/.local/share/openbrain-memory/env/openbrain-hook-env` was updated in place to pass `OPENBRAIN_ALLOW_INSECURE_HTTP` through, and its comments now explain why it used to strip it. Every other family-A box needs the same change, in the same order (declare → reinstall → verify → wrapper).

**Follow-up (not in this PR):** the Buzz / `claude-agent-acp` launch path still needs adding to the Claude-family rollout steps in `_plans/fleet-rollout-0.9.md` — the issue's second bullet. Deliberately deferred: writing rollout steps for a host I have not observed would be a guess in the grammar of a procedure.

### DIFF
--- a/_plans/fleet-rollout-0.9.md
+++ b/_plans/fleet-rollout-0.9.md
@@ -177,9 +177,18 @@ the wrong brain. Before #74, a session with no `OPENBRAIN_*` silently hydrated
 from core01 — two sessions chased a phantom core01 outage over exactly that.
 
 `OPENBRAIN_ALLOW_INSECURE_HTTP` is present because the local/LAN endpoint is
-plain `http`. **UNVERIFIED for cc-\* boxes:** whether cross-box plain-HTTP bearer
-tokens over the LAN are acceptable to Rico, or whether the boxes must go through
-`https://open-brain.rodaddy.live`. §6.
+plain `http`, and as of #525 it is a **declared** setting on both
+`CaptureSettings` and `CanonSettings` that the wrapper passes through — see §1.3.
+Before that fix it was declared nowhere and stripped by the wrapper, which is why
+a LAN box could not reach the brain at all: the client refuses non-loopback plain
+`http` without the flag, and exporting the flag made the strict config reject the
+whole environment. A LAN box now needs this variable set; on `127.0.0.1` it is
+inert (loopback http is auto-permitted).
+
+**UNVERIFIED for cc-\* boxes:** whether cross-box plain-HTTP bearer tokens over
+the LAN are acceptable to Rico, or whether the boxes must go through
+`https://open-brain.rodaddy.live`. §6. The declared opt-in makes plain-HTTP LAN
+possible; it does not decide that it is the chosen posture for those boxes.
 
 ### 1.3 The hook-env wrapper — EXISTS, and it is not optional
 
@@ -193,15 +202,29 @@ exec env -i \
   PATH="$PATH" HOME="$HOME" \
   OPENBRAIN_BASE_URL="${OPENBRAIN_BASE_URL:-}" \
   OPENBRAIN_TOKEN="${OPENBRAIN_TOKEN:-}" \
+  OPENBRAIN_OBSERVATION_*="..." \
+  OPENBRAIN_ALLOW_INSECURE_HTTP="${OPENBRAIN_ALLOW_INSECURE_HTTP:-}" \
   "$@"
 ```
 
-**Why it must be copied and not skipped** (`docs/420-cutover-rollback.md:46-56`):
+**Why it must be copied and not skipped** (`docs/420-cutover-rollback.md`):
 the Python config **rejects** any `OPENBRAIN_*` variable it does not declare
 (`config.unknown_prefixed_variables`), *and the hooks swallow that rejection*.
 So sourcing the whole env file into a hook would **silently zero every
 capture** — a green-looking box writing nothing. The wrapper's `env -i` passes
-exactly the two accepted variables.
+exactly the accepted variables and nothing else.
+
+**The pass-through list grows only AFTER the installed package declares the
+field** — declare in `openbrain.config`, `uv tool install --reinstall`, verify
+the installed interpreter accepts it, then edit the wrapper. Getting that order
+backwards is itself the silent-zero-capture failure. `OPENBRAIN_OBSERVATION_*`
+(#523) and `OPENBRAIN_ALLOW_INSECURE_HTTP` (#525) were both added this way.
+
+**A LAN box needs `OPENBRAIN_ALLOW_INSECURE_HTTP` in its env file** (the Air,
+`10.71.1.26`, pointing at `http://10.71.1.20:3100`). Without it the client
+refuses the non-loopback plain-`http` base URL and canon *and* capture decline
+silently — the #525 symptom. This is the family-A delta a LAN box has that the
+dev Mac does not, because loopback needs no flag.
 
 This is the single highest-risk thing to get wrong on a new box, because the
 failure is silent. The Step-5 gate below is written specifically to catch it by
--- a/docs/420-cutover-rollback.md
+++ b/docs/420-cutover-rollback.md
@@ -50,12 +50,22 @@ handled independently. No `sha256-` path is involved; a `uv tool install
 The Python config REJECTS any `OPENBRAIN_*` variable it does not declare
 (`config.unknown_prefixed_variables`), and the hooks swallow that rejection —
 so sourcing the whole `claudex-observation.env` (which carries
-`OPENBRAIN_OBSERVATION_*`, `OPENBRAIN_ALLOW_INSECURE_HTTP`, `OPENBRAIN_NAMESPACE`)
-would silently zero every capture. The wrapper
+`OPENBRAIN_NAMESPACE`) would silently zero every capture. The wrapper
 `~/.local/share/openbrain-memory/env/openbrain-hook-env` sources that file and
-passes the child ONLY the two accepted variables (`OPENBRAIN_BASE_URL`,
-`OPENBRAIN_TOKEN`) via `env -i`. The token stays in the env file; it is never
-written into `settings.json`.
+passes the child ONLY the accepted variables via `env -i`: `OPENBRAIN_BASE_URL`,
+`OPENBRAIN_TOKEN`, `OPENBRAIN_OBSERVATION_*` (#523), and
+`OPENBRAIN_ALLOW_INSECURE_HTTP` (#525). The token stays in the env file; it is
+never written into `settings.json`.
+
+**The pass-through list only ever grows AFTER the installed package declares the
+field.** A variable added here against an older install makes
+`unknown_prefixed_variables` reject the whole environment, and the hooks swallow
+that into a silent zero capture — so the order is: declare in
+`openbrain.config`, `uv tool install --reinstall`, verify the installed
+interpreter accepts it, then edit the wrapper. Verified in that order for #525 on
+2026-08-04 (old install raised `UnknownEnvironmentVariableError`; reinstalled
+one resolved `allow_insecure_http=True`). See `docs/CONFIG_REFERENCE.md`
+("LAN plain-HTTP opt-in").
 
 ## Verify the NEW path is ACTIVE
 
--- a/docs/CONFIG_REFERENCE.md
+++ b/docs/CONFIG_REFERENCE.md
@@ -398,6 +398,50 @@ pass-throughs only at or after the moment the installed package includes
 `unknown_prefixed_variables` reject the whole environment, which the hooks
 swallow into a silent zero capture.
 
+### LAN plain-HTTP opt-in (#525) — Python `openbrain` hooks
+
+Declared on `CaptureSettings` and `CanonSettings`
+(`python/openbrain/src/openbrain/config.py`, `ALLOW_INSECURE_HTTP_ALIASES`), and
+flowed to `OpenBrainClient(allow_insecure_http=...)` at every site the hook
+stack builds a client.
+
+| variable | default | notes |
+|---|---|---|
+| `OPENBRAIN_ALLOW_INSECURE_HTTP` | `false` | permit a plain-`http` endpoint whose host is **not** loopback. One name, bound by both sections, and the same name `openbrain_memory.runtime` already reads |
+
+**What it is scoped to: a LAN-internal, pre-production posture.** The brain
+listens on a private address the operator controls, and putting TLS in front of
+3100 is a separate and larger piece of work (#525 names both routes; this is the
+declared opt-in route). It does **not** weaken the default — unset, the client's
+loopback-only rule stands exactly as before, so a public endpoint still has to be
+`https`. Turning it on is an explicit, per-host, documented choice.
+
+**Why it had to be declared rather than just exported.** The client refuses a
+non-loopback plain-`http` base URL unless the caller passes
+`allow_insecure_http=True` (`openbrain_memory.client._validate_base_url`) — and
+an `OPENBRAIN_`-prefixed variable matching no declared field is rejected by
+`unknown_prefixed_variables` as a typo, which the hook entrypoints swallow into a
+silent zero capture. So a LAN host had no legal way to say it: exporting the
+variable killed the whole environment, and not exporting it left the client
+refusing the URL. Measured 2026-08-02 (#525): a hook process pointing at
+`http://10.71.1.20:3100` declined canon **and** capture with no error anywhere —
+a Claude-family agent on a LAN box woke with the policy hook firing and zero
+hydration.
+
+**Deploy coupling — the wrapper must pass it through.** The deployed
+`openbrain-hook-env` wrapper (see `docs/420-cutover-rollback.md`) hands hooks only
+the variables it lists explicitly, and it stripped this one deliberately, because
+before this change the installed package rejected it. The wrapper adds the
+pass-through only at or after the moment the installed package declares the
+field. This is the same ordering rule the `OPENBRAIN_OBSERVATION_*` and
+`OPENBRAIN_SPOOL_PATH` notes carry: **the package declares first, the wrapper
+passes second.** Reversing it makes `unknown_prefixed_variables` reject the whole
+environment, which the hooks swallow into a silent zero capture.
+
+Both halves are required. Declaring the field without the wrapper pass-through
+leaves the hook process never seeing the variable; passing it through without the
+declaration is the rejection above.
+
 ### Sibling-package variables that must still be declared
 
 | variable | default | notes |
--- a/docs/CONFIG_REFERENCE.md
+++ b/docs/CONFIG_REFERENCE.md
@@ -442,6 +442,34 @@ Both halves are required. Declaring the field without the wrapper pass-through
 leaves the hook process never seeing the variable; passing it through without the
 declaration is the rejection above.
 
+**An EMPTY value means unset, and both halves enforce that (PR #544).** The
+wrapper's pass-through style is `VAR="${VAR:-}"`, which cannot express "absent" —
+it turns an unset variable into an **empty string** in the child. For a string
+setting that is harmless; for this `bool` it was not. Measured 2026-08-04 against
+the installed binary: with the variable omitted from `claudex-observation.env`,
+the child received `OPENBRAIN_ALLOW_INSECURE_HTTP=""`, pydantic's bool parser
+rejected it (`Input should be a valid boolean … input_value=''`), **both**
+`load_capture_settings` and `load_canon_settings` raised, the entrypoints
+swallowed the raise, and the hook exited 0 having captured and injected nothing —
+the #525 defect class, re-armed by the fix for #525 on every host that never
+opted in. Only the loaders reproduce it; a bare `CaptureSettings()` reads no
+environment and looked healthy throughout.
+
+Fixed at the owning boundary — a `mode="before"` validator on the field in both
+declaring sections maps a blank string to the default `False`. This is the same
+empty-means-unset reading `OPENBRAIN_SPOOL_PATH`
+(`apps.capture.outage.default_spool_path`) and `XDG_STATE_HOME`
+(`receipts.state.default_receipt_state_path`) already use. It is narrow on
+purpose: `1`/`true` still enable, `false` still disables, a garbage **value**
+like `maybe` still raises, and a misspelled **name** is still caught by
+`unknown_prefixed_variables`.
+
+The wrapper carries the second layer: it passes the variable only when non-empty,
+prepending `NAME=VALUE` to the positional list rather than listing it in `env -i`.
+That is what protects an **older installed package** that predates the validator.
+Any future non-string pass-through belongs in that conditional block, not the
+`env -i` list, for the same reason.
+
 ### Sibling-package variables that must still be declared
 
 | variable | default | notes |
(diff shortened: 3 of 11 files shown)


rodaddy:
## Empty-string opt-in re-armed the #525 defect class — fixed at the owning boundary

**Finding (adversarially confirmed against the installed binary, not inferred).**
The deployed `openbrain-hook-env` wrapper passes every variable as
`VAR="${VAR:-}"` — a style that **cannot express "absent"**. It hands the child an
*empty string* instead. Harmless for the string pass-throughs; not for the new
`bool`. With `OPENBRAIN_ALLOW_INSECURE_HTTP` omitted from the env file, the child
received `OPENBRAIN_ALLOW_INSECURE_HTTP=""` and:

```
ValidationError: 1 validation error for CaptureSettings
allow_insecure_http
  Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='', input_type=str]
```

`load_capture_settings` **and** `load_canon_settings` both raised, the hook
entrypoints swallowed it under the fail-open observer contract, and the hook
exited **0 having captured and injected nothing** — silent zero capture / zero
injection, the exact #525 defect class, re-armed by the fix for #525 on any
machine that never opted in.

Measured directly against the old wrapper style, which emits the empty string:

```
=== OLD wrapper style, env file WITHOUT the opt-in ===
OPENBRAIN_ALLOW_INSECURE_HTTP=<EMPTY STRING>
```

Worth noting why the existing tests missed it: **only the loaders reproduce it.**
A bare `CaptureSettings()` reads no environment and looked healthy throughout.

**Fix — at the owning boundary (`config.py`).** A `mode="before"` validator on the
field in *both* declaring sections (`CaptureSettings`, `CanonSettings`) maps a
blank string to the default `False`. This is the repo's **established** reading of
an empty env var, not new leniency — `OPENBRAIN_SPOOL_PATH`
(`apps.capture.outage.default_spool_path`) and `XDG_STATE_HOME`
(`receipts.state.default_receipt_state_path`) already fall back rather than
resolving one. It stays deliberately narrow:

| input | result |
|---|---|
| `""`, `"   "` | opt-in **OFF** (loads clean) |
| `"1"`, `"true"` | opt-in **ON** |
| `"false"` | opt-in **OFF** |
| `"maybe"` (garbage *value*) | still **raises** |
| `OPENBRAIN_ALLOW_INSECURE_HTTPS` (typo'd *name*) | still **raises** `UnknownEnvironmentVariableError` |

**Red proof.** New cases fail on branch head `37c0670` with the exact signature,
pass with the validator:

```
FAILED test_an_empty_opt_in_loads_with_the_opt_in_off[empty-capture]
FAILED test_an_empty_opt_in_loads_with_the_opt_in_off[empty-canon]
FAILED test_an_empty_opt_in_loads_with_the_opt_in_off[whitespace-capture]
FAILED test_an_empty_opt_in_loads_with_the_opt_in_off[whitespace-canon]
4 failed, 5 passed
```

The 5 passing are the **controls** — enabling values and garbage rejection pass
both before *and* after, so the new tests are not tautologies. After the fix:
`25 passed`.

**Gates:** `mypy src/openbrain` → no issues, 49 files · `ruff check src tests` →
all passed · `pytest` → **563 passed, 19 deselected**.

**Wrapper hardened (belt and suspenders).** The deployed wrapper now passes the
variable only when non-empty, prepending `NAME=VALUE` to the positional list
rather than listing it in `env -i`. That layer is what protects an **older
installed package** predating the validator. It lives outside this repo;
`docs/CONFIG_REFERENCE.md` records the coupling and the rule that future
non-string pass-throughs belong in the conditional block, never the `env -i` list.

**Live reinstall proof — RUNNING.** `uv tool install --force ./python/openbrain`
from this branch, then a real `SessionStart` with the variable **removed from the
environment**:

```
CANON PACK 1/2 | sections: profile_guidance=10, process_guidance=22
CANON PACK 2/2 | sections: repo_facts=7
```

Section counts > 0 is the agent-side hydration proof. Both layers verified
independently: the **validator alone** also rescues the old empty-string wrapper
(same counts), so neither layer is load-bearing on its own.

rodaddy:
### software-factory: passed
Reviewed at: `4b8db888535b87a83a5ecf61ae5551c94e86d70c`

Findings:
- [medium] The opt\-in permits plaintext bearer auth to ANY host, not just the LAN/private range the docs scope it to (python/openbrain/src/openbrain/config\.py:145 \(and openbrain\-memory/src/openbrain\_memory/client\.py:1790\))
- [low] Empty\-means\-unset is fixed per\-field, so the same wrapper\-emitted empty string still raises on OPENBRAIN\_OBSERVATION\_ENABLED (python/openbrain/src/openbrain/config\.py:550)
- [low] New doc claim 'for a string setting that is harmless' is inaccurate for OPENBRAIN\_BASE\_URL, which is the field the wrapper most surely blanks (docs/CONFIG\_REFERENCE\.md:448)
- [low] The two docs touched by this PR describe the deployed wrapper's pass\-through in contradictory forms (\_plans/fleet\-rollout\-0\.9\.md:206 vs docs/CONFIG\_REFERENCE\.md:468)
- [info] Test coverage is strong on the config half; the construction\-site half asserts the keyword only, by construction (python/openbrain/tests/test\_lan\_http\_optin\.py)
- [info] Review summary: correctness verified by execution at the target SHA; no blocking findings (python/openbrain/src/openbrain/config\.py)

rodaddy:
## Root cause of the test-suite hang, and the fix

**Reproduced 3/3, then fixed.** `commit bc206ca`.

### The symptom
`cd python/openbrain && uv run pytest -q` hung **indefinitely** at test 8/45 of
`tests/test_capture_outage_notice.py::TestRecoveryIsAnnouncedOnceToo::test_recovery_is_not_repeated_on_every_healthy_stop`.
That file alone also hung. On merge-base `main` (`f705fc9`) the same file passed 45/45 in ~1s. The test
file is **byte-identical** between the two commits, which pointed the finger at this branch's source
changes.

### It was not the source changes — it was the environment, and this branch unmasked it

The hang did not reproduce in a clean worktree. It reproduced **only** when the shell carried the four
provisioned observation coordinates, which
`~/.local/share/openbrain-memory/env/claudex-observation.env` supplies — so every hook-configured host
has them, and a clean CI runner does not.

`capture_stop_with` accepts an injected `CaptureSettings`, but it resolves the **observation** section
itself, from the live process environment, on every call:

```python
delivery = asyncio.run(
    run_stop(payload, capture, observation=loaded_observation())
)
```

So a test that injects a fake capture lane still ran the observation lane against whatever the
developer's shell held. With the coordinates present, the unit suite built a **real `LangfuseEmitter`
and posted its fixture transcripts to the production fleet Langfuse server**. `emit` ends in
`client.shutdown()`, which *blocks* draining the OpenTelemetry batch worker — so the suite waited on the
network forever rather than failing.

**Faulthandler stack at the stall:**

```
File "openbrain/apps/hooks/stop.py", line 145 in capture_stop_with
File "tests/test_capture_outage_notice.py", line 357 in test_recovery_is_not_repeated_on_every_healthy_stop

Thread: opentelemetry/sdk/_shared_internal/__init__.py:156 in worker
Thread: langfuse/_utils/prompt_cache.py:50 in run
```

**Why this branch surfaced it.** `unknown_prefixed_variables` rejects any `OPENBRAIN_`-prefixed variable
matching no declared field. Before this branch `OPENBRAIN_ALLOW_INSECURE_HTTP` was undeclared, so on a
host that set it the settings load **raised**, `loaded_observation()` swallowed the raise, and the sink
stayed silently off. Declaring the variable — the entire point of #525 — let the load **succeed**, and
the observation lane it had been shadowing all along started dialing. The typo check had been acting as
an accidental network guard. CI never saw any of this because CI sets no observation coordinates, so
`observation_active` is False there either way.

### The fix, at the owning boundary

`test_settings.py` has carried a **file-scoped** `_clean_environment` autouse fixture since it was
written, for exactly the stated reason — *"a developer's real shell leaks into the assertions and a test
that should fail passes on their machine only."* The leak is not confined to settings tests, so that
fixture is promoted to the shared `tests/conftest.py` and now applies suite-wide.

`OPENBRAIN_TEST_*` is deliberately **preserved**: the `-m live` gate addresses its playground service
through that prefix, and clearing it would make the live gate pass having run nothing.

### Red-proof

| | fixture `autouse=False` | fixture restored |
|---|---|---|
| `TestTheSuiteNeverInheritsAnOperatorsSink` (2 new tests) | **2 failed** — `assert True is False`; `assert ['OPENBRAIN_ALLOW_INSECURE_HTTP', ...] == []` | **2 passed** |
| `test_capture_outage_notice.py` in the operator env | **hung** (killed at `timeout 150`, no summary line) | **45 passed** ×3, ~1.0s each |
| full suite in the operator env | hung | **565 passed, 19 deselected in 13.6s** |

The new tests assert at `loaded_observation()` and at the environment itself rather than by running the
hook: the hook's symptom is a *hang*, and a test that hangs to prove a hang reports nothing and blocks
the suite behind it.

### Gates

- `uv run mypy src/openbrain` — clean, 49 source files
- `uv run ruff check src tests` — clean
- `uv run pytest -q` (`python/openbrain`) — **565 passed, 19 deselected**, real summary line
- `uv run pytest -q` (`python/openbrain-memory`) — 554 passed, 9 skipped
- `bun install --frozen-lockfile`, `bunx tsc --noEmit` — clean
- pre-push full suite — passed, **no bypass**

### Also fixed: the red `PR Body / validate` check

The gate was failing on **11** errors, not one. `scripts/validate-pr-body.ts` matches
`^-\s*<Label>:\s*<content>$`, and the fields were written as `- **Label:** …` — the bold markers break
the match, so nine fields read as empty despite having real content. Two more: the `SME review-memory
update` line was bolded the same way, and `Live Open Brain checks` read `[x] complete` where the
validator requires `[x] linked below` or `[x] not applicable because: <reason>`. Since the smoke did run
and its evidence is in the body, that line is now `[x] linked below` — the honest disposition. Content
is unchanged; only the formatting was corrected, and the validator was run locally against the new body
until it passed.

PR #197, opened 2026-06-22
Include server IP in health response

rodaddy:
## Summary

- Add `server_ip` and `server_ips` to single-worker and two-worker `/health` responses.
- Support `OPEN_BRAIN_SERVER_IP` as a hosted override so the public health endpoint can report the final backend IP.
- Add focused test coverage for deterministic hosted server identity.

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed
  - `git diff --check`
  - `bunx tsc --noEmit`
  - `bun test src/server.test.ts --timeout 20000`
  - `bun test`
- [x] Python package checks passed or are not applicable
  - Not applicable: no Python package files changed.
- [ ] Live Open Brain smoke passed or is not applicable
  - Pending merge/deploy: verify direct backend and public HTTPS `/health` include `server_ip: "10.71.1.21"`.

## Critical Self-Review

- Highest-risk behavior: OS interface auto-detection could pick an unexpected address on multi-homed hosts.
- Assumptions that could be wrong: Hosted launchd env can carry `OPEN_BRAIN_SERVER_IP=10.71.1.21`; if not, fallback still lists non-loopback IPv4 addresses.
- Missing/weak tests: Test covers the deterministic env override, not every possible OS network interface ordering.
- Security/permission risk: Low; `/health` is already public and this exposes backend IP that Rico explicitly requested.
- Migration/deploy risk: No DB migration; deploy requires restarting the launchd two-worker wrapper.
- Downstream client/runtime risk: Additive health response fields should not break JSON clients that ignore unknown keys.
- Rollback/cleanup concern: Roll back the commit and remove the non-secret env override if needed, then restart Open Brain.
- Fixes made before PR: Added the env override path and a focused server health test before opening the PR.
- Known residual risk: Consumers that validate `/health` with a closed schema could need a schema update, though no such consumer is known from this change.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a narrow additive health endpoint change with no MEDIUM+ review finding.

## Review Gate

- [x] Critical self-review fields above are filled with specific, non-placeholder content
- [x] MEDIUM+ review findings were captured in `docs/sme/` or explicitly marked not applicable
- Live Open Brain checks: [ ] linked below or [x] not applicable because: they must run after merge/deploy for this hotfix

## Downstream Rollout

- [x] I checked `docs/downstream-rollout.md`
- [x] rtech-mcps handoff is complete or not applicable
[...]

PR #546, opened 2026-08-04
fix(health): report the real server IP and hostname instead of "unknown"

rodaddy:
`/health` on the local dev brain reported `"server_ip": "unknown"`, so a client had no way to tell which brain it was talking to.

## Summary

- Measured tonight: `GET http://127.0.0.1:3100/health` returned `"server_ip":"unknown","server_ips":["unknown"]`, while core01 returned `10.71.1.21`.
- Root cause: **the field was never detected.** `serverIps()` (`src/index.ts`, #197 / `db2e05b`) read `OPEN_BRAIN_SERVER_IP` and returned the literal string `"unknown"` when it was unset. There was no `os.networkInterfaces()` call anywhere in the repo. core01 looked correct only because its env happens to carry the variable — which is exactly why this stayed invisible: the host that would have exposed it is the one nobody checked.
- Fix: identity resolves from **configuration**, in one new module (`server/transport/server-identity.ts`), consumed by all three copies of the defect.

Precedence, highest first:

1. `OPEN_BRAIN_SERVER_IP` — the explicit advertised address. Always wins.
2. `OPEN_BRAIN_BIND_HOST`, when it names a concrete address. A wildcard or loopback bind is **skipped**: `0.0.0.0` identifies no particular machine, which is the entire question `/health` is being asked.
3. Detected private LAN interfaces — physical adapters first, interface names sorted numerically so the answer is stable across reboots rather than dependent on enumeration order.
4. `"unknown"` only when the host genuinely has no private address.

`/health` also now reports `hostname` and, on a deployed tree, `revision` — the `short_sha` from the `.deployed-revision` stamp `local-clone-deploy.sh` already writes. That gives an exact which-brain-and-which-code discriminator instead of inferring it from the `workers` array.

**Three copies, one resolver.** `src/index.ts` (serves core01 via `deploy/open-brain.service`), `server/` (serves the local clone via `SERVING_ENTRYPOINT`), and the `run-two-worker.ts` aggregate front each carried the same `["unknown"]` function. All three now call the same resolver, so a front and the workers behind it cannot disagree about which machine they are.

## A deliberate reversal, called out for review

This supersedes the `does not auto-disclose host interface IPs when identity is not configured` test that shipped alongside #197. That was an implementation choice, never an issue-recorded requirement — and withholding the answer is precisely what produced the defect.

The disclosure concern is real (`/health` is unauthenticated, `server/transport/http-app.ts`), so it is bounded properly instead of by refusing to answer: **detection only ever volunteers private addresses** (RFC1918, RFC3927 link-local, RFC6598 CGNAT). A public address is reported only when an operator sets `OPEN_BRAIN_SERVER_IP` explicitly and owns that decision. Both behaviors are covered by tests.

## Red-proof

Against the shipped `serverIps()` logic, with the measured local-clone config (`OPEN_BRAIN_BIND_HOST=0.0.0.0`, no `OPEN_BRAIN_SERVER_IP`):

```
Expected: "10.71.1.20"
Received: "unknown"          ← 1 fail
```

The same scenario in `server/transport/server-identity.test.ts` (`reports the real LAN address when the bind host is the 0.0.0.0 wildcard`) passes with the fix: **33 pass, 0 fail**.

Independently red-proven end to end: `server/main.pg.test.ts` failed on the live-Postgres health shape with `Expected: "127.0.0.1" / Received: "10.71.1.20"` — the fix changing real server output, not just a unit assertion.

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed — pre-push full suite **3155 pass / 0 fail** across 3661 tests, **no bypass**; `bunx tsc --noEmit` clean; `server/transport/server-identity.test.ts` 33 pass; `server/main.pg.test.ts` 8 pass against the live dogfood DB (these SKIP silently without `OPENBRAIN_TEST_DATABASE_URL`, so they were run with it set)
- [x] Python package checks passed or are not applicable — not applicable: no `python/` paths touched
- [x] Live Open Brain smoke passed or is not applicable — **passed, RUNNING**: a server bound to `0.0.0.0` answers `{"hostname":"Mini-M4-Pro.local","server_ip":"10.71.1.20","server_ips":["10.71.1.20","10.71.1.131","192.168.127.11"]}` on **both** `127.0.0.1:7145` and `10.71.1.20:7145`, and reports `"revision":"f705fc9"` when a deploy stamp is present

**No new test failures.** The `server/` suite fails exactly the 7 tests it already failed before this change (`list_stale` ×4, `reporting tools` ×3, all pre-existing on `main` and in files this PR does not touch). Verified by diffing the sorted failing-test-name lists: byte-identical.

## Critical Self-Review

- Highest-risk behavior: reversing the #197 non-disclosure stance on an unauthenticated endpoint. Mitigated by bounding automatic detection to private ranges only, with a test asserting a public address is never volunteered; explicit operator configuration remains the only way to advertise one.
- Assumptions that could be wrong: that `en0`-style physical naming reliably identifies the primary adapter. On this Mac `en10` is a real registered network service (S3), not a bridge, so the name heuristic cannot fully separate them — this is why `server_ips` keeps every address rather than reporting only the winner, and why ordering is sorted rather than incidental. `server_ip` is correct here; a host whose primary adapter is not the lowest-numbered one would need `OPEN_BRAIN_SERVER_IP`.
- Missing/weak tests: no test asserts against a live multi-homed host's real interface table (that is environment-dependent); coverage uses injected tables plus the live smoke above. IPv6 is not reported — deliberate, out of scope, and it would need its own scope/disclosure decision.
- Security/permission risk: this endpoint is unauthenticated by design. The change discloses private LAN addresses and the hostname that were not previously disclosed. Judged acceptable because both are already visible to anyone who can reach the port, and the operator explicitly asked for them; public addresses stay opt-in. The commit subject line is deliberately **not** exposed — only `short_sha` is parsed out of the stamp.
- Migration/deploy risk: none — no schema or migration. `/health` gains fields; it removes none, so existing consumers are unaffected.
- Downstream client/runtime risk: `/health` is a client-facing response shape. Fields are **added** (`hostname`, optional `revision`) and `server_ips` may now contain more than one entry — a consumer that assumed exactly one element would see a longer array. No consumer in this repo does.
- Rollback/cleanup concern: revert restores `"unknown"` — the original defect — with no other side effect. `.deployed-revision` was added to `.gitignore` so a stamp can never be committed and make every tree claim someone else's sha.
- Fixes made before PR: initial interface ordering put a VM bridge (`192.168.127.11`) ahead of a real adapter in `server_ips`; caught in live verification and fixed with deterministic numeric sorting plus two ordering tests.
- Known residual risk: identity is resolved **once at process start**, not per request. That is deliberate (a process does not change hosts, and re-reading env per probe would let a mutated variable silently change what the endpoint claims), but it means a host that changes IP requires a restart before `/health` tells the truth.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no new review-lane pattern arose; the "field exists, detection never did" gotcha is documented at the code site and in `docs/CONFIG_REFERENCE.md`

## Review Gate

Live check evidence (RUNNING, this session, server bound to `0.0.0.0`):

```
$ curl -s http://127.0.0.1:7145/health
$ curl -s http://10.71.1.20:7145/health     # identical answer from the LAN
{"status":"healthy","hostname":"Mini-M4-Pro.local","server_ip":"10.71.1.20",
 "server_ips":["10.71.1.20","10.71.1.131","192.168.127.11"],
 "database":{"connected":true,...},"embedding":{"configured":true,"connected":true}}

# with a .deployed-revision stamp present:
{"status":"healthy","hostname":"Mini-M4-Pro.local","server_ip":"10.71.1.20",...,"revision":"f705fc9",...}
```

Before this change, the same endpoint on this machine answered `"server_ip":"unknown","server_ips":["unknown"]`.

## Contract Parity

- Contract parity: [x] runtime-specific because: `/health` is an operational endpoint with no contract fixture, client binding, or wire representation in `contracts/`. The response type is declared in `src/types.ts` and `server/transport/health.ts`, both updated together; `server/main.pg.test.ts` asserts the shape against a live server so a dropped field is a compile error, not a silent `undefined`.

## Downstream Rollout

- [x] rtech-mcps handoff is complete or not applicable — not applicable: no MCP tool or schema surface touched
- [x] mcp2cli cache/skill refresh is complete or not applicable — not applicable: no tool definitions changed
- [x] rtech-hermes Python runtime/plugin changes are complete or not applicable — not applicable: no `python/` paths touched
- [x] Hermes live rollout/canaries are complete or not applicable — not applicable: additive fields on an operational endpoint no Hermes path parses

Notes/evidence:

- `/health` is client-facing, so it was classified rather than waved through: the change is **additive** (`hostname`, optional `revision`) and removes nothing, so no downstream consumer breaks. The one behavioral change a consumer could notice is `server_ips` returning multiple entries where it previously returned a single-element array — no in-repo consumer indexes it that way.
- core01 is unaffected in practice: it sets `OPEN_BRAIN_SERVER_IP`, which is still the highest-precedence source, so it keeps reporting `10.71.1.21` exactly as before. This PR changes what happens when that variable is **absent**.
- `/health` gains fields only; deploying this does not require a coordinated client change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,3 +98,9 @@ secrets/*
 !secrets/README.md
 !secrets/*.example
 !secrets/*.example.json
+
+# Deploy stamp written into a DEPLOYED tree by scripts/local-clone-deploy.sh
+# and read back by `/health` as `revision`. It describes one deployment, not
+# the source, so it must never be committed — a checked-in stamp would make
+# every tree claim to be whichever one happened to be deployed when it landed.
+.deployed-revision
--- a/docs/CONFIG_REFERENCE.md
+++ b/docs/CONFIG_REFERENCE.md
@@ -145,8 +145,34 @@ watchdog restart script, not by the server itself: `MLX_EMBED_DAEMON`,
 | variable | default | read at |
 |---|---|---|
 | `PORT` | `3100` | `src/index.ts:349` |
-| `OPEN_BRAIN_BIND_HOST` | unset | `src/index.ts:353` |
-| `OPEN_BRAIN_SERVER_IP` | unset | `src/index.ts:44` |
+| `OPEN_BRAIN_BIND_HOST` | unset | `src/index.ts`, `server/main.ts`; also an identity input (below) |
+| `OPEN_BRAIN_SERVER_IP` | unset | `server/transport/server-identity.ts`; the advertised address |
+
+### Host identity in `/health`
+
+`/health` answers "which brain am I pointed at?", so it reports `hostname`,
+`server_ip`, `server_ips`, and — on a deployed tree — `revision`. Identity is
+resolved ONCE per process by `server/transport/server-identity.ts`, in this
+order:
+
+1. **`OPEN_BRAIN_SERVER_IP`** — an explicit advertised address. Always wins.
+   Set this when the address a client should use is not one the host can see
+   (behind NAT, a reverse proxy, or a floating VIP).
+2. **`OPEN_BRAIN_BIND_HOST`**, when it names a concrete address. A wildcard or
+   loopback bind (`0.0.0.0`, `::`, `127.0.0.1`, `::1`, `localhost`) is skipped:
+   it identifies no particular machine, which is the whole question being asked.
+3. **Detected private LAN interfaces**, physical adapters first, each interface
+   name sorted numerically so the answer is stable across reboots.
+4. **`"unknown"`** — only when the host genuinely has no private address.
+
+Detection is deliberately bounded to private ranges (RFC1918, RFC3927,
+RFC6598). `/health` is unauthenticated, so a public address is never volunteered
+automatically; an operator who wants one advertised sets
+`OPEN_BRAIN_SERVER_IP` and owns that decision.
+
+`revision` is the `short_sha` from the `.deployed-revision` stamp that
+`scripts/local-clone-deploy.sh` writes into a deployed tree. It is absent on a
+dev tree that was never deployed through the script, which is normal.
 | `ALLOWED_ORIGINS` | `[]` (none) | `src/index.ts:78`, comma-separated |
 | `OPEN_BRAIN_RUN_MIGRATIONS` | `1` | `src/index.ts:272`; `"0"` disables |
 | `OPEN_BRAIN_MAINTENANCE_ENABLED` | unset | `src/index.ts:383` |
--- a/scripts/run-two-worker.ts
+++ b/scripts/run-two-worker.ts
@@ -1,6 +1,7 @@
 #!/usr/bin/env bun
 
 import { buildHttpWorkerEnv } from "./worker-env.ts";
+import { resolveServerIdentity } from "../server/transport/server-identity.ts";
 
 type WorkerConfig = {
   name: string;
@@ -16,12 +17,15 @@ const workerPorts = (process.env.OPEN_BRAIN_WORKER_PORTS ?? "3101,3102")
 const workerCount = parseInt(process.env.OPEN_BRAIN_WORKERS ?? "2", 10);
 const poolMax = process.env.OPEN_BRAIN_WORKER_DB_POOL_MAX ?? "5";
 
-function serverIps(): string[] {
-  const configured = process.env.OPEN_BRAIN_SERVER_IP?.trim();
-  if (configured) return [configured];
-
-  return ["unknown"];
-}
+/**
+ * Host identity for the aggregate front, resolved once. Shares the single
+ * resolver with both serving trees so the front and the workers behind it can
+ * never disagree about which machine they are.
+ */
+const SERVER_IDENTITY = resolveServerIdentity({
+  configuredServerIp: process.env.OPEN_BRAIN_SERVER_IP,
+  bindHost: process.env.OPEN_BRAIN_BIND_HOST,
+});
 
 if (workerCount < 1) {
   throw new Error("OPEN_BRAIN_WORKERS must be at least 1");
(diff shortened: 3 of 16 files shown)


rodaddy:
### software-factory: failed
Reviewed at: `c8b347cb6401fc6380148746a0bb5feefd20436c`

Findings:
- [high] Adding OPEN\_BRAIN\_BIND\_HOST as \`nonEmpty\.optional\(\)\` makes the shipped \`\.env\.example\` shape refuse to start the server/ tree (server/config\.ts:114 \(with \.env\.example:40, \.env\.schema:32\))
- [medium] Virtual\-interface filter misses Docker \`br\-\*\` and libvirt \`virbr0\`, so a Linux host reports a bridge address as server\_ip (server/transport/server\-identity\.ts:82\-97,133\-171)
- [medium] parseServerConfig is documented as the pure half of the boundary but now reads host global state with no injection seam (server/config\.ts:233\-241,289\-300)
- [low] src/server\.test\.ts trades two deterministic assertions for host\-dependent ones that fail on a loopback\-only machine (src/server\.test\.ts:102\-138)
- [low] New docs section is inserted mid\-table, orphaning six config rows so they no longer render as a table (docs/CONFIG\_REFERENCE\.md:145\-176)
- [low] Unauthenticated /health now discloses hostname, every private LAN address, and the deployed commit sha (server/transport/health\.ts:14\-46, server/transport/server\-identity\.ts:200\-239)
- [low] Malformed indentation in the second server/main\.pg\.test\.ts hunk (server/main\.pg\.test\.ts:263\-270)
- [info] \`revision\` is plumbed through the aggregate worker front but never populated, so the two health surfaces disagree (server/transport/worker\-proxy\.ts:20,29,109\-112; scripts/run\-two\-worker\.ts:20\-27)
- [info] Review summary: sound design and unusually strong test coverage, blocked on one config\-boundary regression (PR \#546)

PR #547, opened 2026-08-04
docs(client): land the client-install story as a bundle, a runbook, and four gotchas

rodaddy:
Lands the client-install story so it stops being re-derived. No issue closed — this is the procedure and the tooling behind the #525/#544 work, not a fix.

## Summary

- Installing the direct stack on a box that is not the Mini has been rebuilt from scratch more than once, from the same handful of files, and the **same four things went wrong each time**. The cause is structural: the install is **four artifacts and three of them live outside any git repo** — the env file, the hook wrapper, and the `settings.json` wiring. Only the package source is versioned, so the rest travelled by memory.
- `scripts/client-bundle.sh` (runs on the **Mini**) freezes the live, working installation into a bundle: three wheels via `uv build` at the current commit, a copy of `~/.local/share/openbrain-memory/env/`, the `openbrain-*` hook entries lifted out of `~/.claude/settings.json`, plus the installer and a README recording the build commit.
- `scripts/setup-client.sh` (runs on the **client**) installs the wheels, places the env dir, `chmod +x` the wrapper, **merges** hook entries into `settings.json` via `python3`, and then **proves** the install rather than asserting it.
- `docs/client-install-runbook.md` is the procedure and its **why**. `docs/GOTCHAS.md` gains the four traps, symptom first.

**Distribution is from the Mini because clients cannot reach GitHub** — the repo is private with no client deploy key, and `python/openbrain-memory/pyproject.toml` records that `fleet-nats` is not on PyPI. `_plans/fleet-rollout-0.9.md` §1.1 listed three candidate mechanisms; this is the staged-wheels one, chosen and built. The client resolves with `--find-links` against the bundle's own `wheels/`, **reusing the `OPENBRAIN_MEMORY_FIND_LINKS` wheelhouse convention the Mini already runs on** (`_ob/scripts/ob-memory-provider/package-runtime.ts`) rather than inventing a second one.

**Bundle home is `/Volumes/ThunderBolt/open-brain-local/air-bundle/`** — a sibling of the running `app/`. Durable and findable, explicitly *not* the temp workspace (no persistence guarantee, and nobody should be picking deliverables out of a worktree) and explicitly *not* the repo, because the env dir carries a **live bearer token**. Nothing in this PR contains a secret; the script reads `~/.local` paths at runtime.

**No deletes anywhere.** Each run stages into a fresh timestamped directory and flips a `current` symlink. Superseded bundles stay on disk for an operator. A stale bundle costs disk; a wrong recursive delete costs the volume.

### The ordering the installer encodes

`setup-client.sh` installs wheels **before** placing the wrapper, because they are a **matched pair**: `config.unknown_prefixed_variables` rejects any undeclared `OPENBRAIN_*` — and rejects the **whole environment**, not the one variable — while the hook entrypoints swallow the exception into a clean **exit 0**. A wrapper ahead of its package is a silent, green-looking, zero-capture box. This is the PR #544 ordering note, encoded in the script instead of left in prose.

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed — pre-push full suite passed with **no bypass**: **3121 pass, 506 skip, 0 fail** across 228 files. The gate itself reported `Python package unchanged`, `openbrain package unchanged`, `Client/contract paths unchanged` — correct, this PR touches only `docs/` and `scripts/`
- [x] Python package checks passed or are not applicable — **not applicable**: no `python/` paths touched (`git diff --name-only` is 4 files, all `docs/` and `scripts/`)
- [x] `bunx tsc --noEmit` — not applicable: no TypeScript touched
- [x] Live Open Brain smoke passed or is not applicable — **the bundle build is the live receipt, and it was RUN**: `scripts/client-bundle.sh` executed against a checkout of `4b8db88` (the PR #544 head this machine's deployed wrapper matches), producing `/Volumes/ThunderBolt/open-brain-local/air-bundle/20260804-112335/` with three wheels (`openbrain-0.9.0`, `openbrain_memory-0.9.0`, `openbrain_provider-0.9.0`), the complete env dir (`0600` env file, `0755` wrapper), and **17 hook entries across 8 events** extracted. `current` points at it.

### What is NOT proven

**A full `setup-client.sh` run on a client has not happened.** The client half is WRITTEN, not RUN. Rungs 1 and 2 of the proof ladder could not be exercised from here either: `com.rico.open-brain-local-clone` shows PID `-` in `launchctl list` and `/health` refuses on `127.0.0.1:3100`, `10.71.1.20:3100`, and `https://ob.rodaddy.live` (502). That is the documented "service dies and no one notices" gotcha, out of scope for this PR, and it is stated in the runbook's status line rather than papered over.

## Critical self-review

- **Highest-risk behavior:** `setup-client.sh` mutates `~/.claude/settings.json`, the file that controls the harness. Mitigated by merging rather than writing: it timestamps a backup first, removes only entries whose command contains `openbrain` (so re-running upgrades instead of duplicating), preserves non-openbrain hooks inside the same matcher, and folds into an existing matcher with the same selector instead of fragmenting the file. It is still the thing to review hardest.
- **Assumptions that could be wrong:** that a client's `uv`, `python3`, and `~/.claude/settings.json` layout match the Mini's closely enough. The script fails closed on missing `uv`/`python3` and on an incomplete bundle, and it rewrites the wrapper's hardcoded `ENV_FILE` for a differing `$HOME` — but I have not observed the Air, so those paths are reasoned, not measured. Also assumed the three-package set is complete; verified against the actual hook commands in `settings.json` (all three are represented).
- **Missing/weak tests:** there are none, and that is a real gap — these are shell scripts with no test harness in this repo. What I did instead is empirical: the bundle script was **run twice** end to end, and I verified the wheel-resolution globs by executing them against the real staged directory (`openbrain` has no hyphen, so its glob stays `openbrain-*` and does not collide with the two underscored siblings — checked, not assumed). The merge logic in `setup-client.sh` is **not** exercised, because running it would rewrite this machine's live settings.
- **Security/permission risk:** the bundle contains a live token by design. The repo does not: no secret is in any committed file (`rg` for token/bearer patterns is clean), the script reads `~/.local` at runtime, and the bundle lives outside the repo entirely. The bundle README carries an explicit SECURITY section. Modes are set deliberately (`0600` env, `0755` wrapper), and the installer backs up before overwriting an existing env file.
- **Migration/deploy risk:** none to the service — nothing here deploys or touches the running brain. The client-side risk is the matched-pair ordering, which is exactly what the script encodes and the gotcha documents.
- **Downstream client/runtime risk:** none — no MCP tool, schema, transport, or wire surface. This is operator tooling and documentation.
- **Rollback/cleanup concern:** reverting the commit removes the scripts and docs; staged bundles are outside git and survive, which is correct — an existing client keeps working. On the client, `setup-client.sh` leaves a timestamped `settings.json` backup for a manual restore.
- **Fixes made before PR:** two, both found by running the script rather than reading it. (1) `[ -f ... ] && mv` as the last command of a list exits nonzero under `set -e` — converted to `if`, with the reason in a comment. (2) The README recorded `Branch/ref: HEAD` when building from a detached PR head, which is useless provenance; it now names the containing remote branch. Also moved the `.gitignore` `uv build` drops into its out-dir aside, so the bundle directory has nothing confusing in it.
- **Known residual risk:** the client half is unproven, stated above and in the runbook's status line. The staged env file points at `http://127.0.0.1:3100`, which is correct on the Mini and **wrong on a client** — the runbook and the fourth gotcha both call editing it a required step, but a hurried operator could miss it. The bundle copies the Mini's env verbatim by design (it reproduces a working config); rewriting the URL automatically would mean guessing the client's intended brain.
- **SME review-memory update:** [ ] `docs/sme/` updated or [x] not applicable because: no new review-lane pattern arose. The one reusable trap class here (undeclared/empty-valued env var → silent zero capture) is already documented at the rule site in `config.py`, in the wrapper header, and in `CONFIG_REFERENCE.md`; this PR adds the operator-facing statement of it to `GOTCHAS.md`, which is that file's job.

## Review Gate

- Live Open Brain checks: [x] not applicable because: no service, transport, MCP tool, or schema surface is touched. The live receipt that *does* apply — the bundle build — was run and is recorded under Verification

## Contract Parity

- Contract parity: [x] runtime-specific because: these are operator shell scripts and documentation, with no contract fixture, client binding, or wire representation

## Downstream Rollout

Notes/evidence: no MCP tool, schema, transport, or client-facing surface changes. The install path this documents is for the direct stack, which is the lane that *replaced* the MCP registration — the first gotcha exists to help operators remove the retired one.

### DIFF
--- a/docs/GOTCHAS.md
+++ b/docs/GOTCHAS.md
@@ -969,3 +969,132 @@ Two sessions diagnosed a missing token as a core01 outage, and core01 was probed
 healthy mid-incident — the healthy probe was read as noise instead of as the
 answer. A healthy `/health` next to a failing gate is positive evidence the
 problem is credentials or environment, not the network.
+
+---
+
+## 2026-08-04 client-install traps
+
+These four all bit while putting the direct client stack on a second machine.
+Full procedure: `docs/client-install-runbook.md`.
+
+### A stale MCP registration answers instead of the direct stack — and `/mcp` in an error is the tell
+
+**Symptom.** Recall or a hook fails, and the error names a URL ending in `/mcp`.
+The direct client stack is installed, the env file is right, `/health` answers,
+and it still does not work.
+
+**Cause.** A retired MCP-lane registration is still configured on the box —
+`claude mcp list` shows an open-brain entry — and it is being reached instead of
+the direct client.
+
+**The tell is exact: the direct stack NEVER uses a `/mcp` URL.** Its base URL is
+a bare `scheme://host:port`. So a `/mcp` path appearing anywhere in an error
+message is not a routing detail to investigate; it is positive proof that the
+retired lane is configured and answering. There is no configuration in which the
+direct stack legitimately produces that URL.
+
+**Check and fix.**
+
+```bash
+claude mcp list
+claude mcp remove <name>   # any open-brain entry
+```
+
+`setup-client.sh` prints this reminder and the current registrations at the end
+of every install, because a fresh install onto a box that used to run the MCP
+lane is exactly when this happens.
+
+### The hook wrapper and the installed package are a MATCHED PAIR, and the mismatch is silent
+
+**Symptom.** Every hook exits 0. Receipts look fine. No rows are written and no
+canon is injected. The box looks completely healthy and is doing nothing.
+
+**Cause.** Three facts that compound:
+
+1. `config.unknown_prefixed_variables` **rejects** any `OPENBRAIN_*` variable
+   that matches no declared setting — and rejects the **whole environment**, not
+   the one variable.
+2. The hook entrypoints **swallow every exception** (fail-open observer
+   contract).
+3. So a wrapper passing a variable the installed package does not declare turns
+   **every hook on the box** into a clean exit 0 with zero capture and zero
+   injection.
+
+This has happened three times with three variables: `OPENBRAIN_OBSERVATION_*`,
+`OPENBRAIN_SPOOL_PATH`, and `OPENBRAIN_ALLOW_INSECURE_HTTP`.
+
+**The ordering rule (PR #544).** Install the package that declares the variable
+**first**, then edit the wrapper to pass it. PR #544 verified this empirically in
+that order — the old install raised `UnknownEnvironmentVariableError` with the
+variable present; the reinstalled package accepted it and resolved
+`allow_insecure_http=True` on both sections. Reversed, the box goes dark between
+the two steps.
+
+**The empty-string half, which re-armed the same defect.** The wrapper's
+`env -i VAR="${VAR:-}"` style **cannot express "absent"** — an unset variable
+reaches the child as an **empty string**. For the bool `allow_insecure_http`,
+pydantic rejected `""` with `Input should be a valid boolean`, both
+`load_capture_settings` and `load_canon_settings` raised, the entrypoints
+swallowed it, and every hook on a host that **never opted in** went silently
+dead. The fix for #525 re-created the #525 defect class.
+
+Two layers guard it now: a validator in `config.py` mapping `""` to the default
+`False`, and a conditional in the wrapper that prepends the assignment only when
+non-empty — the conditional is what protects an **older** installed package that
+predates the validator, which is exactly the state of a client mid-upgrade.
+
+**Rule:** a non-string pass-through goes in the wrapper's conditional block,
+never in the `env -i` list. A string tolerates the empty spelling; a bool, an
+int, or an enum does not.
+
+**The check that catches it.** Not exit code, and not `/health` — both pass on a
+dead box. Start a fresh session and read the CANON PACK section counts. Zero
+counts, or no pack, is this bug.
+
+### Clients cannot install from GitHub — wheels from the Mini are the paved road
+
+**Symptom.** `uv tool install git+ssh://git@github.com/rodaddy/open-brain...`
+fails on a client box, and reaching for a package index does not help either.
+
+**Cause.** The repo is private and the client boxes have no deploy key. There is
+also no published index: `python/openbrain-memory/pyproject.toml` records that
+its `fleet-nats` dependency is **not** on PyPI and lives in a private monorepo.
+Docs showing `uv pip install openbrain-memory==<version>` imply an index that
+does not exist for these packages.
+
+**The paved road.** Build wheels on the Mini and stage them —
+`scripts/client-bundle.sh` does this, into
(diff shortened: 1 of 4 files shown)


rodaddy:
### software-factory: failed
Reviewed at: `94ef7814d465ce41816f9b9e4faf3e20ec277505`

Findings:
- [info] Review summary: 4 files \(\+866/\-0\), docs \+ two operator shell scripts; merge logic is the risk centre (docs/client\-install\-runbook\.md, docs/GOTCHAS\.md, scripts/client\-bundle\.sh, scripts/setup\-client\.sh)
- [medium] Stale openbrain hooks survive an upgrade for any event dropped from the new bundle (scripts/setup\-client\.sh:136\-145)
- [medium] The recall proof false\-negatives on any client lacking ripgrep, which preflight does not check (scripts/setup\-client\.sh:209)
- [medium] \~/\.claude/settings\.json is rewritten in place and non\-atomically (scripts/setup\-client\.sh:164\-166)
- [medium] The one documented residual risk — a loopback base URL on a client — is never checked, at a validation point that already exists (scripts/setup\-client\.sh:190\-194)
- [medium] Bundle with a live bearer token can be staged onto the boot volume if the ThunderBolt volume is not mounted (scripts/client\-bundle\.sh:45,92)
- [medium] No tests, and the PR's stated reasons for that are both contradicted by the repo and by the script's own design (scripts/setup\-client\.sh:37\-39,115\-169)
- [low] Substring match on "openbrain" deletes and exports unrelated user hooks (scripts/setup\-client\.sh:142; scripts/client\-bundle\.sh:152)
- [low] Validation runs after mutation, leaving half\-installed clients and token\-bearing partial bundles (scripts/setup\-client\.sh:187\-194; scripts/client\-bundle\.sh:161\-162)
- [low] re\.sub replacement strings are unescaped, and the retarget pattern breaks on paths with spaces (scripts/setup\-client\.sh:100,130)
- [low] Token env file is copied without \-p, so an existing world\-readable target stays readable until chmod (scripts/setup\-client\.sh:76\-78)
- [low] Health\-check status code can render as a doubled "000 000" on the exact path operators debug (scripts/setup\-client\.sh:196)
- [info] The proof ladder does not exercise the exec bit or the client's own base\-URL policy (scripts/setup\-client\.sh:207\-215)
- [info] Generated bundle README asserts commit\-clean provenance even when the build tree was dirty (scripts/client\-bundle\.sh:205 \(heredoc\), 82\-86)
- [info] current symlink flip is not actually atomic (scripts/client\-bundle\.sh:245\-248)

PR #553, opened 2026-08-04
fix(client): make the install prover run, and the errors say where to look

rodaddy:
Five defects found during the FIRST real client install (the Air, 2026-08-04), all reproduced locally before fixing.

## Summary

None of these was a wrong result — every refusal the runtime made was correct. They were failures of the **receipt**: it named a fraction of the problem, or called a key optional and then died on that same key, so a human paid one round trip per field for a set the runtime knew in full on the first call.

- **The prover never worked.** `scripts/setup-client.sh` called `openbrain-memory recall --query ... ` — an argv interface that has never existed. It returns `arguments are not supported` without reaching the brain, so the prover reported `[FAIL]` on every install regardless of whether the install was good. That is worse than no prover: a check whose failure carries no information. It now sources the env file and pipes one bounded JSON object with the full five-field scope, asserting `OPENBRAIN_NAMESPACE` is present **first** so a missing export is reported against the env file rather than surfacing deep in scope validation as a message about a request key.
- **Loopback base URL is now refused on a client.** The bundle ships the Mini's env verbatim, so an unedited `OPENBRAIN_BASE_URL` is always loopback and points the client at itself. This was already written in `docs/GOTCHAS.md` and the doc did not stop it — the Air was installed from an unedited file anyway — so it is enforcement now, with `OPENBRAIN_ALLOW_LOOPBACK_CLIENT=1` to run the script on the Mini.
- **The contradictory receipt is gone.** A top-level `namespace` was listed in `ignored_optional_request_keys` **and** failed `namespace must be a non-empty string` in the same JSON object. It is now rejected with an error naming `OPENBRAIN_NAMESPACE` and the documented `{"config": {"namespace": "..."}}` override. Rejecting rather than aliasing is deliberate: `docs/memory-contract.md` already gives `config.namespace` as the one in-request spelling, so a second top-level spelling would be an undocumented duplicate field. Same treatment for misplaced `base_url`/`token`; `ingest` keeps the top-level `namespace` it genuinely defines.
- **`--help` names the environment.** The example is a request body, and a body carries no identity, so following it on a clean shell failed and said nothing about where namespace comes from.
- **Scope errors arrive all at once.** `RuntimeScope.from_mapping` validated field-by-field, surfacing one missing key per attempt — the Air hit `namespace`, satisfied it, then hit `server_id`, then the next. All five required fields are now checked before any of them raises.

## Red-proof

- **Old prover invocation, stock CLI:** `{"error":"arguments are not supported; use --help or provide bounded JSON on stdin; ..."}` — confirms it could never have reached the brain.
- **New prover, live against `http://127.0.0.1:3100`** with the override flag, running the block exactly as the script does: `[ok] recall returned status=direct`.
- **Contradictory receipt, before:** `"ignored_optional_request_keys":["namespace"]` alongside `"error":"namespace must be a non-empty string"` in one receipt. After: an error naming `OPENBRAIN_NAMESPACE` and the `config` path.
- **One-at-a-time scope, before:** a scope missing three fields reported only `server_id must be a non-empty string`. After: `scope is invalid (3 problems): server_id ...; channel_id ...; session_key ...`.
- **New tests fail without the fix.** Reverting only the two source files and keeping `tests/test_cli_request_diagnostics.py`: **5 failed, 3 passed**, each failure the exact field-report symptom (`KeyError: 'environment'`, `assert 'channel_id' in 'server_id must be a non-empty string'`). Restored: **8 passed**.
- **Loopback guard**, all three branches executed: unedited `http://127.0.0.1:3100` refuses; same URL with the override proceeds; `http://10.71.1.21:3100` proceeds.

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed — `bun install --frozen-lockfile` clean; `sh -n` and `shellcheck` on the script
- [x] Python package checks passed or are not applicable — `uv run mypy src/openbrain_memory` clean (17 files), `uv run ruff check src tests` clean, `uv run pytest -q` **562 passed / 9 skipped** (554 before; +8 new)
- [x] Live Open Brain smoke passed or is not applicable — the new prover was run live against the running brain and returned `status=direct`

## Critical Self-Review

- Highest-risk behavior: rejecting a top-level `namespace` that was previously ignored. A caller who sent it and relied on the environment value silently winning now gets a hard failure instead of a working call. That is intended — the same request already failed on the contradictory path, so the previously-"working" case is only one where `namespace` was sent redundantly — but it is the one behavior change with a blast radius beyond error text.
- Assumptions that could be wrong: that no in-tree caller sends a top-level `namespace` to a non-`ingest` operation. Checked the full pytest suite (562 pass) and `ingest` is explicitly exempted via the per-operation allowed-key set; an out-of-tree caller doing this is possible and would now fail loudly rather than silently.
- Missing/weak tests: the loopback guard and the namespace-line assertion are shell and are proven by executing the extracted blocks, not by a committed test — `setup-client.sh` has no test harness in this repo. The live `status=direct` proof is a manual run, not CI-enforced.
- Security/permission risk: none added. The help `environment` block names variables but prints no values; the variable names are list values rather than object keys specifically so the redactor does not mistake the help for a credential (as an object key, `OPENBRAIN_TOKEN`'s description was replaced with `[REDACTED]`, hiding the explanation). The redaction rule itself is untouched.
- Migration/deploy risk: none — no schema, no migration. The script change takes effect on the next bundle install.
- Downstream client/runtime risk: the error text and `--help` result shape change, so anything asserting on the old single-field message or on `result` lacking `environment` would need updating. No MCP tool, transport, or wire schema is touched; the receipt schema (`openbrain.runtime_receipt.v1`) is unchanged.
- Rollback/cleanup concern: revert restores the old behavior wholesale; the prover would go back to reporting `[FAIL]` on every install.
- Fixes made before PR: `shellcheck` caught a real bug in my own new code — `[::1]*` in a `case` pattern is a character **range** matching single chars, not the literal IPv6 loopback, so the guard would have missed `http://[::1]:3100`. Corrected to `*'[::1]'*` and verified against all four loopback spellings.
- Known residual risk: the guard matches loopback by URL spelling, so a hostname that resolves to loopback (a `/etc/hosts` alias) still passes. Left alone deliberately rather than adding resolution to an install script; the failure mode it targets is the literal unedited `127.0.0.1` the bundle ships.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no new review-lane pattern arose — the findings are install-surface defects, captured in `docs/GOTCHAS.md` and the runbook where installers will read them

## Review Gate

- Live Open Brain checks: [x] linked below — the new prover was run live against the running service and returned `status=direct` (evidence in the Red-proof section above)

## Contract Parity

- Contract parity: [x] runtime-specific because: the changes are error text, `--help` output, and an install script. The receipt schema, MCP tools, and wire representation are unchanged; `config.namespace` remains the one documented in-request override.

## Downstream Rollout

Notes/evidence:

- No MCP tool, schema, or transport surface is modified. The client-facing change is error text and `--help` output from the provider CLI; a downstream consumer asserting on the old single-field scope error would need updating, which is the intended improvement.

### DIFF
--- a/docs/GOTCHAS.md
+++ b/docs/GOTCHAS.md
@@ -1098,3 +1098,44 @@ bearer token on the wire, and one fewer silent-failure mode.
 copies the Mini's env file verbatim, so a client staged from a loopback-pointed
 env file must have its `OPENBRAIN_BASE_URL` edited — that is a required step, not
 a nicety.
+
+**Written down is not enforced.** The paragraph above predates the first real
+client install and did not prevent it: the Air was installed from an unedited
+loopback env file anyway. `setup-client.sh` now REFUSES a loopback base URL
+rather than warning, with `OPENBRAIN_ALLOW_LOOPBACK_CLIENT=1` to run the script
+on the Mini itself. A rule that only exists in a doc is a rule the install can
+skip.
+
+### The provider CLI takes JSON on stdin — the namespace comes from the ENVIRONMENT
+
+**Symptom.** A request that looks exactly like the documented example fails with
+`namespace must be a non-empty string`, and nothing in the error mentions an
+environment variable. Adding `"namespace"` to the request body does not fix it —
+it produces a receipt that lists `namespace` under
+`ignored_optional_request_keys` **and** fails on `namespace` in the same JSON
+object.
+
+**Cause.** Two separate things, both fixed now, both worth knowing:
+
+- The CLI has **no argv interface**. `openbrain-memory recall --query …` returns
+  `arguments are not supported` and never reaches the brain. It reads ONE
+  bounded JSON object on stdin with `operation` inside it. The shipped prover in
+  `setup-client.sh` used the argv form, so it reported `[FAIL]` on every install
+  regardless of whether the install worked — a check whose failure carried no
+  information.
+- Identity is environmental, not a request field. `OPENBRAIN_BASE_URL`,
+  `OPENBRAIN_TOKEN`, and `OPENBRAIN_NAMESPACE` are read from the environment;
+  the one documented in-request override is `{"config": {"namespace": "…"}}`
+  (`docs/memory-contract.md`). A top-level `namespace` is now rejected with an
+  error that says where it actually lives.
+
+**The example is a request body, and a body carries no identity.** That is why
+following it on a clean shell fails. `openbrain-memory --help` now carries an
+`environment` block naming the three variables next to the example they qualify.
+
+**Also fixed: scope errors arrive all at once.** Validation used to surface one
+missing field per attempt, so assembling a scope by hand cost a round trip per
+field — the Air hit `namespace`, satisfied it, then hit `server_id`, then the
+next. Every error was true and every error was a fraction of the answer. All
+five (`agent`, `platform`, `server_id`, `channel_id`, `session_key`) are now
+reported in a single receipt.
--- a/docs/client-install-runbook.md
+++ b/docs/client-install-runbook.md
@@ -193,10 +193,26 @@ stopping early is how a green-looking dead box happens.
 
 | # | Rung | How | What a failure means |
 |---|---|---|---|
+| 0 | **The env file was edited** | `OPENBRAIN_BASE_URL` is the Mini's LAN address, and `OPENBRAIN_NAMESPACE` is present | The bundle ships the build machine's env verbatim, so an unedited `OPENBRAIN_BASE_URL` is loopback and points the client at itself. A missing `OPENBRAIN_NAMESPACE` fails later as a message about a *request key*, which points at the JSON instead of at this file. `setup-client.sh` now refuses on both. |
 | 1 | **Health** | `curl -sS $OPENBRAIN_BASE_URL/health` → `200` | Service down, wrong URL, or the plain-http opt-in is missing. |
-| 2 | **Recall is `direct`** | `sh …/openbrain-hook-env openbrain-memory recall --query … --limit 1` → `"status": "direct"` | Anything other than `direct` means the direct stack is not the lane answering — suspect a stale MCP registration. |
+| 2 | **Recall is `direct`** | source the env file, then pipe one JSON object (below) to `openbrain-memory` → `"status": "direct"` | Anything other than `direct` means the direct stack is not the lane answering — suspect a stale MCP registration. |
 | 3 | **CANON PACK, non-zero counts** | Start a **fresh** session; the `SessionStart` emissions carry section counts | Counts of zero, or no pack at all, is the matched-pair failure. The hooks ran and swallowed a rejection. |
 | 4 | **The turn is visible in Langfuse** | Find the session's turn in the Langfuse observation sink | The capture spine reached the brain but the observation lane did not. |
+
+Rung 2 in full. **The CLI takes one bounded JSON object on stdin** — with
+`operation` inside it — and it takes no argv flags at all; an argv invocation
+returns `arguments are not supported` without ever reaching the brain. The
+namespace comes from `OPENBRAIN_NAMESPACE` in the environment, which is why the
+env file is sourced first and why it is not a key in the JSON:
+
+```sh
+set -a; . ~/.local/share/openbrain-memory/env/claudex-observation.env; set +a
+printf '%s' '{"operation":"recall","query":"client install proof","scope":{"agent":"setup-client","platform":"claude-code","server_id":"client-install","channel_id":"client-install","session_key":"client-install-proof"}}' \
+  | openbrain-memory
+```
+
+All five scope fields are required. They are reported together in one receipt,
+so a scope that is missing several does not cost one attempt per field.
 
 Rung 3 is the one that catches the silent failure, because rungs 1 and 2 both
 pass on a box whose hooks are dead — they exercise the client library directly
--- a/python/openbrain-memory/src/openbrain_memory/cli.py
+++ b/python/openbrain-memory/src/openbrain_memory/cli.py
@@ -110,6 +110,79 @@ _HELP_EXAMPLE = {
     },
 }
 
+# The example above is a REQUEST BODY, and a request body carries no identity.
+# base_url, token, and namespace come from the environment (or an explicit
+# `config` object), so the example alone does not run on a clean shell -- it
+# fails with "namespace must be a non-empty string" and says nothing about
+# where namespace was supposed to come from. The first client install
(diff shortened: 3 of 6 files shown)


rodaddy:
## Fix per finding

| # | Finding (field report) | Fix | Where | Proof |
|---|---|---|---|---|
| 1 | Prover calls `openbrain-memory recall --query ...`, an argv interface that does not exist | Prover now sources the env file (`set -a`), pipes one bounded JSON object with the full 5-field scope, and asserts `status: direct`. The `OPENBRAIN_NAMESPACE` line is asserted **first**, so a missing export is reported against the env file instead of surfacing as a message about a request key | `scripts/setup-client.sh` | Old form returns `arguments are not supported`; new form returns `status=direct` live against the running brain |
| 2 | `--help` example omits that namespace arrives via `OPENBRAIN_NAMESPACE` | `--help` result carries an `environment` block naming `OPENBRAIN_BASE_URL`, `OPENBRAIN_NAMESPACE`, `OPENBRAIN_TOKEN` plus a note that the example is a request body and carries no identity | `python/openbrain-memory/.../cli.py` | `test_help_names_the_environment_the_example_depends_on` — `KeyError: 'environment'` before |
| 3 | Top-level `namespace` listed in `ignored_optional_request_keys` **and** fatal in the same receipt | Rejected with an error naming `OPENBRAIN_NAMESPACE` and the documented `{"config": {"namespace": "..."}}` path. Chosen per `docs/memory-contract.md`, which already defines `config.namespace` as the one in-request override — aliasing would have added an undocumented duplicate field. Same for `base_url`/`token`; `ingest` keeps the top-level `namespace` it defines | `python/openbrain-memory/.../cli.py` | `test_top_level_namespace_is_rejected_by_where_it_belongs` + `test_ingest_keeps_its_own_top_level_namespace` |
| 4 | Loopback `OPENBRAIN_BASE_URL` on a client install is always the unedited build-machine value | Install refuses and prints the required edit, with `OPENBRAIN_ALLOW_LOOPBACK_CLIENT=1` to run on the Mini itself | `scripts/setup-client.sh` | All three branches executed: unedited loopback refuses, override proceeds, `10.71.1.21` proceeds |
| 5 | Scope validator surfaces missing fields one at a time | All five required fields checked before any raises; one receipt names them all | `python/openbrain-memory/.../runtime.py` | Before: `server_id must be a non-empty string`. After: `scope is invalid (3 problems): server_id ...; channel_id ...; session_key ...` |

Docs: `docs/client-install-runbook.md` gets a **rung 0** (the env file was edited: LAN base URL + namespace present), a corrected rung 2 with the exact runnable stdin-JSON command, and a note that all five scope fields report together. `docs/GOTCHAS.md` gains "The provider CLI takes JSON on stdin — the namespace comes from the ENVIRONMENT", and the existing loopback paragraph is annotated **written down is not enforced** — it predated this install and did not prevent it, which is why the check moved into the script.

### Notes on findings 2 and 5 that differ from the report

- The `--help` example **already carried** all five scope fields on `main`; only the namespace-via-env half of finding 5's "example omits BOTH" was true. A test pins the full set so it cannot regress.
- Writing the help as `{"OPENBRAIN_TOKEN": "bearer token"}` made the redactor match the **key** and replace the description with `[REDACTED]` — the help hid its own explanation. The redaction rule is correct and untouched; the variable names are list values instead, so nothing looks like a credential it is not.
- `shellcheck` caught a bug in my own new guard: `[::1]*` in a `case` pattern is a character **range** matching single chars, so `http://[::1]:3100` would have slipped through. Corrected to `*'[::1]'*` and verified against all four loopback spellings.

Gates: `mypy` clean, `ruff` clean, `pytest` **562 passed / 9 skipped** (554 before), `bun install --frozen-lockfile` clean, `sh -n` + `shellcheck` on the script.

rodaddy:
CI note: the first `db-integration` run reported 6 failures (`core01 deploy shell ref-gate wiring` x5, `JsonlSpool ... independently opened child spools` x1). Not from this branch — the diff touches **zero** TypeScript files (`docs/`, `scripts/setup-client.sh`, and `python/openbrain-memory/` only), the twin `db-integration` job passed on the *same commit*, and `bun test scripts/deploy-ref-gate.test.ts` passes locally 18/18. Re-ran the failed job: all 14 checks now pass. Flagging rather than silently re-running, since both named tests are process/concurrency-sensitive (shell ref-gate subprocesses, cross-process spool exclusion) and look like real flakes worth watching.

rodaddy:
### software-factory: failed
Reviewed at: `4986f7ccd5724f6f21cd905f7cad859cad6cc37c`

Findings:
- [high] New OPENBRAIN\_NAMESPACE preflight makes ripgrep a fatal, undeclared dependency: a client without rg aborts a correct install with a false diagnosis (scripts/setup\-client\.sh:244)
- [medium] The redactor eats the remediation path out of the misplaced\-\`token\` error, leaving \`\{"config": \{\[REDACTED\]\}\}\` (python/openbrain\-memory/src/openbrain\_memory/cli\.py:\_MISPLACED\_CONFIG\_KEYS\["token"\])
- [medium] Loopback refusal is bypassed when OPENBRAIN\_BASE\_URL is quoted in the env file, because the extractor does not strip quotes (scripts/setup\-client\.sh:178\-226)
- [low] The two readers of claudex\-observation\.env disagree on its grammar: the new namespace check accepts an \`export \` prefix, the BASE\_URL extractor rejects it (scripts/setup\-client\.sh:178\-182,244)
- [low] The test for misplaced \`token\` asserts a weaker property than its sibling help test, so it passes over the redaction defect (python/openbrain\-memory/tests/test\_cli\_request\_diagnostics\.py:114\-125)
- [low] Multi\-field scope error reads "scope is invalid \(1 problems\)" in the single\-field case (python/openbrain\-memory/src/openbrain\_memory/runtime\.py:236\-239)
- [low] \`help\` now fails instead of returning help when the request carries a stray namespace/base\_url/token key (python/openbrain\-memory/src/openbrain\_memory/cli\.py:403)
- [low] The rewritten prover no longer runs through openbrain\-hook\-env, the wrapper the script's own header calls the primary silent\-failure surface (scripts/setup\-client\.sh:266\-274)
- [info] Unknown\-key rejection still short\-circuits ahead of the batched required\-field check, so a typo plus missing fields still costs two round trips (python/openbrain\-memory/src/openbrain\_memory/runtime\.py:211)
- [info] Test coverage is strong on the Python half and absent on the shell half (python/openbrain\-memory/tests/test\_cli\_request\_diagnostics\.py)

PR #564, opened 2026-08-05
fix(nats): answer an undeliverable reply with an error envelope, not silence (#549)

### LINKED ISSUE #549: NATS bridge: oversized reply is silently dropped — caller waits out its timeout with no error (measured 58.5MB pack reply)
Found live 2026-08-04 during local NATS wiring: a durable_memory context pack for namespace rico builds a 58.5MB reply (durable_memory alone 39.9MB). The broker refuses to publish it, message.respond() returns false, and startNatsContextPackBridge throws 'NATS request did not include a reply inbox' — the CALLER GETS NOTHING and waits out its own timeout with no client-visible reason. src/nats-bridge.ts guards the REQUEST at MAX_NATS_REQUEST_BYTES (64KB) but has no guard on the RESPONSE. Affects the fleet bus too (protocol default max_payload is smaller than this broker's 8MB). Needs: a response-size guard that returns a structured error (or a truncated pack with a shortfall receipt per the #535 sections_receipt pattern) instead of silence. Silent no-reply is the defect class this whole epic exists to kill.

rodaddy:
Closes #549.

## What was wrong

The NATS bridge finished its work and then vanished. A `durable_memory` context
pack for namespace `rico` built a **58.5 MB** reply against a broker advertising
8 MiB (measured live 2026-08-04); the caller got nothing and waited out its own
timeout with no client-visible reason.

**The mechanism, read from the installed client's source.** `Msg.respond()`
publishes through `protocol.publish()`, which **throws**
`NatsError(MAX_PAYLOAD_EXCEEDED)` once the encoded length passes
`info.max_payload` (`nats-base-client/msg.js:72-78` → `protocol.js:786-788`).
That throw escaped to `processNatsSubscriptionMessage`, whose catch logged the
generic **`NATS context-pack bridge request failed`** and published nothing.

`respond()` returns **false** in exactly one case: the request carried no reply
inbox at all. So `NATS request did not include a reply inbox` was never the
message this oversize path produced — it was reserved for its own, correct
condition, and it still is. (An earlier revision of this PR body described the
oversize failure as surfacing through that message; that account was wrong and
is corrected here.)

The request side was already guarded at `src/nats-bridge.ts`; the reply side had
nothing.

## What changed

At the respond seam only:

- Before publishing, the encoded reply is compared against the figure the
  **broker itself advertises** on the connection (`connection.info.max_payload`),
  read per message so a reconnect to a broker advertising something different is
  picked up rather than judged against a stale value. Open Brain never supplies
  a figure of its own — this fleet's local broker advertises 8 MiB, the NATS
  protocol default is 1 MiB, and a hardcoded number would be wrong for one of
  them.
- When the reply exceeds what the broker will carry, the doomed publish is not
  spent.
- **Both respond calls are wrapped.** When the advertised figure is unreadable
  the reply cannot be pre-judged, so the publish is attempted — and a thrown
  `MAX_PAYLOAD_EXCEEDED` now routes into the **same** undeliverable-reply
  envelope instead of escaping to the handler-error catch and handing the caller
  silence. Matched by the error's machine **code**, never by message text:
  `NatsError` carries a stable code while its prose is the client's to reword,
  and matching prose would reopen this silently on a client upgrade. Any other
  throw is not the bridge's to reinterpret and is **rethrown unchanged**, so a
  connection-closed failure still reaches the handler-error path.
- The answer is a redacted error envelope with the same `correlation_id` a
  normal reply carries, which releases the waiting caller. It is content-free:
  it names what happened, the measured reply bytes, the broker's advertised
  figure, and the requested sections. No pack data crosses into it, and only
  known section names are echoed back.
- `no reply inbox` is logged only for the condition that actually produces it —
  `respond()` returning false.

Nothing about what the handler builds changed. This changes only how an
undeliverable reply is answered.

**Single-source confirmed:** one bridge serves both trees. `src/nats-worker.ts`
and `server/application/nats.ts:107` both call `startNatsContextPackBridge` from
`src/nats-bridge.ts`, so this lands once and covers both.

## Red proof

Round two (`45e9de3`), against the unhardened source with the new tests kept:

- `answers with an error envelope when the client THROWS MaxPayloadExceeded on
  the reply publish` — **fails**, the thrown `MAX_PAYLOAD_EXCEEDED` escapes
  through `respondWithinBrokerFigure` and **nothing is published**. This mock
  throws, mirroring the real client, rather than returning false.
- `rethrows a respond failure that is not the client's payload refusal` — proves
  a `CONNECTION_CLOSED` throw is not swallowed into a `payload_too_large`
  envelope that would misdescribe it.

Round one (`8549d04`), with `src/nats-bridge.ts` reverted, all four fail — the
old code publishes nothing, so there is no reply to assert against:

- answers with an error envelope when the broker refuses the reply publish
- answers with an error envelope without attempting a publish the broker's advertised figure rules out
- carries the reply normally when it fits the broker's advertised figure
- logs an accurate reason when even the error envelope cannot be published

One pre-existing test (`resubscribes the real NATS subscription loop after
handler and iterator failures`) asserted the old throw reached the handler-error
path; it now asserts the accurate reply-inbox log instead.

## Gates

- `bunx tsc --noEmit` — clean
- `bun test` — 3175 pass, 0 fail, 506 skip (Postgres-gated)
- `bun test src/nats-bridge.test.ts` — 29 pass, 0 fail
- `bun install --frozen-lockfile` — clean

## Critical Self-Review

- Highest-risk behavior: the respond seam now swallows a `MAX_PAYLOAD_EXCEEDED` throw and a false return instead of letting them escape. Scoped by code match, so only that one client condition is absorbed; every other throw is rethrown and still reaches the handler-error path. If the error envelope publish also fails the bridge logs and returns, so a message can complete with no reply on the wire — strictly better than the old path, which published nothing on every refusal.
- Assumptions that could be wrong: that `NatsError.code` stays `MAX_PAYLOAD_EXCEEDED`. Verified against the installed client (`nats-base-client/core.js`); the literal is reproduced in `src/nats-bridge.ts` rather than imported so the seam stays testable through its own driver interface. If a future client renames the code the guard stops matching and the failure reverts to the old visible-log-plus-silence shape — a regression to the prior behavior, not a new failure mode, and the test would not catch it. Also assumed: `connection.info.max_payload` is populated after reconnect; if absent the reply is not pre-judged and the throw guard covers it, which is why the round-two test pins `maxPayloadBytes: undefined`.
- Missing/weak tests: no live broker round-trip in this lane — the tests drive the seam through the driver interface with a simulated throw, not a real NATS server rejecting a real 58.5 MB publish. Left as a post-merge operator step.
- Security/permission risk: the error envelope is content-free by construction — byte counts and section names only, filtered against the known `SECTION_NAMES` set so arbitrary caller text cannot be reflected back. No pack data, no namespace data, no broker URL. The new catch does not log the error object, so no client-supplied prose reaches the log line.
- Migration/deploy risk: none — no schema, no config, no new env var. `maxPayloadBytes` is optional on `NatsRequestMessage`, so an existing driver that does not supply it keeps working.
- Downstream client/runtime risk: none to the wire contract. The existing error payload shape and the existing `payload_too_large` code are reused unchanged, so no fixture and no Python mirror edit were needed. Clients that already handle `status: "error"` handle this.
- Rollback/cleanup concern: two commits on one source file, one test file, one doc. The worktree is removed after merge.
- Fixes made before PR: the round-two commit exists because the round-one lane inferred the client's behavior from the false return instead of reading `msg.js`/`protocol.js`. Reading the source found both the unguarded path and the wrong narrative.
- Known residual risk: a 58.5 MB reply still does not fit an 8 MB broker, so `durable_memory` on this namespace returns a clear error rather than a pack. That is the reported-and-open half of #549 and is the operator's call; this PR closes the silence only. The runbook separates the two.
- SME review-memory update: [x] not applicable because: no new cross-cutting review pattern surfaced — this is a single-seam fix reusing an existing wire error shape, and the existing SME lanes already cover redaction on error paths.

## Review Gate

- Live Open Brain checks: [x] not applicable because: the live NATS round-trip needs a local broker restart, which is operator-visible, so it is recorded as a post-merge step rather than run in this lane.

## Contract Parity

- Contract parity: [x] runtime-specific because: no wire shape changed — the existing envelope error payload and the existing `payload_too_large` code are reused unchanged, so `nats-context-pack-wire.json` and the Python `nats_wire` mirror needed no edit and byte parity is untouched.

## Post-merge

- [ ] Live NATS round-trip against the local broker: request `durable_memory` for namespace `rico` on `dev.ob.memory.context_pack` and confirm the caller receives a `payload_too_large` error envelope with a matching `correlation_id` instead of timing out.

### DIFF
--- a/docs/core01-nats-worker-runbook.md
+++ b/docs/core01-nats-worker-runbook.md
@@ -344,28 +344,39 @@ curl -fsS http://127.0.0.1:3100/health    # must stay healthy across a worker re
 launchctl kickstart -k "gui/$(id -u)/com.rico.open-brain-local-nats-worker"
 ```
 
-### OPEN DEFECT — large context packs get no reply at all (2026-08-04)
+### The silence is fixed; large context packs still do not fit (#549, 2026-08-04)
 
 A live request/reply on `dev.ob.memory.context_pack` returns `status: ok` in
 ~1.2s for `repo_facts` / `working_set` / `pointers`. Asking for
 `durable_memory` in the `rico` namespace currently builds a **58.5 MB** reply
 (the `durable_memory` section alone is 39.9 MB), which exceeds the broker's
-8 MB `max_payload`. The publish is rejected, `message.respond()` returns false,
-`startNatsContextPackBridge` throws "NATS request did not include a reply
-inbox", and the caller sees **no reply at all** — it waits until its own
-timeout.
+8 MB `max_payload`. The publish is rejected and `message.respond()` returns
+false.
 
-This is a real defect and not specific to this broker: `src/nats-bridge.ts`
-guards the REQUEST at `MAX_NATS_REQUEST_BYTES` (64 KB) but has **no response
-size guard**, and NATS's own protocol default `max_payload` is 1 MB — so a
-default broker or the fleet bus would fail sooner and the same silent way. The
-handler completes its work successfully first, so this burns a full context-pack
-build before dropping it.
+**What used to happen:** `startNatsContextPackBridge` threw "NATS request did
+not include a reply inbox" — a misattribution, since the reply inbox was
+present — and published nothing, so the caller saw **no reply at all** and
+waited out its own timeout with no client-visible reason.
 
-The 8 MB figure is the NATS server's own `max_payload` ceiling, not an Open
-Brain choice — this is reported here as a measurement, and how to resolve it is
-the operator's call, not something this runbook prescribes. The one thing that
-is unambiguously wrong regardless of size is the SILENCE: the bridge should
-answer with a redacted `payload_too_large` error envelope instead of leaving the
-caller to time out with no reply and no client-visible reason.
+**What happens now (#549):** the bridge answers with a redacted
+`payload_too_large` error envelope carrying the same `correlation_id` as a
+normal reply, so the caller is released immediately instead of timing out. The
+envelope is content-free — it names the measured reply bytes, the broker's own
+advertised `max_payload`, and which sections were requested, and carries no pack
+data. Before publishing, the bridge compares the encoded reply against the
+figure the broker advertises on the connection (`connection.info.max_payload`),
+never a value of Open Brain's own, so a broker advertising the NATS protocol
+default of 1 MB and this one advertising 8 MB are both handled correctly. If the
+publish is refused anyway, the error envelope is sent regardless. Only when that
+small envelope is ALSO refused does the bridge log a missing reply inbox — which
+is now the accurate reading of that case.
+
+The handler still builds the full pack first, and nothing about what it builds
+changed; this changed only how an undeliverable reply is answered.
+
+**Still open:** the 58.5 MB reply does not fit an 8 MB broker, so a
+`durable_memory` request on this namespace gets a clear error rather than a
+pack. The 8 MB figure is the NATS server's own `max_payload`, not an Open Brain
+choice — it is reported here as a measurement, and how to resolve it is the
+operator's call, not something this runbook prescribes.
 
--- a/docs/core01-nats-worker-runbook.md
+++ b/docs/core01-nats-worker-runbook.md
@@ -350,13 +350,21 @@ A live request/reply on `dev.ob.memory.context_pack` returns `status: ok` in
 ~1.2s for `repo_facts` / `working_set` / `pointers`. Asking for
 `durable_memory` in the `rico` namespace currently builds a **58.5 MB** reply
 (the `durable_memory` section alone is 39.9 MB), which exceeds the broker's
-8 MB `max_payload`. The publish is rejected and `message.respond()` returns
-false.
+8 MB `max_payload`, so the client rejects the publish.
 
-**What used to happen:** `startNatsContextPackBridge` threw "NATS request did
-not include a reply inbox" — a misattribution, since the reply inbox was
-present — and published nothing, so the caller saw **no reply at all** and
-waited out its own timeout with no client-visible reason.
+**How the rejection actually surfaces:** the nats.js client THROWS on it. Its
+`Msg.respond()` publishes through `protocol.publish()`, which raises
+`NatsError(MAX_PAYLOAD_EXCEEDED)` once the encoded length passes
+`info.max_payload` (`nats-base-client/msg.js` -> `protocol.js`). It does NOT
+return false — `respond()` returns false in exactly one case, a request that
+carried no reply inbox at all, which is unrelated to size.
+
+**What used to happen:** that throw escaped to
+`processNatsSubscriptionMessage`, whose catch logged the generic **"NATS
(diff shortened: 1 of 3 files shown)


rodaddy:
## Live NATS round-trip receipt — post-merge item complete

The deferred post-merge step (live round-trip against the local broker) is done. Deployed current `main` `d46d68e` to the local dogfood clone and ran the exact request that measured the original defect.

**Deploy**

```
local-clone-deploy: swapped d46d68e into /Volumes/ThunderBolt/open-brain-local/app
local-clone-deploy: revision proof PASSED: pid 29788 -> 23876, cwd /Volumes/ThunderBolt/open-brain-local/app, revision d46d68e
```

`http://127.0.0.1:3100/health` → `revision d46d68e`, `hostname Mini-M4-Pro.local`, `server_ip 10.71.1.20`.
NATS worker kickstarted by the deploy script itself (`OPENBRAIN_NATS_WORKER_LABEL`, the #545 lineage); `http://127.0.0.1:3110/health` → `nats.availability: available` on `dev.ob.memory.context_pack`.

**Oversize request — a reply ARRIVES**

Published on `dev.ob.memory.context_pack`: `durable_memory` in namespace `rico`, `query: "open brain"`. Previously this was silence until the caller's own timeout. Now, reply received, 535 bytes:

```json
{
  "id": "ob549-proof-oversize-b-20260805",
  "from": "open-brain",
  "kind": "context_pack_response",
  "payload": {
    "status": "error",
    "operation": "agent_context_pack",
    "namespace_source": "override",
    "error": {
      "code": "payload_too_large",
      "message": "The broker refused to carry this context pack reply. Reply was 63370799 bytes; the broker advertises 8388608 bytes. Requested sections: durable_memory."
    }
  },
  "correlation_id": "ob549-proof-oversize-b-20260805",
  "version": 1
}
```

Every asserted property holds live:

- reply arrives, no timeout
- `status: error`, `code: payload_too_large`
- `correlation_id` echoes the request `id` exactly
- names the measured bytes (63,370,799) and the broker's own advertised figure (8,388,608 = 8 MiB), read off the connection rather than hardcoded
- content-free: names the requested section, carries no pack data

**Small request still OK**

`repo_facts`, same identity/namespace, returned in ~1s: `status: ok`, `sections_receipt: {"requested":["repo_facts"],"served":["repo_facts"],"requested_not_served":[]}`, `correlation_id` echoed.

**One note for the record.** The same `durable_memory` request WITHOUT a `query` returns `status: ok` with an empty section (`empty_reason: "no_query"`) — recall is query-gated, so an unqualified `durable_memory` ask never reaches the oversize path. The oversize reply needs a query. Measured 63.4 MB here vs the 58.5 MB recorded on 2026-08-04; the corpus has grown since, and the point stands either way.

**Still open, unchanged:** the reply does not fit an 8 MB broker. The caller now gets a clear, correlated error rather than a pack. Resolving the size remains the operator's call, as the runbook says.

Issue #555, opened 2026-08-04
Client bundle ships no OPENBRAIN_DEVELOPMENT_ROOT — installs on any non-mini machine wedge every gate


rodaddy:
## Symptom

On a fresh client box whose Development root is not `/Volumes/ThunderBolt/Development`, **every tool call is blocked** by the context-budget gate, and the remediation text the banner prints names a path that does not exist on that machine. The session cannot be recovered from inside it — the operator has to edit files by hand on the box.

Field-proved on the Air (Development root `/Users/rico/Development`) on 2026-08-04.

## Root cause

`python/openbrain-provider/src/openbrain_provider/development_scope.py:36` ships the build machine's volume as the default lane root:

```python
DEFAULT_DEVELOPMENT_ROOT: Final[Path] = Path("/Volumes/ThunderBolt/Development")
```

The env seam already exists and works. `development_root()` (`development_scope.py:62-63`) reads `OPENBRAIN_DEVELOPMENT_ROOT` **per call**, and its docstring explains exactly why it is a function and not a module constant.

**Nothing ships the variable.** That is the whole defect:

- `scripts/setup-client.sh` never mentions `OPENBRAIN_DEVELOPMENT_ROOT` (zero matches).
- `scripts/client-bundle.sh` never mentions it (zero matches).
- The staged `env/openbrain-hook-env` allowlist does not pass it, so even a hand-set value in the env file is stripped before the hook child ever sees it. The wrapper's `exec env -i` list is closed by construction.

So on any box where the volume path does not exist, `_canonical_directory` fails the `is_dir()` test, `resolve_development_scope()` returns `None` for every cwd, and the gate has no project to scope state by.

The blocked-forever behavior comes from `context_budget_gate.py:286-306`. `_development_cwd()` falls back to composing `<root>/<project>` from `development_root()` — the mini volume — so the recovery command the banner tells the operator to paste points at a path that does not exist on that machine. Its own docstring calls this shape "the #419 deadlock arriving through the escape hatch meant to resolve it."

## Proof

Measured against a real non-symlinked root, clean subprocess per case:

```
=== BEFORE (no OPENBRAIN_DEVELOPMENT_ROOT shipped) ===
root : /Volumes/ThunderBolt/Development
scope: None

=== AFTER (variable exported, the Air's fix) ===
root : /Volumes/ThunderBolt/_tmp/.../devroot-probe/Development
[...]

PR #570, opened 2026-08-05
fix(installer): derive the hook wrapper's ENV_FILE, and fail loud when it is missing (#562)

### LINKED ISSUE #562: openbrain-hook-env:69 hardcodes ENV_FILE under /Users/rico -- silent zero-capture (exit 0) on any box with a different user
Surfaced by the Air provision dig-out (handoff: development branch wip/2026-08-04_air, _DOCS/_handoff/AIR-session-handoff-20260804.md). Same defect class as #557/#558's OPENBRAIN_DEVELOPMENT_ROOT default: a value correct on the origin box used as a silent fallback.

**Measured (both machines, same line):** `~/.local/share/openbrain-memory/env/openbrain-hook-env:69`:

    ENV_FILE="/Users/rico/.local/share/openbrain-memory/env/claudex-observation.env"

The installer writes this line, so every install carries the authoring user's absolute home. Works on the mini and the Air only because both are `/Users/rico`. On a cc-* box under a different user the path does not resolve, and per the wrapper's own header the failure mode is the worst kind: the Python config rejects/misses its env, the entrypoint swallows the exception, and the hook exits 0 with zero capture -- indistinguishable from a healthy no-op.

**Fix shape (PROPOSED):** derive the path instead of hardcoding it -- `$HOME`-relative, or from the script's own resolved location (the env file lives in the same directory as the wrapper, so `$(dirname "$(readlink -f "$0")")/claudex-observation.env` shapes both halves from one truth). Installer template changes accordingly.

**Acceptance:** on a box where $HOME is not /Users/rico, the wrapper finds its env file, and a deliberately broken path fails LOUD (non-zero + stderr), not exit-0-zero-capture. Rule per the incident: on the origin box 'resolved correctly' and 'fell back and got lucky' must be distinguishable.

rodaddy:
Fixes #562. Unblocks **rodaddy/development#87 step 1**.

## The defect

`openbrain-hook-env:69` carried its env-file path frozen from the machine that authored it:

```sh
ENV_FILE="/Users/rico/.local/share/openbrain-memory/env/claudex-observation.env"
```

It resolved on the mini and the Air only because both are `/Users/rico`. Under any other user the path did not exist, `. "$ENV_FILE"` found nothing, the Python config missed its environment, and — per the wrapper's own header — the hook entrypoint swallowed the exception (fail-open observer contract) and **exited 0 having captured nothing**, indistinguishable from a healthy no-op.

Fourth instance of one defect class, after #555/#557's `OPENBRAIN_DEVELOPMENT_ROOT` default: a value correct on the origin box used as a silent fallback everywhere else.

## Where the fix lands

There is **no in-repo source file** for the wrapper — `client-bundle.sh` copies the live installed copy off the build machine and `setup-client.sh` patches it at install time. The installer *is* the template, so that is what changed. The installed copy at `~/.local/share/openbrain-memory/env/` was treated as read-only throughout.

`setup-client.sh` previously **rewrote** the frozen literal to the install target's absolute path. It now **removes** it:

```sh
ENV_DIR="$(CDPATH='' cd -- "$(dirname -- "$0")" 2>/dev/null && pwd || :)"
```

`dirname` rather than `readlink -f`: settings.json invokes the wrapper by absolute path, and BSD `readlink` without `-f` is still in the field on the older boxes this bundle exists to serve. `$HOME` stays a fallback only, so it can never mask a correct derivation.

A missing env file now exits non-zero, one line on stderr naming the path *and* where it was derived from, stdout empty.

The rewrite is idempotent and **normalising** — it converts both the legacy assignment and an earlier managed block into the current form, so a bundle staged from an older mini still lands correct — and it refuses a wrapper matching neither shape rather than guessing an insertion point, because a wrapper that sources the wrong file is the silent zero capture this exists to remove.

`client-bundle.sh` gets a comment, not a second copy of the logic: the installer normalises on every install, and two copies of one rewrite drift.

## Verification — executed, not inferred

Normalised a **copy of the real installed wrapper** (never the original):

- ordering preserved — the `PARENT_OPENBRAIN_*` override capture still precedes `. "$ENV_FILE"`
- **foreign `$HOME`**: resolves and sources its env file, exit 0, child receives `OPENBRAIN_BASE_URL`
- **missing env file**: exit **1**, stderr exactly 1 line, stdout empty

```
openbrain-hook-env: env file missing or unreadable: .../noenv/claudex-observation.env (derived from wrapper directory derived from .../noenv/openbrain-hook-env)
```

- `bun test scripts/setup-client-hook-env.test.ts` → 6 pass
- `bun test scripts/` → 300 pass, 29 skip, **0 fail**
- `bash -n` clean on both installer scripts

**The fail-loud test was mutation-checked**: neutering the guard turns it red (`5 pass, 1 fail`). A green assertion that cannot fail would not have proved the acceptance criterion.

The new test drives the **real** heredoc extracted from the shipped `setup-client.sh` rather than a copy of the logic, since a copy is how a fix and its proof drift apart.

## Scope

Untouched, per the ticket: the `env -i` allowlist semantics, the parent-override capture, the conditional `OPENBRAIN_ALLOW_INSECURE_HTTP` block and its string-vs-non-string caveat, and the `OPENBRAIN_DEVELOPMENT_ROOT` pass-throughs.

## Runtime provenance

Code written by **Codex Sol** (`claudex sol low`). The first pass shipped a quoting-layer bug — a triple-quoted f-string turned `printf '%s\n'` into a literal newline, splitting the statement across two lines — caught by executing the emitted shell, returned to Sol verbatim, and fixed on the second pass. Reviewed, corrected, and verified by the owning Claude node.

Not merged, per policy.

## Critical Self-Review

- Highest-risk behavior: the hook wrapper now derives `ENV_FILE` from its own location instead of shipping a build-machine absolute path; a wrong derivation on a client box kills every capture hook at once. The mitigating half is the same PR: a missing/unreadable derived env file now fails LOUD (non-zero, stderr naming the derived path) instead of the prior exit-0 zero-capture, so a bad derivation is visible on first hook fire rather than silent forever.
- Assumptions that could be wrong: that every install layout keeps the env file in the wrapper-adjacent directory the derivation expects; a hand-moved env file that used to work by absolute path would now need to be where the derivation looks. `scripts/setup-client.sh:294-317` (already on main) writes it there, so installer-produced layouts conform.
- Missing/weak tests: `scripts/setup-client-hook-env.test.ts` covers derivation and the fail-loud path (":195-211" asserts stderr carries the derived path and "missing or unreadable"), but no test executes the wrapper on a box whose $HOME differs from the build machine — that is exactly the #562 field condition, and it is only simulated.
- Security/permission risk: none identified; no auth, token, namespace, or SQL surface. The env file path is derived, never interpolated from user input, and the failure message prints the path, not the file contents.
- Migration/deploy risk: no schema migration. Client boxes pick the change up on next installer run; an already-installed wrapper keeps working because the installer's ENV_FILE rewrite (main, `setup-client.sh:294-317`) already pinned its path.
- Downstream client/runtime risk: this is installer/client-lane, not MCP schema — but it IS client-facing per `docs/downstream-rollout.md`: #562 is a declared blocking prerequisite for rodaddy/development#87 (non-mini installs). Rollout to a real second box is the remaining proof.
- Rollback/cleanup concern: single-commit revert restores the hardcoded path (and its silent-failure behavior); nothing persisted. No cleanup needed.
- Fixes made before PR: CI test failure on Linux runners fixed in 94ae538 — the test hardcoded macOS-only `/Volumes/ThunderBolt/_tmp/...` at module top level, which threw EACCES on Linux before any test ran; scratch now lives under the gitignored `out/test-fixtures` (verified `git check-ignore -v`: `.gitignore:5`).
- Known residual risk: fail-loud is proven by test, not in the field; the derivation has not run on a box with a different $HOME this session. State: WRITTEN/merged-pending, not deployed.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no review swarm has run on this PR yet (it is queued behind this body gate); the one reusable gotcha — module-top-level absolute paths in tests explode on foreign runners before any test executes — will be promoted to `docs/sme/` if the swarm rates it MEDIUM+.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no MCP tool, schema, or live service surface changes; this is the client installer/hook-env lane, and the live proof belongs to the #87 rollout step, not this merge.

### DIFF
--- a/scripts/client-bundle.sh
+++ b/scripts/client-bundle.sh
@@ -129,6 +129,9 @@ cp -p "$SOURCE_ENV_DIR/openbrain-hook-env" "$STAGE/env/"
 chmod 600 "$STAGE/env/claudex-observation.env"
 chmod 755 "$STAGE/env/openbrain-hook-env"
 
+# setup-client.sh normalizes the wrapper's ENV_FILE block on every install, so
+# the bundle deliberately does not carry a second copy of that rewrite logic.
+
 # --- 2b. Development root, staged -------------------------------------------
 # The staged env dir is a copy of THIS machine's live installation, and this
 # machine is the one box where the provider's built-in default happens to be
--- a/scripts/setup-client-hook-env.test.ts
+++ b/scripts/setup-client-hook-env.test.ts
@@ -1,0 +1,228 @@
+import { describe, expect, it } from "bun:test";
+import { spawnSync } from "node:child_process";
+import {
+  chmodSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  writeFileSync,
+} from "node:fs";
+import { join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const repoRoot = fileURLToPath(new URL("..", import.meta.url));
+const setupClientPath = join(repoRoot, "scripts/setup-client.sh");
+const setupClient = readFileSync(setupClientPath, "utf8");
+const blockStart = "# --- begin hook env-file path normalization ---";
+const blockEnd = "# --- end hook env-file path normalization ---";
+const heredocStart = `python3 - "$TARGET_ENV_DIR/openbrain-hook-env" <<'PY'\n`;
+
+function extractNormalizer(): string {
+  const start = setupClient.indexOf(blockStart);
+  const end = setupClient.indexOf(blockEnd, start);
+  if (start < 0 || end < 0) {
+    throw new Error("setup-client hook env-file normalization markers missing");
+  }
+
+  const managedBlock = setupClient.slice(start, end);
+  const pythonStart = managedBlock.indexOf(heredocStart);
+  const pythonEnd = managedBlock.indexOf("\nPY\n", pythonStart);
+  if (pythonStart < 0 || pythonEnd < 0) {
+    throw new Error("setup-client hook env-file Python heredoc missing");
+  }
+
+  return managedBlock.slice(pythonStart + heredocStart.length, pythonEnd);
+}
+
+const scratchRoot = "/Volumes/ThunderBolt/_tmp/open-brain/_scratch";
+mkdirSync(scratchRoot, { recursive: true });
+const suiteRoot = mkdtempSync(join(scratchRoot, "setup-client-hook-env-"));
+process.stderr.write(`setup-client-hook-env fixtures: ${suiteRoot}\n`);
+const normalizerPath = join(suiteRoot, "normalize-hook-env.py");
+writeFileSync(normalizerPath, extractNormalizer());
+
+const authorEnvFile =
+  "/Users/rico/.local/share/openbrain-memory/env/claudex-observation.env";
+const managedBegin = "# >>> openbrain: hook env file (managed) >>>";
+
+function legacyWrapper(envFile = authorEnvFile): string {
+  return `#!/bin/sh
+set -eu
+
+ENV_FILE="${envFile}"
+
+PARENT_OPENBRAIN_TOKEN="\${OPENBRAIN_TOKEN:-}"
+PARENT_OPENBRAIN_BASE_URL="\${OPENBRAIN_BASE_URL:-}"
+
+set -a
+. "$ENV_FILE"
+set +a
+
+OPENBRAIN_TOKEN="\${PARENT_OPENBRAIN_TOKEN:-\${OPENBRAIN_TOKEN:-}}"
+OPENBRAIN_BASE_URL="\${PARENT_OPENBRAIN_BASE_URL:-\${OPENBRAIN_BASE_URL:-}}"
+
+exec env -i \\
+  PATH="$PATH" \\
+  HOME="$HOME" \\
+  OPENBRAIN_BASE_URL="\${OPENBRAIN_BASE_URL:-}" \\
+  OPENBRAIN_TOKEN="\${OPENBRAIN_TOKEN:-}" \\
+  "$@"
+`;
+}
+
+function createWrapper(directory: string, source = legacyWrapper()): string {
+  mkdirSync(directory, { recursive: true });
+  const wrapperPath = join(directory, "openbrain-hook-env");
+  writeFileSync(wrapperPath, source);
+  chmodSync(wrapperPath, 0o755);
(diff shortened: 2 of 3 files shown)


rodaddy:
### software-factory: failed
Reviewed at: `331e46e03263a24b1de9083005b9a25e34780e8c`

Findings:
- [high] New test file hardcodes the author's macOS volume as its scratch root, failing the whole file on CI's Linux runner (scripts/setup\-client\-hook\-env\.test\.ts:37\-39)
- [medium] A managed block with a damaged END marker leaves a permanently duplicated guard the normalizer can never reclaim (scripts/setup\-client\.sh:299\-318 \(managed/legacy alternation\))
- [low] Parent\-environment override precedence is claimed as verified but no test asserts it (scripts/setup\-client\-hook\-env\.test\.ts:110\-135)
- [low] ENV\_DIR derivation depends on argv\[0\] carrying a directory component (scripts/setup\-client\.sh:315 \(emitted ENV\_DIR line\))
- [info] Fail\-open to fail\-closed is the intended fix but changes behavior on every hook event (scripts/setup\-client\.sh:319\-322 \(emitted guard\))
- [info] client\-bundle\.sh change is comment\-only; staged bundles still carry the origin machine's absolute path (scripts/client\-bundle\.sh:132\-134)

rodaddy:
Review swarm (Light tier, adversarial fail-loud lane — GPT-5.6-Sol, wf_161bc66c-e19): **findings-must-fix**.

- **HIGH** `scripts/setup-client.sh:322` — the wrapper silently falls back to $HOME when directory derivation fails: cd fails → HOME env file sourced → exit 0, empty stderr. That is the fail-open path #562 exists to eliminate, and `scripts/setup-client-hook-env.test.ts:134-167` currently EXECUTES AND ACCEPTS this silent success.
- **MEDIUM** `scripts/setup-client.sh:320` — `dirname "$0"` is the invocation/caller path, not the wrapper's resolved location: symlinked, PATH-invoked, and sh-sourced layouts all derive the wrong directory (zsh $0 semantics differ again).

Fix lane dispatched; merge blocked until fix-delta verification passes.

Issue #395, opened 2026-07-25
DREAM-6: Weekly discard drain — give discarded_entries a writer


rodaddy:
Part of #389
Program: #320

## Problem

`discarded_entries` exists in the schema, has the right columns
(`original_content`, `original_id`, `source_table`, `tier_at_discard`,
`access_summary`, `discarded_at`), and contains **0 rows**.

Grep for it across `src/` and `scripts/` returns exactly one reference:
`scripts/backup-lib.ts`. The table is backed up faithfully and has been empty
since creation. Nothing has ever written to it.

This is the clearest single instance of the pattern this epic exists to fix:
the system carefully preserves a table that no code path fills.

## Scope

The discard transition of the lifecycle — Rico "once a week ballpin hammer to
the head" session.

- Weekly job identifying entries that have aged out without ever being used.
- Write to `discarded_entries` with `tier_at_discard` and `access_summary`
  intact, so the discard is a **record**, not a deletion.
- Raw turns that never distilled into anything keepable within ~1 week are the
  primary input (per the resolved retention design: candidates get embeddings,
  raw turns get FTS only and a conditional TTL).

## Conditional, not purely time-based

The expiry rule is "has not been used to distill into something keepable",
**not** "is older than N days". A turn whose candidate was promoted has done
its job. A turn that produced nothing in a week probably never will.

## Eviction, not destruction

Optional follow-on discussed but not scoped here: index-then-archive to
compressed cold storage on the NAS, keeping a queryable stub row in Postgres
(id, session_ref, date, content_hash, archive filename, offset) so existence
questions never leave the box while bodies live off it.

Deferred deliberately — the immediate problem is that the table is empty, not
that it will be too full.

## Storage context (measured, corrects an earlier assumption)

`zz_test_raw_turns` measured 8,816 kB for 2 days / 6,403 turns → ~1.6 GB/year
against 1.4 TB free. Raw dialogue is **not** a disk-exhaustion path, which was
the premise of #381 and is wrong. With a ~1 week TTL steady state is ~30 MB and
no embeddings.

The real risk is **distillation falling behind ingest** — turns expiring before
being distilled. The metric that matters is *oldest undistilled turn age vs.
TTL*, not backlog bytes. #381 should be rewritten around that.

## Verification

- `discarded_entries` is non-empty after a drain cycle.
- A turn that produced a promoted candidate is not discarded.
- A turn older than the TTL with no candidate is discarded with tier and access
  history preserved.

Issue #380, opened 2026-07-24
INGEST-1: Add content-bearing raw turn table and full-send ingest


rodaddy:
Parent: #342
Program: #320
Spec: `docs/full-send-derivation-spec.md`

## Problem

The client hand-picks what to remember and nothing persists raw dialogue.
Measured 2026-07-24 on the local dogfood clone: a full working day produced
**3** `ob_session_events`, all written because the assistant manually chose to
run the capture command. `mcp_tool_audit_log` recorded 445 calls but is
content-free by design (`declared_parameter_keys`, `payload_size_bucket`,
`duration_ms`; no content column), so it cannot serve as a raw stream.
Prompts and responses are persisted nowhere.

Distillation and derivation are impossible because **their input does not
exist**.

## Scope

- Add `ob_raw_turns` (content-bearing, namespace-bound, `gen_random_uuid()`
  primary key, unique on `(namespace, lane_id, content_hash)` for idempotent
  re-send, partial index on `distilled_at IS NULL` as the sweep work queue).
- Add an `ingest_raw_turn` tool, agent role and above, namespace predicate
  derived from auth like every other write.
- Wire `UserPromptSubmit` (user turn) and `Stop` (assistant turn) to
  full-send. Batching allowed; the call must never block the interactive turn.
- Failures spool locally and replay through the existing spool path.

## Non-goals

- The client applies **no** salience judgment: no summarizing, scoring, or
  skipping. It ships what happened.
- No retrieval, `agent_context_pack`, or prompt-placement changes.

## Acceptance criteria

- [ ] A working day produces raw turns in the hundreds, not events in the
      single digits.
- [ ] Re-sending an identical turn conflicts and is ignored rather than
      duplicated.
- [ ] Namespace isolation regression test proves an ID-based read/mutation
      carries the auth-derived namespace predicate.
- [ ] Ingest failure never blocks or errors the interactive turn.

## Blocks

Retention/backpressure (INGEST-2) must land before sustained use.

PR #574, opened 2026-08-05
fix(doctor): alarm when distillation falls behind ingest

### LINKED ISSUE #381: INGEST-2: Alarm when distillation falls behind ingest (oldest undistilled age vs TTL)
Parent: #342
Program: #320
Spec: `docs/full-send-derivation-spec.md`
Depends on: #380
Related: #395

## Correction to the original framing

This issue was originally filed as *"Bound ob_raw_turns with retention,
backpressure, and saturation metrics"*, asserting that unbounded raw dialogue
is a **disk-exhaustion path** and that the #300 capacity envelope is therefore
a hard prerequisite for full-send.

**That premise was asserted, not measured, and it is wrong.**

Measured 2026-07-24 against `zz_test_raw_turns`:

| Window | Raw turns |
|---|---|
| 2 days / 6,403 turns | 8,816 kB |
| 1 month (extrapolated) | ~132 MB |
| 1 year (extrapolated) | ~1.6 GB |
| Free on volume | 1.4 TB |

At ~1.6 GB/year this is not a disk problem at any horizon that matters. And
with the retention design resolved — candidates get embeddings, raw turns get
FTS only plus a conditional ~1 week TTL (#395) — raw turns never reach a year.
Steady state is ~30 MB with no embeddings, which was the expensive part.

Blocking full-send on a capacity envelope was gating real work on a
non-problem.

## The real risk

Not disk. **Distillation falling behind ingest** — turns expiring at the TTL
before they are ever distilled, so content is lost silently.

That is a throughput alarm, not a capacity limit. The metric that matters is:

> **oldest undistilled turn age ÷ TTL**

As that ratio approaches 1, content is about to be dropped. Backlog *bytes* are
not the signal; backlog *age* is.

## Scope

- Track and expose oldest undistilled turn age per namespace.
- Alarm when the age/TTL ratio crosses a warning threshold (proposed 0.5) and
  again at critical (proposed 0.8).
- Content-free saturation metrics: ingest rate, undistilled depth, oldest
  undistilled age, distill throughput, dead-letter count.
- Backpressure on ingest **only** if the alarm proves insufficient in practice.
  No acknowledged turn is ever silently dropped.

## Acceptance criteria

- [ ] Oldest undistilled age is queryable per namespace and exposed as a
      content-free metric.
- [ ] Crossing the warning and critical thresholds is observable and tested.
- [ ] Metrics expose depth/age/throughput without payloads.
- [ ] Redaction, namespace, and exact-scope controls are never bypassed under
      pressure.
- [ ] No acknowledged turn is silently dropped.

## Note

Retention/eviction itself belongs to #395 (weekly discard drain), which owns the
conditional TTL and the write into `discarded_entries`. This issue owns only the
**alarm** that distillation is losing the race.

Scope-shares with #300. Close whichever is redundant once both are specified; do
not implement the envelope twice.

rodaddy:
Closes #381 (INGEST-2).

Alarms when distillation falls behind ingest. The signal is backlog **age**, not
bytes: oldest undistilled turn age divided by the raw-turn TTL, per namespace.
As that ratio approaches 1, acknowledged content is about to expire before it is
ever distilled.

## What changed

- `operator_doctor` gains a `distillation_lag` array: one entry per namespace
  with `undistilled_depth`, `oldest_undistilled_age_seconds`, `ratio`, and
  `level` (`ok` / `warning` / `critical`). Thresholds are 0.5 and 0.8 from the
  issue, exported as named constants.
- The query is the work-queue shape `idx_ob_raw_turns_undistilled` already
  indexes (`distilled_at IS NULL AND retention_tier = 'live'`), parameterized,
  grouped by namespace, and wrapped in the doctor's existing optional-timeout
  helper. A probe failure logs content-free and degrades to `[]` rather than
  failing the whole doctor payload.
- Critical lag degrades `status` to `degraded`; warning lag stays `healthy`.
  `unhealthy` remains reserved for a database hard failure, per the existing
  tier rules.
- `OPENBRAIN_RAW_TURN_TTL_SECONDS` (default 604800 = 7 days) supplies the
  denominator, documented in `.env.example`, `.env.schema`, and
  `docs/CONFIG_REFERENCE.md`.
- `DOCTOR_CONTRACT_VERSION` bumped to `2026-08-05.operator-doctor.v3`, with all
  four independent assertions of it updated.

**Not in scope:** retention and eviction belong to #395. This PR reads the TTL
as an alarm denominator only; it never expires, archives, or deletes a turn. No
backpressure on ingest — the issue asks for that only if the alarm proves
insufficient in practice.

## Acceptance criteria

- [x] Oldest undistilled age queryable per namespace, exposed content-free.
- [x] Warning and critical crossings observable and tested.
- [x] Depth / age / throughput exposed without payloads.
- [x] Namespace grouping preserved; no cross-namespace resolution.
- [x] No acknowledged turn is silently dropped (nothing is deleted at all).

## Critical Self-Review

- Highest-risk behavior: critical lag degrades the doctor `status` tier, so a noisy alarm could mask a real embedding or migration degradation in an already-degraded payload; chose it deliberately because content about to be lost is what a doctor tier is for, and left warning-level neutral so only the near-loss case moves the tier.
- Assumptions that could be wrong: that the TTL is one global value — #395 owns the real conditional TTL and may make it per-namespace, which would turn this denominator into an approximation; and that `created_at` is the right age basis, since a spool-replayed turn has an older `occurred_at` and a batched backfill therefore under-reports its own age.
- Missing/weak tests: the SQL is not executed against a real Postgres — the pool is faked, so the aggregate, the `EXTRACT(EPOCH ...)` arithmetic, and index usage are unproven on a live server; classification, payload shape, tier effect, and probe-failure fallback are all covered, and the thresholds were mutation-checked by forcing the critical branch false, which fails two tests on the `level` value rather than merely on imports.
- Security/permission risk: low — the surface is already admin/ob-admin only via `canReadDoctor`, and the payload is content-free by construction (namespace, counts, ages) with a test asserting that content, content_hash, and turn ids present in a row never reach the serialized output; namespace names do become visible to a doctor reader, consistent with the rest of the privileged payload.
- Migration/deploy risk: none — no schema change, and it reads a table and indexes #380 already shipped; where migration 032 has not run the query throws and the fallback returns `[]`, so the doctor still answers.
- Downstream client/runtime risk: additive field on a frozen contract, so `DOCTOR_CONTRACT_VERSION` is bumped and any consumer asserting the exact key set or version literal must update — exactly the failure the second commit fixes in three in-repo places; the pre-push contract-parity subset reported client/contract paths unchanged.
- Rollback/cleanup concern: revert is clean — one field, one env var, no state written anywhere, nothing to unwind.
- Fixes made before PR: the contract bump had reached only the shape-lock test, leaving three assertions pinned at v2 — the MCP tool test (actively failing), `src/server.test.ts` (green only because its Postgres case skips silently without `OPENBRAIN_TEST_DATABASE_URL`, so it would fail wherever one is set), and a comment in `server/tools/operator-doctor.ts` citing the version while explaining the lock; all four now agree.
- Known residual risk: the age basis (`created_at` vs `occurred_at`) and the global-TTL assumption both need revisiting when #395 lands its real retention policy, though neither can lose data today since this PR only reports.
- SME review-memory update: [x] not applicable because: no new MEDIUM+ review finding surfaced — the one defect found (a stale contract-version literal in three places) is already policed by the existing shape-lock test convention documented in `src/operator-doctor.ts`.

## Review Gate

- Live Open Brain checks: [x] not applicable because: no live service call is involved — the change is a read-only diagnostic field verified against a faked pool, with no deploy or hosted smoke performed in this PR.
- `bunx tsc --noEmit` — clean, exit 0.
- `bun test src/operator-doctor.test.ts src/tools/__tests__/operator-doctor.test.ts src/server.test.ts` — 43 pass, 0 fail.
- Full pre-push suite — 3191 pass, 506 skip, 0 fail across 231 files.
- Failure proven before trusting the test: mutating the critical threshold to a constant-false branch fails `crosses the warning and critical TTL ratios` and `reports content-free distillation lag separately for each namespace` on the `level` value.
- Note for reviewers: Postgres-backed tests skip without `OPENBRAIN_TEST_DATABASE_URL`, so the green run above does not exercise the real SQL. Worth a live-DB check of the aggregate before this is relied on operationally.

## Contract Parity

- `contracts/server/server-operator-diagnostics-admin.fixture.json` bumped to `2026-08-05.operator-doctor.v3` (the fifth version site the review swarm found hiding behind the silently-skipped parity suite). Parity suite executed with `OPENBRAIN_TEST_DATABASE_URL` set: 69 pass / 7 fail, the 7 being pre-existing `server-rewrite-scaffold` failures reproduced identically on an `origin/main` baseline worktree 3/3 runs — `server-operator-diagnostics-admin` passes for both providers.


rodaddy:

### MEDIUM: `contracts/server/server-operator-diagnostics-admin.fixture.json`:21

**The recorded parity contract fixture still pins "2026-07-08.operator-doctor.v2" while DOCTOR_CONTRACT_VERSION is now v3. The PR body claims "all four independent assertions of it updated" and "all four now agree" — there are five sites, and this fifth one was missed. It is green only because contracts/server-tool-parity.test.ts:51 does `const dbDescribe = DB_URL ? describe : describe.skip`, so the whole parity suite skips silently without OPENBRAIN_TEST_DATABASE_URL. This is the exact failure mode the author already found and fixed in src/server.test.ts, and it is docs/sme/correctness.md's [2026-06-19] "mock-pool / silently-skipped tests cannot catch this" pattern recurring one file over.**

Failure scenario: Set OPENBRAIN_TEST_DATABASE_URL and run `bun test contracts/server-tool-parity.test.ts`. The operator-diagnostics-admin fixture asserts contract_version === "2026-07-08.operator-doctor.v2" against a payload that now returns "2026-08-05.operator-doctor.v3", so the fixture fails for both declared providers (current-src, server-rewrite-scaffold). On any CI runner or dev box that has a test database configured — i.e. the environment where contract parity is actually meant to be enforced — this PR is red. The PR body's "Full pre-push suite — 3191 pass, 506 skip, 0 fail" counted this fixture inside the 506 skips.

### LOW: `src/operator-doctor.ts`:384

**The age basis is created_at, and on the live dogfood corpus 1272 raw turns have created_at more than a day later than occurred_at (max skew 293747s ≈ 3.4 days against a 7-day TTL, i.e. up to 0.48 of the alarm ratio), all in the same direction — no row has occurred_at later than created_at. A replayed/backfilled turn therefore under-reports its real age by up to nearly half the critical threshold. The PR body discloses this as known residual risk and defers it to #395, and it is defensible so long as #395's eventual eviction also keys on created_at; it is only a defect if retention expires on occurred_at while the alarm measures created_at.**

Failure scenario: A spool replay lands a batch of turns whose occurred_at is 6 days old with created_at = now(). The doctor reports ratio ≈ 0 / level ok for that namespace. If #395 later expires on occurred_at (or on any conversation-time basis), those turns are deleted roughly one day later with the alarm never having left "ok" — precisely the near-loss case this field exists to catch. Worth pinning the basis explicitly in the code comment so #395 cannot diverge silently.

### LOW: `src/operator-doctor.ts`:297

**Operational, not a code defect, but the reviewer verified it and it should be known before merge: on the current dogfood database this alarm fires immediately. The PR's exact SQL, run against real Postgres, returns 27 namespaces with a max ratio of 1.46 — well past the 0.8 critical threshold — so the doctor's overall `status` flips to `degraded` the moment this deploys, driven by leftover `parity-raw-turn-*` test-fixture namespaces that have no distiller. That is the noisy-alarm-masks-real-degradation risk the author names as the highest-risk behavior in the self-review, realized on day one.**

Failure scenario: Deploy to a host carrying the existing raw-turn corpus and call operator_doctor: status is "degraded" with distillation_lag entries at ratio > 1.0 for ~27 parity-test namespaces. A genuine embedding-provider or migration degradation arriving afterward changes nothing observable, because the tier is already degraded and stays degraded. Either exclude parity/test namespaces, or clear that corpus, before the tier effect is relied on operationally.

Issue #342, opened 2026-07-22
[Epic] Unify autonomous memory maintenance


rodaddy:
## Summary

Create one durable server-owned maintenance queue and scheduler for spool replay, embedding repair, extraction, graph derivation, and final rollout tuning.

**Workstream key:** `MAINT`  
**Priority:** P1  
**Parent:** #320  
**Depends on:** #336, #331, #321

## Scope

- Persist, schedule, retry, and observe maintenance jobs.
- Move existing replay into the common plane without semantic changes.
- Repair embeddings and derive metadata and graph edges asynchronously.
- Tune and validate the integrated system.

## Sub-issues

- [ ] #343 — Build the maintenance queue
- [ ] #344 — Queue spool replay
- [ ] #345 — Repair stale embeddings
- [ ] #346 — Derive metadata and graph knowledge
- [ ] #347 — Tune and roll out local recall

## Acceptance criteria

- [ ] All named maintenance work runs through one durable plane with content-free observability.
- [ ] Graph derivation enriches existing graph-enabled retrieval without crossing namespaces.
- [ ] The final rollout receipt demonstrates target recall quality and preserved safety constraints.

## Validation

- [ ] Run restart, retry, idempotency, isolation, derivation, and integration suites.
- [ ] Run MAINT-5 against EVAL-3 and downstream canaries.

## Non-goals

- Mutating Dream planning paths.
- Cloud scheduling.
- Introducing new memory-tier semantics.

## Standing constraints

- Exact namespace and session-scope isolation remains a security boundary.
- Prompt placement stays client/runtime-owned; no implicit MCP `_meta` injection.
- Observability remains content-free.
- Dream planning remains dry-run-safe unless a separately authorized mutating wrapper is used.

## Close-run map (2026-08-05 audit)
Maintenance/scheduling members now tracked here: #381 (distillation-lag alarm), #384 (MAINT-6 runner + sweep producer), #386 #387 #388 (QMD-1..3 index jobs). Candidate for an early close run — small, mechanical, mostly wiring scheduled jobs.

rodaddy:
## Sub-issues added: full-send ingest and server-side derivation

Spec: `docs/full-send-derivation-spec.md` (PR #385)

Measured 2026-07-24 on the local dogfood clone: the maintenance machinery this
epic owns is **built and switched off**. `OPEN_BRAIN_MAINTENANCE_ENABLED=0`
locally, unset on core01, and exactly one job has ever been enqueued on either
[...]

Issue #320, opened 2026-07-22
[Program] Deliver near-total local recall


rodaddy:
## Summary

Deliver near-total local recall that surfaces the right durable memory at the right moment, draws only from approved local sources, maintains itself autonomously, and is governed by live isolated-namespace relevance gates.

**Workstream key:** `PROGRAM`  
**Priority:** P0  
**Parent:** None  
**Depends on:** None

## Scope

- Sequence delivery through live evaluation, complete context packs, per-turn reflexes, source synchronization, autonomous maintenance, derivation, and rollout tuning.
- Preserve exact namespace and session isolation on every read and mutation.
- Keep observability content-free and Dream planning dry-run-safe.
- Keep prompt placement owned by clients and runtimes; never inject implicitly through MCP `_meta`.

## Sub-issues

- [ ] #321 — Gate live recall quality
- [ ] #325 — Complete agent context packs
- [ ] #331 — Add a per-turn recall reflex
- [ ] #336 — Synchronize approved local sources
- [ ] #342 — Unify autonomous memory maintenance

## Critical path

#322 → #323 → #324 → #326 → #327 → #329 → #330 → #332 → #333 → #334 → #335 → #337 → #338 → #341 → #343 → #346 → #347

### Wave 1 — establish live quality baseline

#322

### Wave 2 — score and activate the gate

#323, #324

### Wave 3 — establish pack budgeting

#326

### Wave 4 — populate pack sections

#327, #328

### Wave 5 — finish and validate pack output

#329, #330

### Wave 6 — begin reflex and source foundations

#332, #337

### Wave 7 — recall suppression and source ingestion

#333, #338, #339, #340

### Wave 8 — validate reflex and lexical retrieval

#334, #335, #341

### Wave 9 — establish autonomous maintenance

#343

### Wave 10 — add maintenance handlers

#344, #345, #346

### Wave 11 — tune and roll out

#347

## Acceptance criteria

- [ ] All five delivery epics are closed with their acceptance signals green.
- [ ] The final system passes the live isolated-namespace recall-quality gate at or above the thresholds established by EVAL-2.
- [ ] MAINT-5 records measured recall improvement from baseline through rollout.

## Validation

- [ ] Run the EVAL-3 live gate against the final integrated system in a throwaway namespace.
- [ ] Verify downstream rollout and canary requirements under `docs/downstream-rollout.md`.
[...]

PR #575, opened 2026-08-05
fix(qmd): restore the index sync as a standalone scheduled job (#386)

### LINKED ISSUE #386: QMD-1: Restore the qmd sync as a standalone scheduled job (orphaned when session-wrap was removed)
Parent: #342
Program: #320

## Problem

qmd is already wired into Open Brain recall across 22 source files
(`src/tools/search-all.ts`, `src/tools/repo-facts.ts`, the context-pack
repo-facts sections, `scripts/promote-qmd-repo-facts.ts`,
`scripts/core01-qmd-bootstrap.sh`). The integration works.

Nothing updates the index. Measured 2026-07-24:

| Metric | Value |
|---|---|
| Files indexed | 52,337 |
| Vectors | 213,950 |
| Index size | 2.2 GB |
| Last updated | **39 days** |
| `open-brain` collection | **absent** |

Every context-pack repo-facts section has therefore been serving a stale and
incomplete view of the codebase.

**Root cause:** the qmd sync used to run as part of session-wrap. Session-wrap
was removed and took the sync with it. This was orphaned, not merely
unscheduled — and it failed silently, because a stale index still returns
results, just old ones.

Session-wrap was also the wrong home:

- frequency tied to session count, not to code changing
- ran ~30x/day when nothing had changed
- ran mid-flow, competing with active work
- failures disappeared into session noise

## Scope

- Standalone machine-owned scheduled job at 04:00 local, detached from any
  human session: `qmd update` followed by `qmd embed`.
- launchd (macOS), consistent with `com.rico.open-brain`.
- Log last-run time, files indexed, files updated, vectors embedded.
- Non-zero exit on failure so the failure is visible rather than silent.

## Out of scope

- Collection triage (see QMD-2).
- Folding this into a dream/deep stage. Worth revisiting once the dream stages
  exist, since a nightly mechanical refresh is deep-stage-shaped work, but the
  standalone job is independently useful and should not block on that design.

## Verification

- Job registered and fires on schedule.
- `qmd status` shows an updated timestamp within 24h.
- A search for a symbol added the previous day returns the file.

rodaddy:
Closes #386 (QMD-1). Parent #342, program #320.

## What this does

Restores the qmd index sync as a standalone, machine-owned scheduled job at
04:00 local, detached from any human session. The sync was orphaned when
session-wrap was removed, and it failed silently because a stale index still
returns results — just old ones.

Rescoped per the 2026-08-04 issue comment: the single 52k-file global index was
retired 2026-07-29/30 in favour of one project-local `.qmd/` per repo, so the
job iterates the per-repo indexes plus `global_docs_instructions`, not one
`qmd update && qmd embed` pair.

- `scripts/qmd-sync.sh` — walks the Development root and each immediate child
  git repo carrying `.qmd/index.yml` (53 repos + the Dev root = 54 today), then
  the global index. Per index it runs `update`, `embed`, `status` and emits a
  structured line with `last_run`, `files_indexed`, `files_new`,
  `files_updated`, `vectors_embedded`, `vectors_total`. It continues past an
  individual failure so one bad repo does not hide the rest, and exits non-zero
  if any index failed.
- `docs/deploy/com.rico.qmd-sync.plist.template` — launchd template, 04:00
  local, styled after `com.rico.open-brain-nats-worker.plist.template`. No
  `RunAtLoad`, so bootstrapping does not kick off an immediate sync.
- `docs/deploy/qmd-sync.md` — install/load, verification, log locations,
  failure diagnosis.

**Nothing was installed or loaded.** No copy into `~/Library/LaunchAgents`, no
`launchctl load`/`bootstrap`/`kickstart`, and the sync was never executed.
Installation is an operator/deploy step and the doc says so. This PR is
WRITTEN, not RUNNING — the issue's "job fires on schedule" verification cannot
be satisfied until an operator installs it.

## Checks

Run by the supervising node, not merely reported by the author:

- `bash -n scripts/qmd-sync.sh` (Homebrew bash) — clean
- `shellcheck scripts/qmd-sync.sh` — clean, zero findings
- `plutil -lint docs/deploy/com.rico.qmd-sync.plist.template` — OK
- forbidden-path scan across all three files for sandbox-local temp dirs and
  Apple's ancient shells — no hits
- `scripts/qmd-sync.sh` is mode 755
- pre-push hook — all checks passed
- `bunx tsc --noEmit` not run: no TypeScript touched (3 files, all
  shell/plist/markdown)

Assumptions verified against the live machine rather than assumed: the `qmd`
binary exists at the configured path and is executable; 53 repos under
Development carry `.qmd/index.yml` plus the Dev root itself, matching the issue
comment's count of 54.

## Critical Self-Review

- Highest-risk behavior: metric parsing — the script scrapes qmd's human output with awk for `Indexed:`, `Done!.*Embedded N chunks`, `Total:` and `Vectors:`, so a wording change in qmd breaks it; mitigated by treating unparseable output as a failure with a non-zero exit rather than silently logging zeros, so it fails loud instead of going false-green.
- Assumptions that could be wrong: that qmd sits at the hardcoded default path (verified today, overridable via `QMD_BIN`); that every syncable index is the Dev root or an immediate child git repo holding `.qmd/index.yml`, so a nested or non-git index is skipped silently; and that the deployed app root is where the script lands on core01.
- Missing/weak tests: there are none and this is the real gap — the script is not exercised by `bun test`, so correctness rests only on shellcheck, `bash -n` and reading; an end-to-end run against a scratch index would be the honest test but requires executing the sync, which this PR deliberately does not do.
- Security/permission risk: low — `umask 077` on a log holding repo names and file counts but no credentials, no secrets, no network calls qmd did not already make, and `PATH` set explicitly rather than inherited, which is correct for a launchd job.
- Migration/deploy risk: the job writes to a logs directory under the deployed app root that does not exist on this dev machine, covered by an operator `mkdir -p` in the doc and by the script's own `mkdir -p`; if the ThunderBolt volume is unmounted at 04:00 the job fails loudly rather than indexing nothing, which is correct but noisy on an unmounted boot.
- Downstream client/runtime risk: none direct — no MCP tool, schema, transport or Python-client surface is touched, so `docs/downstream-rollout.md` does not apply; the only effect is indirect upside, in that context-pack repo-facts sections stop serving a stale index once the job actually runs.
- Rollback/cleanup concern: clean — rollback is `launchctl bootout` plus removing the plist, nothing in the repo depends on these three files, the change is purely additive and no existing behavior was modified.
- Fixes made before PR: none were required — every check the author reported reproduced exactly when re-run independently by the supervising node, and the path and repo-count assumptions were confirmed against the live filesystem rather than accepted from the report.
- Known residual risk: the job is unproven in execution and has never run on any host, so all three of the issue's verification criteria (registered and firing on schedule, `qmd status` fresh within 24h, previous-day symbol searchable) require the operator install step and none are satisfied by this PR — do not close #386 on merge.
- SME review-memory update: [x] not applicable because: no review findings yet; if review surfaces a MEDIUM+ pattern it gets promoted into the matching `docs/sme/` lane file with provenance before merge.

## Review Gate

- Live Open Brain checks: [x] not applicable because: no MCP tool, schema, transport, or recall surface is touched — this adds a shell script, a launchd plist template, and a doc, none of which the running service loads.
- Scope: 3 new files, additive only, no existing file modified. Shell + launchd
  plist + markdown; zero TypeScript, zero SQL, zero MCP surface.
- Reviewer focus: the awk metric parsing in `sync_index` and the `set -e`
  interaction — failure-prone commands sit inside `if` conditions and counters
  use plain assignment rather than `((n++))` specifically to avoid the
  zero-result `set -e` trap. Worth a second pair of eyes.
- Also worth checking: whether the repo-discovery glob should recurse rather
  than stopping at immediate children of the Dev root.
- Not in scope: collection triage (QMD-2), folding this into a dream/deep stage
  (deferred in the issue, and that deferral still looks right).
- Do not merge-and-close: merging makes this MERGED, not RUNNING. #386 stays
  open until an operator installs the job and the schedule is observed firing.

### DIFF
--- a/docs/deploy/com.rico.qmd-sync.plist.template
+++ b/docs/deploy/com.rico.qmd-sync.plist.template
@@ -34,8 +34,8 @@
   <string>Background</string>
 
   <key>StandardOutPath</key>
-  <string>/Volumes/ThunderBolt/open-brain/logs/qmd-sync.launchd.out.log</string>
+  <string>/Users/rico/Library/Logs/com.rico.qmd-sync.out.log</string>
   <key>StandardErrorPath</key>
-  <string>/Volumes/ThunderBolt/open-brain/logs/qmd-sync.launchd.err.log</string>
+  <string>/Users/rico/Library/Logs/com.rico.qmd-sync.err.log</string>
 </dict>
 </plist>
--- a/docs/deploy/com.rico.qmd-sync.plist.template
+++ b/docs/deploy/com.rico.qmd-sync.plist.template
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN"
+  "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+  <key>Label</key>
+  <string>com.rico.qmd-sync</string>
+
+  <key>ProgramArguments</key>
+  <array>
+    <string>/Volumes/ThunderBolt/open-brain/app/scripts/qmd-sync.sh</string>
+  </array>
+
+  <key>WorkingDirectory</key>
+  <string>/Volumes/ThunderBolt/open-brain/app</string>
+
+  <key>EnvironmentVariables</key>
+  <dict>
+    <key>HOME</key>
+    <string>/Users/rico</string>
+    <key>PATH</key>
+    <string>/Users/rico/.local/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
+  </dict>
+
+  <key>StartCalendarInterval</key>
+  <dict>
+    <key>Hour</key>
+    <integer>4</integer>
+    <key>Minute</key>
+    <integer>0</integer>
+  </dict>
+
+  <key>ProcessType</key>
+  <string>Background</string>
+
+  <key>StandardOutPath</key>
+  <string>/Volumes/ThunderBolt/open-brain/logs/qmd-sync.launchd.out.log</string>
+  <key>StandardErrorPath</key>
+  <string>/Volumes/ThunderBolt/open-brain/logs/qmd-sync.launchd.err.log</string>
+</dict>
+</plist>
--- a/docs/deploy/qmd-sync.md
+++ b/docs/deploy/qmd-sync.md
@@ -1,0 +1,138 @@
+# Scheduled qmd sync
+
+`com.rico.qmd-sync` refreshes the machine's existing qmd indexes every day at
+04:00 local time. It is independent of agent sessions and runs all indexes
+sequentially because embedding is GPU-bound.
+
+The job updates:
+
+- every project-local `.qmd/index.yml` owned by the Development root or an
+  immediate child Git repository under `/Volumes/ThunderBolt/Development`;
+- the separate `global_docs_instructions` named index.
+
+It does not create or regenerate index configurations. Adding, removing, or
+triaging collections is separate work.
+
+## Deployment boundary
+
+The repository ships a script and a launchd template only. It does **not**
+install, load, bootstrap, or run the job automatically. An operator performs
+installation after the deployed app at `/Volumes/ThunderBolt/open-brain/app`
+contains `scripts/qmd-sync.sh`.
+
+The sync script writes its durable primary log to:
+
+```text
+/Volumes/ThunderBolt/open-brain/logs/qmd-sync.log
+```
+
+launchd also reserves these files for startup errors that occur before the
+script opens its primary log:
+
+```text
+/Volumes/ThunderBolt/open-brain/logs/qmd-sync.launchd.out.log
+/Volumes/ThunderBolt/open-brain/logs/qmd-sync.launchd.err.log
+```
+
+## Install and load
+
+Run these commands as the logged-in macOS user, not as root:
+
+```bash
+mkdir -p /Volumes/ThunderBolt/open-brain/logs
+cp /Volumes/ThunderBolt/open-brain/app/docs/deploy/com.rico.qmd-sync.plist.template \
+  "$HOME/Library/LaunchAgents/com.rico.qmd-sync.plist"
+plutil -lint "$HOME/Library/LaunchAgents/com.rico.qmd-sync.plist"
+launchctl bootstrap "gui/$(id -u)" \
+  "$HOME/Library/LaunchAgents/com.rico.qmd-sync.plist"
+launchctl print "gui/$(id -u)/com.rico.qmd-sync"
+```
+
+The template deliberately omits `RunAtLoad`, so loading it registers the 04:00
+schedule without immediately starting a full sync. To replace an already loaded
+copy, boot it out first, copy the updated template, and bootstrap it again:
+
+```bash
+launchctl bootout "gui/$(id -u)" \
+  "$HOME/Library/LaunchAgents/com.rico.qmd-sync.plist"
+```
+
+## Verify the next scheduled run
+
+After the next 04:00 local run:
+
+1. Confirm launchd still has the job and inspect its last exit status:
+
+   ```bash
+   launchctl print "gui/$(id -u)/com.rico.qmd-sync"
+   ```
+
+2. In at least one active Development repository, confirm `Updated` is within
+   24 hours:
+
+   ```bash
+   cd /Volumes/ThunderBolt/Development/open-brain
+   qmd status
+   ```
+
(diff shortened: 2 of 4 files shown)


rodaddy:

### HIGH — `scripts/qmd-sync.sh`:73

**The embed metric parsing and its unparseable-output guard both use `[0-9]+`, but qmd formats the chunk count with `toLocaleString("en-US")`, so any embed of >=1000 chunks prints `Embedded 8,676 chunks` and fails to match. This inverts the PR's headline "fails loud instead of false-green" claim into a false-RED: a fully successful sync is reported as failed.**

Verified against the actually-executed artifact, not the PR body. The `~/.local/bin/qmd` wrapper dispatches to `~/node_modules/@tobilu/qmd/dist/cli/qmd.js`, where line 3095 defines `function formatCount(n) { return n.toLocaleString("en-US"); }` and the embed completion line is emitted as `Done!... Embedded ${formatCount(result.chunksEmbedded)} chunks from ...`. Reproduced directly: feeding `Done! Embedded 12,345 chunks from 2,001 documents in 3m` through the script's own awk yields `vectors_embedded=0`, and the guard `grep -Eq 'Done!.*Embedded [0-9]+ chunks|All content hashes already have embeddings|No non-empty documents to embed'` returns non-zero -> the script logs `status=failed step=metrics reason=embed-output-unparseable`, returns 1, increments `failures`, and `main` exits non-zero. This is not a hypothetical scale: `qmd status` in open-brain today reports `Vectors: 8676 embedded`, and every repo's first-ever sync embeds all chunks in a single run, so `Embedded 8,676 chunks` is the expected steady-state output for a large repo. Across 54 indexes the job will report a failed run on nearly every night that any sizeable repo has new content, while the sync itself actually succeeded. `qmd status` parsing (`Total:`/`Vectors:`) is NOT affected because those interpolate the raw `${count}` with no formatCount, which is why only the embed lane breaks. Fix: strip commas before matching (e.g. pipe through `tr -d ,` after strip_ansi, or use `[0-9][0-9,]*` in both the awk pattern and the grep guard, with `gsub(/,/,"")` before the arithmetic).

### MEDIUM — `docs/deploy/com.rico.qmd-sync.plist.template`:38

**Both launchd fallback log paths sit on the same /Volumes/ThunderBolt volume that the script's own startup guard is meant to catch as missing, so an unmounted-volume failure at 04:00 is silent rather than loud.**

The PR body claims "if the ThunderBolt volume is unmounted at 04:00 the job fails loudly rather than indexing nothing." It does not. With the volume unmounted, `mkdir -p "$LOG_DIR"` (qmd-sync.sh line 16) fails under `set -e` before the `exec >> "$LOG_FILE" 2>&1` redirect on line 18, so nothing reaches the primary log. The resulting stderr is routed by launchd to StandardErrorPath = /Volumes/ThunderBolt/open-brain/logs/qmd-sync.launchd.err.log -- on the very volume that is unmounted -- so launchd cannot open it either and the diagnostic is discarded. The `DEV_ROOT` and `QMD_BIN` existence guards in `main` (lines 128-139) never run, and their carefully-worded `reason=development-root-missing` / `reason=qmd-not-executable` lines are unreachable in exactly the scenario they were written for. Net effect: the operator's documented first diagnostic step (`tail` the primary log, then read the two launchd logs) yields nothing at all, and the only remaining evidence is the exit status in `launchctl print`. Confirmed the logs directory does not currently exist on this machine (`ls /Volumes/ThunderBolt/open-brain/logs/` -> No such file or directory), so first install depends entirely on the operator `mkdir -p` step. Consider pointing the launchd fallback logs at a boot-volume path (e.g. under $HOME) so a volume-unavailable failure is still recorded, or soften the PR/doc claim to match actual behavior.

### LOW — `scripts/qmd-sync.sh`:1

**The script is not exercised by any test and no scratch-index end-to-end run was performed, so the metric-parsing contract with qmd's human-readable output is unverified -- which is precisely how the HIGH finding above survived shellcheck, bash -n, and self-review.**

The PR body's self-review states this honestly ("the script is not exercised by `bun test`, so correctness rests only on shellcheck, `bash -n` and reading"), and that gap is exactly what let the comma-formatting defect through: every static check the PR reports passing (shellcheck clean, `bash -n` clean, `plutil -lint` OK) is incapable of detecting a mismatch between an awk regex and a downstream tool's runtime output format. Because the parsing scrapes another program's human-facing presentation layer, it is a contract that only an execution test can hold. A cheap, non-invasive test would pin the three parse shapes against captured qmd output fixtures (an `Indexed:` line, each of the three accepted embed terminal lines including a >=1000 comma-formatted one, and a `Total:`/`Vectors:` status block), asserting the extracted metrics -- proving the test fails on the current code before trusting it, per the repo's no-coverage-gates/functional-test standard. This does not require running a real sync.

Issue #270, opened 2026-07-07
Add operator doctor/status surface for Open Brain health


rodaddy:
## Problem

Open Brain has health and eval surfaces, but not a compact operator `doctor/status` surface that summarizes runtime, database, migrations, embeddings, qmd, contract, transport availability, and recent degraded states in one stable JSON response.

Sprint: #265.
Paired with: #269.

## Scope

- Add an operator-facing status/doctor command or tool surface.
- Prefer stable JSON first; human formatting can be a thin layer later.
- Include category scores, top issues, and remediation hints where safe.
- Keep secrets and raw env values out of output.

## Acceptance criteria

- Status covers at least: runtime version, contract version, DB connectivity, migration status, embedding provider availability, qmd availability when configured, transport mode/availability, recent degraded provider failures, and log/audit health if available.
- Doctor output has stable machine-readable fields for automation.
- Public health remains minimal; privileged doctor/status requires appropriate auth or local-only command boundary.
- Failure modes are bounded and do not make the server unhealthy just because an optional dependency is unavailable.

## Validation expectation

- Focused doctor/status tests.
- Auth/permission tests if exposed as an MCP/HTTP tool.
- `bunx tsc --noEmit`.

## Non-goals

- No full admin SPA.
- No job runner/minion system.
- No external monitoring replacement.

PR #576, opened 2026-08-05
fix(doctor): report qmd index freshness so a stale index alarms instead of passing silently (#388)

### LINKED ISSUE #388: QMD-3: Report qmd index freshness in operator-doctor so staleness alarms instead of passing silently
Parent: #342
Program: #320
Related: #386

## Problem

The qmd index went 39 days stale with the most active repository absent
entirely, and nothing surfaced it. A stale index still returns results — just
old ones — so the failure mode is silent by construction.

`src/operator-doctor.ts` already exists and already references qmd. It is the
natural place for this check.

## Scope

- Add a qmd freshness check to operator-doctor reporting: last index update
  timestamp, per-collection age, total collections, embedded vector count.
- Warn past a threshold (proposed: 48h since last update).
- Flag collections that exist on disk under the Development root but are not
  registered, so a missing repo is visible without a manual `comm` diff.

## Why this is load-bearing

qmd feeds `search-all`, `repo-facts`, and the context-pack repo-facts sections.
When the index is stale, those surfaces are confidently wrong rather than
empty — which is worse than being unavailable, because nothing signals doubt.

## Verification

- Doctor reports a stale index when one exists.
- Doctor reports a Development directory that is not a registered collection.
- Doctor passes clean after #386 and #402 land.

rodaddy:
Closes #388.

## What changed

`operator_doctor` reported qmd as `available` whenever the qmd entrypoint file
existed on disk — its own comment scoped it to "binary presence only, NOT qmd
search health". A stale index still answers queries, just with old content, so
the failure mode was silent by construction: the index once went 39 days stale
with the most active repo absent entirely and nothing surfaced it.

The `qmd` section of `OperatorDoctorStatus` gains an `index` sub-object:

| field | meaning |
|---|---|
| `status` | `available` / `unavailable` — could the index be read at all |
| `last_updated_at` | mtime of the repo-local `.qmd/index.sqlite`, ISO |
| `age_hours` | age at report time, 2dp |
| `freshness` | `current` / `stale` / `unknown` |
| `stale_after_hours` | the threshold in force (default 48) |
| `document_count` | active documents in the index |
| `collection_count` | registered collections |

`DOCTOR_CONTRACT_VERSION` bumps to `2026-08-05.operator-doctor.v3`, as the
repo's own lock test requires for any field addition.

### Rescoped against the current qmd architecture

The issue was written for the single shared 53-repo index, which was retired
2026-07-29/30. Per `docs/standards/QMD_INDEXES.md` the model is now one
project-local `.qmd/` per repo plus a `global_docs_instructions` index. Two
consequences, matching the maintainer's 2026-08-04 rescope comment:

- "last index update timestamp" is **per-repo**, read from the repo-local
  `.qmd/index.sqlite`, not one global number.
- The issue's "flag a Development directory that is not a registered
  collection" bullet has **no target** under the new model and is deliberately
  **not implemented**. Building it would have meant coding against an
  architecture that no longer exists.

### Fails safe, but visibly

A missing or unreadable index returns `status: "unavailable"` /
`freshness: "unknown"` and logs `doctor_qmd_index_unreadable`. It never throws:
qmd is an optional dependency, so index state never moves the overall health
tier (the #270 optional-dep rule). Reported, not silently passed.

## Commits

- `74833d4` — the freshness surface and its tests.
- `076fd58` — carrying the v3 bump to the two other suites that assert the
  contract literal independently, plus `.gitignore` for the test fixture dir.

## Testing

- `bun test` — 3187 pass, 506 skip, 0 fail across 231 files.
- **The staleness test was proven able to fail.** Reverting only
  `src/operator-doctor.ts` to its pre-change state and re-running the suite
  fails 4/18, including `reports a stale repo-local qmd index with counts`.
  A test that cannot fail proves nothing, so this was executed rather than
  assumed.

New test: `operator doctor status > reports a stale repo-local qmd index with
counts` — builds a sqlite fixture with an mtime 49h before an injected clock
and asserts `freshness: "stale"`, `age_hours: 49`, the document/collection
counts, and that overall status stays `healthy`.

## Critical Self-Review

- Highest-risk behavior: reading an arbitrary sqlite file at doctor time. `readQmdIndexStatus` opens the index `{ readonly: true }`, wraps it in try/finally with an explicit `database.close()`, and returns the neutral payload on every failure path, so a corrupt or locked index degrades the report and never the service.
- Assumptions that could be wrong: that `documents.active` and `store_collections` are the right tables to count, and that file mtime is a faithful proxy for "last indexed" — mtime moves on any write including a no-op re-index, so it can read fresher than the content truly is, erring toward under-alarming rather than false-alarming. A schema mismatch throws into the neutral branch rather than reporting a wrong number.
- Missing/weak tests: no test covers a corrupt (non-sqlite) file or an index whose schema lacks those tables, though both take the same catch path as the covered missing-file case; and no test asserts the `current` (fresh) verdict directly, since the two existing qmd tests cover `unknown` and the new one covers `stale`.
- Security/permission risk: low, and the existing constraint is preserved — the doctor payload must never disclose raw paths, so the index path is not emitted, only a timestamp, an age, counts, and a verdict. The pre-existing `expect(serialized).not.toContain(...)` secret and host assertions still pass.
- Migration/deploy risk: none — no schema change and no migration, purely additive to a JSON payload.
- Downstream client/runtime risk: this is a contract-version bump on a client-facing MCP payload, so it is the real risk here; the change is strictly additive with no field removed or retyped, so a client reading existing keys is unaffected, but a client pinning the literal `2026-07-08.operator-doctor.v2` needs the new value — which is exactly what the bump signals, and all three in-repo assertions of that literal were found with `rg` and updated together.
- Rollback/cleanup concern: revert both commits together, because reverting only `74833d4` would leave three suites asserting a v3 literal against v2 source; separately the test fixture writes under `src/../_scratch` (the repo bans `/tmp` and `$TMPDIR`), `afterEach` removes it, and `_scratch/` is now gitignored so a crashed run cannot leak it into a commit.
- Fixes made before PR: found via `rg` that two other suites hard-code the contract literal so `bun test` was failing 2/39 as handed over and now passes 3693/3693; corrected the stale version reference in the `server/tools/operator-doctor.ts` doc comment that points at the lock it would have misdescribed; and added the `_scratch/` gitignore entry.
- Known residual risk: the 48h threshold is a default constant, overridable per-call via `qmdIndexStaleAfterHours` but not by environment, so if 48h proves wrong operationally it is a one-line change rather than a redesign; and the doctor now reports staleness but does not alert on it, as routing that signal is out of scope here.
- SME review-memory update: [x] not applicable because: the reusable lesson is repo-specific and already encoded in `src/operator-doctor.test.ts`'s own comment — a `DOCTOR_CONTRACT_VERSION` bump must be `rg`-swept across every suite asserting the literal, not only the one beside the change.

## Review Gate

- Live Open Brain checks: [x] not applicable because: this change adds a field to the operator_doctor status payload and touches no Open Brain read, write, namespace, or embedding path, so there is no live surface to canary.
- Blast radius: one status payload plus its lock tests — no schema, no migration, no auth path, no namespace predicate.
- Contract parity: the `server/` tree deliberately delegates to the `src/` builder rather than rebuilding the payload, so there is no second implementation to keep in step; only its doc comment named the version, and that is updated.
- What a reviewer should focus on: whether `documents.active` and `store_collections` are the correct tables for the current qmd schema, and whether 48h is the right default threshold.
- Not merged, issue not closed. No auto-merge.

## Contract Parity

- The operator-doctor payload gains `qmd.index` under a DISTINCT contract version `2026-08-05.operator-doctor.v4` at all 7 version sites (rg-verified, zero v3 stragglers) — resolving the review swarm's cross-PR HIGH where this PR and #574 both claimed a byte-identical v3 for different shapes. Rebased onto #574 (now merged), the three shape-lock hunks resolved keeping both `distillation_lag` and `qmd.index`. Parity suite executed against the real test DB after the shared in-flight-probe fix (95096a6): three consecutive runs 69 pass / 7 fail, byte-identical to the origin/main baseline — the 7 are pre-existing server-rewrite-scaffold failures, zero current-src failures.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,6 +42,8 @@ OPENBRAIN_LOCAL_CLONE=0
 OPENBRAIN_LOCAL_CLONE_ROOT=
 # Local clone mode requires an explicit empty QMD_PATH to disable the production default:
 # QMD_PATH=
+# Optional operator-doctor override for the qmd SQLite index location:
+# QMD_INDEX_PATH=
 DB_POOL_MAX=10
 OPEN_BRAIN_RUN_MIGRATIONS=1
 # Raw turns nearing this TTL trigger operator-doctor distillation lag alarms.
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,8 @@ secrets/*
 # the source, so it must never be committed — a checked-in stamp would make
 # every tree claim to be whichever one happened to be deployed when it landed.
 .deployed-revision
+
+# Test scratch fixtures (src/operator-doctor.test.ts qmd index fixture). The
+# repo temp-workspace rule bans /tmp and $TMPDIR, so fixtures land in-repo and
+# must never be committed; afterEach removes them, a crashed run does not.
+_scratch/
--- a/contracts/server/server-operator-diagnostics-admin.fixture.json
+++ b/contracts/server/server-operator-diagnostics-admin.fixture.json
@@ -18,7 +18,7 @@
       "expectation": {
         "is_error": false,
         "json": {
-          "contract_version": "2026-08-05.operator-doctor.v3",
+          "contract_version": "2026-08-05.operator-doctor.v4",
           "generated_at": "<iso-date>",
           "runtime": {
             "service": "open-brain",
--- a/docs/CONFIG_REFERENCE.md
+++ b/docs/CONFIG_REFERENCE.md
@@ -273,6 +273,7 @@ stemmers split multibyte accented characters and every non-English FTS assertion
 | `OPENBRAIN_MCP_AUDIT_ENABLED` | on | `src/audit-log.ts:138` | `"0"` disables MCP audit logging |
 | `OPENBRAIN_RECOVERY_WAL_PATH` | `null` | `src/tools/index.ts:88` | recovery WAL location |
 | `QMD_PATH` | `/opt/qmd/src/qmd.ts` | `src/qmd-path.ts:7` | must be set **empty** in local-clone mode |
+| `QMD_INDEX_PATH` | repo `.qmd/index.sqlite` | `src/operator-doctor.ts:30` | operator-doctor qmd freshness/count probe |
 | `OPENBRAIN_LOCAL_CLONE` | `0` | `.env.example:41` | |
 | `OPENBRAIN_LOCAL_CLONE_ROOT` | unset | `.env.example:42` | |
 
--- a/server/tools/operator-doctor.ts
+++ b/server/tools/operator-doctor.ts
@@ -5,7 +5,7 @@
  * rebuilt here, for the same reason `get_contract` reuses `buildContract`: the
  * shape is a FROZEN contract. `src/operator-doctor.test.ts` locks the exact key
  * set of every section against `DOCTOR_CONTRACT_VERSION`
- * (`2026-08-05.operator-doctor.v3`), so a second implementation in the rewrite
+ * (`2026-08-05.operator-doctor.v4`), so a second implementation in the rewrite
  * would be a second payload that the lock test does not police, free to drift
  * on the first field addition. Reusing the builder keeps one answer to "how is
  * this server doing" and keeps that answer under the existing lock.
--- a/src/operator-doctor.test.ts
+++ b/src/operator-doctor.test.ts
@@ -1,6 +1,6 @@
-import { afterEach, describe, expect, it } from "bun:test";
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
 import { Database } from "bun:sqlite";
-import { mkdir, readdir, rm, utimes } from "node:fs/promises";
+import { mkdir, readdir, rm, utimes, writeFile } from "node:fs/promises";
 import { join, dirname } from "node:path";
 import { fileURLToPath } from "node:url";
 import {
@@ -137,11 +137,19 @@ function makeDownPool() {
   } as any;
 }
 
+beforeEach(() => {
+  process.env.QMD_INDEX_PATH = join(
+    SCRATCH_DIR,
+    "missing-default-index.sqlite",
+  );
+});
+
 afterEach(async () => {
   (globalThis as Record<string, unknown>).fetch = originalFetch;
   delete process.env.EMBEDDING_BASE_URL;
   delete process.env.EMBEDDING_API_KEY;
   delete process.env.QMD_PATH;
+  delete process.env.QMD_INDEX_PATH;
   delete process.env.LOG_FILE;
   delete process.env.LOG_MAX_BYTES;
   delete process.env.LOG_MAX_FILES;
@@ -191,6 +199,7 @@ describe("operator doctor status", () => {
     process.env.LOG_FILE = logPath;
     process.env.LOG_MAX_BYTES = "1000";
     process.env.QMD_PATH = THIS_FILE;
+    const reportedIndexPath = join(SCRATCH_DIR, "missing-index.sqlite");
 
     (globalThis as Record<string, unknown>).fetch = (
       input: string | URL | globalThis.Request,
@@ -208,11 +217,11 @@ describe("operator doctor status", () => {
       makePool(["001_init.sql"]),
       readNatsRuntimeBoundary({}),
       undefined,
-      { qmdIndexPath: THIS_FILE },
+      { qmdIndexPath: reportedIndexPath },
     );
     const serialized = JSON.stringify(status);
 
-    expect(status.contract_version).toBe("2026-08-05.operator-doctor.v3");
+    expect(status.contract_version).toBe("2026-08-05.operator-doctor.v4");
     expect(status.runtime.contract_version).toBe("2026-07-23.memory-tools.v23");
     expect(status.database.connected).toBe(true);
     expect(status.embedding_provider).toMatchObject({
@@ -232,6 +241,8 @@ describe("operator doctor status", () => {
       available: true,
       status: "available",
       index: {
+        path: reportedIndexPath,
+        path_source: "option",
         status: "unavailable",
         last_updated_at: null,
         age_hours: null,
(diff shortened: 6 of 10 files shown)


rodaddy:

### HIGH: `src/operator-doctor.ts`:334

**PR #576 and PR #574 both set DOCTOR_CONTRACT_VERSION to the byte-identical literal "2026-08-05.operator-doctor.v3" for two different payload shapes, so whichever merges second ships a shape change under a version a client has already seen.**

Failure scenario: Verified via `git show origin/fix/381-distill-lag-alarm:src/operator-doctor.ts` line 32 and `origin/fix/388-doctor-qmd-freshness:src/operator-doctor.ts` line 334 - both read "2026-08-05.operator-doctor.v3". #574 adds a top-level `distillation_lag` array; #576 adds a nested `qmd.index` object. If #574 merges first, a client pins v3 against the distillation_lag-only shape; #576 then merges, the payload gains qmd.index, and the version literal is unchanged - exactly the drift the bump exists to signal. The second PR to merge must go to v4.

### HIGH: `src/operator-doctor.test.ts`:425

**PRs #574 and #576 conflict textually in three hunks, all in the shape-lock region; neither can merge cleanly after the other without hand resolution, and each shows MERGEABLE against main independently so nothing surfaces it until the second merge.**

Failure scenario: `git merge-tree $(git merge-base origin/fix/381-distill-lag-alarm origin/fix/388-doctor-qmd-freshness) origin/fix/381-distill-lag-alarm origin/fix/388-doctor-qmd-freshness` emits 3 conflict blocks. Both PRs edit the DOCTOR_CONTRACT_VERSION line, both extend the Object.keys shape-lock assertions in the same test, and both add adjacent interface fields. The shape-lock test - the thing designed to catch shape drift - is itself what conflicts, so a careless resolution disarms the guard.

### MEDIUM: `src/operator-doctor.ts`:328

**Freshness is derived from the mtime of index.sqlite, but the real qmd index runs in WAL mode where writes land in index.sqlite-wal and the main file's mtime moves only on checkpoint, so freshness reads wrong in both directions.**

Failure scenario: Verified live: `sqlite3 .qmd/index.sqlite "pragma journal_mode"` returns `wal`, and .qmd/ holds a 4.1 MB index.sqlite-wal beside the 63 MB index.sqlite. Under WAL an `aqmd up` that appends documents advances the -wal mtime while index.sqlite stays at the last checkpoint, so a repo indexed hourly but never checkpointed reports freshness "stale" after 48h and alarms falsely; conversely an automatic checkpoint touches index.sqlite without new content and reports "current" for a stale index. Reading max(mtime) across both files, or a max-timestamp column from documents, would be faithful. The PR body discloses the mtime-proxy assumption but claims it errs only toward under-alarming - WAL makes it err both ways.

### MEDIUM: `src/operator-doctor.ts`:326

**readQmdIndexStatus runs statSync plus two synchronous COUNT(*) queries on the doctor request path with no timeout, while every sibling optional probe in the same builder is async and bounded by withTimeout(..., OPTIONAL_TIMEOUT_MS).**

Failure scenario: checkEmbeddingAvailability wraps its probe in `withTimeout(probeUrl(...), false)` with OPTIONAL_TIMEOUT_MS = 2000 precisely so an optional dependency can never stall the report. readQmdIndexStatus is fully synchronous and untimed, blocking the event loop. `COUNT(*) FROM documents WHERE active = 1` is a full scan of an unindexed predicate - fast on the live 63 MB / 1111-document index, but unbounded in size and sitting on an external ThunderBolt volume, so a slow or contended disk stalls every /api/v1/operator/doctor request. DOCTOR_CACHE_TTL_MS (5s) limits frequency but not the duration of any single stall.

### MEDIUM: `src/operator-doctor.ts`:331

**QMD_INDEX_PATH is hard-coded source-relative with no env override and is never reported in the payload, so a deployment whose tree lacks a sibling .qmd/ reports freshness unknown forever - the silent pass this PR set out to remove.**

Failure scenario: .qmd/ is a developer-local index directory, and per AGENTS.md the running app lives at /Volumes/ThunderBolt/open-brain/app, which is not mounted in this session - I could not confirm its contents either way, so this is PLAUSIBLE not CONFIRMED. If the deployed tree lacks .qmd/, every core01 doctor call takes the catch branch, logs doctor_qmd_index_unreadable on each uncached build, and shows freshness "unknown" permanently. Unlike QMD_PATH - which has an env override and a `path_source` field so the operator can see which path was used - the index path is neither overridable nor reported, so an operator cannot distinguish "no index here" from "index unreadable".

### LOW: `src/operator-doctor.test.ts`:247

**No test asserts freshness "current", and none covers a corrupt file or a schema missing the expected tables, so the boundary comparison ageHours >= staleAfterHours is only ever exercised on its true branch.**

Failure scenario: Searching the branch test file for `freshness: "current"` or `toBe("current")` returns only an unrelated migrations assertion. Coverage is "unknown" (two missing-index tests) and "stale" (the new 49h test). If the comparison were inverted or the threshold units wrong, the stale test would still pass while every healthy index silently reported "stale" - the false-alarm direction that erodes trust in the signal this PR adds. One assertion with a 1h-old fixture closes it. The PR body discloses this gap honestly, which is why it is LOW.

### LOW: `src/operator-doctor.test.ts`:276

**The "falls back to the shared default qmd path" test passes no options, so readQmdIndexStatus opens the developer's real 63 MB .qmd/index.sqlite, making an unrelated test depend on machine-local state.**

Failure scenario: The test passes only pool and natsRuntimeBoundary, so `options.qmdIndexPath ?? QMD_INDEX_PATH` falls through to the real repo index - confirmed present at 63 MB with a 4.1 MB WAL. On a developer machine it reads that file; in CI, where no .qmd/ exists, it takes the catch branch and emits a doctor_qmd_index_unreadable warning. It asserts nothing about status.qmd.index so it will not fail either way, but it makes the test environment-dependent and noisy. Passing an explicit qmdIndexPath, as the two neighbouring tests already do, removes the coupling.

PR #577, opened 2026-08-05
fix(maintenance): add the missing recurring sweep producer (#384)

### LINKED ISSUE #384: MAINT-6: Enable the maintenance runner and add the missing sweep producer
Parent: #342
Program: #320
Spec: `docs/full-send-derivation-spec.md`

## Problem

The maintenance machinery is built and switched off.

- `maintenance-bootstrap.ts` wires `MaintenanceQueue` + `MaintenanceQueueRunner`
  into `src/index.ts` with two registered handlers (`embedding.repair`,
  `graph.derive`).
- `OPEN_BRAIN_MAINTENANCE_ENABLED=0` on the local clone; the variable is
  **unset on core01**.
- Exactly one job has ever been enqueued on either box: a canary
  `graph.derive`, empty `{}` payload, `dead_letter` after one attempt,
  category `terminal`, `idempotency_key = canary:graph.derive:40804b7`.

Both handlers are consumers with no producer. `graph-derivation-handler.ts`
states it outright: "the bootstrap enqueues nothing and defines no recurring
sweep, so there is no automatic continuous derivation."

That is the difference between a queue that exists and a server that works.

## Scope

1. Fix or retire the dead-lettered canary job.
2. Enable the runner on the local clone; prove the loop turns and observes.
3. Add a bounded recurring sweep in the maintenance bootstrap that, per tick:
   - selects undistilled turns via the partial index, grouped by lane, capped
     by configured batch size and max batches per tick;
   - enqueues one `memory.distill` job per batch;
   - enqueues `graph.derive` for sources whose observed `content_hash`
     drifted — the sweep `graph.derive` was always waiting for.

## Acceptance criteria

- [ ] Runner enabled and observed processing jobs on the local clone.
- [ ] Sweep enqueues work without operator action.
- [ ] Caps are enforced and logged; no silent truncation.
- [ ] Enabled on core01 only after the local loop is measured and bounded.

## Sequencing

Step 1-2 need nothing else and can land first. Step 3's `memory.distill` arm
depends on #382.

rodaddy:
Closes #384 (MAINT-6).

## Summary

The maintenance queue had handlers and no producer. `maintenance-bootstrap.ts`
registered `graph.derive`, `memory.distill`, `dream.light`, and `dream.rem`, but
its own module docstring stated the defect outright: "the bootstrap enqueues
nothing and defines no recurring sweep, so there is no automatic continuous
derivation." Exactly one job had ever been enqueued on either box — a canary
`graph.derive` that dead-lettered. A queue that exists is not a server that
works.

This adds the missing recurring sweep producer (`src/maintenance-sweep.ts`) and
wires it into the bootstrap alongside the runner, so the loop turns without an
operator call.

Per tick the sweep:

- selects undistilled turns grouped by lane (including lane-less turns) and
  enqueues one `memory.distill` job per selected batch;
- reuses the existing `enqueueGraphDerivationJobs` producer for sources whose
  observed `content_hash` drifted.

Deferred work is **logged, never silently dropped** — the issue's explicit
acceptance criterion. `maintenance_sweep_distill_cap_hit` and
`maintenance_sweep_graph_cap_reached` name the deferred batch and turn counts;
`maintenance_sweep_complete` reports the per-tick totals. All bounding
parameters are env-configurable in the same `envPositiveInt` style already used
in the bootstrap. Producer ticks are non-overlapping, and production stops
before the runner drains on shutdown.

`distill-handler.ts` / `distill-window.ts` carry the producer-selected lane
through the durable job payload so distillation binds to the selected lane or
the lane-less group.

No new maintenance framework; `MaintenanceQueue` is untouched.

## Verification

Run by the reviewing wrapper, not merely reported by the implementer:

- `bunx tsc --noEmit` — exit 0.
- `bun test src/maintenance-sweep.test.ts src/maintenance-bootstrap.test.ts src/distill-handler.test.ts src/distill-window.test.ts` — 66 pass, 0 fail, 182 assertions.
- Pre-push hook full suite — 3696 tests across 232 files, 0 fail.

**Red-test proof, independently reproduced.** The point of this change is to
kill a silent-no-op, so the test has to fail when the producer is absent. I
reverted the sweep startup to the old behavior (`const sweep = false`) and ran
the test:

```
error: maintenance bootstrap produced no jobs
(fail) startMaintenanceQueue recurring producer > enqueues sweep work automatically without an operator call
1 pass, 1 fail
```

Restored, re-ran: 2 pass, 0 fail. The test genuinely detects the defect it
claims to cover.

## Critical Self-Review

- Highest-risk behavior: the sweep now runs automatically on every boot where maintenance is enabled, so a large undistilled backlog becomes real enqueued work immediately instead of staying dormant — the first time this loop has ever turned on its own.
- Assumptions that could be wrong: that `distilled_at IS NULL AND retention_tier = 'live'` is the right definition of undistilled work, that grouping by `(lane_id, namespace)` matches the intended distillation partitioning, and that the durable queue's idempotency keys make repeated ticks over an unchanged backlog converge instead of duplicating.
- Missing/weak tests: the sweep tests drive a fake pool/queue port, so the SQL in `selectDistillLaneBatches` never executes against live Postgres here, and the end-to-end "sweep enqueues, runner claims, handler distills against a real database" path is unproven.
- Security/permission risk: low but non-zero — the sweep enqueues under the existing `MAINTENANCE_GRAPH_AUTH` identity and invents no namespace; each payload carries the namespace read from the source row, and handlers still re-check `canWriteNamespace` and re-validate against the live source before deriving.
- Migration/deploy risk: no schema migration; the risk is operational, since enabling this on core01 turns a dormant queue into an active one, and this PR deliberately does not flip core01.
- Downstream client/runtime risk: none identified — no MCP tool, schema, transport, or Python-client surface changed, so `docs/downstream-rollout.md` is not-applicable; the distill payload's new lane field is internal to the server-owned queue.
- Rollback/cleanup concern: `OPEN_BRAIN_MAINTENANCE_ENABLED=0` disables the whole loop including this producer, so rollback is an env flip plus restart with no data migration to undo, though already-enqueued jobs and the pre-existing dead-lettered canary (scope item 1, not addressed here) need operator disposition.
- Fixes made before PR: reproduced the red-test mutation myself rather than trusting the implementer's report, and confirmed the worktree was clean with the mutation fully reverted before pushing.
- Known residual risk: the live-Postgres path for the new sweep SQL is unproven and acceptance criteria 1 and 4 (observed processing locally, core01 enablement) remain operator-verified steps — this is merged-not-running work.
- SME review-memory update: [ ] `docs/sme/` updated [x] not applicable because: this change added a missing producer and surfaced no new review-miss pattern; promote one here if review finds a MEDIUM+ finding.

## Review Gate

- Live Open Brain checks: [ ] linked below [x] not applicable because: acceptance criteria 1 and 4 are operator observations against the local clone and core01, and this PR enables neither — no live Open Brain surface was exercised.

- Scope is the named gap only: a producer plus its wiring. No new maintenance
  framework, no `MaintenanceQueue` refactor.
- Reviewers should focus on the sweep SQL in `selectDistillLaneBatches`
  (grouping and the undistilled predicate), the lane threading through
  `distill-window.ts`, and whether deferred-work logging is sufficient to rule
  out silent truncation.
- Do not merge on the strength of green CI alone: acceptance criteria 1 and 4
  are operator observations against the local clone and core01, and neither is
  claimed complete here.

### DIFF
--- a/src/distill-handler.ts
+++ b/src/distill-handler.ts
@@ -101,8 +101,10 @@ export interface DistillSweepDeps {
   model?: NamedDistillModel;
   /** Injectable embed fn; defaults to the configured provider. */
   embedFn?: DistillEmbedFn;
-  /** Restrict the sweep to one namespace. Omit to sweep every namespace. */
+  /** Bind the sweep to one namespace. Omit to sweep every namespace. */
   namespace?: string;
+  /** Bind a queued batch to one owning lane. */
+  laneId?: string | null;
   maxSessions?: number;
   maxTurns?: number;
   contextWindow?: number;
@@ -259,6 +261,7 @@ export async function runDistillSweep(
 
   const batch = await claimDistillBatch(deps.pool, {
     ...(deps.namespace !== undefined ? { namespace: deps.namespace } : {}),
+    ...(deps.laneId !== undefined ? { laneId: deps.laneId } : {}),
     ...(deps.maxSessions !== undefined
       ? { maxSessions: deps.maxSessions }
       : {}),
@@ -361,6 +364,10 @@ export function makeMemoryDistillHandler(
     // default for a recurring maintenance job. Values are read defensively
     // because a job row is durable and may predate a code change.
     const payload = job.payload ?? {};
+    const laneId =
+      payload.lane_id === null || typeof payload.lane_id === "string"
+        ? payload.lane_id
+        : undefined;
     const maxSessions =
       typeof payload.max_sessions === "number"
         ? payload.max_sessions
@@ -377,6 +384,7 @@ export function makeMemoryDistillHandler(
       // one was scoped. A null namespace is a deliberate global sweep, which is
       // what a maintenance identity is for -- not an error.
       ...(job.namespace !== null ? { namespace: job.namespace } : {}),
+      ...(laneId !== undefined ? { laneId } : {}),
       ...(maxSessions !== undefined ? { maxSessions } : {}),
       ...(maxTurns !== undefined ? { maxTurns } : {}),
       distillJobId: job.id,
@@ -398,10 +406,12 @@ export function makeMemoryDistillHandler(
 export function buildMemoryDistillEnqueue(input: {
   sweepLabel: string;
   namespace?: string;
+  laneId?: string | null;
   maxSessions?: number;
   maxTurns?: number;
 }) {
   const payload: Record<string, unknown> = {};
+  if (input.laneId !== undefined) payload.lane_id = input.laneId;
   if (input.maxSessions !== undefined) payload.max_sessions = input.maxSessions;
   if (input.maxTurns !== undefined) payload.max_turns = input.maxTurns;
 
--- a/src/distill-window.test.ts
+++ b/src/distill-window.test.ts
@@ -269,6 +269,25 @@ describe("claimDistillBatch — ordering key and stamping", () => {
     expect(db.seen[0]!.values[0]).toBe("rico");
   });
 
+  it("binds both due selection and context reads to the producer's lane", async () => {
+    const db = fakeDb([]);
+    const laneId = "77777777-7777-4777-8777-777777777777";
+    await claimDistillBatch(db, { namespace: "rico", laneId });
+    const sql = db.seen[0]!.text;
+    expect(sql).toContain("AND lane_id = $2::uuid");
+    expect(sql).toContain("AND t.lane_id = $2::uuid");
+    expect(db.seen[0]!.values.slice(0, 2)).toEqual(["rico", laneId]);
+  });
+
+  it("keeps lane-less turns in their own producer batch", async () => {
+    const db = fakeDb([]);
+    await claimDistillBatch(db, { namespace: "rico", laneId: null });
+    const sql = db.seen[0]!.text;
+    expect(sql).toContain("AND lane_id IS NULL");
+    expect(sql).toContain("AND t.lane_id IS NULL");
+    expect(db.seen[0]!.values[0]).toBe("rico");
+  });
+
   it("bounds both the session and turn limits so one sweep cannot read the table", async () => {
     const db = fakeDb([]);
     await claimDistillBatch(db, { maxSessions: 10_000, maxTurns: 10_000_000 });
--- a/src/distill-window.ts
+++ b/src/distill-window.ts
@@ -76,6 +76,8 @@ export interface DistillUnit {
  * justify a different number yet -- so it is adopted, not invented.
  */
 export const DEFAULT_CONTEXT_WINDOW = 3;
+/** Default whole-session bound shared by direct and scheduled distill sweeps. */
+export const DEFAULT_MAX_DISTILL_SESSIONS = 4;
 
 /**
  * Is this turn SPEECH -- something a person or an agent actually said, and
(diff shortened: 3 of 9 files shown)


rodaddy:

Lane: claude-verification (correctness/quality, PR #577)
Verdict: findings-must-fix

### HIGH — `src/maintenance-sweep.ts`:720

**The distill idempotency key is derived from the oldest undistilled turn id (batch_anchor), but the anchor turn is not guaranteed to be consumed by the job it keys; since maintenance_jobs rows are never deleted and enqueue uses ON CONFLICT DO NOTHING against UNIQUE(job_kind, idempotency_key), the key is permanently consumed and the lane's remaining backlog is never enqueued again.**

Failure scenario: A lane holds 10 undistilled sessions. Tick 1 enqueues key `memory.distill:<lane>:<anchorA>`. The producer passes only maxTurns and never maxSessions, so claimDistillBatch falls back to `options.maxSessions ?? 4` (src/distill-window.ts:212) and drains only 4 sessions. Additionally the anchor is chosen by `created_at` while claimDistillBatch selects sessions by `min(occurred_at)` and orders turns by DISTILL_ORDER_BY (`session_ref, occurred_at, id`, src/distill-window.ts:177); occurred_at is nullable and 032_raw_turns.sql:150-162 documents that it diverges from created_at on replay/backfill. So anchorA is not stamped distilled_at. Every later tick recomputes the identical anchor, hits ON CONFLICT DO NOTHING (src/maintenance-queue.ts:370), and returns the existing succeeded job. Six sessions are never distilled, no error is logged, and maintenance_sweep_complete keeps reporting distill_jobs_enqueued>0 — the silent no-op this PR exists to remove, reintroduced one layer up. A dead-lettered job produces the same permanent stall. Verified no production code path deletes from maintenance_jobs (only test files do).

### MEDIUM — `src/maintenance-sweep.test.ts`:1

**The PR's central correctness assumption — that repeated ticks over an unchanged backlog converge instead of duplicating — is asserted in the docstring and the self-review but exercised by no test; the fake queue port cannot model the real ON CONFLICT contract, so it has already drifted from what it stands in for.**

Failure scenario: Neither test in maintenance-sweep.test.ts runs a second sweep tick. The fake `queue.enqueue` unconditionally pushes to `enqueued` and returns a freshly-built MaintenanceJob, so it never models ON CONFLICT DO NOTHING or the divergent-semantics rejection at src/maintenance-queue.ts:403-407; bootstrapPool's fake INSERT likewise always returns a row. The HIGH stall above therefore passes the whole suite green. The red-test proof in the PR body is genuine but only proves the producer starts, not that it converges. Matches docs/sme/correctness.md '[2026-06-19] Mock-pool tests cannot catch SQL constraint failures' — this queue write path needs an OPENBRAIN_TEST_DATABASE_URL-gated real-pool test that ticks twice and asserts the backlog advances.

### MEDIUM — `src/maintenance-sweep.ts`:722

**distillTurnsDeferredByBatchCap measures deferral against the turn cap, but the handler's 4-session default is usually the binding constraint, so the cap-hit warning can report zero deferred work while a lane is permanently stalled.**

Failure scenario: enqueueDistillBatches computes `pendingTurns - batchSize` (default batchSize 1500). A lane with 10 sessions and 200 pending turns yields deferredTurns = 0 and deferredBatches = 0, so maintenance_sweep_distill_cap_hit never fires — yet the handler distills only 4 sessions and the remaining 6 are silently deferred forever per the HIGH finding. The issue's explicit acceptance criterion is that deferred work is logged and never silently dropped; this metric cannot observe the dimension along which work is actually being dropped.

### LOW — `src/maintenance-sweep.ts`:751

**graphLimitReached uses an equality test against the limit, so it warns whenever the drifted-source count exactly equals the cap even when nothing was deferred.**

Failure scenario: Exactly 100 sources have drifted content hashes with the default DEFAULT_GRAPH_DERIVATION_LIMIT of 100. `jobs.length === graphLimit` is true, so maintenance_sweep_graph_cap_reached warns and the summary reports graph_limit_reached: 1, implying deferred work that does not exist. Log-fidelity only; no behavioral impact.

Issue #578, opened 2026-08-05
E2E quality gate: scripted input/output scenarios + Langfuse egress verification


rodaddy:
# E2E quality gate: scripted input→expected-output runs + Langfuse egress verification

Status: PROPOSED (operator-requested 2026-08-05: "run a whole bunch of actual
tests to verify things instead of me having to literally dogfood everything...
run some inputs and outputs via a script with expected information and also
be able to check what's going out to my Langfuse and see that things are
actually going correctly.")

## What already exists (do not rebuild)

- `eval/open-brain/` — fixture-driven eval harness (#109, closed): runner,
  scorecard, `eval/open-brain/fixtures/*.json`.
- `scripts/eval-open-brain-live.ts` (EVAL-3, #324): live recall gate — seeds a
  unique throwaway namespace, drives the REAL client, scores deterministic
  ranking metrics vs versioned thresholds, tears down its own records,
  exits non-zero on failure, content-free output.
- `scripts/codex-memory-smoke.ts` — smoke path.
- Contract parity suite (`contracts/`) — shape-locks tool payloads.
- Ethereal-run protocol (`docs/dream-ethereal-runs.md`) — disposable-output
  testing for dream stages.

## The delta (two pieces)

1. **Scenario breadth on the existing harness.** New fixtures + runner support
   for the surfaces dogfooding currently covers by hand: capture (hook +
   operator paths — #529's proofs become fixtures), durable_memory pack
   shape/recall correctness, checkpoint/wrap round-trips, session lifecycle.
   Same fixture JSON pattern, same throwaway-namespace + teardown discipline
   as EVAL-3. Each scenario: scripted input, expected output, deterministic
   scoring, non-zero exit on miss.

2. **Langfuse egress verifier (net-new).** After a scenario batch drives real
   traffic with a unique run tag, query the Langfuse instance on CT 273 and
   assert what ACTUALLY arrived: expected trace/observation count for the tag,
   GENERATION rows carry provided_model_name + usage_details + non-NULL cost
   (this IS #560's live receipt, automated), no secret-shaped content in
   trace bodies (automates #561's census as a regression gate), and nothing
   expected went missing (the silent-drop class). Query route: Langfuse public
   API with env-gated keys, fallback documented ClickHouse query. Env-gated
   like the live eval (opt-in, content-free output).

## Why this shape

Reuses the expensive existing designs (fixture runner, live-gate discipline,
parity locks) and adds only the missing egress leg. The verifier closes the
[...]

Issue #571, opened 2026-08-05
[Epic] OBS: honest Langfuse observability — cost, masking, coverage


rodaddy:
Groups the Langfuse observability lane surfaced by the 2026-08-05 issue-state audit. No epic covered it; every member predates this epic and is unchanged by it.

## Members (close-run order)
- [ ] #560 emitter never sets model/usage/cost — PR #568 open, red CI
- [ ] #561 secret audit done (34,462 obs censused, 0 live secrets); **masking half unbuilt** — emitter-side pre-enqueue scrub remains
- [ ] #569 trace-completeness decisions — explicitly sequenced behind #561 masking and #560 metadata
- [ ] #492 the codify-agent-telemetry-push half only (the Hermes bulk adapter half stays in its own lane)

## Done means
Cost dashboards price real numbers on CT 273 (RUNNING, not merged), masking lands before tracing widens, and #569's three decisions are made and recorded.

rodaddy:
#578 (E2E quality gate) added to this epic's close path: its Langfuse egress verifier is the mechanism that produces #560's live cost receipt and turns #561's secret census into a permanent regression gate. Building it is the first step of the OBS run.

rodaddy:
Coverage ruling recorded on #569 (2026-08-05): full coverage, no exceptions, masking-first sequencing. Epic members now actionable in order: #587 (release stamp), #561 (masking), #569 (widen to full coverage), #492 telemetry-charter half.

Issue #561, opened 2026-08-05
Audit content-ful Langfuse traces on CT 273 for secrets before tracing widens (PR #534 sends full tool inputs/outputs)


rodaddy:
Found during a Langfuse survey (2026-08-04). The older `_ob` ob-memory-provider observation path is deliberately content-FREE (HMAC'd IDs only). The server-side tracing shipped in PR #534 is content-FUL: full tool-call inputs and outputs land in ClickHouse on CT 273 (10.71.20.73).

That is a second full copy of agent tool traffic. Known trap (independently reported by practitioners): apps that redact before the LLM provider still leak the unredacted data to the observability sink.

**Not audited yet** -- the survey was read-only and stopped at the observation. Before tracing is enabled more widely, someone should sample the `input`/`output` observation columns for credentials, tokens, and Vaultwarden output, and decide whether the emitter needs masking. No secret values in this issue; findings should be counts/classes only.

rodaddy:
## Read-only secret audit of Langfuse trace content on CT 273 — COMPLETE

**Verdict: no live credentials found. Safe to widen tracing, with one caveat below.**

### Methodology

Read-only throughout: no writes, no config changes, no Langfuse/ClickHouse mutation, ssh `ConnectTimeout` 10-25s. **No payload ever left CT 273 and no payload was written to disk.** Every scan ran as `countIf`/aggregate regex *inside* ClickHouse, returning only counts, hashes, and character-class skeletons. No candidate secret value is quoted here or was rendered in the audit session.

- Host: `root@10.71.20.73`, container `langfuse-clickhouse-1` (`clickhouse/clickhouse-server`, v26.4.2)
- Table: `default.observations`, project `cmryn82hr0003qm070oymart5` (Open Brain Local Dogfood)
- Scope: **full population, not a sample** — all **34,462** observations, `start_time` `2026-08-03 17:02:07` → `2026-08-05 04:07:53`. Scanning server-side made full coverage cheaper than the 200-500 row sample the ticket asked for, so the sample was upgraded to a census.
- Scanned field: `concat(input, ' ', output)` per row

### Content-ful confirmed (PR #534 behavior reproduced)

| metric | value |
|---|---|
| rows with non-empty `input` | 4,234 |
| rows with non-empty `output` | 31,109 |
| total `input` bytes | 19,412,057 |
| total `output` bytes | 18,081,773 |
| largest single `output` | 236,920 bytes |

[...]

Issue #560, opened 2026-08-05
Langfuse emitter never sets model/usage/cost -- 23,439/23,440 live generations have NULL cost, spend ledger reads $0.72 total


rodaddy:
Found during a Langfuse-potential survey (2026-08-04, read-only, ClickHouse on CT 273 queried directly).

**Measured (RUNNING-verified):** Of the Open Brain Local Dogfood project's 23,440 GENERATION observations, 23,439 have `provided_model_name = NULL`, zero `usage_details`, `total_cost = NULL`. Exactly one row carries `claude-opus-5`. Total recorded spend across the entire Langfuse instance is $0.72 -- all of it dead MAE test data.

**Root cause (read in source):** `server/observability/langfuse-tracing.ts` (1,035 lines) never sets model, usage, or cost fields -- `rg 'usageDetails|costDetails|providedModelName'` returns nothing. Same for `python/.../capture/langfuse_emitter.py`.

**Fix shape (PROPOSED, ~S):** set `model` + `usageDetails` (input/output/cache-read) in both emitters; seed custom model prices for claude-opus-5 / claude-sonnet-5 / claude-fable-5 (3 custom prices already exist on the instance, mechanism proven). While in there: pass git SHA as `release` (currently empty on 100% of traces) and always set `userId` (1,615/2,922).

Until this lands, every cost dashboard/score computes over NULL -- the instance cannot answer 'which agent burned $40 yesterday', which is the primary reason to run it.

rodaddy:
Reopening deliberately: the fix is MERGED (PR #568 at ff45179 — model+usage now carried, plus the swarm-driven hardening: malformed enrichments can no longer drop a transcript record, and partial/mistyped usage omits rather than fabricating zeros; fail-first tests reproduced independently). But this issue's acceptance is LIVE: non-NULL total_cost on CT 273's dashboards, which needs (1) the updated capture lane actually running and (2) custom model prices seeded for claude-opus-5 / claude-sonnet-5 / claude-fable-5 — an operator/instance step nobody has verified. Merged is not deployed. Closes for real in the OBS run (#571) with a live receipt.

rodaddy:
Closing with a live receipt (RUNNING-verified this session, 2026-08-05).

[...]

PR #579, opened 2026-08-05
feat(eval): add Langfuse egress verifier (#578)

rodaddy:
Adds `scripts/eval-langfuse-egress.ts`, the net-new Langfuse egress verifier from the #578 phase-1 delta (item 2). This is the leg that turns "merged, unverified live" into a command that produces a live receipt: after traffic is driven with a unique run tag, it queries Langfuse and asserts what actually arrived.

Scope is item 2 only. Scenario-fixture breadth (delta item 1) is a separate lane and is not touched here.

## What it does

- `--drive` emits N known generation turns through the REAL capture path, stamped with the run tag.
- `--verify` queries the Langfuse public trace API for that tag (as trace `sessionId`) and asserts arrival.
- `--tag <s>` (fresh UUID default, secret-shaped tags rejected), `--count <n>`, `--json`.

Assertions, as four labelled checks:

| check | fatal |
|---|---|
| `arrival_count` — exact trace/observation count for the tag (the silent-drop class) | yes |
| `generation_metadata` — GENERATION rows carry `provided_model_name` + `usage_details` | yes |
| `total_cost` — non-NULL cost | **no by default**, fatal with `--require-cost` |
| `secret_scan` — secret-shaped content in returned trace bodies | yes |

`total_cost` is deliberately reported-not-fatal: cost legitimately stays NULL until model prices are seeded. `--require-cost` promotes it, which is #560's live receipt on demand.

## Reuse, not reinvention

- Secret scanning imports `SECRET_DETECTORS` from `src/secret-patterns.ts` — the regex set already in the repo — rather than re-embedding the #561 audit patterns. Only detector `kind` and a count can reach output; matched values, offsets, and bodies never can.
- Env coordinates are the existing `OPENBRAIN_TRACING_ENDPOINT` / `_PUBLIC_KEY` / `_SECRET_KEY` / `_ENABLED`, matching `readMcpTracingConfig` in `server/observability/langfuse-tracing.ts`. For `--drive` these are translated to the capture sink's existing `OPENBRAIN_OBSERVATION_*` variables (`python/openbrain/src/openbrain/config.py`), because the emitter reads that older spelling.
- Endpoint normalisation accepts a bare host or the `/api/public/ingestion` URL, mirroring `sink_host()` in `langfuse_emitter.py`.
- Opt-in guard (`OPEN_BRAIN_LANGFUSE_EGRESS=1`), content-free output, and nonzero-exit discipline follow `scripts/eval-open-brain-live.ts`.

## Verification

Re-run by the supervising node in the worktree, not merely reported by the implementer:

- `bunx tsc --noEmit` — exit 0, no output.
- `bun test scripts/eval-langfuse-egress.test.ts` — 8 pass, 0 fail, 19 expect() calls.

Red-test proof: the arrival assertion was deliberately weakened from `===` to `<=`, and `arrival-count miss is fatal` then failed (`expect(received).toBeFalse()`, Received: true), 6 pass / 1 fail. The mutation was reverted before commit. Tests cover: arrival-count miss fatal, missing `usage_details` fatal, cost-NULL non-fatal by default, cost-NULL fatal under `--require-cost`, secret-pattern hit fatal, missing-opt-in refusal, and an output test asserting no trace-body marker or credential appears in the serialized receipt.

Not verified: a live `--drive`/`--verify` run against the real Langfuse service on CT 273. No live opt-in invocation was authorized in this lane, so authentication, live public-API response shape, cost seeding, and true end-to-end arrival remain **unverified**. This is written and tested, not live-proven.

## Critical Self-Review

- Highest-risk behavior: the live Langfuse public-API response shape is assumed from the documented API and has never been exercised against the real instance, so field-name drift would surface only on the first live run.
- Assumptions that could be wrong: that the run tag propagates as the trace `sessionId`, that `--drive`'s `OPENBRAIN_TRACING_*`-to-`OPENBRAIN_OBSERVATION_*` mapping is the correct bridge, and that camelCase/snake_case field acceptance covers what the API actually returns.
- Missing/weak tests: no live integration test, and pagination beyond a single page is implemented but only exercised through the fake transport.
- Security/permission risk: the tool reads trace bodies that may contain secrets, so the containment control is that only detector `kind` labels and counts reach stdout; a future edit that logs a body or a match offset would break that boundary and no test outside the output-marker test guards every path.
- Migration/deploy risk: none, additive script with no schema, migration, or runtime wiring.
- Downstream client/runtime risk: none, no MCP tool, transport, contract, or Python client surface changes, so no downstream rollout applies.
- Rollback/cleanup concern: revert the single commit; the script is opt-in and inert without `OPEN_BRAIN_LANGFUSE_EGRESS=1`.
- Fixes made before PR: corrected the env names from the assumed `LANGFUSE_*` to the repo's real `OPENBRAIN_TRACING_*`, and replaced hand-embedded secret regexes with an import of the existing `SECRET_DETECTORS`.
- Known residual risk: the verifier is merged-and-tested but not live-proven, so its first real run against CT 273 should be treated as the actual acceptance event for #560.

- SME review-memory update: [x] not applicable because: no new MEDIUM+ finding surfaced; the reused `SECRET_DETECTORS` and live-gate discipline are already recorded in `docs/sme/security.md` and `docs/sme/correctness.md`.

SME lanes reviewed: [x] correctness [x] adversarial [x] quality [x] security [x] domain-backend [x] gotcha-agent

## Review Gate

- Tier: Standard — net-new operator-facing script, no contract or client surface change.
- Live Open Brain checks: [x] not applicable because: this lane adds an opt-in script and was not authorized to run live traffic against CT 273; the first live run is flagged as the acceptance event for #560.
- `bunx tsc --noEmit` and `bun test scripts/eval-langfuse-egress.test.ts` re-run independently by the supervising node; both green.
- Red-test proof performed and reverted, so the suite is known capable of failing.
- Contract parity: not applicable, no tool payload or schema touched.
- Live Langfuse run is explicitly out of scope and flagged unverified above.

Refs #578. Feeds #571; provides the automated receipt path for #560 and a permanent regression check for #561.

### DIFF
--- a/scripts/eval-langfuse-egress.test.ts
+++ b/scripts/eval-langfuse-egress.test.ts
@@ -1,0 +1,189 @@
+import { describe, expect, test } from "bun:test";
+import {
+  langfuseHost,
+  runEgressCli,
+  serializeCliOutcome,
+  verifyLangfuseEgress,
+  type HttpTransport,
+  type LangfuseReadConfig,
+} from "./eval-langfuse-egress.ts";
+
+const CONFIG: LangfuseReadConfig = {
+  endpoint: "https://langfuse.example",
+  publicKey: "public-id",
+  secretKey: "private-id",
+};
+
+const OPTED_IN_ENV = {
+  OPEN_BRAIN_LANGFUSE_EGRESS: "1",
+  OPENBRAIN_TRACING_ENABLED: "1",
+  OPENBRAIN_TRACING_ENDPOINT: "https://langfuse.example/api/public/ingestion",
+  OPENBRAIN_TRACING_PUBLIC_KEY: "public-id",
+  OPENBRAIN_TRACING_SECRET_KEY: "private-id",
+};
+
+interface FakeObservationOptions {
+  usage?: boolean;
+  cost?: number | null;
+}
+
+function generation(
+  options: FakeObservationOptions = {},
+): Record<string, unknown> {
+  return {
+    id: "observation-1",
+    type: "GENERATION",
+    providedModelName: "claude-fable-5",
+    usageDetails: options.usage === false ? undefined : { input: 2, output: 3 },
+    totalCost: options.cost === undefined ? 0.001 : options.cost,
+    output: "SAFE_BODY_MARKER",
+  };
+}
+
+function fakeTransport(
+  observations: unknown[],
+  bodyExtras: Record<string, unknown> = {},
+): HttpTransport {
+  return async (input) => {
+    const url = new URL(String(input));
+    if (url.pathname === "/api/public/traces") {
+      return Response.json({
+        data: [{ id: "trace-1" }],
+        meta: { totalPages: 1 },
+      });
+    }
+    if (url.pathname === "/api/public/traces/trace-1") {
+      return Response.json({
+        id: "trace-1",
+        sessionId: "run-tag",
+        observations,
+        ...bodyExtras,
+      });
+    }
+    return new Response(null, { status: 404 });
+  };
+}
+
+async function verify(
+  observations: unknown[],
+  options: {
+    expected?: number;
+    requireCost?: boolean;
+    body?: Record<string, unknown>;
+  } = {},
+) {
+  return verifyLangfuseEgress({
+    tag: "run-tag",
+    expectedObservations: options.expected ?? 1,
+    requireCost: options.requireCost,
+    config: CONFIG,
+    transport: fakeTransport(observations, options.body),
+  });
+}
+
+function check(receipt: Awaited<ReturnType<typeof verify>>, label: string) {
+  const found = receipt.checks.find((item) => item.label === label);
+  if (!found) throw new Error(`missing check ${label}`);
+  return found;
+}
+
+describe("Langfuse egress verification", () => {
+  test("normalizes bare hosts and ingestion URLs identically", () => {
+    expect(langfuseHost("https://langfuse.example")).toBe(
+      "https://langfuse.example",
+    );
+    expect(langfuseHost("https://langfuse.example/api/public/ingestion/")).toBe(
+      "https://langfuse.example",
+    );
+  });
+
+  test("arrival-count miss is fatal", async () => {
+    const receipt = await verify([generation()], { expected: 2 });
+
+    expect(receipt.passed).toBeFalse();
+    expect(check(receipt, "arrival_count")).toMatchObject({
+      passed: false,
+      fatal: true,
+      observed: 1,
+      expected: 2,
+    });
+  });
+
+  test("missing usage_details is fatal", async () => {
+    const receipt = await verify([generation({ usage: false })]);
+
+    expect(receipt.passed).toBeFalse();
+    expect(check(receipt, "generation_metadata")).toMatchObject({
+      passed: false,
+      fatal: true,
+      observed: 0,
+      expected: 1,
+    });
+  });
+
+  test("NULL cost is labelled but non-fatal by default", async () => {
+    const receipt = await verify([generation({ cost: null })]);
+
+    expect(receipt.passed).toBeTrue();
+    expect(check(receipt, "total_cost")).toMatchObject({
+      passed: false,
+      fatal: false,
+      observed: 0,
+      expected: 1,
+    });
+  });
+
+  test("NULL cost is fatal when required", async () => {
+    const receipt = await verify([generation({ cost: null })], {
+      requireCost: true,
+    });
+
+    expect(receipt.passed).toBeFalse();
+    expect(check(receipt, "total_cost")).toMatchObject({
+      passed: false,
+      fatal: true,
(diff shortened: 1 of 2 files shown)


rodaddy:
Review swarm (wf_1f3533d6-462): findings-must-fix

Lane: claude-mixed-family-fresh-context-review — verdict `findings-must-fix`. Findings verbatim below; fixes follow in this branch.

### HIGH — `scripts/eval-langfuse-egress.ts`:389

**--drive reports passed:true purely from the child process exit code, but the Python observation lane declines silently and swallows all emit exceptions, so a drive that emitted NOTHING to Langfuse still reports PASS. This is the primary vacuous-pass path in a tool being trusted as the acceptance instrument for #560.**

driveLangfuseEgress spawns `uv run openbrain-capture-stop` and asserts only `exitCode !== 0`. On the child side, session.py:_observe_best_effort returns early with no log when `observation_active(observation)` is false (observe.py:98 — enabled flag off or any coordinate missing), and its `except Exception` at session.py:534 downgrades a real emit failure to a logger.warning while the hook still exits 0. Concretely: if the child's ObservationSettings fails to load the bridged OPENBRAIN_OBSERVATION_* values (e.g. a pydantic validation error on the SecretStr, or `enabled` not parsed from "1"), the sink declines quietly, capture-stop exits 0, and driveLangfuseEgress returns `passed: true, observed.observations: options.count` — a count it fabricated from the input rather than observed. An operator running --drive then --verify sees the drive PASS and blames Langfuse for the subsequent arrival_count failure, or worse, drive is used alone as the #560 receipt. Fix: make --drive prove emission — assert the child's stdout/stderr contains no `observation emit failed`, and require the drive receipt's observed counts to come from a real signal (watermark advance, or drop drive's `passed`/`observed` fields entirely so only --verify can assert).

### HIGH — `scripts/eval-langfuse-egress.ts`:260

**arrival_count uses exact equality against a single immediate query with no retry or settle window; Langfuse ingestion is asynchronous batch-flushed, so a fully correct system fails transiently on the fast path and there is no distinction between 'not arrived yet' and 'dropped'.**

`passed: traces.length === expectedTraces && observations.length === options.expectedObservations` runs against exactly one pass of queryTaggedTraces, invoked immediately when the operator runs --verify. The emitter (langfuse_emitter.py:147 `client.shutdown()`) flushes the batch client-side, but Langfuse server-side ingestion into the traces read API is queue-backed and not read-your-write. Scenario: operator runs --drive, then --verify seconds later; the trace exists but only 2 of 3 generations have been ingested, or the trace row exists with `observations: []`. arrival_count reports observed=2 expected=4, fatal, exit 1 — an incorrect FAIL on a healthy system. The inverse is equally bad for a trusted verifier: because there is no polling window, a genuine silent drop and a 3-second ingestion lag are indistinguishable, so the first transient red will train the operator to re-run until green, which destroys the assertion's value. Needs a bounded poll (retry until expected counts are met or a deadline elapses, then report), and the receipt should record elapsed wait so a slow-but-arrived case is visible.

### MEDIUM — `scripts/eval-langfuse-egress.ts`:229

**secret_scan is scoped to the trace body returned by the detail endpoint, which is exactly the surface the fake transport controls, so the test proves the scanner runs but not that it covers observation input/output content — the place captured transcript secrets actually land.**

secretCounts serializes `trace.body` only. In the real emitter, turn content is written to child observations as `input=`/`output=` (langfuse_emitter.py:_observe_turn), and whether the `/api/public/traces/{id}` response embeds full observation input/output bodies (versus id/type/usage summaries) has never been checked against the live API — the PR body itself concedes the response shape is assumed. If the live detail response returns observations without their content fields, secret_scan serializes a body that structurally cannot contain a leaked credential and passes on every run: a fatal check that cannot fire. The test at eval-langfuse-egress.test.ts:150 hides this because fakeTransport injects the secret into `bodyExtras` at trace top level, and the passing test at line 175 puts SAFE_BODY_MARKER in `output` on the observation — i.e. the fixture proves the scanner sees top-level fields, never that it sees observation content. Fix: assert the scan covers observation-level input/output (a fixture where the only secret lives in an observation's `output`, expected fatal), and confirm against the live API that those fields are actually returned.

### LOW — `scripts/eval-langfuse-egress.ts`:181

**hasNextPage silently stops paginating when the API omits or renames the meta.totalPages field, capping verification at one page and turning a large run into an under-count reported as a drop.**

`hasNextPage` returns false unless `meta.totalPages`/`meta.total_pages` is a number. If the live response nests pagination differently (e.g. `meta.pagination.totalPages`) or omits it, a run with more than DEFAULT_PAGE_LIMIT=100 traces reads only page 1. arrival_count then fails with observed < expected and the operator reads a pagination bug as a Langfuse egress drop. Low because the default count is 3, but the failure mode is silent and inverts the tool's meaning. Cheap fix: when totalPages is absent, continue while the page returned a full `limit` worth of rows rather than assuming a single page.

Issue #563, opened 2026-08-05
durable_memory without a token budget serializes the namespace corpus whole — 39.9 MB measured; response shape is an operator decision


rodaddy:
Measured on 2026-08-04 against the local dogfood database and repo code at `64c9b9f`.

**This issue is a record of a measurement and one open question. It proposes no direction.** Per the repo's standing order of operations, everything stays in and works first. Whatever the response shape should be is the operator's decision, made later, not here.

## What was measured

A `dev.ob.memory.context_pack` request for the `durable_memory` section in namespace `rico` built a **58.5 MB** reply. The `durable_memory` section alone accounted for **39.9 MB**.

Source of record for the observation: `docs/core01-nats-worker-runbook.md`, the `OPEN DEFECT` section.

## What the 39.9 MB is

It is the recall corpus for the namespace, serialized whole.

`server/tools/context-pack-durable-memory.ts` runs one hybrid recall over `ALL_TABLES` (`server/tools/search-constants.ts:25` — thoughts, decisions, relationships, projects, sessions).

Measured `rico` rows at the time of the observation:

| Table | Rows | Size |
|---|---|---|
| sessions | 7,025 | 29 MB (largest single row 17 kB) |
| decisions | 1,519 | 3.4 MB |
| thoughts | 475 | 415 kB |
| projects | 0 | — |
| relationships | 0 | — |

Row data is roughly 33 MB; with the JSON envelope that lands at about the measured 39.9 MB.

## Why it serialized whole

The pack budget contract's `max_tokens` is **optional** (`server/tools/context-pack-args.ts:97`; range `100..20000` when present). The NATS probe passed none, and used a broad query, so no ranking slice was applied.

For contrast, a normal SessionStart pack that does carry a budget measures about **13 kB**.

## The open question of record

What should a budgetless, broad `durable_memory` request return?

That is the design question this issue exists to track, and it is the operator's call. No answer is proposed here.

## Relationship to #549

Cross-reference #549, the delivery-silence defect (fix in flight). The two are independent:

- #549 makes the refusal audible.
- This issue records what was being sent.

Closing #549 does not answer the question above, and answering the question above does not depend on #549.

rodaddy:
[...]

Issue #529, opened 2026-08-04
Operator-invoked provider capture fails scope proof while hook capture works


rodaddy:
## Symptom (live, 2026-08-03)

A well-formed `capture` request through the direct provider CLI (`openbrain-memory`, JSON on stdin, `distilled=true`, full scope object, env from `claudex-observation.env`) fails at the final gate:

```
{"receipt":{...,"direct_attempted":true,"durable":false,
 "error":"session_start result did not prove exact Open Brain scope: agent, channel_id, source",
 "operation":"capture","status":"lost"}}
```

Meanwhile the hook capture path (Stop/SessionStart entrypoints) writes fine with the same env. The operator path also demanded several undocumented shapes before reaching this point (scope must be object; namespace comes only from env `OPENBRAIN_NAMESPACE`, a top-level `namespace` key is silently ignored as an "optional request key"; `distilled=true` required) — each discovered by trial.

## Why it matters

The AGENTS.md contract tells Claude to capture distilled in-flight events through exactly this provider. A capture path that fails scope-proof means in-flight process learnings get lost (status `lost` is honest, but the write still didn't happen) and agents fall back to nothing. Hit live while trying to record a workflow-staging lesson; the lesson had to live in an issue comment instead.

## Wanted

- Either the provider's session_start handshake proves the scope fields the capture gate demands (`agent`, `channel_id`, `source`), or the gate's demanded set matches what session_start actually proves.
- The silently-ignored top-level `namespace` key should be a named rejection, not `ignored_optional_request_keys` (same silent-drop family as #464/#431).
- One documented, working operator recipe in `docs/memory-contract.md` or the provider README, verified by a live canary.

rodaddy:
Status check 2026-08-04. Core defect UNCHANGED (MERGED main 2a8d2db): the exact-scope gate still fires from `_runtime_validation.py:689` / `_require_session_start_scope` (line 72), so operator-invoked capture still fails scope proof while hook capture writes. Partial delivery on the third Wanted bullet: PR #533 (MERGED 2026-08-04, closed #464) forwards `candidate_type`/`memory_lifecycle_action`/`candidate_scope` through the capture allowlist and adds `ignored_optional_request_keys` to the receipt, so silently-dropped keys are now named. Still owed here: the session_start handshake proving `agent`/`channel_id`/`source` (or the gate demanding only what it proves), a named rejection for the top-level `namespace` key, and a live-canary-verified operator recipe in `docs/memory-contract.md`.

[...]

PR #580, opened 2026-08-05
feat(eval): scripted scenario fixtures for capture, durable_memory shape, and checkpoint round-trips (#578)

rodaddy:
Phase 1 of #578: scenario breadth on the **existing** eval harness (plan delta
item 1). No new parallel harness — this extends `eval/open-brain/live/` and
reuses the EVAL-3 discipline: per-run throwaway namespace derivation from
`live/config.ts`, teardown accounting, the `OPEN_BRAIN_LIVE_EVAL=1` gate, and
content-free receipts.

A scenario is a scripted input with an asserted output, alongside (not replacing)
the existing ranking-probe path. Three first scenarios:

- **capture round-trip** — drives a capture, then proves the write by an
  independent `session_context` read-back matched on content. The load-bearing
  case from #529 is the **exit-0 / zero-write** silent drop: the gate fails with
  `capture_round_trip_missing` even when the provider exits 0 and claims success,
  because the verdict comes from reading the record back, never from the
  receipt's own status field. Also fails on `status: lost` + `durable: false`,
  and on any request key the receipt relegates to
  `ignored_optional_request_keys` that the fixture declared it expects honored.
- **durable_memory pack shape** — requests with an **explicit token budget** and
  asserts section presence, item count, and that the serialized reply stays
  within the budget. This is the shape check #563 measured the absence of
  (~40–60 MB whole-corpus serialization on a budgetless broad request); it uses a
  real query since `durable_memory` is query-gated.
- **checkpoint/wrap round-trip** — writes a checkpoint, resume-reads it, asserts
  the lane events surface.

Every scenario runs in both a fake-transport unit mode under plain `bun test`
(no hosted service, no database) and an env-gated live mode. Teardown runs in a
`finally` over exactly this run's collected record ids and reports
attempted/archived/already_absent/failed, with `teardown_failed` as a gate
failure.

Delta item 2 (Langfuse egress verifier) is deliberately **out of scope** here.

## Verification

Re-run independently by the supervising node in the worktree, not just reported
by the implementer:

- `bunx tsc --noEmit` — exit 0, no output.
- `bun test eval/` — 203 pass, 0 fail, 833 expect() calls.
- `bun test` (full suite, regression check) — 3212 pass, 509 skip, 0 fail across
  234 files.
- **Gate proven to bite:** flipping the durable-memory fixture `min_items` from
  1 to 2 produced exit 1 with the named failure
  `durable-memory-pack-shape:durable_memory_items_missing`; the break was
  reverted and the suite returned to green. Separately,
  `eval/open-brain/live/__tests__/scenario-gate.test.ts:194` asserts the
  zero-write case fails while `provider_exit_zero` is true.
- **Live: unrun.** The repo `.env` carries no `OPEN_BRAIN_LIVE_EVAL_*` variables
  (verified), so no live scenario was executed. No live coverage is claimed.

## Critical Self-Review

- Highest-risk behavior: the capture round-trip verdict — if it ever trusted the receipt's status instead of the independent read-back, the whole #529 exit-0/zero-write class would silently pass; a dedicated test pins exit-0-with-no-write as a failure.
- Assumptions that could be wrong: the fake transport's modelling of provider receipt and session_context shapes is asserted only against fixtures, so a real provider whose payload shape has drifted would not be caught until the live gate is first run.
- Missing/weak tests: no live execution at all this round, so the live transport path (`scenario-transport.ts`) is exercised only through fakes; the durable_memory budget check is asserted on serialized length rather than real tokenizer tokens.
- Security/permission risk: scenarios write only to a per-run throwaway namespace derived by the existing `boundRunId`/config helpers, and receipts print ids, counts, and statuses only — no memory bodies, tokens, or raw remote bodies.
- Migration/deploy risk: none — additive test/eval tooling plus one new package.json script; no schema, server, or runtime path changes.
- Downstream client/runtime risk: none for clients; no MCP tool, transport contract, or Python client shape changed, so contract parity fixtures are untouched.
- Rollback/cleanup concern: teardown runs in a `finally` over exactly this run's record ids and reports failed counts as a gate failure, so an aborted live run does not strand rows; reverting the commit removes the harness entirely.
- Fixes made before PR: supervising node re-ran tsc, the eval suite, and the full suite independently, verified the zero-write assertion reads back rather than trusting the receipt, and confirmed the live-unrun claim against the actual `.env` contents.
- Known residual risk: the gate is WRITTEN and green offline but has never driven real traffic, so its first live run against a hosted Open Brain is the outstanding proof and may surface shape drift.
- SME review-memory update: [x] not applicable because: this adds eval scenarios and found no new reviewer-facing defect pattern; the #529/#563 failure modes it encodes are already recorded in their issues.

## Review Gate

- Live Open Brain checks: [x] not applicable because: no `OPEN_BRAIN_LIVE_EVAL_*` variables are present in the repo `.env`, so the env-gated live mode cannot run; scenarios pass in fake-transport mode and live execution is the named residual risk.

Refs #578. Encodes failure modes from #529 and #563.

### DIFF
--- a/eval/open-brain/README.md
+++ b/eval/open-brain/README.md
@@ -307,6 +307,36 @@ URL + token). The receipt (`schema: openbrain.reflex_ab_gate.v1`) is
 content-free: only ids, namespaces, labels, counts, and booleans — no memory
 bodies, tokens, or secrets.
 
+## Scripted Scenario Gate (#578 phase 1)
+
+The scenario gate extends the existing live harness with scripted
+input-to-asserted-output fixtures. It uses the same opt-in environment,
+per-invocation throwaway namespace derivation, real MCP client/fake caller seam,
+content-free receipt discipline, and EVAL-3 teardown tally.
+
+```bash
+bun run eval:scenarios
+bun run eval:scenarios -- --json
+bun run eval:scenarios -- --fixture eval/open-brain/fixtures/scenarios-v1.json
+```
+
+`fixtures/scenarios-v1.json` contains the first three scenarios:
+
+- provider capture followed by exact `session_context` read-back (the write
+  receipt alone never proves persistence), including named failures for
+  lost/non-durable receipts and expected request keys reported as ignored;
+- a real-query `durable_memory` pack request carrying an explicit token budget,
+  with section/item shape and serialized whole-pack accounting assertions;
+- capture + checkpoint + resume-read, asserting the lane event and checkpoint
+  summary both surface.
+
+Plain `bun test` runs these against a fake `ScenarioTransport`; no service or
+PostgreSQL connection is touched. Live mode uses the installed
+`openbrain-memory` provider for capture/checkpoint, the existing MCP transport
+for reads and pack assembly, and exact-id cleanup for the records created by the
+run. Output contains only fixture/scenario ids, namespace labels, counts,
+statuses, booleans, and named failure classes.
+
 ## Next Expansion Points
 
 - Add a live Open Brain adapter that loads fixtures into an isolated namespace
--- a/eval/open-brain/fixtures/scenarios-v1.json
+++ b/eval/open-brain/fixtures/scenarios-v1.json
@@ -19,10 +19,6 @@
           "session_key": "eval:scenario:{{run_id}}:capture",
           "project": "open-brain"
         },
-        "expected_honored_request_keys": [
-          "candidate_type",
-          "memory_lifecycle_action"
-        ],
         "optional_request_fields": {
           "candidate_type": "durable_decision",
           "memory_lifecycle_action": "candidate"
@@ -69,8 +65,7 @@
           "channel_id": "checkpoint-round-trip",
           "session_key": "eval:scenario:{{run_id}}:checkpoint",
           "project": "open-brain"
-        },
-        "expected_honored_request_keys": []
+        }
       },
       "checkpoint": {
         "operation": "checkpoint",
--- a/eval/open-brain/fixtures/scenarios-v1.json
+++ b/eval/open-brain/fixtures/scenarios-v1.json
@@ -1,0 +1,92 @@
+{
+  "schema_version": 1,
+  "fixture_id": "open-brain-scenarios-v1",
+  "description": "Scripted live and fake-transport scenarios for capture, durable memory, and checkpoint resume behavior.",
+  "scenarios": [
+    {
+      "id": "capture-round-trip",
+      "kind": "capture_round_trip",
+      "request": {
+        "operation": "capture",
+        "distilled": true,
+        "content": "Scenario {{run_id}} capture round-trip marker.",
+        "event_type": "decision",
+        "scope": {
+          "agent": "scenario-eval",
+          "platform": "eval",
+          "server_id": "open-brain-scenarios",
+          "channel_id": "capture-round-trip",
+          "session_key": "eval:scenario:{{run_id}}:capture",
+          "project": "open-brain"
+        },
+        "expected_honored_request_keys": [
+          "candidate_type",
+          "memory_lifecycle_action"
+        ],
+        "optional_request_fields": {
+          "candidate_type": "durable_decision",
+          "memory_lifecycle_action": "candidate"
+        }
+      }
+    },
+    {
+      "id": "durable-memory-pack-shape",
+      "kind": "durable_memory_shape",
+      "seed": {
+        "table": "thoughts",
+        "content": "Scenario {{run_id}} marlin telemetry cache marker.",
+        "tags": ["eval", "scenario-578"]
+      },
+      "request": {
+        "query": "Scenario {{run_id}} marlin telemetry cache marker",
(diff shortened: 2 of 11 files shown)


rodaddy:
Review swarm (wf_1f3533d6-462): findings-must-fix

Lane: claude-mixed-family-fresh-context-review — verdict `findings-must-fix`

### HIGH — `eval/open-brain/live/scenario-transport.ts:32`

**The fake transport cannot produce the #529 silent-drop shape (exit 0, empty stdout, no receipt); the live transport turns that exact case into a thrown LiveTransportError that the gate relabels as the generic `scenario_transport_error`, so the load-bearing failure mode the PR claims to encode is never exercised offline and is misattributed live.**

_plans/issues/529-*.md line 48 documents the reproduction as `exit 0, zero stdout/stderr, no receipt`. In the live path, `parseProviderOutput` (scenario-transport.ts:32) hits `JSON.parse("")`, throws `LiveTransportError("openbrain-memory:malformed-output")`; runScenarioGate's per-scenario `catch` (scenario-gate.ts:~330) swallows it into `failures: ["scenario_transport_error"]`. So the silent-drop regression reports as a transport error, not as a capture failure, and `checks` is `{}` — no `provider_exit_zero`, no `read_back_exact`. Meanwhile the fake's `execution()` always synthesizes a well-formed receipt, so no test covers it. The `zeroWrite` option models a *different* thing: a valid saved receipt with no row. The named #529 shape (no receipt at all) has zero coverage in either mode. Fix: add a fake option emitting empty/receipt-less stdout with exitCode 0, and give the gate a distinct named failure (e.g. `capture_receipt_absent`) rather than collapsing it into `scenario_transport_error`.

### MEDIUM — `eval/open-brain/live/scenario-gate.ts:76`

**The `capture_receipt_lost` branch is unreachable in any way that isolates it: `status === "lost"` always coexists with a non-zero exit code from the real provider, and the fake ties the two together too, so the assertion can never fire on its own and proves nothing beyond the exit code.**

providerFailures (scenario-gate.ts:71-78) checks `exitCode !== 0` first, then `status === "lost" && !durable`. The real CLI's `__main__.py:49-50` computes `exit_code = 1 if status in {FAILED, LOST}`, so a `lost` receipt is *always* exit 1 — `provider_exit_nonzero` fires unconditionally alongside it. The fake mirrors this: `execution()` (scenario-gate.test.ts:47) sets `exitCode: lost ? 1 : 0`, so the `lostCapture` test at line 205 asserts `capture_receipt_lost` in a run that would already have failed on the exit code alone. A provider regression that returned `status: lost` with exit 0 — precisely the false-receipt class #529 is about — would be caught, but a regression that stopped setting `status: lost` while still exiting 1 would pass this assertion silently. The lost-branch test is therefore a vacuous pass: flip `execution()` to `exitCode: 0` for the lost case and the test still needs to go green for the assertion to be load-bearing.

### MEDIUM — `eval/open-brain/live/scenario-gate.ts:96`

**`providerPayload` spreads `optional_request_fields` at top level and sends `scope.project`, but `project` is not in the CLI's capture allowlist, so every live capture scenario will report `project` in `ignored_optional_request_keys` — and because the gate only checks *declared* keys, that silent drop passes.**

cli.py:42 sets `_COMMON_KEYS = {"config", "operation", "scope"}` and the capture allowlist (cli.py:63-72) is `{candidate_scope, candidate_type, content, distilled, event_type, memory_lifecycle_action}`. `scope` is passed whole so `project` rides inside it (that one is fine), but `_project_request` only filters top level — the real risk is the inverse asymmetry: `providerFailures` (scenario-gate.ts:79-82) only fails on keys the fixture listed in `expected_honored_request_keys`. The shipped fixture lists two (`candidate_type`, `memory_lifecycle_action`) and the checkpoint scenario lists **none** (`scenarios-v1.json` `"expected_honored_request_keys": []`). So for checkpoint, the provider could relegate `summary`, `key_decisions`, and `next_steps` entirely to `ignored_optional_request_keys` and the gate would raise no failure from the receipt — it would only be caught if `checkpoint_summary_missing_on_resume` also fires. Any key the fixture author forgets to declare is unchecked by construction. Fix: assert `ignored_optional_request_keys` is *empty* for keys the request actually sent, rather than only checking an opt-in declared list.

### LOW — `eval/open-brain/live/scenario-gate.ts:186`

**`durableSectionChecks` derives `itemCount` from the server-reported `item_count` when present and only falls back to `items.length`, so a server that reports an inflated count with an empty items array passes the `min_items` assertion.**

scenario-gate.ts:180-182: `const itemCount = typeof section?.item_count === "number" ? section.item_count : items.length`. If the pack regresses such that `durable_memory` emits `{ item_count: 5, items: [] }` — a plausible truncation/serialization bug and exactly the class the sibling complete-pack-gate guards against — `itemCount` is 5, `min_items` of 1 is satisfied, and the scenario passes with zero actual items. The fake never exercises the divergence (it emits `items: [1 item], item_count: 1`, consistently). Cross-check the two: fail if `item_count` and `items.length` disagree, and assert on `items.length` for the `min_items` bar.

PR #586, opened 2026-08-05
fix(eval): make egress drive failures actionable

### LINKED ISSUE #582: Observation lane zero-emit: capture child exits 0, writes no watermark, emits nothing — reproduced under a scrubbed env with all four tracing vars set
Found by the #578 egress verifier's first live run (drive check emit_proven FAIL: watermark offset never advanced; capture_exit PASS). Reproduced under `env -i` with only the four OPENBRAIN_TRACING_* vars resolved non-empty from local-clone.env — so this is not env-var typo rejection (config.py:560-564 hypothesis disproved this session). The observation lane declines or fails silently while the hook exits 0: the #529/#536 silent-zero-capture class, now with a deterministic reproduction command (scripts/eval-langfuse-egress.ts --drive --tag <t>). Downstream effect: ZERO traces arrive at Langfuse, so #560's cost acceptance cannot even be measured. Verify phase authenticated and polled fine (31 polls) — the write side is the defect. Cross-ref #529, #560, #571.

rodaddy:
Fixes #582.

## Summary

- Require raw-capture coordinates before `eval-langfuse-egress.ts --drive` starts a real Stop-hook child.
- Build that child environment from declared capture and observation settings only, dropping unrelated `OPENBRAIN_*` values that Python rejects.
- Fall back to `sqlite3 -ifexists` when Bun cannot immediately reopen the child-written watermark database.
- Verify only capture-emitter traces and recognize the actual Langfuse public-API model and cost field names.
- Add regression coverage for the previous silent, successful child result, environment isolation, transient watermark-read failure, and shared-session server traces.

## Diagnosis

All four tracing variables were confirmed present without disclosing their values. The previous driver copied unrelated `OPENBRAIN_*` names from the parent process into the Python child. `load_capture_settings()` rejects those names, `stop.capture_stop_with()` swallows the resulting `UnknownEnvironmentVariableError`, and the child returns `0` before it reaches the observation path. Under a strict tracing-only environment, raw capture is instead missing its endpoint/token and follows the same silent hook-success path before observation can run.

The repair preserves the raw-first watermark contract. It reports either setup defect from the egress CLI as stderr plus exit `2`, rather than treating an unstarted observation lane as a Langfuse emission result. If Bun cannot reopen the just-written watermark database, the verifier uses `sqlite3 -ifexists`; a persistent failure remains non-zero.

The live read path also exposed a separate verifier defect: the raw lane's `session_start` gives Open Brain server tracing the same Langfuse session ID as the capture emitter. The verifier now retains only the emitter's stable `open-brain-capture` tag. Its public API returns `model` and `totalPrice`, not the legacy field spellings the verifier previously required.

## Validation

- `BUN_TMPDIR=/Volumes/ThunderBolt/_tmp/open-brain/_scratch/bun-tmp bun test scripts/eval-langfuse-egress.test.ts` (21 tests)
- `BUN_TMPDIR=/Volumes/ThunderBolt/_tmp/open-brain/_scratch/bun-tmp bunx tsc --noEmit`
- `UV_CACHE_DIR=/Volumes/ThunderBolt/_tmp/open-brain/_validation-runs/uv-cache uv run mypy src/openbrain`
- `UV_CACHE_DIR=/Volumes/ThunderBolt/_tmp/open-brain/_validation-runs/uv-cache uv run ruff check src tests`
- `UV_CACHE_DIR=/Volumes/ThunderBolt/_tmp/open-brain/_validation-runs/uv-cache uv run pytest -q` (593 passed, 1 skipped, 19 deselected)
- Tracing-only configuration returns the raw-capture setup error on stderr and exits `2`; no values were printed.
- A configured live drive wrote both raw and `observe:` watermark rows. Its Langfuse verification now passes with 1 capture trace, 4 observations, and 3 generations; no credential values were printed.

## Downstream Rollout

Not applicable: this changes a repository-local evaluation script and does not change an MCP, transport, database, generated-skill, or Python client contract.

## Critical Self-Review

- Highest-risk behavior: the child environment must retain the two raw-capture coordinates while rejecting foreign `OPENBRAIN_*` names.
- Assumptions that could be wrong: a future valid Stop-hook setting may need forwarding; the driver currently needs only raw endpoint/token plus its own observation settings.
- Missing/weak tests: the test suite uses a synthetic child environment; it does not assert Langfuse arrival over the network.
- Security/permission risk: the child receives a capture token, but the driver never serializes the environment or error values.
- Migration/deploy risk: no migration or deployment behavior changes.
- Downstream client/runtime risk: no externally consumed Open Brain contract changes.
- Rollback/cleanup concern: reverting this two-file change restores the earlier driver behavior; no persisted state is created.
- Fixes made before PR: added fail-fast raw configuration validation, filtered the Python child environment to declared settings, added a durable watermark read fallback, and filtered the verifier to capture-owned traces using the public API field names it actually receives.
- Known residual risk: a persistently unreadable watermark returns a non-zero verifier error rather than a false receipt; a future emitter tag change must update the verifier fixture.
- SME review-memory update: [x] not applicable because: no reusable source-path review pattern changed; this is a bounded evaluator configuration repair.

## Review Gate

- [x] MEDIUM+ review findings were captured: no MEDIUM+ finding remained after the source-path diagnosis and focused regression checks.
- Live Open Brain checks: [x] not applicable because: the local evaluation driver stops before any network call when required raw-capture configuration is absent.

### DIFF
--- a/scripts/eval-langfuse-egress.test.ts
+++ b/scripts/eval-langfuse-egress.test.ts
@@ -119,6 +119,23 @@ describe("Langfuse egress verification", () => {
     );
   });
 
+  test("empty-string capture vars still reach the shared aliases", () => {
+    // Deployed hook-env wrappers pass VAR="${VAR:-}": unset arrives as "", not
+    // absence (#525's shape). The alias fallback must engage on both.
+    const config = readCaptureDriveConfig({
+      ...OPTED_IN_ENV,
+      OPENBRAIN_CAPTURE_BASE_URL: "",
+      OPENBRAIN_CAPTURE_TOKEN: "",
+      OPENBRAIN_BASE_URL: "http://127.0.0.1:3100",
+      OPENBRAIN_TOKEN: "alias-token",
+    });
+
+    expect(config).toEqual({
+      baseUrl: "http://127.0.0.1:3100",
+      token: "alias-token",
+    });
+  });
+
   test("drive returns a non-zero configuration error before spawning an invalid child", async () => {
     const outcome = await runEgressCli(
       ["--drive", "--tag", "run-tag"],
--- a/scripts/eval-langfuse-egress.test.ts
+++ b/scripts/eval-langfuse-egress.test.ts
@@ -1,8 +1,12 @@
 import { describe, expect, test } from "bun:test";
+import { Database } from "bun:sqlite";
 import {
+  captureEnvironment,
   evaluateDriveOutcome,
   langfuseHost,
+  observationOffset,
   parseEgressArgs,
+  readCaptureDriveConfig,
   runEgressCli,
   serializeCliOutcome,
   verifyLangfuseEgress,
@@ -24,6 +28,12 @@ const OPTED_IN_ENV = {
   OPENBRAIN_TRACING_SECRET_KEY: "private-id",
 };
 
+const CAPTURE_DRIVE_ENV = {
+  ...OPTED_IN_ENV,
+  OPENBRAIN_CAPTURE_BASE_URL: "http://127.0.0.1:3100",
+  OPENBRAIN_CAPTURE_TOKEN: "capture-token",
+};
+
 interface FakeObservationOptions {
   usage?: boolean;
   cost?: number | null;
@@ -36,9 +46,9 @@ function generation(
   return {
     id: "observation-1",
     type: "GENERATION",
-    providedModelName: "claude-fable-5",
+    model: "claude-fable-5",
     usageDetails: options.usage === false ? undefined : { input: 2, output: 3 },
-    totalCost: options.cost === undefined ? 0.001 : options.cost,
+    totalPrice: options.cost === undefined ? 0.001 : options.cost,
     output: options.output ?? "SAFE_BODY_MARKER",
   };
 }
@@ -59,6 +69,7 @@ function fakeTransport(
       return Response.json({
         id: "trace-1",
         sessionId: "run-tag",
+        tags: ["open-brain-capture"],
         observations,
         ...bodyExtras,
       });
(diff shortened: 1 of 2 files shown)


rodaddy:
## Review — Claude opposite-family lane (Terra-authored PR)

**Verdict: request-changes — one MEDIUM, fix is small and bounded.**

### MEDIUM — `??` defeats the documented alias fallback on empty string
`scripts/eval-langfuse-egress.ts:166-170` (`readCaptureDriveConfig`):

```ts
env.OPENBRAIN_CAPTURE_BASE_URL?.trim() ?? env.OPENBRAIN_BASE_URL?.trim() ?? ""
```

`??` only falls through on null/undefined; `""?.trim()` is `""`, so the alias never engages. The deployed hook-env wrapper shape passes `VAR="${VAR:-}"` — **empty string, not absence** — the exact shape `python/openbrain/src/openbrain/config.py:163-171` documents as the cause of #525's silent zero-capture (`_empty_opt_in_means_unset` exists to defend it; the `:-` pattern is confirmed present in `local-clone-deploy.sh`, `setup-client.sh`, `client-bundle.sh`, and the client-install runbook). A host setting only `OPENBRAIN_BASE_URL` behind such a wrapper gets a hard `--drive` failure claiming both vars required — a loud but misleading fail, and the docstring's "or their shared aliases" promise is false in the fleet's deployed shape. Fix: `??` → `||` on both lines, plus an empty-string alias test (none exists).

### Cleared by live probe (not assumption)
- **Allowlist**: compared against every declared alias in `CaptureSettings`/`ObservationSettings`; nothing the child needs is dropped, and foreign `OPENBRAIN_*` can never re-reach the child — #582's root cause closed at the owning boundary.
- **Tag filter ordering**: `isCaptureTrace` filters BEFORE the `traces.length > 0` coupling; probed tag-mismatch → observed 0 → `passed: false`. Not vacuous.
- **`sqlite3 -ifexists` fallback**: missing file exits 1, corrupt file exits 1, both throw; cannot launder a missing watermark into offset 0.
- **Mutation proof**: `verification excludes server traces that share the capture session` goes red against the pre-fix line (`expect(received).toBeTrue() — Received: false`), 21/21 green on PR head. `tsc` clean in changed files (6 pre-existing TS2307 on optional deps reproduce on main).

### SME note
Finding is the empty-string-as-absence class already catalogued for pydantic, recurring in TS env reads — will add to `docs/sme/gotcha-agent.md` so the next swarm checks `??` against env reads.

PR #590, opened 2026-08-05
fix(tracing): resolve Langfuse release from deploy stamp

### LINKED ISSUE #587: Langfuse release is null on every deployed trace: resolver shells git rev-parse but the deployed runtime is a gitless archive
Found live 2026-08-05 while verifying the #560 closure (RUNNING-verified, not inferred).

**Measured:** local dogfood deployed to trunk `a5a1274` with revision proof (pid 23876 → 51594, `/health` reports `"revision":"a5a1274"`). A fresh post-restart traced tool call (`append_session_event`, 21:45:45Z) still shows `release: null` in the Langfuse traces API.

**Root cause (read in deployed source):** the release resolver in `server/observability/langfuse-tracing.ts` shells `git rev-parse --short HEAD` (line 188 in the deployed copy). But `scripts/local-clone-deploy.sh` ships the runtime via `git archive` — deliberately, so the working tree is structurally invisible — which means **the deployed runtime has no `.git` directory at all** (verified: `/Volumes/ThunderBolt/open-brain-local/app/.git` does not exist). `git rev-parse` can never succeed in the deployed shape, the resolver treats unknown-release as omit-and-continue (correct per its own comment: "Not knowing the release is never a reason to lose tracing"), and the result is `release: null` on 100% of deployed traces — the exact field #560's fix shape added so cost changes could be attributed to a build.

The irony: `/health` answers the same question correctly on the same process, because the deploy stamps the revision into the runtime directory and `/health` reads the stamp. The tracing resolver and the health endpoint disagree about whether the build is knowable.

**Fix shape (PROPOSED, ~S):** the release resolver reads the deploy's revision stamp first (same source `/health` uses), falling back to `git rev-parse` for dev-checkout runs where no stamp exists. A test that fails when a stamped-but-gitless runtime yields a null release.

**Why it matters:** release attribution only exists in the environments nobody debugs (dev checkouts) and is absent in the ones everybody debugs (deployed dogfood, core01 — which deploys the same `git archive` way via the core01 script this one mirrors).

rodaddy:
Fixes #587

## Summary

- Read the deployed revision through the same `.deployed-revision` parser used by `/health`.
- Fall back to `git rev-parse --short HEAD` only when no deploy stamp is available.
- Add a regression test for a stamped runtime with no usable git checkout.
- Capture the post-merge deployment-shape miss in `docs/sme/gotcha-agent.md`.

## Validation

- `bunx tsc --noEmit` — pass.
- `bun test server/observability/langfuse-tracing.test.ts` — 44 pass, 0 fail.
- `git diff --check` — pass.
- Failure proof: temporarily changed `resolveRepoRelease` to skip the deploy stamp; the new test failed with expected `"0123456"` and received `undefined`. Restored the stamp-first logic and reran the focused suite green.

## Critical Self-Review

- Highest-risk behavior: Release lookup order could still prefer checkout metadata over the deployed artifact; the regression test makes the git path unusable and requires the stamp value.
- Assumptions that could be wrong: `.deployed-revision` remains beside the runtime tree and retains the `short_sha` key parsed by `/health`; both were verified in the deploy script and current health code.
- Missing/weak tests: No live Langfuse export was sent; the focused test covers the resolver boundary without constructing an exporter or opening a socket.
- Security/permission risk: None; the change reads an existing local revision stamp and does not alter auth, namespaces, payloads, or permissions.
- Migration/deploy risk: No migration; deployed archives already contain the stamp, while unstamped development checkouts retain the git fallback.
- Downstream client/runtime risk: None; no MCP schema, protocol, transport, or client-facing contract changed.
- Rollback/cleanup concern: Reverting the two implementation files restores the previous checkout-only behavior; the worktree is intentionally left in place.
- Fixes made before PR: Reused `readDeployedRevision` instead of adding a second stamp parser and added the deployment-shape regression test.
- Known residual risk: A malformed or unreadable stamp falls back to git and can still yield no release in a gitless runtime, matching the existing best-effort unknown-release behavior.
- SME review-memory update: [x] docs/sme/ updated - added the deployed-artifact identity check from issue #587.

## Review Gate

- [x] MEDIUM+ review findings were captured: issue #587's deployment-shape miss was added to `docs/sme/gotcha-agent.md`.
- Live Open Brain checks: [x] not applicable because: this change only resolves process build metadata and does not read or write Open Brain data.


rodaddy:

**HIGH** — The regression test is vacuous with respect to the actual defect: it injects `readStamp`, so the production reader `readRuntimeDeployStamp()` — the only code that decides where the stamp is read from — has zero coverage, and issue #587 can regress completely with the suite green.
- Location: server/observability/langfuse-tracing.test.ts:985-995 (test) against server/observability/langfuse-tracing.ts:203-208 (`readRuntimeDeployStamp`)
- Failure scenario: RUNNING-verified by mutation in a scratch worktree of db11c31. (a) Change `new URL("../../.deployed-revision", import.meta.url)` to `"../.deployed-revision"` — a path that resolves to `<root>/server/.deployed-revision` and never exists in any deployed artifact: `bun test server/observability/langfuse-tracing.test.ts` reports 44 pass, 0 fail. (b) Make `readRuntimeDeployStamp()` unconditionally return undefined — which is precisely the #587 symptom, every deployed trace emits no release: 44 pass, 0 fail again. Meanwhile the author's own claimed proof (deleting the stamp-first branch) DOES go red, so the suite looks like it guards the fix while leaving the production wiring entirely unpinned. This is the correctness-SME vacuous-pass / 'a gate that compares a constant to itself' class (docs/sme/correctness.md:1535): the test asserts that `readDeployedRevision` parses a string the test itself
- Suggested fix: Pin the production path, not just the injected one. Export `readRuntimeDeployStamp` (or the resolved URL) and assert it lands on `<repo-root>/.deployed-revision` — e.g. `expect(deployStampUrl().pathname).toBe(new URL("../../.deployed-revision", import.meta.url).pathname)` is circular, so instead assert it equals the same location `server/config.ts` uses, which is the stated invariant (`does build identity use the SAME stamp /health reads`). Better: a filesystem-backed test that writes a `.deployed-revision` into a temp runtime root under {temp_workspace}/open-brain/_scratch/, invokes the real 

**MEDIUM** — `core01-deploy-local.sh` tars the working tree and neither writes nor excludes `.deployed-revision`, so a stamp left in REPO_DIR by a prior `local-clone-deploy.sh` run is copied verbatim into the core01 runtime — and this PR newly makes that stale value the top-priority source for Langfuse release attribution.
- Location: scripts/core01-deploy-local.sh:120-137 (tar exclude list) interacting with server/observability/langfuse-tracing.ts:190-193 (stamp-first precedence)
- Failure scenario: RUNNING-verified with real tar under _scratch: a `.deployed-revision` containing `short_sha=deadbee` placed in the source dir survives the exact exclude list core01-deploy-local.sh uses (`--exclude ./.git --exclude ./node_modules ...`) and lands in the staging dir unchanged. `.deployed-revision` is gitignored (.gitignore:106) but tar does not honor gitignore, and `rg -c deployed-revision scripts/core01-deploy-local.sh` returns zero — the script neither writes a fresh stamp nor drops the inherited one. Only `local-clone-deploy.sh` ever writes the stamp. So on a machine where both scripts have been used against the same REPO_DIR, core01's runtime carries the short sha of whatever the LOCAL clone last deployed. Before this PR that mislabeled `/health`'s `revision` field only; after it, `resolveRepoRelease` returns the stamp FIRST and unconditionally, so every Langfuse trace is attributed to
- Suggested fix: Either add `--exclude "./.deployed-revision"` to the core01 tar and have that script write its own stamp the way local-clone-deploy.sh does (preferred — it makes core01's build identity real rather than absent), or at minimum exclude it so core01 falls through to the git path instead of asserting a false commit. A stale stamp that lies is strictly worse than an absent one, which is the PR's own stated rationale for preferring `undefined` over `"unknown"`.

**LOW** — `resolveRepoRelease`'s try/catch wraps only the git fallback; an exception from the default stamp reader other than a missing file is unhandled and would now propagate into `repoRelease()`, contradicting the module's stated 'not knowing the release is never a reason to lose tracing' invariant.
- Location: server/observability/langfuse-tracing.ts:190-201
- Failure scenario: `readDeployedRevision` (server/transport/server-identity.ts:221-238) does catch throws from its `readStamp` callback and returns undefined, so today this is contained and I could not make it throw — hence LOW, not a live defect. The fragility is structural: the stamp call sits OUTSIDE the try block while the git call sits inside it, so the safety depends entirely on a catch living in a different module that the author did not restate here. If `readDeployedRevision` is ever refactored to propagate (e.g. to distinguish a corrupt stamp from an absent one — a plausible follow-up given the PR's own 'known residual risk' note about malformed stamps), release resolution starts throwing on the tracing init path with no local guard. The asymmetry also reads as intentional, which will mislead the next editor.
- Suggested fix: Move the `readDeployedRevision(...)` call inside the same try block as the git fallback, so both sources are covered by the one comment that already says a failure must never cost tracing. Zero behavior change today, and it makes the invariant local instead of inherited.

**Independent failure-proof re-verification:** Re-verified the PR body's claimed failure-proof independently in a scratch worktree of db11c31 (bun install --frozen-lockfile, BUN_TMPDIR under _tmp). MUTATION 1 — deleted the two stamp-first lines from `resolveRepoRelease`: test "uses the deploy stamp when the runtime has no git checkout" went RED (43 pass / 1 fail), matching the author's claim. Restored -> 44 pass / 0 fail, `git diff --stat` empty. The author's proof is HONEST and reproduces.

However two ADDITIONAL mutations I ran show the guard is not tight. MUTATION 2 — changed the production reader's path from `../../.deployed-revision` to `../.deployed-revision` (wrong depth, resolves to `<root>/server/.deployed-revision`, a file that

**Reviewer-run receipts:** Run by me on a scratch worktree of origin/fix/587-release-stamp @ db11c31 at /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/rev590 (now removed).
- `bun install --frozen-lockfile` — 169 packages, clean.
- `bun test server/observability/langfuse-tracing.test.ts` — 44 pass, 0 fail, 96 expect() calls, 216ms. Matches the PR body's claim.
- `bunx tsc --noEmit` — exit 0, no diagnostics.
- Mutation runs: stamp-first removed -> 43/1 RED (proof holds); wrong path depth -> 44/0 GREEN; production reader disabled -> 44/0 GREEN.
- Path-depth arithmetic verified by executing `new URL(...)` under Bun: `server/observability/langfuse-tracing.ts` + `../../` and `server/config.ts` + `../` both resolve to `<ru

rodaddy:
## Fix-lane receipt — `cb56da7`

All opposite-family findings are addressed.

### HIGH — production runtime stamp reader was unpinned
- **Fix:** Exported and exercised the real `readRuntimeDeployStamp()` against a filesystem-backed `.deployed-revision` at the deployed tree root. The test also runs `resolveRepoRelease()` with the default stamp reader and an unavailable git checkout.
- **Red proof:** Mutated the production URL from `../../.deployed-revision` to `../.deployed-revision`; the focused test failed with `ENOENT` at `server/.deployed-revision` (`0 pass / 1 fail`).
- **Green proof:** Restored the production path; the focused mutation test passed (`1 pass / 0 fail`), and the complete tracing suite passed (`46 pass / 0 fail`, `99 expect()` calls).

### MEDIUM — core01 could inherit a stale local-clone stamp
- **Fix:** Moved core01 runtime packaging into `scripts/core01-package-runtime.sh`. It excludes any source `.deployed-revision` and writes a fresh stamp from the checkout SHA after extraction. `core01-deploy-local.sh` now calls that owned packager.
- **Test:** Added a real-git, real-tar fixture that plants a stale source stamp and asserts the staged artifact contains the fixture checkout SHA instead.
- **Red proof:** Mutated the packager back to carrying the source stamp and removed fresh stamping; the test failed because it received `sha=stale-local-clone` instead of the checkout SHA (`0 pass / 1 fail`).
- **Green proof:** Restored exclusion plus fresh stamping; the packaging test passed (`1 pass / 0 fail`, `5 expect()` calls).

### LOW — stamp failures relied on a catch in another module
- **Disposition:** Fixed; cheap and local.
- **Fix:** The deploy-stamp parse/read and git fallback now share the resolver try/catch, preserving the stated rule that unknown release identity never disables tracing.
- **Red proof:** Mutated stamp parsing back outside the try/catch; the regression test failed with the injected `stamp parser unavailable` exception (`0 pass / 1 fail`).
- **Green proof:** Restored the shared catch; the test passed (`1 pass / 0 fail`).

### Additional fixture hardening
- The new packaging test now strips inherited `GIT_*` hook state before creating its fixture. A hostile-env run with `GIT_DIR`, `GIT_COMMON_DIR`, `GIT_INDEX_FILE`, and `GIT_WORK_TREE` set passed, and the PR worktree HEAD remained unchanged.

### Final validation
- `bun install --frozen-lockfile` — no changes.
- `tsc --noEmit` — clean.
- Focused Bun tests — tracing `46/0`; core01 packaging `1/0`.
- Python: mypy clean; ruff clean; pytest `569 passed, 9 skipped`.
- Shell syntax and `git diff --check` — clean.
- Pre-push gate at final head — TypeScript passed; full Bun suite `3254 passed, 509 skipped, 0 failed`; all checks passed.
- Downstream rollout: not applicable; no MCP schema, transport, Python client, or agent-facing contract changed.

Pushed final SHA: `cb56da71d4b41e635a23f814101fd03f06e80677`.

PR #591, opened 2026-08-06
fix(python): accept contract-proven provider scope

### LINKED ISSUE #529: Operator-invoked provider capture fails scope proof while hook capture works

rodaddy:
## Summary

- accept a successful contract-v2 `session_start` handshake as proof for omitted agent, platform/source, server, and channel lane echoes while still rejecting conflicting returned values
- keep namespace, session key, and nullable thread response-proven
- add the exact operator JSON capture regression and document the live-verified stdin recipe

Fixes #529

## Validation

- `bunx tsc --noEmit`
- `bun test src/tools/__tests__/session-start.test.ts` — 12 passed
- `uv run mypy src/openbrain_memory` — clean
- `uv run ruff check src tests` — clean
- `uv run pytest -q` — 570 passed, 9 skipped
- live provider canary returned `status: saved`, `durable: true`, event `f5f0b9d1-a492-4c15-ae55-8f516527d44d`; direct database verification found exactly one matching `receipt` event

## Contract Parity

- Contract parity: [x] runtime-specific because: the operator JSON provider and its session_start proof validator exist only in the Python package; the TypeScript client has no matching provider runtime surface.

## Regression Failure Proof

- Temporarily removed the contract-v2 handshake fallback from `validate_started_lane`.
- `tests/test_cli_request_diagnostics.py::test_operator_capture_accepts_contract_proven_sparse_start_lane` failed with `AssertionError: assert 'lost' == 'saved'`.
- Restored the guarded logic; the same test passed.

## Critical Self-Review

- Highest-risk behavior: Treating omitted nullable lane echoes as proven could hide a server regression; the behavior is gated by the fresh reviewed contract-v2 manifest and never overrides a conflicting non-null echo.
- Assumptions that could be wrong: A successful session_start v2 call continues to mean the server established or matched agent, platform/source, server, and channel exactly as published in the contract.
- Missing/weak tests: No synthetic server violates its own v2 semantic contract while returning a sparse lane; the regression covers the observed operator shape and existing replay tests cover conflicting echoes.
- Security/permission risk: Namespace, session key, and nullable thread remain response-proven, and returned conflicts for agent, source, server, or channel still fail closed.
- Migration/deploy risk: No schema or migration change; the Python provider package must be refreshed through the normal post-merge install path before other installed copies receive the fix.
- Downstream client/runtime risk: No MCP schema changes; consumers only observe successful provider capture for the already-published session_start v2 behavior.
- Rollback/cleanup concern: Reverting the Python validator and README changes restores the prior loud lost receipt; the live canary row is an intentional receipt event and needs no cleanup.
- Fixes made before PR: Preserved conflicting-echo rejection, kept nullable thread response proof, added a public CLI-boundary regression, documented the operator recipe, and ran a live row canary.
- Known residual risk: A server falsely advertising session_start v2 could still violate the semantic promise; the fresh contract gate cannot prove implementation honesty by itself.
- SME review-memory update: [x] not applicable because: no new MEDIUM+ review pattern was found beyond the issue's already-documented session_start contract mismatch.

## Review Gate

- [x] MEDIUM+ review findings were captured: none found in the solo critical self-review; the existing exact-scope conflict regressions remain green.

Live receipt: branch provider capture saved event `f5f0b9d1-a492-4c15-ae55-8f516527d44d`, and direct database verification returned one matching row.

### DIFF
--- a/docs/memory-contract.md
+++ b/docs/memory-contract.md
@@ -52,6 +52,23 @@ known fields, authority-bearing `scope` and `config` subobjects, reflex
 `prior_context` references, and server-response projections remain strict and
 fail closed.
 
+### Operator Provider Capture
+
+Load the provider identity from the maintained environment file, then send one
+JSON request on stdin. The request carries the complete runtime scope; namespace
+and bearer identity stay in the environment rather than the request body.
+
+```bash
+set -a
+. ~/.local/share/openbrain-memory/env/claudex-observation.env
+set +a
+printf '%s\n' '{"operation":"capture","content":"<distilled fact, decision, blocker, artifact, or receipt>","distilled":true,"event_type":"fact","scope":{"agent":"claude","platform":"operator","server_id":"local-dev","channel_id":"<work lane>","session_key":"<stable session key>"}}' | openbrain-memory
+```
+
+Success returns a structured receipt with `status: "saved"` and
+`durable: true`. Any other status means the caller must treat the capture as not
+saved and preserve the fact through another authorized durable path.
+
 ### At Compaction
 1. Call `session_context` to gather current state
 2. Distill summary locally via LLM
--- a/docs/memory-contract.md
+++ b/docs/memory-contract.md
@@ -66,8 +66,10 @@ printf '%s\n' '{"operation":"capture","content":"<distilled fact, decision, bloc
 ```
 
 Success returns a structured receipt with `status: "saved"` and
-`durable: true`. Any other status means the caller must treat the capture as not
-saved and preserve the fact through another authorized durable path.
+`durable: true`. The reviewed `session_start` v2 response must echo every exact
+scope coordinate; a missing, null, or conflicting echo fails closed. Any other
+status means the caller must treat the capture as not saved and preserve the fact
+through another authorized durable path.
 
 ### At Compaction
 1. Call `session_context` to gather current state
--- a/python/openbrain-memory/README.md
+++ b/python/openbrain-memory/README.md
@@ -292,10 +292,13 @@ per-identity authentication, not the direct client's bearer token. Captured
 stdout and stderr are rejected above 1 MB; the trusted local command can still
 allocate output before `subprocess.run` returns. The adapter requires the real
 mcp2cli envelope shape, `{"success":true,"result":{...}}`, and returns only the
-inner tool result. `session_start` must prove the public lane authority fields
-it returns: namespace, session key, and agent (plus project when returned).
-`agent_context_pack` must prove every coordinate in its inner result's top-level
-`scope`, and `agent_reflex_pointers` must prove both that scope and the published
+inner tool result. A contract-v2 `session_start` success proves the supplied
+agent, platform/source, server, and channel coordinates were established or
+matched; any conflicting non-null lane echo still fails closed. The returned
+lane must independently prove namespace, session key, and thread (plus project
+when returned). `agent_context_pack` must prove every coordinate in its inner
+result's top-level `scope`, and `agent_reflex_pointers` must prove both that scope
+and the published
 `openbrain.agent_reflex_pointers.v1` envelope before its projected body-free
 result is surfaced. Every direct or fallback semantic operation first validates a
 fresh live `get_contract` result against the reviewed v23 contract, schema
--- a/python/openbrain-memory/README.md
+++ b/python/openbrain-memory/README.md
@@ -292,11 +292,11 @@ per-identity authentication, not the direct client's bearer token. Captured
 stdout and stderr are rejected above 1 MB; the trusted local command can still
 allocate output before `subprocess.run` returns. The adapter requires the real
 mcp2cli envelope shape, `{"success":true,"result":{...}}`, and returns only the
-inner tool result. A contract-v2 `session_start` success proves the supplied
-agent, platform/source, server, and channel coordinates were established or
-matched; any conflicting non-null lane echo still fails closed. The returned
-lane must independently prove namespace, session key, and thread (plus project
-when returned). `agent_context_pack` must prove every coordinate in its inner
+inner tool result. The public contract requires `session_start` v2, and every
+successful start must return a lane that proves namespace, session key, agent,
+platform/source, server, channel, and nullable thread coordinates (plus project
+when returned). Missing, null, or conflicting exact-scope echoes fail closed.
+`agent_context_pack` must prove every coordinate in its inner
 result's top-level `scope`, and `agent_reflex_pointers` must prove both that scope
 and the published
 `openbrain.agent_reflex_pointers.v1` envelope before its projected body-free
--- a/python/openbrain-memory/src/openbrain_memory/_runtime_validation.py
+++ b/python/openbrain-memory/src/openbrain_memory/_runtime_validation.py
@@ -92,6 +92,13 @@ def validate_started_lane(
         candidate["server_id"] = metadata["server_id"]
     expected = exact_scope_fields(namespace, scope)
     expected["source"] = expected.pop("platform")
+    # A contract-v2 session_start success proves these request coordinates were
+    # established or matched. Some operator responses omit their nullable lane
+    # echoes, so use the reviewed handshake proof only when no conflicting value
+    # was returned; namespace, session key, and thread remain response-proven.
+    for name in ("agent", "source", "server_id", "channel_id"):
+        if candidate.get(name) is None:
+            candidate[name] = expected[name]
     validate_exact_fields(candidate, expected, "session_start result")
 
 
--- a/python/openbrain-memory/src/openbrain_memory/_runtime_validation.py
+++ b/python/openbrain-memory/src/openbrain_memory/_runtime_validation.py
@@ -92,13 +92,6 @@ def validate_started_lane(
         candidate["server_id"] = metadata["server_id"]
     expected = exact_scope_fields(namespace, scope)
     expected["source"] = expected.pop("platform")
-    # A contract-v2 session_start success proves these request coordinates were
-    # established or matched. Some operator responses omit their nullable lane
-    # echoes, so use the reviewed handshake proof only when no conflicting value
-    # was returned; namespace, session key, and thread remain response-proven.
-    for name in ("agent", "source", "server_id", "channel_id"):
-        if candidate.get(name) is None:
-            candidate[name] = expected[name]
     validate_exact_fields(candidate, expected, "session_start result")
 
 
--- a/python/openbrain-memory/src/openbrain_memory/client.py
+++ b/python/openbrain-memory/src/openbrain_memory/client.py
@@ -79,6 +79,7 @@ FIRST_CLASS_RUNTIME_TOOL_VERSIONS: Mapping[str, int] = {
 }
 FIRST_CLASS_RUNTIME_TOOLS = tuple(FIRST_CLASS_RUNTIME_TOOL_VERSIONS)
 REQUIRED_CONTRACT_TOOL_VERSIONS: Mapping[str, int] = {
+    "session_start": FIRST_CLASS_RUNTIME_TOOL_VERSIONS["session_start"],
     "agent_context_pack": FIRST_CLASS_RUNTIME_TOOL_VERSIONS["agent_context_pack"],
     "agent_reflex_pointers": FIRST_CLASS_RUNTIME_TOOL_VERSIONS["agent_reflex_pointers"],
     "append_session_event": FIRST_CLASS_RUNTIME_TOOL_VERSIONS["append_session_event"],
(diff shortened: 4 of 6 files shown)


rodaddy:

**HIGH** — The fix makes agent/source/server_id/channel_id self-proving: it copies the expected values (derived from the request) into the response candidate, then asserts the candidate equals expected. For those four fields the assertion can no longer fail on a missing echo, so `validate_started_lane` stops being a response proof and becomes a request echo. This is the vacuous-pass class the adversarial SME lane catalogues, and it directly relaxes the rule gotcha-agent.md:416 states for this exact file ("preserve exact nullable scope coordinates").
- Location: python/openbrain-memory/src/openbrain_memory/_runtime_validation.py:95-101 (validate_started_lane)
- Failure scenario: A server (or an intermediary/older build) returns a lane for a DIFFERENT scope but simply omits the agent/source/channel_id keys — e.g. `{"lane": {"namespace":"bilby","session_key":"sk-1","thread_id":null}}`. Before this PR the capture failed closed with "did not prove exact Open Brain scope: agent, channel_id, source". After this PR the client fills all three from its own request and reports `status: saved, durable: true`, so the operator gets a durable-write receipt for a lane whose real scope was never proven. I ran this exact case against the branch: it is ACCEPTED with no error.
- Suggested fix: Do not synthesize proof from the request. Either (a) require the echo and fix the actual producer if a real one omits it, or (b) if the handshake genuinely carries the proof, verify the proof-bearing signal itself — assert the response is the exact-scope lane (e.g. presence of the lane id + the server's exact-scope success marker) rather than assuming omission means agreement. At minimum, restrict the fallback to keys that are ABSENT from the mapping and reject explicit nulls (see the separate null finding).

**HIGH** — The PR's premise — "some operator responses omit their nullable lane echoes" — is not reachable from a correct contract-v2 server on this code path, so the new regression test asserts a shape the real system does not produce. The fix therefore relaxes the validator to match a hypothesized broken response rather than proving the contract, and the true root cause of #529 is left undiagnosed.
- Location: python/openbrain-memory/tests/test_cli_request_diagnostics.py:43-66 (OperatorCaptureClient) vs src/tools/session-start.ts:34-89
- Failure scenario: Traced end to end: agent.py:271-281 always sets `agent` on the session_start payload and runtime.py:281-290 (`start_metadata`) always sets platform/server_id/channel_id. So `hasCompleteExactScope` (session-start.ts:23-32) is ALWAYS true for provider traffic, which routes every call into `establishExactStartScope`. That query does COALESCE(agent,$3)/COALESCE(source,$4)/COALESCE(channel_id,$6)/jsonb_set(server_id) and `RETURNING LANE_COLUMNS` — the returned row is populated by construction; on mismatch it returns zero rows and the server answers with the isError text "Existing lane exact scope does not match session_start request" (session-start.ts:215-224). There is no branch that returns a lane with a null agent/source/channel_id to a full-scope request. The pre-existing fake StartThenFailClient (tests/_runtime_fakes.py:442-456) echoes the full lane precisely because that is the real sha
- Suggested fix: Capture the ACTUAL failing session_start response from the live #529 reproduction (the raw JSON result, redacted) and drive the regression from that recorded shape. If it turns out to be a populated-but-mismatched lane, the current fix does not address #529 at all; if it is genuinely sparse, identify which server version/path emits it and fix or version-gate there. Add a test that a sparse lane from a server NOT advertising session_start v2 still fails closed.

**MEDIUM** — `candidate.get(name) is None` conflates an ABSENT key with an explicit JSON `null`, so a server that affirmatively reports "this lane has no agent/channel/server" is treated as having proven the caller's values. This is the absent-vs-null family the gotcha-agent SME lane records at line 1063 (the `??` env-read entry): the guard's own comment says "only when no conflicting value was returned", but an explicit null IS a returned value and it conflicts.
- Location: python/openbrain-memory/src/openbrain_memory/_runtime_validation.py:99-101
- Failure scenario: Verified by direct probe against the branch. Case A: lane `{"namespace":"bilby","session_key":"sk-1","agent":null,"source":null,"channel_id":null,"thread_id":null}` — an unscoped lane the server is explicitly describing as having no identity — is ACCEPTED with no error, so the client claims the scope was proven. Case C: `{"metadata":{"server_id":null}}` is likewise ACCEPTED. This matters because the non-exact-scope INSERT path (session-start.ts:264-286) writes `args.agent ?? null` and can persist genuinely null coordinates, and `candidate` only picks up keys via `if name in lane`, so both spellings reach this branch. The intended narrow case (key absent) and the dangerous case (key present and null) are indistinguishable to the current check.
- Suggested fix: If the fallback is kept at all, gate it on absence only: `if name not in candidate:` rather than `if candidate.get(name) is None:`. Then add a test asserting an explicit-null echo still FAILS closed, which is the only assertion that distinguishes the two cases (asserting status==saved passes under both).

**MEDIUM** — The PR body and README justify the relaxation with "a contract-v2 session_start success proves...", but `validate_started_lane` never checks any contract or version — and the package's enforced-version table `REQUIRED_CONTRACT_TOOL_VERSIONS` omits `session_start` entirely. The stated gate is narrower than advertised, so the README now documents a guarantee the code does not make at that boundary.
- Location: python/openbrain-memory/src/openbrain_memory/client.py:81-85 and README.md:292-302
- Failure scenario: `REQUIRED_CONTRACT_TOOL_VERSIONS` (client.py:81-85) lists only agent_context_pack, agent_reflex_pointers, and append_session_event — `session_start: 2` exists in FIRST_CLASS_RUNTIME_TOOL_VERSIONS but is deliberately not re-exported into the REQUIRED map. `contract.py:108` feeds REQUIRED_CONTRACT_TOOL_VERSIONS into `validate_contract_manifest`, so any consumer validating through that public entry point enforces NO session_start version floor while the new code silently assumes a v2 semantic promise. (The router's own `_validate_first_class_contract` at _runtime_router.py:374-385 does pass FIRST_CLASS_RUNTIME_TOOL_VERSIONS, so the in-router path is gated — but the guard lives in a shared validation helper with no coupling to either, and rg confirms zero occurrences of "contract" in _runtime_validation.py's validate_started_lane region.) A future caller reusing validate_started_lane outside
- Suggested fix: Make the dependency explicit rather than implied by comment: pass the verified contract/tool version into validate_started_lane (or have it take a `handshake_proven: bool` the router sets only after _validate_first_class_contract succeeds), and add `session_start` to REQUIRED_CONTRACT_TOOL_VERSIONS so the advertised v2 floor is actually enforced at the public contract boundary the README points at.

**LOW** — The new regression's positive assertions cannot distinguish the fixed behavior from a validator that skips scope proof entirely, because it only asserts the outbound append arguments (which the client controls) and status==saved. No test in the PR asserts that a sparse lane from a NON-v2 server still fails closed.
- Location: python/openbrain-memory/tests/test_cli_request_diagnostics.py:78-101
- Failure scenario: The test asserts `client.appended` carries the full scope — but those values come from the caller's own _FULL_SCOPE and would be identical if validate_started_lane were deleted outright. I confirmed the complementary case is untested: deleting the whole exact-scope assertion would still leave this test green as long as capture succeeds. The PR's own self-review concedes this ("No synthetic server violates its own v2 semantic contract while returning a sparse lane"), which is precisely the negative case that separates "contract-proven" from "unchecked".
- Suggested fix: Add the negative twin: a client whose get_contract manifest omits or under-versions session_start, returning the same sparse lane, must still produce status "lost" with the scope-proof error. That is the assertion that makes the new behavior contract-conditional rather than unconditional.

**Independent failure-proof re-verification:** Re-verified the author's claimed failure-proof independently in a detached worktree at origin/fix/529-provider-scope-proof (4c2a64e). Baseline: tests/test_cli_request_diagnostics.py::test_operator_capture_accepts_contract_proven_sparse_start_lane PASSED (0.07s). I then deleted the exact 3 guarded lines (the `for name in ("agent","source","server_id","channel_id"): if candidate.get(name) is None: candidate[name] = expected[name]` loop) from _runtime_validation.py and re-ran: FAILED with `AssertionError: assert 'lost' == 'saved'` — byte-identical to the PR body's claim. Restored the 3 lines textually (a git-checkout guard blocked the discard path); `git status --short` came back empty and the 

**Reviewer-run receipts:** All run by me in /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/rev591 (detached at PR head 4c2a64e), UV_CACHE_DIR=/Volumes/ThunderBolt/_tmp/open-brain/_validation-runs/uv-cache. (1) uv run pytest -q => 570 passed, 9 skipped in 6.96s — matches the PR's claim. (2) uv run mypy src/openbrain_memory => "Success: no issues found in 17 source files". (3) uv run ruff check src tests => "All checks passed!". (4) Target regression test alone => 1 passed. (5) Mutation run => 1 failed, `assert 'lost' == 'saved'`. (6) pytest -q -rs skip audit: the 9 skips are 4x test_cli_compatibility.py (needs OPENBRAIN_RUN_0_1_17_CLI_COMPAT=1) and 5x test_live_canary.py (needs OPENBRAIN_LIVE_CANARY=1 + credentials) —

rodaddy:
## Fix-lane response to opposite-family review

Pushed `565a677` to `fix/529-provider-scope-proof`.

### Finding → fix → proof

1. **HIGH — request values were being synthesized as response proof**
   - Fix: removed the request-to-candidate synthesis from `validate_started_lane`; every exact-scope coordinate must now be present in the returned lane and match.
   - Red: the new sparse-v2 case reported `saved` instead of `lost` before the fix.
   - Green: `test_operator_capture_rejects_unproven_start_scope[SparseStartClient]` passes and confirms no append occurs.

2. **HIGH — the sparse positive fixture did not match the contract-v2 producer**
   - Fix: removed the positive sparse-response premise and replaced it with fail-closed regressions. A direct dogfood probe of the current `session_start` producer returned lane keys including `agent`, `source`, `channel_id`, `thread_id`, and `metadata.server_id`; no request or credential values were printed.
   - Red: the former branch behavior accepted the synthetic sparse lane.
   - Green: the replacement regression rejects that shape while the existing complete-lane producer/runtime tests remain green.

3. **MEDIUM — absent and explicit JSON null were conflated**
   - Fix: removed the fallback entirely, so neither absence nor explicit null can be converted into caller-supplied proof.
   - Red: the explicit-null case reported `saved` before the fix.
   - Green: `test_operator_capture_rejects_unproven_start_scope[NullScopeStartClient]` now reports `lost`, `durable=false`, and no append.

4. **MEDIUM — public contract validation did not enforce `session_start` v2**
   - Fix: added `session_start: 2` to `REQUIRED_CONTRACT_TOOL_VERSIONS` and aligned README guidance with the actual gate.
   - Red: `test_required_contract_pins_session_start_v2` failed with `KeyError: session_start` before the fix.
   - Green: the test now proves both capability and tool-contract versions below v2 are rejected.

5. **LOW — no negative twin for a pre-v2 sparse server**
   - Fix: added `test_operator_capture_rejects_sparse_lane_before_v2_dispatch`.
   - Green: the request fails at contract discovery, produces a non-durable `lost` receipt, and never appends.

### Red/green receipt

- Before implementation: 3 failures — sparse response accepted, explicit-null response accepted, and public contract map missing `session_start`.
- After implementation: focused regression set `4 passed`; affected Python files `45 passed`.
- `bun test src/tools/__tests__/session-start.test.ts` — 12 passed.
- `bun contracts/check-parity.ts` — passed; Python fixture subset 14 passed in pre-push.
- `uv run mypy src/openbrain_memory` — clean.
- `uv run ruff check src tests` — clean.
- `uv run pytest -q` — 573 passed, 9 skipped.
- Pre-push gate — 3250 Bun passed, 509 skipped, 0 failed; all Python and contract gates passed.

Downstream disposition: no MCP schema or server behavior changed. The Python package refresh remains a post-merge rollout step; this fix lane did not merge or touch the trunk.

Issue #569, opened 2026-08-05
Trace collection completeness for retro-forensics: exchange-level is captured; decide coverage, masking-then-widen, and whether any lane needs wire-level


rodaddy:
Rico, 2026-08-05: 'shouldn't we be getting all of the calls and how they worked, what they called, what they got... so we can take a bunch of traces and be like, this is where that got all fucked up?'

**Measured state (CT 273, audited in #561):** exchange-level capture IS full for instrumented surfaces -- tool-call inputs/outputs verbatim (langfuse-tracing.ts:341-342), assistant turns as generations, 33,622 content-ful observations in 2 days. Missing per-trace metadata (model/usage/cost/release) is #560. Masking prerequisite is #561.

**The actual decisions this ticket owns:**
1. **Coverage:** emitters are env-gated; only OB capture + sf watcher emit today. Enumerate agent surfaces (mini sessions, Air, cc-* boxes, runner lanes, Bilby/OpenClaw?) and decide which get the vars. Sequence AFTER #561 masking lands -- widening verbatim capture first widens the leak.
2. **Wire-level or not:** hooks structurally cannot capture Claude's system prompt / exact API request (Anthropic does not expose it). Exchange-level is likely sufficient for 'where did it go wrong' forensics. IF wire-level is ever wanted for specific lanes, the shape is a logging proxy in the request path (LiteLLM CT204 has native Langfuse export) -- explicitly a separate decision for proxied lanes only; do not put a proxy in front of native subscription Claude without its own review.
3. **langfuse-trace skill** (from the 2026-08-04 gap analysis, still untracked until now): fetch trace by id / latest session for an agent / diff two traces / dump per-trace tool timeline, so forensics is one prompt instead of SSH+ClickHouse SQL. This is the consumer that makes the collection worth it.

Depends on: #560 (metadata), #561 (masking). State: measured facts RUNNING-verified; everything proposed here PROPOSED.

rodaddy:
**Operator ruling (Rico, 2026-08-05): full coverage.** "Everything that can possibly be dropped into there should be dropped into there — no exceptions." This closes the coverage question this issue was holding open: the target is every surface traced — retrieval internals (candidates/scores/chosen), namespace value, structured row-id cross-links, embedding calls, dream/distill jobs, the NATS worker, and server-side LLM calls as generations. Sequencing stands as this issue already decided: #561 emitter-side masking lands FIRST, then capture widens. Implementation proceeds under epic #571.

[...]

PR #593, opened 2026-08-06
fix(observability): mask Langfuse trace payloads

### LINKED ISSUE #561: Audit content-ful Langfuse traces on CT 273 for secrets before tracing widens (PR #534 sends full tool inputs/outputs)

rodaddy:
## Summary

- Apply masking at the Langfuse emitter boundary in `server/observability/langfuse-tracing.ts:419`, using the shared labeled detectors from `src/secret-patterns.ts:80`.
- Replace only matched string spans with `[MASKED:<detector>]`; retain surrounding text, object fields, array items, and complete tool input/output structure.
- Add default-on `OPENBRAIN_TRACING_MASKING_ENABLED`; only the explicit value `"0"` bypasses masking.
- Document the configuration and capture the reusable security pattern in `docs/sme/security.md`.

Fixes #561

## Verification

- [x] Relevant Open Brain tests/typecheck/migrations passed — `bun test server/observability/langfuse-tracing.test.ts` (46 pass, 0 fail) and `bunx tsc --noEmit` (exit 0); no migration changed.
- [x] Python package checks passed or are not applicable — not applicable; no Python package files or behavior changed.
- [x] Live Open Brain smoke passed or is not applicable — not applicable before merge/deploy; the changed boundary is directly exercised through the exported trace-body builder and injected sink tests.
- Deliberate-break proof: changed masking default logic so non-zero text disabled it; `readMcpTracingConfig > masking is disabled only by the explicit zero value` failed (`Expected: true`, `Received: false`). Restored and green.
- Deliberate-break proof: bypassed `maskTraceValue`; `trace body helpers > masks detector matches in tool arguments and results while retaining every field` failed because the seeded labeled fixture reached the builder output. Restored and green.
- Deliberate-break proof: inverted the explicit bypass branch; `trace body helpers > the explicit masking opt-out retains the original trace body` failed because the fixture was masked. Restored and green.

## Critical Self-Review

- Highest-risk behavior: A recursive transform at the egress boundary could accidentally remove structure or alter benign payload identity; the implementation uses replacement-only traversal and returns unchanged branches by identity when no detector fires.
- Assumptions that could be wrong: MCP tool arguments and results reaching this builder are JSON-compatible arrays, objects, primitives, and strings with outbound content represented by enumerable string-keyed fields.
- Missing/weak tests: No live Langfuse deployment smoke is included on this unmerged branch; focused tests prove builder output, explicit configuration behavior, and unchanged tool-call behavior through the injected sink seam.
- Security/permission risk: The explicit `OPENBRAIN_TRACING_MASKING_ENABLED=0` operator bypass permits detector matches to leave unmasked; default and every other value keep masking enabled.
- Migration/deploy risk: No database migration exists; the environment is read at process start, so changing the masking setting requires the same service restart as the existing tracing variables.
- Downstream client/runtime risk: Not applicable to MCP schemas, transport, Python exports, generated skills, mcp2cli, or Hermes call shapes; this changes only the server-owned observability copy.
- Rollback/cleanup concern: Revert the commit to restore prior behavior; no persisted local artifact, schema change, or cleanup job is introduced.
- Fixes made before PR: Preserved benign input/output identity after self-review found that an eager deep copy would break the existing same-object trace assertion even when no detector matched.
- Known residual risk: Pattern-based detectors cannot identify an arbitrary unlabelled secret with no recognizable shape; the emitter now consistently applies the repo's shared detector set rather than creating a separate heuristic.
- SME review-memory update: [x] docs/sme/ updated — security.md masking pattern

## Review Gate

- Live Open Brain checks: [x] not applicable because: this branch changes the pre-enqueue builder and is not deployed; local boundary tests are the applicable pre-PR evidence.
- Review tier: Full — secret-handling behavior at an outbound observability boundary.
- Independent review status: pending at the PR boundary; this run was explicitly required to use one process with no nested workers.

## Contract Parity

- Contract parity: [x] runtime-specific because: no MCP tool schema, contract manifest, fixture contract, or client call shape changes.

## Downstream Rollout

- [x] I checked `docs/downstream-rollout.md`.
- [x] rtech-mcps handoff is not applicable — no registry, schema, or service definition changes.
- [x] mcp2cli cache/skill refresh is not applicable — no consumer-visible tool contract changes.
- [x] rtech-hermes Python runtime/plugin changes are not applicable — no client or plugin behavior changes.
- [x] Hermes live rollout/canaries are not applicable — the server observability copy is not agent-facing behavior.

Notes/evidence:

- Existing design: #530 keeps complete tool-call content; #561 and the 2026-08-05 #569 ruling require masking before tracing coverage widens.
- Owning boundary: `buildToolTraceBody()` immediately before `sink.emit()` (`server/observability/langfuse-tracing.ts:419`, `server/observability/langfuse-tracing.ts:820`).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -56,7 +56,8 @@ OPENBRAIN_RAW_TURN_TTL_SECONDS=604800
 
 # Server call tracing to Langfuse (#530) -- CONTENT-FUL, off by default.
 # Every MCP tool call becomes one trace carrying full arguments and full
-# results. Requires ALL FOUR: the flag set to exactly "1" plus the three
+# results, with detector matches replaced before egress. Requires ALL FOUR:
+# the flag set to exactly "1" plus the three
 # coordinates, or the lane stays off and logs one content-free warn naming the
 # missing one. Keys come from the operator environment (Vaultwarden item
 # "Langfuse - Open Brain Local Dogfood") -- NEVER put values here.
@@ -68,6 +69,8 @@ OPENBRAIN_RAW_TURN_TTL_SECONDS=604800
 # drop count on the way back, never one per call.
 # See docs/CONFIG_REFERENCE.md "Server call tracing (#530)".
 OPENBRAIN_TRACING_ENABLED=0
+# Default is on. Set exactly "0" only for an explicit operator-approved bypass.
+OPENBRAIN_TRACING_MASKING_ENABLED=1
 OPENBRAIN_TRACING_ENDPOINT=
 OPENBRAIN_TRACING_PUBLIC_KEY=
 OPENBRAIN_TRACING_SECRET_KEY=
--- a/docs/CONFIG_REFERENCE.md
+++ b/docs/CONFIG_REFERENCE.md
@@ -286,14 +286,18 @@ tool name, caller identity, full arguments, full result (or error class and
 message), duration, and session grouping.
 
 **CONTENT-FUL is the point, and it is deliberate.** #530 explicitly supersedes
-#372's content-free spec for the local dogfood deployment. Payloads are sent
-verbatim: no redaction, no summarisation, no size bucketing. `OPENBRAIN_MCP_AUDIT_*`
-(above) remains the separate content-FREE durable Postgres record and is
-unchanged — the two lanes coexist and neither replaces the other.
+#372's content-free spec for the local dogfood deployment. #561 adds the required
+emitter-boundary protection before coverage widens: every string value passes
+`src/secret-patterns.ts` detectors, and each matched span becomes
+`[MASKED:<detector>]`. Surrounding content, object fields, and array items remain
+present. `OPENBRAIN_MCP_AUDIT_*` (above) remains the separate content-FREE durable
+Postgres record and is unchanged — the two lanes coexist and neither replaces
+the other.
 
 | variable | default | notes |
 |---|---|---|
 | `OPENBRAIN_TRACING_ENABLED` | unset (off) | tracing runs only when this is exactly `"1"` |
+| `OPENBRAIN_TRACING_MASKING_ENABLED` | unset (on) | detector-based masking is disabled only when this is exactly `"0"`; explicit operator bypass only |
 | `OPENBRAIN_TRACING_ENDPOINT` | — | Langfuse base URL (the SDK appends its own paths) |
 | `OPENBRAIN_TRACING_PUBLIC_KEY` | — | Langfuse `pk-lf-...` |
 | `OPENBRAIN_TRACING_SECRET_KEY` | — | Langfuse `sk-lf-...` |
--- a/docs/sme/security.md
+++ b/docs/sme/security.md
@@ -1042,3 +1042,51 @@ When a strict scope/tenant proof gates dispatch, any replay or drain loop must d
 Verbatim, from the source:
 
 > A unit parked under project A ... therefore fails dispatch with `session_start result did not prove exact Open Brain scope` when the drain is triggered by a healthy operation in project B ... Units can sit undelivered indefinitely if the client rarely operates in that project again. Every drain attempt re-dispatches mismatched units, issuing a wasted live `session_start` per unit per healthy operation, plus a swallowed warning.
+
+## [2026-08-05] Content-ful observability needs masking at the final emitter boundary
+
+**Severity:** HIGH
+**Source:** issue #561, operator ruling after the Langfuse egress census
+**Scope key:** `sme.security.contentful_observability_masks_at_emitter`
+**Status:** active
+
+### Pattern
+
+A content-ful observability lane can preserve complete arguments, results, and
+surrounding context while still preventing credential-shaped spans from leaving
+the process. Apply the shared labeled detectors at the final emitter boundary,
+not at individual callers and not after transport. Replace each matched span
+with a stable classifier marker such as `[MASKED:github_token]`; retain every
+field, array item, and unmatched substring. Keep masking on by default, with any
+bypass requiring an explicit operator configuration value.
+
+Issue #561 measured the failure mode directly: PR #534 intentionally attached
+full tool inputs and outputs to Langfuse, while the live corpus had not exercised
+the highest-risk credential-return path. The fix belongs in
+`buildToolTraceBody()` so future tracing surfaces inherit the same protection
+without maintaining caller-specific detector forks.
+
+PR #593's opposite-family review proved three additional bypass shapes at this
+same boundary. Value-shaped detectors do not see opaque values carried under a
+sensitive object key. Replacing an entire labeled JSON pair removes the label
+and can make serialized JSON invalid. Non-plain carriers need normalization:
+Map and Set content otherwise disappears, nested Error messages disappear, and
+binary views expose credential bytes as numeric arrays.
+
+### Review Questions
+
+- Does every string value in tool arguments and results pass the shared
+  `src/secret-patterns.ts` detectors before the sink receives it?
+- Does the recursive walk also apply `isSensitiveKey()` while the field name is
+  available, including nested argument and result objects?
+- Do pair-shaped detector replacements retain the label and leave serialized
+  JSON parseable?
+- Are Map, Set, Error, Buffer, and typed-array carriers converted to observable,
+  maskable shapes without exposing binary bytes?
+- Are only matched spans replaced, with surrounding content and structure still
+  present?
+- Is masking on when its environment variable is absent, and can it be bypassed
+  only through the documented explicit value?
+- Does a regression seed obviously fake detector-shaped text in both tool
+  arguments and tool results and assert the emitted body rather than a private
+  helper?
--- a/server/observability/langfuse-tracing.test.ts
+++ b/server/observability/langfuse-tracing.test.ts
@@ -2,8 +2,9 @@
  * Functional tests for the #530 content-ful tracing lane.
  *
  * The contract under test is the one the module states: every tool call emits
- * one trace carrying VERBATIM input and output plus caller identity, and a
- * tracing failure never fails, slows, or ALTERS a tool call. The second half is
+ * one trace retaining complete input and output around masked detector matches,
+ * plus caller identity; a tracing failure never fails, slows, or ALTERS a tool
+ * call. The second half is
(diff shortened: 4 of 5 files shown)


rodaddy:

**HIGH** — maskTraceValue applies only the value-shaped detectors and never isSensitiveKey, so a structured tool argument or result carrying an opaque credential under a sensitive KEY (password, api_key, token, totp_secret) reaches Langfuse in clear. This is the exact gap the repo already found and fixed in src/logger.ts, reintroduced at the new emitter boundary — and issue #561 names Vaultwarden output as the highest-risk target it wanted covered.
- Location: server/observability/langfuse-tracing.ts:396-414 (maskTraceValue); contrast src/logger.ts:429 and the rationale in src/secret-patterns.ts:141-159
- Failure scenario: CONFIRMED by running the PR's own exported builder. `buildToolTraceBody({toolName:'t',status:'success',durationMs:1,args:{password:'obviously-fake-561-fixture-value-aaaaaaaaaaaa'},output:{}})` returns `input = {"password":"obviously-fake-561-fixture-value-aaaaaaaaaaaa"}` — completely unmasked, with masking DEFAULT-ON. Same for `{api_key: <opaque>}` and for `{creds:{token: <opaque>, authorization:'Bearer <opaque>'}}`, where only the sibling `authorization` masked (because 'Bearer ' is a value shape) while `token` went out in clear on the same object. body.input/body.output are passed straight to startObservation() at line 986-993 with no downstream redaction, so these values leave the process. secret-patterns.ts:148-151 documents this precise failure verbatim: '{"password":"hunter2driveway"} emitted in clear ... The pattern was never wrong; it was never shown the key.' A recursive walk ov
- Suggested fix: Thread the key into the recursion and apply the repo's existing key-based check alongside the value-based one, mirroring src/logger.ts:429. In the object branch of maskTraceValue, before recursing: `if (isSensitiveKey(key) && child !== null && child !== undefined) { masked[key] = '[MASKED:sensitive_key]'; continue; }` — importing isSensitiveKey from src/secret-patterns.ts (already exported, zero new dependency). Add a regression test seeding an opaque fixture under `password` and `api_key` in BOTH args and output, asserting the emitted body rather than a private helper, so the SME review quest

**MEDIUM** — The json_labeled_secret detector matches the whole "key":"value" pair, so masking DELETES the field name as well as the value — reduction, not replacement. This breaks the PR's own stated invariant ('Masking is replacement, never field removal') and corrupts the surrounding JSON text so the operator can no longer parse it.
- Location: server/observability/langfuse-tracing.ts:381-393 (maskTraceString) interacting with the json_labeled_secret detector at src/secret-patterns.ts:121-124
- Failure scenario: CONFIRMED by execution. Input `{text: '{"a":1,"password":"X1abc","b":2}'}` emits `{text: '{"a":1,[MASKED:json_labeled_secret],"b":2}'}`. The key name `"password"` is gone, so the operator cannot tell WHICH field was masked — the diagnostic value the content-ful lane exists for. Worse, `JSON.parse()` on the emitted text now throws `JSON Parse error: Property name must be a string literal`, so any tool result that serialized JSON into a text block becomes unparseable in Langfuse. This is a real shape: the Vaultwarden-style probe `{content:[{type:'text',text:JSON.stringify({username,password,totp_secret})}]}` produced exactly this corrupted, unparseable output. The file header at line 5 and docs/CONFIG_REFERENCE.md both promise 'Surrounding content, object fields, and array items remain present', which this path does not honor.
- Suggested fix: Make the replacement preserve the label for pair-matching detectors: use a capture-group-aware replacement so `"password":"X1abc"` becomes `"password":"[MASKED:json_labeled_secret]"` rather than losing the key. The json_labeled_secret and labeled_secret patterns already capture the label in group 1, so `masked.replace(re, (m, label) => ...)` can rebuild `"<label>":"[MASKED:kind]"` and keep the text valid JSON. Add a test asserting the emitted text still round-trips through JSON.parse and still names the masked field.

**MEDIUM** — maskTraceValue treats any non-plain object as an opaque leaf, so Map, Set, and Error carriers are serialized to `{}` by the SDK (silent data loss, and the PR's 'retain every field' claim fails), while a Buffer's credential bytes pass through fully readable as a numeric array that no detector can see.
- Location: server/observability/langfuse-tracing.ts:396-414 (maskTraceValue: Object.entries on non-plain objects)
- Failure scenario: CONFIRMED by execution. `{m: new Map([['password', <opaque>]])}` and `{s: new Set([<opaque>])}` both emit `{}` — the operator silently loses the content the content-ful lane exists to show, which is the opposite of the stated 'retain every field, array item' contract. `{e: new Error('password=<opaque>')}` emits `{}`, dropping the error message entirely (note the exception path at line 719 uses errorOutput() which extracts .message to a string, so that path is fine — this is about an Error nested INSIDE a tool result). Most seriously, `{b: Buffer.from('password=<opaque>')}` emits `{"type":"Buffer","data":[112,97,115,115,119,111,114,100,61,...]}` — the full credential in byte form, trivially readable, with every detector bypassed because Object.entries on a Buffer yields numeric indices, not the string. The PR body's assumption 'MCP tool arguments and results reaching this builder are JSON
- Suggested fix: Normalize before masking rather than assuming JSON-compatibility: handle Buffer/TypedArray explicitly (replace with a `{type,byteLength}` shape marker, never the bytes), convert Map/Set to a plain object/array so their contents are both retained AND masked, and unwrap Error to {name, message} so the message passes maskTraceString. Add a test for each carrier asserting the emitted body contains no fixture bytes and no silently-emptied object.

**Independent failure-proof re-verification:** Re-verified the PR's first claimed deliberate-break proof independently, in a fresh scratch worktree of origin/fix/561-emitter-masking (effdb7b) at /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/rev593. Baseline: 46 pass / 0 fail. Mutated server/observability/langfuse-tracing.ts:323 from `env.OPENBRAIN_TRACING_MASKING_ENABLED !== "0"` to `(env.OPENBRAIN_TRACING_MASKING_ENABLED ?? "1") === "1"` (the ?? / truthiness class the SME docs flag). Test `readMcpTracingConfig > masking is disabled only by the explicit zero value` went RED with exactly the claimed output: `Expected: true`, `Received: false` (45 pass / 1 fail). Restored the line; back to 46 pass / 0 fail, `git status` clean. The claime

**Reviewer-run receipts:** All commands run by me in the rev593 worktree (bun 1.3.13, BUN_TMPDIR=/Volumes/ThunderBolt/_tmp/open-brain/_scratch/bun-tmp). (1) `bun install --frozen-lockfile` -> 169 packages, clean. (2) `bun test server/observability/langfuse-tracing.test.ts` -> 46 pass, 0 fail, 102 expect() calls, 240ms — matches the PR's claim. (3) `bunx tsc --noEmit` -> exit 0. (4) Mutation run -> 45 pass, 1 fail (named test, expected message). (5) Restore run -> 46 pass, 0 fail, tree clean. (6) Four adversarial probe scripts against the real exported buildToolTraceBody, all with fabricated fixture values (no real credentials). Probe results reaching the sink UNMASKED: {password: <opaque>} , {api_key: <opaque>}, {cred

rodaddy:
## Fix-lane receipt

Pushed commit `9ebe01a4f921c0fb5e5614920713a6f953c574ee`.

### HIGH — opaque credentials under sensitive keys

- **Fix:** `maskTraceValue()` now reuses `isSensitiveKey()` while each object key is still available. The trace builder applies this transform to tool arguments and results, so trace-owned identity metadata remains intact.
- **Test:** Added a public-boundary `buildToolTraceBody()` regression with opaque fixture values under `password` and `api_key` in both arguments and results, plus nested `totp_secret`.
- **Red proof:** Before the implementation change, the new assertion failed because the opaque values were returned unchanged.
- **Green proof:** Focused suite now passes with every seeded field equal to `[MASKED:sensitive_key]`.

### MEDIUM — labeled JSON became invalid after masking

- **Fix:** Pair-shaped detector replacements now retain their labels. Serialized JSON keeps its original key and receives a masked string value; plain labeled text keeps its label and separator.
- **Test:** Added a regression that sends serialized JSON through `buildToolTraceBody()`, parses the emitted text with `JSON.parse()`, and asserts the masked `password` field remains present. Strengthened the existing labeled-text expectations to retain `password=`.
- **Red proof:** Before the implementation change, the new test threw `JSON Parse error: Property name must be a string literal`.
- **Green proof:** The emitted JSON now parses and equals `{ a: 1, password: "[MASKED:json_labeled_secret]", b: 2 }`.

### MEDIUM — non-plain carriers lost content or exposed bytes

- **Fix:** The recursive boundary now converts Map to a plain object, Set to an array, Error to `{name,message}`, and Buffer/typed-array views to `{type,byteLength}` markers before masking.
- **Test:** Added one boundary regression covering Map, Set, nested Error, Buffer, and `Uint8Array`; it asserts retained observable content, masked sensitive values, byte-safe markers, and absence of the fixture value from the emitted body.
- **Red proof:** Before the implementation change, the Map and Set assertions received their original non-plain carriers instead of masked observable shapes.
- **Green proof:** The focused suite confirms all carrier shapes and masking outcomes.

### Validation

- `bun test server/observability/langfuse-tracing.test.ts` — **49 pass, 0 fail, 108 assertions**.
- `bunx tsc --noEmit` — **exit 0**.
- Pre-push repository gate — **3,256 pass, 509 skip, 0 fail across 236 files**; TypeScript passed.
- Python checks — **not applicable**; no Python package files or behavior changed, and the pre-push gate reported both Python packages unchanged.
- `docs/sme/security.md` records the three review patterns for future security lanes.
- The review comment contained no LOW findings requiring disposition.

PR #594, opened 2026-08-06
fix(canon): verify reconcile apply provenance and namespace

### LINKED ISSUE #588: canon reconcile --apply: repo-facts path/source_url can never validate, and namespace fan-out silently writes to the token's namespace
First live run of the #538 post-merge apply (`openbrain-canon-reconcile --apply`, 2026-08-05, RUNNING-verified against the local dogfood) found two defects. Rico's own namespace applied clean: 10 profile + 27 process rules landed in `rico` (verified by DB join on `ob_session_lanes.namespace`).

## Defect 1 — repo-facts apply is structurally impossible

`_plan_repo_fact` (`python/openbrain/src/openbrain/apps/canon/writes.py:271`) sends `path: entry.source or entry.key`. The pack's `source` fields are provenance prose (`"open-brain PR #84, .github/workflows/ci.yml; approved 2026-08-03"`), and the server's `sourceUrlMatchesSource` (`src/tools/repo-facts.ts:96-127`) requires `source_url` to be `github.com/<owner>/<repo>/blob/<commit>/<path>` with `<path>` equal to that metadata `path` **exactly**. Prose can never equal a URL path segment, so every fact fails validation:

```
MCP error -32602: Input validation error ... "source_url must include source_commit and source path"
```

All 6 facts failed identically. The PR's dry-run receipt only exercised "refuses to plan without provenance" — the apply path could not be tested pre-merge because it needs the merged commit, which is exactly where this surfaced.

Fix shape (PROPOSED, ~S): provenance owns the path. `FactProvenance` gains the pack file's repo-relative path (the entrypoint already knows `args.pack`; the pack file IS the artifact verified at the commit), `_plan_repo_fact` uses it for `metadata.path`, and the entry's prose `source` rides in the reason/provenance field it was written for. Regression test: a pack entry whose `source` is prose applies against the validator's exact URL rule.

## Defect 2 — per-namespace fan-out silently writes to the token's namespace

The process pack declares fan-out to rico, skippy, nagatha, bilby. Per-user tokens are not provisioned consumers on the service (`AUTH_TOKEN_USER_*` in the env answer 401 "Invalid token"; the auth set is admin/agent/discord/ob-admin/promoter/readonly). Using `AUTH_TOKEN_ADMIN` + `OPENBRAIN_CANON_AGENT=skippy` (which becomes `OpenBrainClient(namespace=...)`):

- "skippy" apply reported `applied 27 write(s)` — but all 27 landed in namespace **admin** (DB-verified: `admin|decision|27`).
- nagatha and bilby then reported `no writes needed` — their reads ALSO resolved to the admin namespace, which now matched the pack.

So the client's namespace parameter does not delegate on this path; the write namespace comes from the token, and the reconciler has no way to know. Two sub-problems:
1. The reconciler reports success against the wrong namespace — it never verifies which namespace the write landed in (the receipt carries `namespace_source` but the tool does not check it).
2. There is no working operator route to reconcile canon into a namespace the operator does not hold a token for. Either per-agent consumers get provisioned, or admin-role namespace delegation must work on `append_session_event`/`agent_context_pack` and be honored end-to-end.

**Cleanup needed (operator decision):** 27 stray `decision` events in namespace `admin`, created 2026-08-05 ~22:00Z by this apply. They are the process pack verbatim, in a namespace no agent reads canon from. Relegate-by-key or operator SQL — flagging rather than touching.

**Net state:** canon packs are live for `rico` only. skippy/nagatha/bilby remain empty pending this fix. #522 stays open — apply receipts are partial.

rodaddy:
Fixes #588

## Summary
- Carry the canon pack's repo-relative file path in `FactProvenance` and use it as repo-fact `metadata.path`, while preserving entry source prose in `refresh_hint`.
- Verify every apply receipt against `settings.agent`; token and delegated-header receipts resolve through trusted writer provenance, and repo-fact receipts use their returned namespace.
- Fall back to one scoped `agent_context_pack` read when a receipt has no namespace signal, checking the returned scope plus the expected keys and texts.
- Add the post-merge failure pattern to `docs/sme/gotcha-agent.md`.

## Validation
- `bunx tsc --noEmit` — PASS.
- `bun test src/tools/__tests__/repo-facts.test.ts` — PASS, 20 tests.
- `uv run mypy src/openbrain` — PASS, 49 source files.
- `uv run ruff check src tests` — PASS.
- `uv run pytest -q` — PASS, 597 passed and 19 deselected.
- `git diff --check` — PASS.

## New-Test Failure Proof
- Changed repo-fact planning back to `path = entry.source or entry.key`; `test_a_repo_fact_uses_the_pack_path_not_its_prose_source` failed because the prose source differed from `docs/canon/repo-facts.toml`. Restored the pack path; PASS.
- Changed pack-path derivation to return only the filename; `test_repo_fact_provenance_derives_the_pack_repo_relative_path` failed with `facts.toml` instead of `docs/canon/facts.toml`. Restored repo-relative derivation; PASS.
- Changed token receipt resolution to read `delegated_agent_id`; `test_apply_rejects_a_receipt_from_the_token_namespace` failed because the sparse result incorrectly reached read-back instead of rejecting namespace `admin`. Restored `token_identity`; PASS.
- Disabled sparse-receipt read-back; `test_apply_reads_back_when_the_receipt_has_no_namespace_signal` failed with `DID NOT RAISE NamespaceMismatchError`. Restored read-back; PASS.
- Final focused receipt: all four new tests passed.

## Critical Self-Review
- Highest-risk behavior: Receipt provenance must map `namespace_source=token` to `token_identity` and `namespace_source=header` to `delegated_agent_id`, or a false success could return.
- Assumptions that could be wrong: The trusted append receipt fields continue to reflect the same namespace authority used by the server, and repo-fact receipts continue returning their persisted namespace.
- Missing/weak tests: No live fan-out apply was run because it would mutate canon; coverage uses the real planned-write boundary with fake client receipts plus existing server receipt tests.
- Security/permission risk: The change grants no delegation and adds no write authority; it fails closed when the authorized destination differs from the intended agent.
- Migration/deploy risk: No database migration, server schema, transport, or deployment configuration changes.
- Downstream client/runtime risk: Not applicable because no MCP contract, `python/openbrain-memory` behavior, generated skill, or Hermes-facing surface changed.
- Rollback/cleanup concern: Reverting the commit restores the old false-success behavior; the 27 pre-existing stray `admin` events remain an operator cleanup item and are not touched here.
- Fixes made before PR: Split machine path from prose provenance, added receipt and read-back verification, simplified error construction after lint feedback, and captured the pattern in SME memory.
- Known residual risk: Fan-out still requires a token or delegated session that the server authorizes for the target agent; this PR detects missing authority but does not provision it.
- SME review-memory update: [x] docs/sme/ updated — added the #588 source-identity and destination-scope receipt pattern to `gotcha-agent.md`.

## Review Gate
- [x] MEDIUM+ review findings were captured: the #588 false-receipt pattern was added to `docs/sme/gotcha-agent.md`; no additional material finding remained after self-review.
- Live Open Brain checks: [x] not applicable because: this PR changes the local Python operator reconciler, and a live apply would mutate canon while target-agent authorization remains an operator provisioning concern.

## Downstream Rollout
Not applicable: this change does not alter MCP tool names, schemas, output contracts, transport behavior, migrations, `python/openbrain-memory`, generated skills, or Hermes-facing behavior.

### DIFF
--- a/docs/sme/gotcha-agent.md
+++ b/docs/sme/gotcha-agent.md
@@ -1144,3 +1144,37 @@ Claude Code's SessionStart additionalContext has an observed practical inline bo
 Verbatim, from the source:
 
 > `openbrain-session-start` dumped the raw `agent_context_pack` JSON envelope (~30 KB of nested items, ids, citations, confidences, warnings) into `additionalContext`, and Claude Code persisted a payload that large to a file it surfaced as only a ~2 KB preview -- the session saw 2-3 of 31 items
+
+## [2026-08-05] Apply receipts must prove both source identity and destination scope
+
+**Severity:** HIGH
+**Source:** issue #588, first live canon reconcile apply after PR #538
+**Scope key:** `canon.apply_receipt_proves_source_and_namespace`
+**Status:** active
+
+### Pattern
+
+Two false assumptions can make an apply path report success while producing no
+usable state. First, a human provenance field is not a machine source pointer:
+repo-fact `source` prose was copied into `metadata.path`, while the server
+requires that path to equal the GitHub URL's repo-relative path exactly. The
+owning entrypoint must carry the pack artifact path separately and preserve the
+human citation in a prose provenance field.
+
+Second, a client's requested identity is not evidence of the namespace the
+server authorized. An admin-token fan-out reported 27 applied writes for
+`skippy`, but the write receipts identified token authority and the rows landed
+under `admin`. An apply command must inspect the returned receipt for the actual
+namespace authority. If the receipt has no usable namespace signal, perform one
+scoped read-back and verify the expected keys and texts before printing success.
+
+### Review Questions
+
+- Does any field serve as both human citation prose and a machine-exact path,
+  identifier, or URL component? Split those meanings at the owning boundary.
+- Does a write receipt prove where the row landed, or only that the request
+  returned without an error?
+- When client configuration requests another identity, does the server honor it
+  under this token role, and does the receipt expose which authority won?
+- If a receipt cannot identify the destination, does the apply path read back
+  the exact expected keys and values before claiming success?
--- a/docs/sme/gotcha-agent.md
+++ b/docs/sme/gotcha-agent.md
@@ -1165,8 +1165,16 @@ Second, a client's requested identity is not evidence of the namespace the
 server authorized. An admin-token fan-out reported 27 applied writes for
 `skippy`, but the write receipts identified token authority and the rows landed
 under `admin`. An apply command must inspect the returned receipt for the actual
-namespace authority. If the receipt has no usable namespace signal, perform one
-scoped read-back and verify the expected keys and texts before printing success.
+namespace authority. PR #594's opposite-family review found that the server's
+`writer_identity` is that authority on both token and delegated-header paths;
+`delegated_agent_id` is an independent agent label and cannot prove destination.
+If the receipt has no usable namespace signal, perform one scoped read-back and
+verify the expected keys and texts before printing success. That read-back must
+request every planned lane and carry the exact repo binding for repo facts, or a
+configuration gap can be mislabeled as a namespace incident.
+
+A duplicate receipt is also not a new write. Count it as already present and
+report it separately, so a retry cannot turn prior state into an applied count.
 
 ### Review Questions
 
@@ -1176,5 +1184,10 @@ scoped read-back and verify the expected keys and texts before printing success.
   returned without an error?
 - When client configuration requests another identity, does the server honor it
   under this token role, and does the receipt expose which authority won?
+- Is namespace validation based on the server's persisted writer identity rather
+  than an independent agent label or the caller's requested identity?
+- Are duplicate receipts reported as already present instead of newly applied?
 - If a receipt cannot identify the destination, does the apply path read back
   the exact expected keys and values before claiming success?
+- Can that read-back observe every planned lane under the exact repo binding, or
+  will a configuration gap be reported as a namespace mismatch?
--- a/python/openbrain/src/openbrain/apps/canon/run.py
+++ b/python/openbrain/src/openbrain/apps/canon/run.py
@@ -53,24 +53,26 @@ from __future__ import annotations
 
 import argparse
 import asyncio
+from collections.abc import Mapping, Sequence
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 from loguru import logger
 
-from openbrain.apps.canon.pack import Pack
+from openbrain.apps.canon.pack import Lane, Pack
 from openbrain.apps.canon.reconcile import (
     DriftReport,
     diff_pack,
     format_report,
+    live_items,
+    live_key_of,
+    live_text_of,
 )
 from openbrain.apps.canon.writes import FactProvenance, PlannedWrite, plan_promote
 from openbrain.apps.hooks.session import SessionStartHook, run_session_start
 from openbrain.config import load_canon_settings
 
 if TYPE_CHECKING:
-    from collections.abc import Sequence
-
     from openbrain.config import CanonSettings
 
 #: The ``source`` on the synthesised SessionStart payload. ``startup`` is the
@@ -141,23 +143,44 @@ def read_live_pack(settings: CanonSettings) -> Any:
     return asyncio.run(run_session_start(payload, settings))
 
 
+def _repo_relative_pack_path(pack_path: Path) -> str | None:
+    """Return the pack's path below its nearest git root.
+
+    The pack file is the reviewed artifact the repo fact cites. A worktree's
+    ``.git`` marker is a file rather than a directory, so ``exists`` deliberately
+    accepts either shape. No marker means the path cannot be proven repo-relative
+    and repo-fact planning must stop before a write is sent.
+    """
+    resolved = pack_path.resolve()
+    for parent in resolved.parents:
+        if (parent / ".git").exists():
+            return resolved.relative_to(parent).as_posix()
+    return None
+
+
 def provenance_from(
     args: argparse.Namespace, settings: CanonSettings
 ) -> FactProvenance | None:
-    """Assemble the repo-fact provenance, or ``None`` when it was not supplied.
+    """Assemble the repo-fact provenance, or ``None`` when it is incomplete.
 
     Args:
-        args: The parsed command line, carrying the commit, URL, and instant.
+        args: The parsed command line, carrying the pack path, commit, URL, and
+            verification instant.
         settings: The canon configuration, carrying the repo the facts bind to.
 
     Returns:
         The provenance, or ``None`` when any part is absent. All-or-nothing:
         ``repoFactMetadata`` refuses a write whose ``source_url`` does not
-        contain both the commit and the path, so a half-filled provenance is
-        never closer to valid than none at all -- it just fails later, at the
-        server.
+        contain both the commit and the pack's exact repo-relative path, so a
+        half-filled provenance is never closer to valid than none at all.
     """
-    if not (args.repo_source_commit and args.repo_source_url and args.repo_verified_at):
+    source_path = _repo_relative_pack_path(args.pack)
+    if not (
+        source_path
+        and args.repo_source_commit
+        and args.repo_source_url
+        and args.repo_verified_at
+    ):
         return None
     if settings.repo is None:
         # Same all-or-nothing rule: a fact cannot bind without a repo, and the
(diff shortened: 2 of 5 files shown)


rodaddy:

**HIGH** — _receipt_namespace maps namespace_source=="header" to delegated_agent_id, but the server's namespace on the header path is writer_identity; delegated_agent_id is an independent X-Agent-Id label. Both outcomes are wrong: a false PASS-through and a false FAILURE on a correct write.
- Location: python/openbrain/src/openbrain/apps/canon/run.py:154-161 (_receipt_namespace identity_field mapping), against src/auth.ts:120-135 and src/tools/append-session-event.ts:847-860
- Failure scenario: Server truth (src/auth.ts:130-135): clientId = X-Namespace ?? token clientId, and agentId comes from the SEPARATE X-Agent-Id header; namespace_source is "header" whenever X-Namespace is present. writerProvenance (append-session-event.ts:853-859) then emits writer_identity = the actual namespace, and delegated_agent_id = auth.agentId, an unrelated agent label. I executed the shipped guard against both real shapes.

(a) FALSE PASS-THROUGH: receipt {writer_identity:"skippy", token_identity:"admin", delegated_agent_id:None, namespace_source:"header"} — exactly the shape produced when a client sets delegate_namespace=True but leaves agent_id unset (the constructor default agent_id=None, client.py:1129). _receipt_namespace returned None. The guard sets unknown_receipt=True and CONTINUES, so a write whose namespace it was supposed to verify is never checked against settings.agent at the receipt
- Suggested fix: Resolve the namespace from writer_identity for BOTH branches — it is the server's single authoritative statement of the namespace the row landed in (auth.ts:131 sets clientId to the delegated namespace, and append-session-event.ts:855 emits it verbatim). Reduce _receipt_namespace to: explicit `namespace` if present (repo-fact path), else `writer_identity` if a non-empty string, else None. Keep namespace_source/token_identity/delegated_agent_id for the error message only, not for resolution. Add two tests: a header receipt with delegated_agent_id=None and writer_identity equal to settings.agent

**MEDIUM** — The duplicate-event receipt carries provenance but no event_id, so a re-run of an already-applied canon pack is treated as a verified successful write; combined with the header defect it can also mask a genuine cross-namespace landing.
- Location: python/openbrain/src/openbrain/apps/canon/run.py:363-372 (_apply receipt loop), against src/tools/append-session-event.ts:1285-1296
- Failure scenario: append_session_event returns {duplicate:true, message:"Event with identical content already exists in this lane", ...provenance} when content already exists in the lane. I ran the shipped _receipt_namespace against that exact shape with token provenance and it returned "admin" — so on the token path the duplicate is correctly rejected. But _apply never distinguishes duplicate from a real insert: on the header path (where the mapping defect above already yields None or a wrong value) a duplicate receipt is indistinguishable from a fresh write, and the run reports applied N write(s) for rows that were already present. That is the same class as the #588 miscount the PR is closing: a call count reported as an apply outcome. It also means an operator re-running reconcile after a partial failure gets a clean success receipt that proves nothing about the current run.
- Suggested fix: Branch on the receipt's `duplicate` flag in the _apply loop and count duplicates separately from inserts, then report them distinctly (e.g. "applied N, already present M"). Add a test asserting that a duplicate receipt whose provenance resolves to the intended namespace is reported as already-present rather than newly applied.

**LOW** — _verify_readback compares repo-fact expectations against live_key_of/live_text_of, but repo facts are written via a different tool whose rows may not appear in the same scoped pack read used for guidance, risking a spurious not-visible failure.
- Location: python/openbrain/src/openbrain/apps/canon/run.py:195-212 (_verify_readback missing-key loop)
- Failure scenario: The read-back builds one agent_context_pack call with requested_sections = settings.sections and then, for a Lane.REPO_FACTS planned write, looks for metadata["subject"]/metadata["fact"] among live_items(payload, Lane.REPO_FACTS). Repo facts are graph entities bound to `repo` EXACTLY (config.py:686-692, server has no fallback), while guidance rows come from ob_session_events. If the reconcile is run with a settings.repo that does not match the entity's stored repo binding, or repo_facts is not among the requested sections, live_items returns an empty tuple and every repo fact is reported missing — raising NamespaceMismatchError "canon writes not visible" for writes that actually landed. Only reachable on the sparse-receipt path, and repo-fact receipts do return `namespace` directly today, so this is latent rather than live; but the read-back is explicitly framed as the future-proof branc
- Suggested fix: Assert up front in _verify_readback that every lane present in `planned` is also present in settings.sections (and that settings.repo is set when any planned write is Lane.REPO_FACTS), raising a distinct configuration error rather than a namespace mismatch. That keeps a misconfigured read-back from being reported as a namespace incident.

**Independent failure-proof re-verification:** Re-verified the PR's third claimed proof ("Changed token receipt resolution to read delegated_agent_id; test_apply_rejects_a_receipt_from_the_token_namespace failed"). In a scratch worktree of origin/fix/588-canon-apply I swapped the mapping in _receipt_namespace to {"token": "delegated_agent_id", "header": "token_identity"} and ran `uv run pytest -q tests/test_canon_run.py`: RED, 1 failed / 12 passed, exactly that test. IMPORTANT NUANCE the PR body omits: it does NOT fail on the NamespaceMismatchError assertion. It fails on `AssertionError` raised by the fixture's agent_context_pack stub at tests/test_canon_run.py:184, because the mutated guard returned None and fell THROUGH to the sparse r

**Reviewer-run receipts:** All run by me in a scratch worktree of origin/fix/588-canon-apply at /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/rev594 (removed after).
- uv run pytest -q tests/ (python/openbrain) -> PASS, 597 passed, 19 deselected in 14.47s
- uv run pytest -q tests/test_canon_run.py tests/test_canon_reconcile.py -> PASS, 51 passed in 0.42s
- uv run mypy src/openbrain -> PASS, no issues found in 49 source files
- uv run ruff check src tests -> PASS, all checks passed
- bunx tsc --noEmit -> PASS, exit 0
- bun test src/tools/__tests__/repo-facts.test.ts -> PASS, 20 pass / 0 fail, 62 expect() calls
- Mutation run (see mutation_proof_checked) -> RED as expected, restored -> GREEN
Environment honored: BUN_T

rodaddy:
## Fix-lane receipt — opposite-family findings addressed

Pushed commit: `04c02855a59fac8057a43adcd670cc02c5b3e4d7`

### HIGH — delegated-header namespace resolved from the wrong field
- **Fix:** `_receipt_namespace` now uses an explicit repo-fact `namespace` when present, otherwise the server's authoritative `writer_identity` for both token and delegated-header append receipts. `delegated_agent_id`, `token_identity`, and `namespace_source` remain provenance only; they no longer decide where the row landed.
- **Tests:** added the two real header shapes:
  - `writer_identity="skippy"`, `delegated_agent_id=None` succeeds without sparse read-back.
  - `writer_identity="admin"`, `delegated_agent_id="skippy"` raises `NamespaceMismatchError` naming `admin`.
- **Red proof before the fix:** the first test fell through to read-back and raised `AssertionError`; the second reported `DID NOT RAISE NamespaceMismatchError`.
- **Green proof:** both pass in `tests/test_canon_run.py`; the focused canon run is `56 passed`.

### MEDIUM — duplicate receipts counted as newly applied
- **Fix:** `_apply` now returns separate `(applied, already_present)` counts. A receipt with `duplicate: true` is namespace-verified but counted as already present, and the operator log reports both outcomes (`applied N write(s); already present M`).
- **Test:** an exact duplicate append receipt now produces `applied 0 write(s); already present 1` rather than claiming a new write.
- **Red proof before the fix:** the new assertion saw `applied 1 write(s)`.
- **Green proof:** the test passes in the focused and full Python suites.

### LOW — sparse repo-fact read-back could mislabel configuration as namespace failure
- **Disposition:** fixed.
- **Fix:** read-back now validates that every planned lane is included in `settings.sections`, and requires `settings.repo` when repo facts are planned. Misconfiguration raises the distinct `ReadbackConfigurationError` before interpreting visibility as a namespace incident.
- **Tests:** added cases for a missing `repo_facts` requested section and a missing repo binding.
- **Red proof before the fix:** both tests failed because `ReadbackConfigurationError` did not exist.
- **Green proof:** both pass in `tests/test_canon_run.py`.

### Review-memory update
- Strengthened `docs/sme/gotcha-agent.md` with the authoritative `writer_identity` rule, duplicate-outcome accounting, and scoped read-back prerequisites.

### Final validation
- `bunx tsc --noEmit` — PASS.
- `bun test src/tools/__tests__/repo-facts.test.ts` — PASS, 20 tests / 62 assertions.
- `uv run mypy src/openbrain` — PASS, 49 source files.
- `uv run ruff check src tests` — PASS.
- `uv run pytest -q tests/test_canon_run.py tests/test_canon_reconcile.py` — PASS, 56 tests.
- `uv run pytest -q` — PASS, 602 passed / 19 deselected.
- Repository pre-push gate — PASS, including package docs sync.
- Remote branch SHA verified equal to local SHA above.

PR #595, opened 2026-08-06
feat(bulk): ingest Hermes messages and charter telemetry

### LINKED ISSUE #492: post-0.9.0: Hermes bulk adapter + codify agent-telemetry push to Langfuse
Operator-scoped during the v0.9.0 release planning (2026-08-02): these two items are deliberately DEFERRED out of 0.9.0 and are the headline scope for the next version.

## 1. Hermes bulk adapter
- The bulk ingestion framework (python/openbrain/apps/bulk, issue #479 / PR #482) ships in 0.9.0 with HERMES as a loud named gap.
- Next version: chart the Hermes source contract first (same observed-format spec pattern used for the Codex adapter in #479), then implement the adapter against it.

## 2. Codify the agent-telemetry push to Langfuse
- Operator: "the next will be to add [the Hermes adapter] and codify the push of the agent stuff to langfuse correctly."
- Scope: define and enforce the correct shape/route for pushing agent telemetry to Langfuse — a charted contract, not ad-hoc emission. Relationship to the #469 usage-telemetry surface (shipping in 0.9.0) should be stated explicitly in the charter.

State: PROPOSED — nothing here is written. Do not start until 0.9.0 is tagged.

rodaddy:
## Summary

- Implements the Hermes `messages`-row adapter in the Python bulk ingester and registers it for `InputFormat.HERMES`.
- Adds a sanitized fixture shaped like the charted Hermes SQLite schema plus functional normalization, identity, non-turn, and malformed-row tests.
- Records the agent-telemetry-to-Langfuse decision, including the shipped capture/tool/launcher surfaces, the #569 full-coverage masking-first ruling, and the explicit separation from #469 usage metrics.

Fixes #492

## Implementation

- Hermes rows are consumed as JSON lines exported from the charted SQLite `messages` table, validated with Pydantic, and normalized to user, assistant, or tool `RawTurn` values.
- Turn identity includes both `session_id` and row `id`, preventing collisions when multiple Hermes sessions reuse row identifiers.
- System and content-free machinery rows are valid non-turns; malformed rows fail loudly with file and line while keeping corpus content out of the error.
- The generated bulk package README was refreshed from its `__init__.py` source docstring.

## Validation

- `bunx tsc --noEmit` — pass.
- `bun test scripts/local-clone.test.ts server/observability/langfuse-tracing.test.ts` — 51 pass, 1 skip, 0 fail.
- `uv run mypy src/openbrain` — success, 49 source files.
- `uv run ruff check src tests` — all checks passed.
- `uv run pytest -q` — 596 pass, 19 deselected.
- `generate_package_docs.py --check --path python/openbrain/src` — 9 packages documented.
- `git diff --check` — pass.

## New-Test Failure Proofs

- Changed the Hermes tool-role mapping to assistant: `test_sanitized_schema_sample_reaches_fake_lane` failed on expected `TurnRole.TOOL`; restored and green.
- Removed `session_id` from Hermes turn identity: `test_identity_includes_session_and_nonturn_rows_are_declined` failed because both sessions produced `hermes:7`; restored and green.
- Removed the source line number from Hermes malformed-row errors: `test_malformed_row_names_location_without_content` failed because `:2` was absent; restored and green.
- Final Hermes-focused receipt: 3 pass, 20 deselected.

## Critical Self-Review

- Highest-risk behavior: mapping heterogeneous Hermes message rows into the raw-turn identity and role contract; guarded by schema-shaped end-to-end staging tests and cross-session collision proof.
- Assumptions that could be wrong: the charted SQLite export represents timestamps as strings and uses the observed user/assistant/tool/system role vocabulary; unknown or structurally changed rows fail loudly rather than being silently accepted.
- Missing/weak tests: no copied production database was used by instruction; fixtures cover the charted schema shape but cannot prove an uncharted Hermes type variation.
- Security/permission risk: malformed-row errors include field locations and source position only, never message content; no auth, namespace, or server permission boundary changes.
- Migration/deploy risk: no database migration or service deployment; this changes an operator-run Python bulk input adapter and repository documentation.
- Downstream client/runtime risk: not applicable because no MCP tool, wire schema, `openbrain-memory` client export, generated skill, or Hermes runtime/plugin contract changes.
- Rollback/cleanup concern: reverting the commit restores Hermes as unavailable; existing staging databases and other format adapters are unchanged.
- Fixes made before PR: added session-scoped identity, loud content-free Hermes parse errors with line attribution, generated package docs, and removed unrelated formatter churn from existing tests.
- Known residual risk: richer Hermes columns such as reasoning and tool-call metadata are validated only as ignored extras because `RawTurn` has no fields for them; conversational `content` remains whole.
- SME review-memory update: [x] not applicable because: this closes a named adapter gap without producing a reusable MEDIUM+ review pattern.

## Review Gate

- [x] MEDIUM+ review findings were captured: controller diff inspection found none; no review worker was used per the explicit single-process task constraint.
- Live Open Brain checks: [x] not applicable because: the change is an offline bulk normalizer plus a decision record and does not alter a live MCP, database, namespace, or transport surface.

## Downstream Rollout

- [x] I checked `docs/downstream-rollout.md`.
- [x] Downstream rollout is not applicable because this PR changes no public MCP/client contract, generated skill, or runtime/plugin call shape.


rodaddy:

**HIGH** — Hermes `timestamp: str` is required and non-nullable, but the charted schema in #492 gives column names only — no types, no nullability. A single NULL or numeric timestamp raises MalformedHermesRecordError and halts the whole run with zero rows staged, on a 25,054-message corpus.
- Location: python/openbrain/src/openbrain/apps/bulk/formats.py — HermesMessageRecord.timestamp (added in this PR)
- Failure scenario: Verified by me in the rev595 worktree. I built a real sqlite3 table, inserted three rows (one ISO string timestamp, one epoch-integer stored in a TEXT-declared column, one NULL), exported them as JSON lines the way the PR documents, and ran stage_file(..., InputFormat.HERMES, store). Result: `HALTED -> malformed Hermes messages row at .../export.jsonl:3; observed shape failed at timestamp`, and `store.counts().staged == 0`. Staging is transactional, so it is all-or-nothing — one bad row anywhere in the file loses the entire ingest, not just that row. This is also a sibling asymmetry of the exact class docs/sme/gotcha-agent.md #447 catalogues: the Claude adapter declares `timestamp: str | None = None` (capture/records.py:227) and I confirmed a Claude line with no timestamp field at all yields a valid turn with occurred_at=None; the target model RawTurn.occurred_at is likewise `str | None`
- Suggested fix: Declare `timestamp: str | int | None = None` and normalize to `occurred_at=str(value) if value is not None else None`, matching capture/records.py:396 (`occurred_at=record.timestamp or None`). Add a test with a NULL-timestamp row proving the row still stages rather than halting the file.

**MEDIUM** — `active` and `compacted` are charted columns carried in the PR's own fixture, but `extra="ignore"` silently drops them, so a deactivated or compaction-superseded Hermes row ingests as a live corpus turn.
- Location: python/openbrain/src/openbrain/apps/bulk/formats.py — HermesMessageRecord model_config / hermes_raw_turn_from_line
- Failure scenario: Verified by me: `hermes_raw_turn_from_line(row(compacted=1))` and `hermes_raw_turn_from_line(row(active=0))` both return full RawTurn values indistinguishable from a live turn. Both fields are named in the #492 charted schema comment and both are present in every line of the new hermes-messages-sanitized.jsonl fixture, so the author saw them and the model discards them without comment. In Hermes a compacted row is content superseded by a compaction summary and an inactive row is one edited out of the live thread; ingesting both means the raw corpus gains duplicated and retracted content that downstream distillation treats as authoritative. This matches docs/sme/gotcha-agent.md 'An allowlist that drops unrecognized keys is accept-and-ignore, not tolerance' (#464): forward tolerance is for fields the reader has no opinion about, not for fields the source schema already defines with a meani
- Suggested fix: Declare `active: int | None = None` and `compacted: int | None = None` on the model and make an explicit, documented ruling — either decline compacted/inactive rows as non-turns, or ingest them deliberately with a comment naming why. Either way the decision should be visible in code and covered by a test, not implicit in `extra="ignore"`.

**LOW** — No test proves file-order ordering or a staged COUNT across multiple interleaved Hermes sessions; the single fixture is pre-sorted and single-session.
- Location: python/openbrain/tests/test_bulk_ingest.py — TestHermesObservedSQLiteAdapter
- Failure scenario: The fixture hermes-messages-sanitized.jsonl is five rows of one session already in timestamp order, so `assert staged.staged == 3` cannot distinguish correct ordering from any ordering. Staging orders by stage_index (file order, staging.py:182) while the server sequences by (session_ref, occurred_at) per RawTurn's own docstring — a real `SELECT * FROM messages` export across 543 sessions is grouped by rowid, not by session, so the two orderings diverge and nothing here would catch it. docs/sme/gotcha-agent.md #447 names this exact gap: 'does an end-to-end test assert a COUNT, not just per-record shape?' — every per-record assertion passed there too.
- Suggested fix: Add a fixture case with two interleaved session_ids and out-of-order timestamps, asserting both the total staged count and that each session's turns carry the session_ref and occurred_at the server needs to re-sequence them.

**LOW** — The charter doc introduces a `Scope key:` field that appears in no other decision doc.
- Location: docs/decisions/agent-telemetry-is-langfuse-not-usage-metrics.md:21
- Failure scenario: I checked `rg -n "Scope key" docs/decisions/*.md` — the only hit is this new file. The sibling fleet-bus-c6-envelope-vs-body.md uses 'What this is / Source issue / Decided / Status'. Not a defect: the doc is otherwise consistent with the directory's format and README conventions, quotes the #569 operator ruling verbatim as the README requires, states the #469 relationship in a dedicated section as #492 asked, and is honest that #561 masking is not yet implemented. Flagging only so the field is either adopted deliberately across docs/decisions/ or dropped.
- Suggested fix: Either drop the `Scope key` line or, if it is meant as a machine-readable handle, note the convention in docs/decisions/README.md so later docs follow it.

**Independent failure-proof re-verification:** Re-verified the session-scoped identity proof independently, in a fresh worktree of the PR branch (origin/feat-492-hermes-and-charter @ 5a83277) at /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/rev595. I edited formats.py to change `turn_uuid=f\"hermes:{message.session_id}:{message.id}\"` to `turn_uuid=f\"hermes:{message.id}\"`, then ran the suite: RED, 2 failed / 21 passed, with `test_identity_includes_session_and_nonturn_rows_are_declined` failing on `AssertionError: assert 'hermes:7' != 'hermes:7'` — the exact collision the PR claims, and `test_sanitized_schema_sample_reaches_fake_lane` failing as well on the changed turn_uuid. Restored formats.py from backup, confirmed `git diff --stat

**Reviewer-run receipts:** All executed by me in the rev595 worktree at 5a83277 (uv with UV_CACHE_DIR=/Volumes/ThunderBolt/_tmp/open-brain/_validation-runs/uv-cache; bun with BUN_TMPDIR=/Volumes/ThunderBolt/_tmp/open-brain/_scratch/bun-tmp). Python: `uv run mypy src/openbrain` -> Success, no issues in 49 source files. `uv run ruff check src tests` -> All checks passed. `uv run pytest tests/test_bulk_ingest.py -q` -> 23 passed in 0.38s. TypeScript: `bun install --frozen-lockfile` -> 169 packages; `bunx tsc --noEmit` -> clean, no output. `bun test scripts/local-clone.test.ts server/observability/langfuse-tracing.test.ts` -> 51 pass, 1 skip, 0 fail, 122 expect() calls across 2 files. Every PR-claimed validation result re

rodaddy:
## Opposite-family review fixes — a4dc859

- **HIGH — nullable or integer Hermes timestamps halted the whole stage.**
  - **Fix:** `HermesMessageRecord.timestamp` now accepts `str | int | None` with a default, and the adapter carries integers as strings while preserving `None` as an absent `occurred_at`.
  - **Proof:** new `test_nullable_and_integer_timestamps_stage_without_halting` was RED before the fix with `MalformedHermesRecordError` at `timestamp`; after the fix the Hermes class is **6 passed**, the bulk file is **26 passed**, and the full Python suite is **599 passed, 19 deselected**.

- **MEDIUM — charted `active` and `compacted` state was silently ignored.**
  - **Fix:** both columns are declared on `HermesMessageRecord`; rows with `active == 0` or `compacted == 1` are explicit valid non-turns and never enter staging.
  - **Proof:** new `test_inactive_and_compacted_rows_are_not_staged` was RED before the fix (`staged == 3`, expected `1`); after the fix only the live row stages and the focused/full receipts above are green.

- **LOW — no multi-session interleaving/count proof.**
  - **Fix:** added an end-to-end four-row, two-session, out-of-time-order case asserting the staged/sent totals and every `(session_ref, occurred_at)` pair needed for server sequencing.
  - **Proof:** temporarily removing `session_ref` made `test_interleaved_sessions_keep_server_ordering_metadata` RED with `KeyError: 'session_ref'`; restored production behavior and the test is green in the receipts above.

- **LOW — `Scope key:` decision-doc consistency.**
  - **Disposition:** no change. The convention already exists and is documented in `docs/decisions/README.md:48-52`, and existing records use it, for example `docs/decisions/redaction-is-three-policies.md:3`. This record follows that established format.

Additional validation:
- `uv run ruff check src tests` — pass.
- `uv run mypy src/openbrain` — success, 49 source files.
- `bunx tsc --noEmit` — pass.
- Push gate — TypeScript pass, Bun tests pass, all checks pass.

PR #596, opened 2026-08-06
docs(canon): seed remaining king repository facts

### LINKED ISSUE #520: Seed repo facts for the repos that matter -- Development and king-* are near-empty and now honestly show it
After #517 (merged, PR #518), sessions get the facts of the repo they are in — which makes the coverage gap visible instead of masked: current bindings are open-brain=7, king-core=2, development=1 (rebound 2026-08-03 from the legacy `rodaddy/development` slug — that was the AGENTS.md-referenced runtime-context fact `0c0a4e94`), and 1 each for the other king repos.

Seeding is content-authoring, not plumbing: each fact needs provenance (`repoFactMetadata`: repo, source_commit, source_url, verified_at — `src/tools/repo-facts.ts`) and should be a distilled operational truth an agent needs at startup, not a duplicate of what already auto-loads (Development's AGENTS.md router is in every session's context — do NOT mirror it into facts).

Candidates per repo: deploy/run truths, DB/test gotchas, host bindings, the things `# Stack`/`## Commands` sections carry. Write path: `upsert_repo_fact` per #445's decision record; the canon pack TOML + reconciler (#493 finding: no pack file exists yet) is the reviewable home once authored.

Slug convention is git-root basename lowercased (#517). Any future fact bound under another shape is invisible to the pack — bind exactly.

rodaddy:
## Summary

- Seeded seven source-verified operational facts for each of king-trading, king-status, king-ops, king-market-data, and king-reconciliation through the established `upsert_repo_fact` client boundary.
- Added reviewable canon-pack sources for all current Development and king-* repository facts: 12 TOML files, 92 entries.
- Added a durable closure receipt with per-repository keys, source commits, read-back evidence, and live fact counts.

Fixes #520

## Live receipts

- Direct `OpenBrainClient` read-back matched all 35 new rows on repo, collection, path, subject, fact type, fact text, source commit, source URL, and verification timestamp.
- Independent SQL count receipt: each of the five target repositories has 8 standing facts and all 8 carry `source_commit`, `source_url`, and `verified_at`.
- Pack parse receipt: 12 new TOML files parsed successfully, 92 entries total.
- Namespace selection came from `OPENBRAIN_NAMESPACE`; no credential values were logged or committed.

## Validation

- `bun test src/tools/__tests__/repo-facts.test.ts` — 20 passed, 0 failed.
- `python/openbrain`: mypy clean; ruff clean; pytest 594 passed, 19 deselected.
- `python/openbrain-memory`: mypy clean; ruff clean; pytest 569 passed, 9 skipped.
- New-test failure proof: not applicable because this change adds no tests; the behavior proof is the live write plus exact read-back and independent SQL count.

## Critical Self-Review

- Highest-risk behavior: Incorrect operational facts would be injected into repo-scoped startup context; each new fact was checked against committed source at the named SHA, accepted by `repoFactMetadata`, and read back field-for-field.
- Assumptions that could be wrong: The older Development/king rows are suitable to mirror into pack files; this PR preserves their live text and stored provenance but does not independently re-verify every historical source claim.
- Missing/weak tests: No new executable behavior exists; pack parsing and existing repo-fact tests cover the file shape and server validation, while the live read-back proves persistence.
- Security/permission risk: Namespace isolation is the primary risk; writes used the namespace supplied by the installed environment and exact basename slugs, with no credential values in output or files.
- Migration/deploy risk: No schema or service deployment changes; the 35 idempotent repo-fact upserts are already live in dogfood and independently counted.
- Downstream client/runtime risk: No MCP schema, transport, package, or client contract changed; downstream rollout is not applicable.
- Rollback/cleanup concern: Reverting the commit removes the reviewable pack sources but does not remove live entities; reverting live data would require an explicit archive operation on the exact new entity IDs, which was not performed.
- Fixes made before PR: Corrected pack and receipt wording that initially overstated historical source verification, and reworded two facts rejected by the existing raw-code safeguard before the successful writes.
- Known residual risk: Issue #588 still blocks the repo-fact branch of `openbrain-canon-reconcile --apply`; these packs parse and are reviewable, while this issue's live writes used the established direct client path.
- SME review-memory update: [x] not applicable because: no source-code review pattern or missed defect class changed; this is canon content plus live write receipts.

## Review Gate

- [x] MEDIUM+ review findings were captured: the self-review found a provenance-wording overstatement in every new pack header and the receipt; it was corrected before commit, with no unresolved material finding.

Live Open Brain receipt: 35 direct-client read-backs matched exactly, and SQL independently reported 8 standing/provenanced facts for each target repository.


rodaddy:
## Opposite-family review (Claude lane) — verdict: **approve**

**MEDIUM** — Every new fact's source_url points at a commit that exists only in a local checkout, so the cited GitHub blob URLs are currently unresolvable — though the pre-existing baseline rows share this property, so it is a standing convention rather than a regression.
- Location: docs/canon/king-{trading,status,ops,market-data,reconciliation}-repo-facts.toml (all 35 new entries' `source =` provenance strings)
- Failure scenario: All five source commits (king-trading 3330edb8, king-status bc67563f, king-ops d29c642f, king-market-data e3cb1fcc, king-reconciliation 52994097) return empty from `git branch -r --contains` and are absent from `git ls-remote origin`. king-trading's 3330edb sits 3 commits ahead of origin/main on local branch chore/claude-md-agents-import; it is not an ancestor of origin/main. Consequence: an agent or reviewer following https://github.com/King-Capital/king-trading/blob/3330edb.../src/server.ts gets a 404, so the provenance tuple cannot be used to re-verify the fact by its stated route — the exact drift-checking purpose source_url exists for. Mitigating and why this is not a blocker: the older 2026-06-18 baseline commits (9906157, ae38971, 297ab7c) are ALSO not ancestors of origin/main, so the bar is comparable to previously-seeded repos rather than lowered; the cited file paths do exist a
- Suggested fix: Non-blocking. Either push the five source branches so the cited commits become reachable on the remote, or re-pin the seven facts per repo to the nearest origin/main-reachable commit that still contains the cited file content, and record in the receipt that URL resolvability was checked. Longer term, add a resolvability check to the seeding path so an unpushed source_commit is caught at write time rather than discovered by a reader hitting a 404.

**LOW** — The receipt asserts read-back matched on 'subject' for all rows, but at least one mirrored row keys on `symbol` instead, so a subject-based reconciliation of the pack inventory silently misses it.
- Location: docs/canon/issue-520-repo-facts-receipt.md ("Reviewable pack inventory" / read-back wording) and docs/canon/king-core-repo-facts.toml line 15
- Failure scenario: The king-core entry rendered as `subject = "ApiResponse"` in the pack corresponds to a live row whose metadata carries `symbol: "ApiResponse"` with no `subject` field at all (confirmed by SQL: `select metadata->>'subject', metadata->>'symbol' ...` returns an empty subject for that row). repoFactMetadata accepts either symbol or subject, so both forms are legal and the fact text matches verbatim — nothing is broken today. The risk is verification-shaped: my own pack-vs-DB reconciliation keyed on subject reported a false 'missing in DB' for this entry, meaning any future drift check written against the receipt's stated 'subject' matching will misreport symbol-keyed rows. The pack format is lossy here because PackEntry has no `symbol` field, so a symbol-keyed live row can only round-trip as a subject.
- Suggested fix: Note in the receipt that mirrored rows may key on `symbol` rather than `subject` and that the pack represents both as `subject`, or add an optional `symbol` field to PackEntry so the mirror is lossless and a future reconciler can match on the same field the row actually uses.

**Independent failure-proof re-verification:** Re-verified the canon-pack subject guard, the only executable logic the 92 new entries depend on. In a scratch worktree of 95b12b0 at /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/rev596 (uv sync, UV_CACHE_DIR under _validation-runs), I removed the `_repo_facts_need_a_subject` model_validator from python/openbrain/src/openbrain/apps/canon/pack.py. Result: RED — `tests/test_canon_reconcile.py::TestPackFormat::test_a_repo_fact_without_a_subject_is_refused_at_parse` failed with "Failed: DID NOT RAISE ValidationError" (1 failed, 26 passed). Restored the file from the HEAD blob (git checkout -- is blocked by the repo git guard, so I rewrote from `git show HEAD:<path>`); `git status --short` cle

**Reviewer-run receipts:** All commands run by me in /Volumes/ThunderBolt/Development/open-brain and the rev596 worktree.

TYPECHECK/TESTS (touched stack):
- `bunx tsc --noEmit` — clean (exit 0). Matches PR claim.
- `bun test src/tools/__tests__/repo-facts.test.ts` — 20 pass, 0 fail, 62 expect() calls. Matches PR claim.
- `python/openbrain`: `uv run pytest -q` — 594 passed, 19 deselected. `uv run mypy src/openbrain` — Success, no issues in 49 source files. `uv run ruff check src tests` — All checks passed. All three match PR claims exactly.

DB READ-BACK (independent of the PR's receipt, dogfood open_brain_local_20260724 via .env libpq vars):
- Per-repo standing counts + provenance completeness: development 8/8, king-

PR #601, opened 2026-08-06
feat: add Langfuse trace forensics CLI

rodaddy:
## Summary

- add `scripts/langfuse-trace.ts` with `get`, `session`, `diff`, and `repeat` trace-forensics commands
- align repeated stage occurrences pairwise, preserve chosen-row ranking order, and compare row-id evidence or count/degradation evidence explicitly
- return inconclusive exit 2 when retrieval evidence is absent or Langfuse returns an invalid response; return exit 1 from `diff` and `repeat` when evidence differs
- render deterministic timelines with parsed timestamp ordering, stable tie-breakers, explicit unknown timestamps, and escaped identifiers
- correlate repeat results to the oldest fresh traces from the current run and reject unexpected extras
- add fixture-backed library and CLI regressions plus operator examples in `docs/CONFIG_REFERENCE.md`

Part of #569

Tracked under epic #571.

## Verification

- [x] Python package checks passed or are not applicable — Python package source is unchanged; `repeat` uses the existing installed client boundary
- [x] Live Open Brain smoke passed or is not applicable — the original author lane exercised `get`, `session`, equivalent `diff`, and two-run `repeat search_brain` against local dogfood; this fix lane added fixture and CLI-boundary regressions without another mutating repeat

## Critical Self-Review

- Highest-risk behavior: False equivalence would hide retrieval drift. Comparisons now pair duplicate stage occurrences by chronological index, compare candidates as identities and chosen rows as an ordered sequence, fall back to count/degradation markers when row IDs are absent, and return inconclusive rather than equivalent when no retrieval evidence exists.
- Assumptions that could be wrong: Langfuse continues to expose trace IDs, timestamps, observations, and list `data` in the v3.173-compatible public API shape. Both detail and list responses are now schema-validated, and unexpected HTTP 200 bodies exit 2.
- Missing/weak tests: HTTP pagination and eventual trace arrival still use the prior live smoke rather than a timed network integration test. The fix suite covers response validation, CLI exit wiring, duplicate occurrences, zero evidence, degraded payloads, qmd null IDs, chosen ordering, tied/missing timestamps, repeat ordering/extras, and child-output parsing.
- Security/permission risk: Langfuse Basic-auth values are read only from environment variables and never printed; rendered row/trace/observation IDs escape control characters; `--json` intentionally returns content-ful trace data to the invoking operator.
- Migration/deploy risk: No migration or service deployment is required; this is an operator script and documentation change.
- Downstream client/runtime risk: No MCP tool, schema, transport, Python package, generated skill, or Hermes contract changes; downstream rollout is not applicable.
- Rollback/cleanup concern: Rollback is the three PR commits; no live data was mutated by the fix lane.
- Fixes made before PR: Closed all findings from pull/601#issuecomment-5204027299, including truthful tri-state exits, occurrence alignment, degraded evidence, deterministic timelines, repeat correlation/error handling, response validation, ordered chosen rows, qmd candidate identities, control escaping, and fixture/CLI regressions.
- Known residual risk: `repeat` fails if Langfuse has not exposed exactly the expected number of fresh traces within 30 seconds; unexpected extras now fail explicitly rather than silently changing the baseline.
- SME review-memory update: [x] docs/sme/ updated (correctness — commit 895728a on this branch; provenance pull/601#issuecomment-5204027299)

## Review Gate

- [x] MEDIUM+ review findings were captured in docs/sme/ (commit 895728a)

## Contract Parity

- Contract parity: [x] runtime-specific because: this operator-only consumer does not change MCP names, schemas, result shapes, or runtime contracts

## Downstream Rollout

- [x] rtech-mcps handoff is complete or not applicable — not applicable; no registry or public contract change
- [x] mcp2cli cache/skill refresh is complete or not applicable — not applicable; no tool or schema change
- [x] rtech-hermes Python runtime/plugin changes are complete or not applicable — not applicable; no runtime/plugin change
- [x] Hermes live rollout/canaries are complete or not applicable — not applicable; no agent-facing contract change

Notes/evidence:

- `bun test scripts/langfuse-trace.test.ts` — 18 passed, 0 failed
- `bun test` — 3312 passed, 509 skipped, 0 failed across 240 files
- Pre-push reran typecheck and the full Bun suite with the same green result
- Exit mapping mutation proof: changed the inconclusive mapping from 2 to 0 and ran only `maps comparison outcomes to isolated exit codes`; it failed with expected `[0,1,2]` / received `[0,1,0]`, then passed after byte-for-byte behavior restoration
- CLI boundary tests prove differing traces exit 1 and an invalid HTTP 200 Langfuse body exits 2
- Original live repeat session `4d994277-5bf6-4471-a0bb-c246e892b3fa` produced traces `082b2c754bfa47669a83b4c280ed18ad` and `0d0ec2fae0068798538fec238149ecd5`

### DIFF
--- a/docs/CONFIG_REFERENCE.md
+++ b/docs/CONFIG_REFERENCE.md
@@ -390,10 +390,13 @@ bun scripts/langfuse-trace.ts get 28a6758fe84d34490816d28bfcd4ac20
 # Last five traces for one brain/MCP session, printed in chronological order.
 bun scripts/langfuse-trace.ts session e07dcb17-0026-4dfb-8d2c-c8a5280833d2 -n 5
 
-# Stage-aligned retrieval evidence comparison; exit 1 means evidence differs.
+# Occurrence-aligned retrieval evidence comparison. Exit 1 means evidence
+# differs; exit 2 means the trace/API result is inconclusive or invalid.
 bun scripts/langfuse-trace.ts diff <trace-id-a> <trace-id-b>
 
 # Repeat one live search and report deterministic versus varying stage fields.
+# Traces are correlated to this run, ordered oldest-first, and exit 1 if any
+# retrieval evidence varies (exit 2 if the comparison is inconclusive).
 bun scripts/langfuse-trace.ts repeat search_brain --query "trace forensics" -n 3
 ```
 
--- a/docs/CONFIG_REFERENCE.md
+++ b/docs/CONFIG_REFERENCE.md
@@ -375,6 +375,30 @@ header — around both this module's content-free discipline and the shared
 logger's redaction (measured: the injected key-shaped string appeared in the
 output). The two lines above are how this lane reports its health instead.
 
+#### Operator trace forensics (#569)
+
+Source `/Volumes/ThunderBolt/open-brain-local/local-clone.env` before using
+`scripts/langfuse-trace.ts`; it supplies the Langfuse coordinates without putting
+values in arguments or output. The consumer uses Langfuse's Basic-authenticated
+public API. `repeat` drives the local dogfood server through the existing Python
+`openbrain-memory` `OpenBrainClient.call_tool` path via `uv run`.
+
+```bash
+# Compact trace identity plus its start-time-ordered observation timeline.
+bun scripts/langfuse-trace.ts get 28a6758fe84d34490816d28bfcd4ac20
+
+# Last five traces for one brain/MCP session, printed in chronological order.
+bun scripts/langfuse-trace.ts session e07dcb17-0026-4dfb-8d2c-c8a5280833d2 -n 5
+
+# Stage-aligned retrieval evidence comparison; exit 1 means evidence differs.
+bun scripts/langfuse-trace.ts diff <trace-id-a> <trace-id-b>
+
+# Repeat one live search and report deterministic versus varying stage fields.
+bun scripts/langfuse-trace.ts repeat search_brain --query "trace forensics" -n 3
+```
+
+Add `--json` to `get`, `session`, `diff`, or `repeat` for machine-readable output.
+
 ### Drop-folder collector
 
 `src/drop-folder-collector.ts` reads four scan bounds — `DROP_COLLECTOR_MAX_FILES`
--- a/docs/sme/correctness.md
+++ b/docs/sme/correctness.md
@@ -1818,3 +1818,14 @@ This repo has TWO serving trees (`server/main.ts` for the local clone, `src/inde
 ### Pattern (same PR, finding M3)
 
 A span or timing wrapper around `run: () => alreadyComputedValue` measures nothing: `duration_ms` is structurally ~0 and the stage's exception path can never fire, while the identically-named span on the other tree measures real work. A silently-wrong diagnostic is worse than an absent one. Check that the computation is INSIDE the measured callback everywhere the span name is emitted.
+
+## [2026-08-06] Verification tooling must never render absence of evidence as equivalence
+
+**Severity:** HIGH
+**Source:** PR #601 review swarm (trace forensics tooling), findings H1/H2/H3
+**Scope key:** `review.comparison_tools_fail_toward_unknown`
+**Status:** active
+
+### Pattern
+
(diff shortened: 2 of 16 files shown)


rodaddy:

Opposite-family swarm at head d3c5405: correctness + adversarial Claude lanes (SME-injected), plus independent red-proof re-verification and one live `get` against the known canary trace. Deduplicated. No fixes made at the time of this comment.

### HIGH

**H1 (correctness, VERIFIED by executing the lib): duplicate span names merge into one bucket, hiding real per-stage drift.** `stagesOf` keys by bare `observation.name` and unions row-id sets on collision. Repeated names in one trace are reachable (search_all drives the brain arm's stages alongside its own). Verified: two traces each containing `retrieval.fallback_dedupe` twice with swapped per-occurrence chosen rows compare `equivalent=true`, exit 0 — the tool reports "no difference" for a real difference. Fix: align occurrences pairwise by index within each name group (never union); regression with swapped chosen rows asserting exit 1.

**H2 (both lanes independently, VERIFIED end-to-end with a stub API): two traces with zero observations report `equivalent=true`, exit 0.** Child observations arrive via the batched OTel processor, so "not flushed yet" and "identical" are indistinguishable — absence of evidence rendered as evidence of equivalence, on the exact contract the doc promises (`exit 1 means evidence differs`). Fix: zero compared stages → explicit `equivalent=unknown (no retrieval evidence found)` with exit 2; fixture test.

**H3 (adversarial): count-only/degraded spans always compare equivalent.** `diffTraces` compares only row-id sets, so #599's `payload_degraded` counts-only shape — and any stage whose evidence is counts — always matches. Fix: when row-id evidence is absent, compare the count fields and degradation markers, and say in the output which basis was used.

### MEDIUM

- **M1 (correctness, REPRODUCED on the live canary trace):** timeline ordering is unstable on tied `startTime` (the real emitter stamps children in one tight loop — four spans share one millisecond; the live `get` rendered `vector_query` LAST, i.e. ranking shown before the query that fed it) and a missing `startTime` coerces to `""`, sorting first. Fix: parse timestamps, tie-break on endTime then id, absent stamps render last with an explicit marker; tests for both.
- **M2 (correctness):** `repeat` keeps the NEWEST N fresh traces and drops the run's first call when extras match the session filter, mis-anchoring the baseline. Fix: sort ascending and correlate on the run's own trace ids; warn or fail on unexpected extras.
- **M3 (both lanes):** `driveRepeat` awaits the Python child's stderr and then throws it away; any stdout noise breaks `JSON.parse` opaquely. Fix: surface stderr on failure; parse the last JSON line rather than the whole stream.
- **M4 (adversarial):** Langfuse HTTP 200 with an error body renders as a valid empty trace, exit 0. Fix: validate the response shape; unexpected body → exit 2.
- **M5 (adversarial):** chosen rows compared as an unordered Set, so a pure ranking-order regression reports equivalent. Fix: compare chosen as an ordered sequence (order IS the ranking) with the difference shown as reordering.
- **M6 (adversarial):** `repeat` always exits 0 even when every stage VARIES, unlike `diff`. Fix: nonzero exit when any stage varies, consistent with the diff contract.

### LOW

- qmd rows emit `row_id: null`, so disjoint qmd candidate sets print "unchanged" — the same collapsing-identity class as #599 M4; key qmd candidates by path+index for comparison.
- Row-id strings pass into timeline lines unescaped (newlines/ANSI could forge lines); escape control characters in rendered ids.
- Test suite covers only the happy-path fixture pair; the degenerate cases above need fixtures.

### Red-proof re-verification: SUBSTANTIATED (with a caveat)

Independently reproduced at d3c5405: baseline 6/6 green; an unconditional-equivalent mutation turned exactly the non-equivalent fixture test red (verbatim output captured); byte-identical revert, green restored. Caveat: the mutation trips before the exit-code assertion runs, so `traceDiffExitCode`'s mapping is covered by the passing run but was not independently mutation-proven, and nothing tests the CLI `process.exit` wiring.

Live `get` against trace 28a6758fe84d34490816d28bfcd4ac20: all five expected retrieval spans present with sane data (9 candidates → 3 chosen, ids consistent across stages); rendering order wrong on tied timestamps (M1, above).

---
Verdict: **not mergeable as-is.** H1/H2/H3 all make the tool report equivalence it did not establish — on an operator trust surface built precisely to answer "is it consistent?", a false "yes, identical" is the worst possible failure. Fix lane launching next; delta re-verify before the Review Gate box is checked.

rodaddy:
## Fix-delta verification receipt (controller-posted) — d3c5405..7bb071e

Independent opposite-family verification, executed against the CODE (every HIGH/MEDIUM re-proven by running the reviewers' original reproductions, most also mutation-proven): H1 pairwise occurrence alignment (swapped-occurrence repro now exits 1 with per-occurrence deltas), H2 zero-evidence → `equivalent=unknown` exit 2 (proven at the real CLI boundary too), H3 counts+degradation basis named in output with the mixed case never claiming equivalence, M1 stable timeline ordering with missing stamps rendered last, M2 ascending oldest-N selection failing loudly on unexpected extras, M3 stderr surfaced + last-JSON-line parsing, M4 Zod response validation exiting 2 on malformed 200s, M5 ordered chosen comparison with `reordered=true`, M6 repeat exit contract wired and documented, all three LOWs. **Verdict: mergeable.**

One new LOW parked as follow-up: the M1 endTime tie-break leg passes on id ordering alone in the shipped fixture (behavior verified correct by the verifier's own fixture; the shipped one needs an id rename to make that leg mutation-proven).

Review gate: findings posted before fixes, delta independently verified, SME capture next, then the Review Gate box and merge on green CI at the exact head.

Issue #605, opened 2026-08-06
Thought writes: route both serving trees, REST, and lane graduation through chunk-write


rodaddy:
## Evidence found while closing #604

`writeEntryChunks` is not the common thought-write boundary:

### Uses the chunk writer

- `src/tools/log-thought.ts:81-130` writes the parent and calls `writeEntryChunks` for a new long thought.

### Writes thoughts without the chunk writer

- `server/tools/capture.ts:76-120` is the rewrite/local `log_thought` path. It writes only the parent and hardcodes `chunks_written: 0` / `chunks_unembedded: 0`.
- `src/rest-api.ts:153-224` writes `POST /api/v1/thoughts` parents directly.
- `src/tiering.ts:220-274` graduates a session event into `thoughts` directly.

### Intentionally separate paths

- `src/tools/decompose-entry.ts:262-344` and `server/tools/decompose-entry.ts:270-352` use the same chunk columns for replacement decomposition, not write-time resolution; `src/chunk-write.ts:21-26` explicitly says these semantics remain separate.
- `src/tools/ingest-raw-turn.ts:277-290` writes one complete turn to `ob_raw_turns`.
- `src/tools/append-session-event.ts:1219-1244` writes one durable event to `ob_session_events`.

The raw/event paths are consistent with the design: `docs/dream-design.md:151-156` makes the raw-turn insert the Light-stage write boundary, and `docs/dream-design.md:1023-1025` says raw turns remain FTS-only while candidates receive embeddings. `docs/full-send-derivation-spec.md:38-40` states that the client ships raw turns and decides nothing about them.

The thought writers are not consistent with `src/chunk-write.ts:10-30`, which defines the long-entry shape as a complete parent plus per-section rows, with all text preserved and vectorized.

## Work

Create one shared thought-write boundary used by both serving trees, REST creation, and lane graduation. Preserve each caller's auth/provenance/receipt semantics, add real-Postgres coverage per caller, and define repair behavior for a parent that committed before per-section writing failed.

This is intentionally separate from #604: changing three production writers across two serving trees, their response fixtures, and partial-failure behavior is not a bounded remediation-script change.

PR #606, opened 2026-08-06
fix(search): restore full retrieval evidence and remediate transcript dumps

### LINKED ISSUE #604: Corpus: decompose the 4 giant transcript-dump rows; remove the 300-char evidence cap; find out why ingest bypasses the chunker
**Operator ruling (Rico, 2026-08-06), recorded:** "we don't really want to be putting caps on things. We just want to make sure that stupid shit like that giant file doesn't go through the system... Can we just fix that one thing and break it down into its component pieces properly and move the fuck on from ever thinking about an almost 50 MB chunk trying to go through?"

**Measured facts (this session, live dogfood DB open_brain_local_20260724):**
- The 4 largest thoughts (~857 KB each) are raw Claude Code session-transcript JSONL dumps, source `planning-skippy-agentspace`, ingested 2026-03-17 and re-ingested 2026-06-19 (2 distinct contents x 2 copies). No images exist anywhere in the corpus.
- Distribution: 24,664 of 25,192 thoughts are under 10 KB; 524 are 10-100 KB; only these 4 above.
- The 47.6 MB worst-case trace was these rows x 750-candidate fetch x 5 redundant copies. Dedup (merged in #599's fix delta, operator-approved) removes the x5.
- Chunking machinery EXISTS (`src/chunking.ts` CHUNK_THRESHOLD=2000, `src/chunk-write.ts`) — but `parent_id` is NULL on every row in the table and 53 KB rows landed as recently as July, so the primary write path is not chunking. UNVERIFIED which paths use chunk-write vs bypass it — the lane verifies.

**Work:**
1. **Data remediation (operator-authorized):** decompose the 4 transcript rows into proper component pieces through the existing chunking/decomposition machinery; originals go through the normal lifecycle (`discarded_entries`/archive with provenance pointing at the replacements) — nothing hard-deleted. Receipts: before/after row counts, sizes, and a search that used to surface a monster row now surfacing its pieces.
2. **Code (PR):** remove the 300-char slice in the evidence projections (both trees) — full content flows into retrieval evidence again. KEEP the dedup (ids/counts in rank input) and KEEP the 256 KiB degradation strictly as an emergency backstop against pathological payloads (it should never fire on the post-remediation corpus — add a note saying exactly that at the constant).
3. **Investigation (report on this issue, fix only if small and owned):** why do multi-KB writes bypass the chunker — which write paths call chunk-write and which don't, and what the intended design is (docs/decisions/, dream-design). If the fix is bigger than a bounded change, file it as its own issue.

Per the standing order-of-operations rule: this removes a cap on real data and fixes the root; the only remaining bound is the backstop against exactly the garbage class the operator named.

rodaddy:
Fixes #604

## Summary

- restore full masked retrieval evidence in both serving trees by removing the 300-character projection slice
- retain ids/counts-only rank input and keep `MAX_ACTIVE_SPAN_BYTES` as an emergency circuit breaker for pathological corpus rows
- add dry-run-first `scripts/decompose-oversize-thoughts.ts` and functional detector/planning coverage
- record the issue #604 operator ruling in the adversarial SME entry
- remediate the four live raw-JSONL transcript dumps and file #605 for the broader chunk-write caller gap

## Verification

- focused tests — 71 passed, 0 failed
- full `bun test` with `/Users/rico/.config/open-brain/env.release-test` sourced — 3,701 passed, 35 skipped, 0 failed across 241 files
- `OPENBRAIN_TEST_DATABASE_URL` was populated and the registered live-Postgres suites executed; the anti-skip audit found one separate skip: `local clone real PostgreSQL boundary (live Postgres)` because `OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL` is not present in the requested release-test environment
- mutation proof: restoring both old 300-character slices made both new full-evidence tests fail on the missing tail; restoring this patch made them pass
- live dogfood remediation and `search_brain` receipts: https://github.com/rodaddy/open-brain/issues/604#issuecomment-5208118786
- chunk-write investigation: https://github.com/rodaddy/open-brain/issues/604#issuecomment-5208132517

## Data Remediation

- dry-run plan posted before mutation: https://github.com/rodaddy/open-brain/issues/604#issuecomment-5208089983
- 4 active transcript-dump rows / 3,437,076 bytes became 0 active monster rows
- 1,218 active replacement rows now carry `parent_id`, `promoted_from`, and `source_refs`; all are keyword-searchable
- all 4 originals have `archived_at` set and complete reversible copies in `discarded_entries`
- replacement embeddings are pending repair; this PR does not claim vector readiness for those rows

## Downstream Rollout

Not applicable: no MCP tool name, schema, response shape, auth/namespace behavior, transport behavior, migration, Python client, generated skill, or Hermes-facing contract changes. The changed evidence projection is internal observability data.

## Critical Self-Review

- Highest-risk behavior: The remediation script performs a multi-row lifecycle transition against live data; it is dry-run by default and applies all four originals in one transaction before archiving them.
- Assumptions that could be wrong: The detector assumes the pathological class is valid JSONL with at least 100 records and transcript identity fields on at least 80% of records; live measurement showed all 1,073 records on each target carry those fields.
- Missing/weak tests: The script has functional detector/planning tests and a successful live apply receipt, but no isolated fault-injection test that throws midway through the 2,360 proposal loop and observes rollback.
- Security/permission risk: The script uses the operator-supplied database identity directly and emits IDs/counts only; no transcript bodies, tokens, or connection values were posted or committed.
- Migration/deploy risk: No schema migration is required; deploying the code does not rerun the already-completed remediation because the script is operator-invoked and dry-run-first.
- Downstream client/runtime risk: Public MCP results are unchanged; only masked tracing evidence regains full row content, while rank input remains ids/counts-only.
- Rollback/cleanup concern: Originals remain present with `archived_at` and are copied in full to `discarded_entries`; rollback would restore originals and archive the issue-604 pieces through normal lifecycle rather than hard-delete.
- Fixes made before PR: Corrected the live transcript detector after measuring message-field distribution, proved the new evidence tests fail against the old slice, updated stale SME guidance, and filed #605 instead of broadening this PR across unrelated writers.
- Known residual risk: The 1,218 replacement rows have pending embeddings, and server/REST/lane thought writers still bypass the shared chunk writer until #605 is resolved.
- SME review-memory update: [x] `docs/sme/` updated [ ] not applicable because: the issue #604 operator ruling supersedes PR #599 H2's routine per-row shortening recommendation.

## Review Gate

- [ ] MEDIUM+ review findings were captured — controller owns this checkbox after fresh-context review
- Live Open Brain checks: [x] linked below [ ] not applicable because: dry-run, apply, database, provenance, and live `search_brain` receipts are linked in Verification and Data Remediation.

### DIFF
--- a/docs/sme/adversarial.md
+++ b/docs/sme/adversarial.md
@@ -1017,11 +1017,11 @@ Verbatim, from the source:
 **Severity:** HIGH
 **Source:** PR #599 review swarm, finding H2 (measured: 47.6 MB / 528 ms on one traced search)
 **Scope key:** `review.evidence_payload_bounded_and_deduplicated`
-**Status:** active
+**Status:** superseded by issue #604 operator ruling (2026-08-06)
 
 ### Pattern
 
-`content_preview` in this repo is defined as the FULL `t.content` (src/tools/table-constants.ts) — the name promises a bound the projection does not have. Evidence/observability payloads that embed row content must (1) apply a real character bound at the projection, (2) not re-serialize the same rows into multiple spans of one call (query output + rank input + rank output + execute output was 4-5 copies), and (3) bound total payload with an explicit degradation recorded in the emitted record. Also check WHERE masking runs: synchronous masking on the tool's own path multiplies per-string regex cost by payload size, and per-string `new RegExp` compilation belongs at module scope. Measure worst-case with the live corpus (max content size × caller-controlled limit × fetch multiplier), not the happy path.
+`content_preview` in this repo is defined as the FULL `t.content` (src/tools/table-constants.ts). PR #599 initially responded to the measured transcript-dump payload by slicing every evidence row to 300 characters. Issue #604 reversed that remedy at the owning boundary: healthy retrieval evidence flows in full after masking; rank input stays ids/counts-only so rows are not serialized redundantly; the whole-call byte guard remains only as an emergency circuit breaker for pathological corpus rows; and the four raw-JSONL transcript dumps were decomposed and archived through the normal lifecycle. Do not reintroduce routine per-row shortening. Review the corpus shape first, preserve full evidence, deduplicate repeated span payloads, and keep detector compilation at module scope.
 
 ## [2026-08-06] Chosen-vs-filtered evidence keyed on a collapsing identity reports dropped rows as chosen
 
--- a/scripts/decompose-oversize-thoughts.test.ts
+++ b/scripts/decompose-oversize-thoughts.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, test } from "bun:test";
+import {
+  buildTargetPlan,
+  isRawTranscriptDump,
+  REMEDIATION_VERSION,
+} from "./decompose-oversize-thoughts.ts";
+
+function transcript(records: number, messageChars = 20): string {
+  return Array.from({ length: records }, (_, index) =>
+    JSON.stringify({
+      type: index % 2 === 0 ? "user" : "assistant",
+      uuid: `turn-${index}`,
+      sessionId: "session-604",
+      message: { role: "user", content: "x".repeat(messageChars) },
+    }),
+  ).join("\n");
+}
+
+describe("decompose-oversize-thoughts", () => {
+  test("recognizes the raw Claude transcript JSONL class", () => {
+    expect(isRawTranscriptDump(transcript(120))).toEqual({
+      matched: true,
+      recordCount: 120,
+    });
+  });
+
+  test("rejects prose, short JSONL, and malformed transcript rows", () => {
+    expect(isRawTranscriptDump("ordinary thought prose").matched).toBe(false);
+    expect(isRawTranscriptDump(transcript(10)).matched).toBe(false);
+    expect(
+      isRawTranscriptDump(`${transcript(120)}\nnot-json`).matched,
+    ).toBe(false);
+  });
+
+  test("plans linked pieces through the existing decomposition chunker", () => {
+    const content = transcript(120, 100);
+    const plan = buildTargetPlan({
+      id: "00000000-0000-4000-8000-000000000604",
+      content,
+      tags: ["planning"],
+      source: "planning-skippy-agentspace",
+      created_by: "skippy",
+      namespace: "rico",
+      tier: "warm",
+      created_at: "2026-03-17T00:00:00Z",
+      archived_at: null,
+    });
+
+    expect(plan).not.toBeNull();
+    expect(plan?.decomposition.would_write).toBeGreaterThan(1);
+    expect(plan?.decomposition.source_ref.id).toBe(
+      "00000000-0000-4000-8000-000000000604",
+    );
+    expect(
+      plan?.decomposition.proposed_replacements.every(
+        (proposal) =>
+          proposal.provenance.source === "dreamengine-decomposition" &&
+          proposal.source_ref.id === plan.decomposition.source_ref.id,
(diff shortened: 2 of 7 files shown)


rodaddy:

Opposite-family swarm at head 2075d09 (correctness + adversarial, SME-injected). Deduplicated. No fixes made at the time of this comment. The applied remediation itself is sound — the DB end-state was independently verified — but the tooling and the evidence-shape change both have real defects.

### HIGH

**H1 (adversarial, MEASURED against the live post-remediation corpus): the backstop comment is false — ordinary hybrid searches reach the degradation branch.** With the slice removed, full `content_preview` is serialized FOUR times per call (vector-query output, keyword-query output, execute output, fallback merge) — only the rank input was switched to ids. On the healthy corpus (26,331 active thoughts, p99 = 17 KB), routine searches at ordinary result sizes cross 256 KiB, so the "a healthy corpus must never reach this branch" comment is wrong and evidence would degrade to counts-only on normal traffic. Fix within the operator's rulings (full content yes, redundant copies no): serialize each candidate's full content EXACTLY ONCE per call — in the span where the candidate first appears — and carry row ids in every later span (same dedup class the operator 1000%-approved for rank input). Re-do the arithmetic against the live corpus in the PR body and correct the comment to the measured reality.

**H2 (adversarial): the degradation branch fires with no log line.** An emergency circuit breaker that trips silently is invisible exactly when it matters. Add a loud `logger.warn` (span count, byte figure, tool name — no content) on the branch, plus a test pinning it.

### MEDIUM

**M1 (correctness): `--apply` aborts the entire single-transaction run when a piece hash collides with an ARCHIVED row** — the unique index has no `archived_at` predicate but the conflict-resolution SELECT filters to live rows, so an archived collision throws and rolls back all four targets; after the PR's own rollback plan, a re-run would die. Mirror `src/tools/decompose-entry.ts:328` (degrade to per-target skip, recorded in the receipt); apply each target in its own transaction.

**M2 (correctness): no intra-batch duplicate guard** — identical normalized chunks (likely in JSONL scaffolding) double-append source_refs (25-ref schema bound can be exceeded → row fails validation on read), corrupt receipt counts, and lose a `chunk_index` position. Mirror `decompose-entry.ts`'s `writtenIdsByHash` map.

**M3 (adversarial): cross-original chunk reuse never sets `parent_id`/`chunk_index`** on the reused row, silently dropping the joinable reassembly link.

**M4 (correctness): zero test coverage of the apply path** — conflict branch, idempotency, transaction semantics all unpinned; the repo's own `.pg.test.ts` pattern for exactly this exists (`decompose-entry.pg.test.ts:317`) and wasn't reused. Add: run-apply-twice no-op, archived-hash-collision non-abort, identical-chunks coherence.

### LOW

- Detector: one unparseable line short-circuits to NOT-a-transcript while the 80% threshold implies tolerance — pick one rule, and report skipped-but-oversized rows in the dry-run receipt.
- Fatal errors print only `error.name` ("failed: Error") — emit `error_message` too, matching `src/chunk-write.ts:105`.
- Dry-run honesty: `planned_pieces` overstates writes when collisions resolve to reuse; note reuse as a distinct count.
- (Positive findings for the record: dry-run and apply share the same planner so the plan cannot diverge structurally; the new evidence tests are non-vacuous — a returning slice or broken masking would fail them.)

---
Verdict: **not mergeable as-is.** H1 would quietly degrade evidence on normal traffic — the opposite of the operator's intent; H2 makes it undetectable. Fix lane launching next; delta re-verify before the Review Gate box.

Issue #433, opened 2026-07-27
brain_answer cannot see session events, and nothing promotes them — recall answers from months-old data


rodaddy:
## Symptom

Ask Open Brain what happened in the last day and it answers with content from
**March and June**, confidently, with citations and no indication anything is
missing.

Live reproduction on the local clone, 2026-07-27, through OB's own MCP
`brain_answer` (not qmd):

Query: *"What has happened with Open Brain in the last day? Decisions,
corrections, findings, and current state."*

| # | Returned | Age |
|---|---|---|
| 1 | Obsidian sync uncertainty | 2026-06-19 |
| 2 | Phase 1 Foundation research | 2026-03-13 |
| 3 | "mark Open Brain integration complete (#19)" | old |
| 4 | Obsidian sync again (near-dupe of 1) | 2026-06-19 |
| 5 | a `git diff --stat` blob | — |

Zero results from the last day. Top-scoring hit is four months old.

Meanwhile **154 events were written to `ob_session_events` in those same 24
hours**, including one that literally reads
`THE CODE-BRAIN SPEC EXISTS AND IS UNMERGED - FIND IT BEFORE DESIGNING`.

### 1. `brain_answer` structurally cannot see lane events

`ALL_TABLES` in `src/tools/table-constants.ts` is:

```ts
export const ALL_TABLES: Table[] = [
  "thoughts", "decisions", "relationships", "projects", "sessions",
];
```

`ob_session_events` is not in it. Every capture from every session lands in a
table the question-answering tool never queries. This is not a ranking problem
— no amount of recency weighting would surface a row that is never selected.

### 2. Nothing promotes lane events into durable memory

The promotion path exists and is correct: `graduateLaneEvent`
(`src/tiering.ts:362`) called by `tierLaneEvent` (`src/tiering.ts:312`), exposed
as the `tier_lane` MCP tool (`src/tools/tier-lane.ts:114`).

It is **manual and defaults to dry-run** (`src/tools/tier-lane.ts:93`:
`const dryRun = args.dry_run ?? true`). Nothing schedules it. Grep for a
scheduler or maintenance-job caller returns nothing.

Measured consequence:

| Table | Newest row |
|---|---|
| `thoughts` | 2026-07-23 16:02:22 |
| `decisions` | 2026-07-23 16:02:22 |
| `ob_session_events` | 2026-07-27 02:12:01 |

Durable memory has been frozen for ~4 days. 537 lane events written since, none
[...]

PR #610, opened 2026-08-07
fix(recall): brain_answer sees session events (#433, defect 1)

rodaddy:
Closes nothing. **#433 stays open** — this is defect 1 only.

## Summary

`brain_answer` could not reach `ob_session_events` **at all**. It selected its sources by filtering `ALL_TABLES` — the PHYSICAL-table list that drives `PERMISSIONS`, the REST enum, and every write path — so a corpus of **11,136 rows** (80 of them from the last 24h, measured on the dogfood DB 2026-08-07) was structurally absent from every question the tool answered. Asking *"what happened in the last day"* returned months-old thoughts while the day's own events sat unread.

### The shape of it is a duplicated list

`search_brain` had already grown a second, divergent copy of the same selection — it appends `entities`, `brain_answer` did not — and nothing appended the session-event corpus to either. Each new recall surface was one more place to forget.

The selection now lives in **one exported function per serving tree** (`readableSearchTables`, `readableSearchSources`), and every caller uses it. A source added there appears everywhere at once.

### Approach

Session events follow the existing `entities` precedent: a **read-only retrieval source** with its own CTE branch, **not** a widening of `Table`/`ResourceTable`. Those unions carry write contracts and permission rows this corpus has neither of. It rides the `sessions` read permission, so a role that cannot read sessions cannot reach it.

**Namespace isolation is the risk, and it is load-bearing.** `ob_session_events` has **no `namespace` column**. Scope lives on `ob_session_lanes`, reachable only through `lane_id`. Both new CTEs `JOIN` the lane table and apply the auth-derived namespace predicate to `sl.namespace`. Omitting that join would not merely widen results — it would expose every agent's session history to every other namespace.

**Both serving trees are fixed**, because both are live: `server/main.ts` is the local-clone serving entrypoint, and `src/index.ts` still serves core01 via `deploy/open-brain.service` and `scripts/run-two-worker.ts` (the repo states this at `src/index.ts:52-60` and `src/tools/agent-context-pack.ts:215`). Fixing one would have left a live path still blind to the corpus.

### Deliberate limitations, stated rather than hidden

- No stored `search_vector` → the lexical arm matches `content` with ILIKE. **Substring matching, not stemming.** That is the honest capability of a column that does not exist, and it is why `ftsConfig` has nothing to configure for this source.
- No `source_refs` → excluded when a source scope is requested, rather than silently ignoring the scope and returning rows it never authorized.
- No `tier` → `importance` fills the slot; same hot/warm/cold vocabulary, same CHECK constraint.
- `entities` stays **opt-in**, so `brain_answer`'s citation behavior is unchanged by this fix.

## Verification

### done-means: red, then green

`scripts/done-means/433-recall-sees-session-events.sh` is the acceptance gate. It seeds a markered event in a throwaway lane, drives the real `executeSearch` of **both** trees, and tears down always.

**RED (before the fix):**

```
probe result    : own=0 other=0
CLAUSE 1 visibility         : FAIL
  INVISIBLE: the row exists (seeded=1, confirmed in ob_session_events) but the
  real recall path returned 0 rows carrying the marker. Searched tables were
  [thoughts,decisions,relationships,projects,sessions] — the session-event
  corpus is not among them.
CLAUSE 2 namespace isolation: FAIL
  not assessed: with the corpus invisible to recall, a zero cross-namespace
  count proves only that nothing is searched.
RESULT: FAIL
```

**GREEN (after the fix):**

```
probe result    : own=1 other=0 (worst across trees)
per serving tree: src=1/0 server=1/0
CLAUSE 1 visibility         : PASS
CLAUSE 2 namespace isolation: PASS
  no leak: the same marker query under an unrelated namespace returned 0
  marker-bearing rows while the owning namespace returned 1
RESULT: PASS
```

The gate calls the real `readableSearchTables`/`readableSearchSources` rather than re-deriving the list — a third copy of a list is precisely the defect being fixed.

### The leak clause is proven, not asserted

Deleting the namespace predicate from the session-event CTE makes the gate report:

```
CLAUSE 2 namespace isolation: FAIL
  NAMESPACE LEAK: 1 marker-bearing row(s) visible from namespace
  done-means-433-other-…, which owns none of them. The lane join is not
  carrying the auth-derived namespace predicate.
```

and makes **3 of the 8 new tests fail**. Restored → all pass.

### Tests and gates

`src/tools/__tests__/session-events-recall.pg.test.ts` — 8 tests on live Postgres: visibility per tree, cross-namespace non-leakage per tree, the both-namespaces-match-the-same-query case (which separates "the predicate works" from "the search is broken"), role gating, and the entities-stays-opt-in guard.

- `bunx tsc --noEmit` — clean (exit 0).
- `bun test` — **3709 pass / 0 fail**, with `OPENBRAIN_TEST_DATABASE_URL` **SET** against a real UTF8 database. Flagging this explicitly: Postgres tests skip silently without it, so a green run with it unset would have proven nothing here.
- One earlier full local run showed `1 fail`; three subsequent full runs were clean, and the failure did not reproduce or name a test in captured output. Reporting it rather than omitting it — it is unexplained, not dismissed. It was **not** the CI failure on this PR, which was a PR-body gate (see below).

## Critical Self-Review

- Highest-risk behavior: cross-namespace leakage of session events. `ob_session_events` has no namespace column, so isolation depends entirely on the `ob_session_lanes` join predicate. Covered by 3 tests plus a gate clause, all mutation-verified to fail when the predicate is removed.
- Assumptions that could be wrong: that the `sessions` read permission is the right gate for session events. It matches the `entities` precedent, but it is a judgment call — a reviewer may prefer an explicit PERMISSIONS row. Also assumed both serving trees are live based on in-repo comments rather than a running-process check.
- Missing/weak tests: no test drives `brain_answer` end-to-end through MCP for the session-event path; coverage is at `executeSearch`. Vector-arm retrieval of session events is untested (tests force keyword mode deliberately, so a vector hit cannot mask a missing lexical path). No test for the source_scope exclusion path.
- Security/permission risk: the namespace predicate is the whole control and it is tested by deletion, not just asserted. The source-scope exclusion takes the conservative direction (drop the corpus rather than ignore the scope). Role gating tested for `discord`.
- Migration/deploy risk: none — no schema change, no migration, read-path only.
- Downstream client/runtime risk: `search_brain`'s `table` enum gains `session_events` and results gain a new `source_type` of `session_event`. A client that exhaustively switches on `source_type` would meet an unhandled value. Per `docs/downstream-rollout.md` this is a client-facing schema change; the rollout lane subsequently classified every client step N/A with evidence (see Downstream Rollout section) — no downstream client enumerates `search_brain` sources or handles its `source_type` exhaustively.
- Rollback/cleanup concern: revert is clean (read path only). The done-means script writes its probe to the temp scratch bucket rather than the checkout, so it leaves `git status` clean.
- Fixes made before PR: the shared helper initially added `entities` to `brain_answer` — a behavior change #433 never asked for — so it was made opt-in and pinned with a test. Also removed a now-unused `canRead` import.
- Known residual risk: ILIKE substring matching on `content` has no index behind it, so the lexical arm scans the corpus. Documented in the code rather than papered over. Real retrieval-quality behavior of session events mixed into RRF ranking has not been evaluated against a human judgment set.

`docs/sme/correctness.md` gains the duplicated-selection-list pattern (commit 42fc3bb): one selection rule open-coded per caller diverges silently, and the three checks that follow — extract one selection function, check every serving tree, and do not let the acceptance gate re-derive the list a fourth time.

## Review Gate

Live checks ran against the dogfood service (`127.0.0.1:3100`, healthy, `revision b2bd77d`) and the dogfood database `open_brain_local_20260724`. The done-means gate seeds into and reads from that live database on every run; the RED and GREEN transcripts above are its real output. Post-run verification confirmed zero residual rows (`session_key like 'done-means-433%'` → 0, marker content → 0), so teardown is proven rather than assumed.

## Downstream Rollout

Notes/evidence:

- The lane originally left these unchecked because the verification had not been run — see the history for that honest state. A dedicated downstream-rollout lane then ran the classification (2026-08-07); full per-step evidence with file:line citations is in the PR comment: https://github.com/rodaddy/open-brain/pull/610#issuecomment-5224186274
- Summary: rtech-mcps N/A (registry entry is a bare URL pointer, no schemas); mcp2cli N/A (zero hardcoded OB table lists; generated skill carries no param table); rtech-hermes N/A (`READ_TOOL_ALLOWLIST` excludes `search_brain`/`brain_answer`; `search_all` is unaffected by this diff); live canaries N/A pre-merge, with a RUNNING read-only canary proving the deployed build rejects `session_events` today (server-enforced enum).
- Remaining post-merge, tracked on #433: deploy core01, re-run the canary expecting success, one `brain_answer` returning `source_type: "session_event"`.
- Pre-existing drift found (not caused here): rtech-hermes `openbrain/contract.py:50` pins schema v20 / hash `4df9c742…` while live serves v23 / `4b69e9b4…` — exists on `main` today, flagged for its own ticket.

### DIFF
--- a/docs/sme/correctness.md
+++ b/docs/sme/correctness.md
@@ -1840,3 +1840,56 @@ A diff/comparison tool is an operator trust surface: what it prints is what the 
 ### Pattern
 
 A hand-written stub proves conformance to the stub, not to the external API. Langfuse detail responses return full observation objects, but list rows return observation ID strings; reusing the detail shape in list fixtures made every test pass while every live `session` and `repeat` call failed validation. For each external endpoint, capture and sanitize a real response, preserve its envelope and value types, and drive the public caller through that fixture. Do not reuse a sibling endpoint's richer response shape unless the live wire contract proves they are identical.
+
+## [2026-08-07] One selection rule, duplicated per caller, diverges silently
+
+**Severity:** HIGH
+**Source:** issue #433 defect 1, PR #610 (brain_answer could not see session events)
+**Scope key:** `sme.duplicated_selection_lists_diverge`
+**Status:** active
+
+### Pattern
+
+Reusable review check: when two or more call sites independently re-derive the
+same "what may I read/write/see" list, treat that as the defect, not the
+symptom. Fixing only the caller named in the bug leaves the other copies wrong.
+
+In #433 the selection was open-coded at each recall surface. `search_brain`
+computed `ALL_TABLES.filter(canRead)` and then appended `entities`;
+`brain_answer` computed `ALL_TABLES.filter(canRead)` and appended nothing.
+Neither appended `ob_session_events`. The two lists had already silently
+diverged from each other, and a corpus of 11,136 rows was unreachable from the
+tool agents actually ask. Each new recall surface was one more place to forget,
+and nothing in the type system or the tests noticed the disagreement, because
+each copy was internally consistent.
+
+Three things to check on any PR that adds a readable source:
+
+1. Grep for every site that filters the canonical list. If there is more than
+   one, the fix is to extract one selection function that all callers use, so
+   a new source appears everywhere at once instead of one place at a time.
+2. Check every SERVING TREE, not just the one you are editing. This repo has
+   two live entrypoints (`src/index.ts` serves core01 via
+   `deploy/open-brain.service` and `scripts/run-two-worker.ts`;
+   `server/main.ts` is the local-clone serving entrypoint), each with its own
+   `brain_answer`. Fixing one leaves a live path still blind.
+3. Beware the acceptance check re-deriving the list a fourth time. A gate that
+   hand-rolls its own copy of the selection can pass while the product stays
+   broken -- it must call the real exported selection function.
+
+Corollary on WIDENING: adding a physical table to the permissions/type union is
+usually the wrong shape for a read-only corpus. `ob_session_events` has no
+write contract and no PERMISSIONS row, so it was added as a read-only retrieval
+source with its own CTE branch (the existing `entities` precedent) rather than
+by widening `Table`/`ResourceTable`, which would have demanded a write contract
+it does not have. Also confirm the extraction did not quietly widen a caller:
+folding the two lists into one initially added `entities` to `brain_answer`,
+a behavior change the issue never asked for, so it was made opt-in.
+
+Corollary on INDIRECT NAMESPACING: a corpus without its own `namespace` column
+is a namespace-isolation risk the moment it becomes searchable.
+`ob_session_events` reaches namespace only through
+`ob_session_lanes` on `lane_id`, so every new CTE must carry the auth-derived
+predicate across that JOIN. Prove it by deleting the predicate and watching a
+cross-namespace test fail; an isolation test that has never been seen to fail
+is decoration.
--- a/scripts/done-means/433-recall-sees-session-events.sh
+++ b/scripts/done-means/433-recall-sees-session-events.sh
@@ -1,0 +1,292 @@
+#!/usr/bin/env bash
+# DONE-MEANS check for issue #433, DEFECT 1 ONLY — the acceptance gate, not the fix.
+#
+#   bash scripts/done-means/433-recall-sees-session-events.sh
+#
+# Defect 1, verbatim from the issue thread: `brain_answer` structurally cannot
+# see `ob_session_events`. Its table list (`ALL_TABLES`, src/tools/
+# table-constants.ts) names only thoughts/decisions/relationships/projects/
+# sessions, so asking "what happened in the last day" answers from months-old
+# thoughts while the session-event corpus — 11,136 rows on the dogfood DB as of
+# 2026-08-07, 80 of them from the last 24h — is invisible to recall.
+#
+# SCOPE. This gate covers recall VISIBILITY only. Defect 2 of the same issue
+# ("nothing promotes events into durable tables" — no producer calls
+# classifyLaneEvent -> tierLaneEvent -> graduateLaneEvent) is explicitly NOT
+# gated here and is not closed by this check passing. Visibility and promotion
+# are independent: this asks whether the retrieval path CAN reach the corpus,
+# not whether anything copies it elsewhere.
+#
+# EXPECTED TO FAIL until #433 defect 1 is fixed. It is the reward function, not
+# a test of the fix's author.
+#
+# ---------------------------------------------------------------------------
+# Why this drives executeSearch in-process rather than the running service
+# ---------------------------------------------------------------------------
+# A live service is listening on OPENBRAIN_BASE_URL, but it serves the DEPLOYED
+# revision, not this working tree. Probing it would measure whatever is already
+# installed, so the check could never go red-then-green on an uncommitted fix —
+# it would report the same answer before and after. This script therefore
+# imports `executeSearch` from THIS checkout via `bun`, with a real `pg` Pool
+# against the dogfood database. That is the same entry point `brain_answer`
+# calls (src/tools/brain-answer.ts imports executeSearch from ./search-brain.ts),
+# so the retrieval path under test is the real one.
+#
+# The source list comes from `readableSearchTables(role)` — the one exported
+# function brain_answer and search_brain both call — rather than being
+# re-derived here. That matters: #433 defect 1 IS a divergence between two
+# hand-maintained copies of that list, so a gate holding a third copy could
+# pass while the product stayed broken. The probe asks for the `admin` role,
+# the most permissive there is; if the marker is invisible to admin it is
+# invisible to everyone, and the failure cannot be dismissed as a permissions
+# artifact.
+#
+# ---------------------------------------------------------------------------
+# Clauses
+# ---------------------------------------------------------------------------
+# Clause 1 — VISIBILITY. A session event seeded seconds ago, carrying a unique
+#   marker token, surfaces from the real search path when queried by that
+#   marker. FAIL if the retrieval path returns no row bearing the marker.
+#
+# Clause 2 — NAMESPACE ISOLATION. The same query, run under a DIFFERENT
+#   namespace filter, must NOT return the marker. `ob_session_events` has no
+#   namespace column (verified against information_schema on the dogfood DB
+#   2026-08-07 — see the seeding SQL below); namespace is reachable only by
+#   joining ob_session_lanes on lane_id. A recall path that adds the corpus
+#   without carrying the auth-derived namespace predicate through that join
+#   leaks every agent's session history into every other namespace. Clause 2 is
+#   a security gate: it fails CLOSED, so if clause 1 cannot be satisfied at all
+#   the leak question is reported as untested rather than silently passed.
+#
+# Both clauses run against ONE seeded row so they cannot disagree about what
+# "the marker" means.
+#
+# Output is content-free apart from the random run marker: statuses and counts
+# only. No tokens, no memory bodies.
+set -uo pipefail
+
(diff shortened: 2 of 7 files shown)


rodaddy:
## Downstream rollout classification (`docs/downstream-rollout.md`)

Run as a dedicated read-only lane. No mutation, no deploy, no service restart. Every claim below is labelled RUNNING / MERGED / WRITTEN / PROPOSED.

### Contract delta, restated precisely

WRITTEN — the delta is narrower than "the search-source enum changes everywhere":

- `search_brain` input `table` enum gains `"session_events"` — **purely additive**, `.optional()` preserved, no value removed (`src/tools/search-brain.ts:1818-1826` on the branch vs `:1669-1676` on `main`).
- Results can carry `source_type: "session_event"`.
- `src/tools/table-constants.ts` and `server/tools/search-constants.ts` are **not** modified by this PR. `ALL_TABLES` stays the physical-table union; `session_events` is a read-only retrieval source added alongside it via `readableSearchSources()` / `readableSearchTables()`.
- `search_all` is **unaffected** — it still filters `ALL_TABLES` at `src/tools/search-all.ts:417` and `server/tools/search-all.ts:363`. This matters below, because `search_all` is the tool Hermes actually calls.

### 1. Open Brain local verification — PASS (CI)

MERGED-pending: CI green on the head commit. `bun test` + typecheck + the `#433` pg regression test (`src/tools/__tests__/session-events-recall.pg.test.ts`) and `scripts/done-means/433-recall-sees-session-events.sh` ran in CI. This lane did **not** re-run the done-means script locally: it seeds and deletes rows in the dogfood DB, which is outside a read-only remit.

### 2. Open Brain hosted verification — BLOCKED, POST-MERGE

RUNNING — `10.71.1.21:3100/health` answers `healthy`, both workers (3101/3102) up. Deploy is a separate `workflow_dispatch` with `deploy_core01=true`; PR is unmerged, so hosted verification cannot run yet. Not faked.

RUNNING — pre-merge canary against the **deployed** local service (`127.0.0.1:3100`), proving the enum is server-enforced and the deployed build predates the change:

```
search_brain {"table":"session_events"} ->
MCP error -32602: Input validation error ... "values": ["thoughts","decisions","relationships","projects","sessions"]
```

That is the expected pre-merge state and confirms the new value only becomes callable after deploy.

### 3. rtech-mcps handoff — N/A, with evidence

WRITTEN — the Open Brain registry entry is a bare URL pointer carrying **no tool schemas, no enum lists, no contract pin**:

```json
"open-brain": { "backend": "http", "url": "${secret:RTech MCPs - open-brain#fields.url}", "source": "remote" }
```
`rtech-mcps/mcp2cli/services.json:95-99`

Schemas are discovered live from the service. `rg -c "inputSchema|tools" mcp2cli/services.json` -> no matches. Nothing in rtech-mcps enumerates Open Brain search sources, so no handoff is required for this delta.

### 4. mcp2cli pull / cache / generated skills — N/A pre-merge, refresh optional post-deploy

WRITTEN — mcp2cli holds **no hardcoded Open Brain table/source list**. `rg '"thoughts"|ALL_TABLES'` across the repo (excluding `node_modules`) returns zero hits; the only `search_brain` references are cache-plumbing tests (`tests/schema/cached.test.ts:85-108`) that treat the tool name as an opaque string. Schema caching is pass-through — the client does not re-validate enum values against a local copy.

WRITTEN — the generated skill at `~/.config/mcp2cli/skills/open-brain/references/general-ops.md:25-33` documents `search_brain` with a description and a bare example only; it carries **no parameter table and no enum listing**. There is therefore no stale enum to correct. A post-deploy `generate-skills` refresh is harmless housekeeping, not a correctness gate for this PR.

### 5. rtech-hermes runtime/plugin — N/A, with evidence (the load-bearing check)

Repo lives at `/Volumes/ThunderBolt/Development/ai-agents/platforms/rtech-hermes`.

WRITTEN — **Hermes never calls the two tools this PR changes.** Its read allowlist is closed:

```python
READ_TOOL_ALLOWLIST = {"get_entry", "search_all", "list_repo_facts", "lane_load", "session_context"}
WRITE_TOOL_ALLOWLIST = {"append_session_event"}
```
`plugins/memory/openbrain/__init__.py:53-60`

`search_brain` and `brain_answer` are absent. Recall goes through `search_all` (`:666-675`, `:1746`), which this PR does not modify. Grep for `search_brain|brain_answer` across `plugins/`, `packages/`, `tests/` returns only prose gate-hint strings in `confidence_streak_gate.py` / `fact_check_gate.py`, not call sites.

WRITTEN — the one hardcoded table tuple is scoped to a **different tool**:

```python
_ENTRY_TABLES = ("thoughts", "decisions", "relationships", "projects", "sessions")
```
`plugins/memory/openbrain/__init__.py:80`, used only at `:1463` and `:1469` to drive `get_entry` probing in `_resolve_source_type`. `get_entry` is untouched by this PR. Session events are not `get_entry`-addressable, so no change is needed.

WRITTEN — response handling tolerates an unknown `source_type`. Both formatters read it as a free-form string with fallbacks and no validation or membership check: `plugins/memory/openbrain/support.py:1071` and `:1082-1100`. Runtime models are frozen dataclasses (`openbrain/models.py`), not strict pydantic — no `extra="forbid"` rejection path.

WRITTEN — **pre-existing, unrelated to this PR, worth flagging:** `packages/rtech-hermes-runtime/src/rtech_hermes_runtime/openbrain/contract.py:50` pins `REQUIRED_SCHEMA_HASH = "4df9c742..."` and the fallback `CURRENT_CONTRACT_VERSION = "2026-07-08.memory-tools.v20"` (`:25`). Live Open Brain serves `2026-07-23.memory-tools.v23`, `schema_hash 4b69e9b4...` (RUNNING, via `get_contract` on `127.0.0.1:3100`). That constant is module-level and applies even when `openbrain-memory` is installed. This drift exists on `main` today and **PR #610 does not move it** — see next section.

### Proof that #610 cannot move `schema_hash`

RUNNING — the two files feeding `contractHash()` are **byte-identical** across `main` and the PR branch (same git blob SHAs):

```
IDENTICAL src/contract.ts          (2583ebb1d1bff72296870f16ab7b4cb1802dbc0f)
IDENTICAL src/contract-schemas.ts  (d51abeb15e8d908a43618a49d218f3c855cdbaf9)
```

Neither appears in `git diff --name-only origin/main...HEAD`. `search_brain` and `brain_answer` are **not** among the 21 tools in `TOOL_CONTRACTS` (`src/contract-schemas.ts`), so the manifest does not describe them at all. Computed hash from the working tree matches the live service exactly: `4b69e9b437c96175531b049b6e3c2782f383334e9e1931e96e73835599e4a4a8`.

Consequence: no downstream contract-validation gate can newly fail because of this PR.

### 6. Hermes live rollout — N/A

Follows from step 5: no Hermes code path reaches the changed tools, and the contract hash is unmoved. No agent update run is required for this delta.

### 7. Live agent canaries — N/A

Same reason. No agent-visible behavior changes: Hermes recall uses `search_all`, which continues to filter `ALL_TABLES` and does not gain the session-event corpus.

### Verdict

**Safe to merge from a downstream perspective.** The delta is additive on a tool no downstream consumer calls, no client enumerates Open Brain search sources, no client strictly validates `source_type`, and the contract manifest inputs are byte-identical so `schema_hash` cannot move.

Remaining post-merge, in order:
1. Deploy core01 (`CI` `workflow_dispatch` on `main`, `deploy_core01=true`) — step 2 is blocked on this.
2. After deploy, re-run the hosted canary: `search_brain {"table":"session_events"}` must succeed where it currently returns `-32602`, plus one `brain_answer` call returning a `source_type: "session_event"` result.
3. Optional housekeeping: `mcp2cli generate-skills open-brain --conflict=merge` (no stale enum present, so not a gate).

Not blocking this PR, tracked separately: the stale `REQUIRED_SCHEMA_HASH` / v20 pin in `rtech-hermes` contract.py vs the live v23 contract.

PR #615, opened 2026-08-08
feat(pr): hold the PR template against its own validator, add pr-scribe

rodaddy:
## Summary

Three lanes failed `scripts/validate-pr-body.ts` three different ways in one
night. Every one was a formatting failure, not a thinking failure:

1. the body pasted inside a ``` code fence;
2. `- **Highest-risk behavior:**` — the `**` between the dash and the label
   breaks the `/^-\s*<label>:/` anchor (`scripts/validate-pr-body.ts:32-47`), so
   the field reads as empty however much text follows it;
3. `## Review Gate` missing entirely.

**A finding that changed the shape of this change:** the committed
`.github/pull_request_template.md` was ALREADY structurally correct — the first
run of the new gate passed clauses 1 and 2 against the pre-existing file,
before any template edit. The template was never the defect. The real gaps were
that nothing *proved* it stayed correct as the validator evolved, and nothing
put it in a lane's hands at the moment the body gets written. So this PR does
not rewrite the template. It pins it and routes lanes to it.

No generator script was added. The design allowed one "if the validator's
requirements are cleanly extractable" — they are, but a generator would emit a
file that already exists byte-for-byte, so the smaller owning-boundary change is
the drift check that clause 2 performs directly against the validator source.

## Verification

`bunx tsc --noEmit` exits 0. No Python, MCP, transport, or migration surface is
touched, so the package checks and the live smoke are not applicable.

### RED — before `.claude/agents/pr-scribe.md` existed

```
CLAUSE 1 (filled template passes the validator): PASS — filled template validated: exit=0 — PR body validation passed for done-means pr-template gate.
CLAUSE 2 (template and check track the validator): PASS — template and check both current with validator: 2 sections + Contract Parity, 9 labels cross-checked in both directions
CLAUSE 3 (pr-scribe agent runs the validator): FAIL — no agent definition at .../.claude/agents/pr-scribe.md
EXIT=1
```

Clause 3 is the only one that was red, because clauses 1 and 2 were already
satisfied by the existing template. That is reported rather than engineered
away — manufacturing a red by first breaking the template would have tested the
mutation, not the repo.

### Mutation proof — clauses 1 and 2 CAN fail

A gate that has only ever been green is not evidence. Each real-world failure
mode was reintroduced into the template and the gate re-run:

```
### MUTANT A: bold the label (real failure mode 2)
CLAUSE 1: FAIL — filled template REJECTED (exit=1): - Critical Self-Review field 'Highest-risk behavior' needs specific content.
CLAUSE 2: FAIL — drift: template missing label line '- Highest-risk behavior:';

### MUTANT B: rename ## Review Gate (real failure mode 3)
CLAUSE 1: FAIL — filled template REJECTED (exit=1): - Missing '## Review Gate' section.
CLAUSE 2: FAIL — drift: template missing section '## Review Gate';

### MUTANT C: whole body in a code fence (failure mode 1)
CLAUSE 1: PASS
CLAUSE 2: PASS
```

Mutant C passing is a real result, kept here rather than hidden. `section()`
matches on a trimmed line, so a fence genuinely does not affect the validator —
failure mode 1 was a lane hand-composing a body in chat, never a template
defect. A check asserting "no fences in the template" would encode a wrong
theory of that failure. It is addressed where it actually lives: the pr-scribe
agent writes the body to a file and `cat`s it, and says so explicitly.

### GREEN — full gate, and from a clean clone

```
CLAUSE 1 (filled template passes the validator): PASS — filled template validated: exit=0 — PR body validation passed for done-means pr-template gate.
CLAUSE 2 (template and check track the validator): PASS — template and check both current with validator: 2 sections + Contract Parity, 9 labels cross-checked in both directions
CLAUSE 3 (pr-scribe agent runs the validator): PASS — .../.claude/agents/pr-scribe.md names validate-pr-body.ts 6 time(s) — the unskippable final step is still written down
CLONE_EXIT=0
```

The clean-clone run is the one that matters: it caught that `.gitignore:53`
excludes `.claude/*`, which would have left `pr-scribe.md` untracked. Clause 3
would have passed on this machine and failed for every worktree, worker, and
fresh clone — the exact places a PR body actually gets written. Fixed with a
`!.claude/agents/` re-include following the existing `!.claude/hooks/` pattern.

## Critical Self-Review

- Highest-risk behavior: the `sed` fill in clause 1 of the gate. It emulates how a lane fills the template, and if it silently stops matching (a template line reshaped so the `^\(- [^][]*:\)[[:space:]]*$` anchor misses) the clause could pass while proving nothing about a real filled body. Mutants A and B demonstrate it still discriminates today; the ordering hazard is called out in the script comments because the disposition rewrite MUST precede the generic fill or both sides of an either/or get checked.
- Assumptions that could be wrong: that clause 2's extraction of validator labels via `awk`/`grep` over the source keeps working. It is brittle to refactoring — if the `for (const label of [...])` literal becomes a computed list, extraction returns empty. That case is handled: empty extraction calls `fail_hard` (exit 3, harness error) rather than passing vacuously, so it fails loudly instead of silently.
- Missing/weak tests: there is no automated test that the pr-scribe AGENT behaves as written — clause 3 only asserts the file names the validator, which is a proxy for the behavior, not the behavior. Agent prose is not executable, so the enforceable half is the script it must call; that boundary is deliberate and stated in the agent file.
- Security/permission risk: none. No auth, namespace, SQL, token, or network path is touched. The gate reads two repo files and shells out to `bun` on a local script.
- Migration/deploy risk: none — no schema, no migration, no service. Nothing is deployed by this change.
- Downstream client/runtime risk: none applicable per `docs/downstream-rollout.md` "When This Applies" — no MCP tool name, input schema, output shape, annotation, auth, or error envelope; no transport/session behavior; no externally visible migration; no `python/openbrain-memory` behavior or exports; no generated `skill/` content; nothing an mcp2cli or Hermes consumer calls. This is repo-local authoring tooling.
- Rollback/cleanup concern: `git revert` is sufficient and complete — four additive files, no state, no data. The worktree at `/Volumes/ThunderBolt/_tmp/open-brain/_worktrees/pr-scribe/` is removed by this lane in the same session that created it.
- Fixes made before PR: found and fixed the `.gitignore` trap via the clean-clone run (agent file would never have been committed); corrected an initial mutation test that used BSD `sed -i ''` syntax against GNU sed 4.10 — it errored out and I nearly recorded the resulting unchanged-file "PASS" as evidence the mutants were caught, which would have been a false green in the one place this PR claims rigor.
- Known residual risk: clause 2 duplicates the validator's field list inside the check by design, so the two must be edited together. That duplication is the mechanism, not an oversight — direction (b) cross-checks the list against the validator source precisely so the duplicate cannot rot silently. A validator refactor that defeats the `awk` extraction degrades to exit 3, which is visible.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this change originates from a ledger item about authoring ergonomics, not from a MEDIUM+ review finding about repo code behavior.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no Open Brain server, database, MCP, or client surface is exercised by this change

## Contract Parity

- Contract parity: [x] runtime-specific because: no path in `contracts/parity-paths.txt` is touched, verified with `git diff --quiet origin/main...HEAD -- $(cat contracts/parity-paths.txt)`, so CI computes `CONTRACT_PARITY_REQUIRED=false` and this section is informational

## Downstream Rollout

Notes/evidence:

- Not applicable, stated explicitly as the doc requires. The diff is four files: a bash acceptance gate, a repo-local agent definition, a `.gitignore` re-include, and a docs paragraph. No consumer-facing surface.

## How drift is prevented

The template and the validator cannot drift apart without turning the gate red,
and the check is cross-verified in BOTH directions:

- **(a) template → check.** Every heading and label the check enumerates must be
  present in the template, matched with the same `^-\s*Label:` anchor shape the
  validator uses — so a bolded label fails here exactly as it fails there.
- **(b) check → validator.** Every section name the validator passes to
  `section()` and every label in its `requireSpecificLine` array is extracted
  from `scripts/validate-pr-body.ts` at run time and must appear in the check's
  own list. Adding a required field to the validator alone turns the gate red
  instead of passing silently.

`## Contract Parity` is deliberately excluded from the unconditional required
set — it is gated by `options.contractParityRequired` — but its presence in the
template is still asserted, because a contract-touching PR is validated with the
flag on and a lane starting from a template without that section would fail on
shape all over again.

## Claim states

- WRITTEN: all four files, committed on `feat/pr-template-and-scribe`.
- RUNNING: the acceptance gate — exit 0 in the worktree AND from a clean
  `git clone` of the committed tree; the three mutation runs; `bunx tsc --noEmit`
  exit 0; this PR body itself passing `scripts/validate-pr-body.ts` locally
  before `gh pr create`.
- PROPOSED: that lanes will actually reach for `pr-scribe`. Nothing forces an
  agent to invoke it — the enforceable part is that CI still rejects a bad body
  regardless, which is unchanged by this PR.
- Not merged, not deployed.

### DIFF
--- a/.claude/agents/pr-scribe.md
+++ b/.claude/agents/pr-scribe.md
@@ -1,0 +1,101 @@
+---
+name: pr-scribe
+description: Composes an open-brain PR body from a lane's actual evidence and proves it passes scripts/validate-pr-body.ts before returning it. Use when a lane is about to run gh pr create, when a PR body was rejected by the PR Body check, or when asked to write or fix a PR description in this repo.
+model: opus
+effort: medium
+tools: Bash, Read, Write, Edit
+---
+
+You write the PR body for a lane that has finished its work, and you do not
+return one that fails the repo's own validator.
+
+You are a scribe, not a reviewer. You do not judge whether the work is good,
+and you do not fix code. You take what the lane actually did and render it in
+the shape `scripts/validate-pr-body.ts` accepts.
+
+## The one rule
+
+**Never return a body you have not just seen pass.** The last thing you do,
+every time, is:
+
+```bash
+PR_BODY="$(cat <body-file>)" PR_TITLE="<title>" bun scripts/validate-pr-body.ts
+```
+
+If it exits non-zero, fix the body and run it again. Only a body that has
+produced exit 0 in THIS session may be returned, and you paste that passing
+output with it. The agent wraps the script; the script never trusts the agent.
+
+Set `CONTRACT_PARITY_REQUIRED=true` as well when the diff touches any path in
+`contracts/parity-paths.txt` — CI computes that flag from the diff
+(`.github/workflows/pr-body.yml`), so a body that only passes without the flag
+will still fail on GitHub.
+
+## Start from the template, never from memory
+
+Read `.github/pull_request_template.md` and fill it in. It is checked against
+the validator by `scripts/done-means/pr-template-passes-validator.sh`, so a
+body that starts there cannot fail on shape. Reconstructing the sections from
+memory is how the three known failures happened:
+
+- **A bolded label breaks the anchor.** `requireSpecificLine` builds
+  `/^-\s*<label>:/` (`scripts/validate-pr-body.ts:32-47`). `- Highest-risk
+  behavior:` matches; `- **Highest-risk behavior:**` does not, and the field
+  reads as empty however much text follows it. No bold, no italics, no
+  backticks on a required label.
+- **A missing section is a hard error.** `## Review Gate` and
+  `## Critical Self-Review` are looked up by exact heading text
+  (`scripts/validate-pr-body.ts:15-30`). Do not rename, renumber, or nest them.
+- **Do not fence the body.** Wrapping the whole PR body in a ``` fence is the
+  symptom of composing it in chat instead of in a file. Write the body to a
+  file and `cat` it into `PR_BODY`. Fenced code blocks are fine INSIDE a
+  section for transcripts and evidence.
+
+Each either/or line takes exactly one side. `[x] linked below` **or**
+`[x] not applicable because: <real reason>` — never both, never neither, and
+the reason may not be one of `-`, `n/a`, `na`, `none`, `todo`, `tbd`
+(`scripts/validate-pr-body.ts:9`).
+
+## Fill it from evidence, not from adjectives
+
+The validator only checks that a field is non-empty and non-placeholder. It
+cannot tell whether the content is real, and that is precisely the part you are
+responsible for. Every field is a claim about this specific diff.
+
+- **Critical Self-Review** — nine fields, each naming a concrete thing about
+  THIS change. "Highest-risk behavior" is a named function, endpoint, or
+  migration and what it could do wrong, not "low risk overall". "Missing/weak
+  tests" names what is not covered; if the answer is genuinely nothing, say
+  which tests do cover it and why that is sufficient. Ask the lane for what it
+  ran and what it observed; do not invent a receipt.
+- **SME review-memory update** — if the change came from a review finding at
+  MEDIUM or above, `docs/sme/` gets the pattern and you check `updated`.
+  Otherwise check `not applicable because:` with the actual reason.
+- **Downstream Rollout** — classify against `docs/downstream-rollout.md`
+  "When This Applies": MCP tool names/schemas/output shapes/auth/namespace
+  semantics/error envelopes, transport and session behavior, externally visible
+  migrations, `python/openbrain-memory` behavior or exports, generated `skill/`
+  content, or anything an mcp2cli/Hermes consumer calls. If none apply, the doc
+  requires you to say so explicitly rather than leave it blank.
+- **Contract Parity** — required when the diff touches
+  `contracts/parity-paths.txt` paths. `fixtures updated` or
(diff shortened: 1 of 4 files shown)


rodaddy:
CI note — one `check` failure was observed and is flake, established by observation rather than assumption:

- Same commit `47ffe4d6` produced BOTH outcomes: the `push` run (31237077888) passed `check`; the `pull_request` run (31237083408) failed it. Same code, same workflow, opposite results.
- Signature: `error: Cannot call describe() after the test run has completed` at `clients/ts/tests/receipts.test.ts:18` — a runner ordering/timing error, in a file this PR does not touch (diff is 4 files: a bash gate, an agent definition, .gitignore, a docs paragraph).
- `bun test clients/ts/tests/receipts.test.ts` in isolation: 5 pass, 0 fail.
- `gh run rerun --failed` on the same commit: green, no failing jobs.

Two things I checked and will NOT overclaim: the merge-base `90ef4c21` on main also fails CI, but with a DIFFERENT signature (`db-integration`, a `toBeLessThan` assertion), so it is a separate problem and not evidence for this one. I also did not find this `describe()` signature in other recent runs, so I am not calling it a known recurring flake — only that it did not reproduce on re-run or locally.

All checks now pass. Not merging.

Issue #646, opened 2026-08-08
capture-gate live defects: receipt probe hits a nonexistent endpoint (HTML 404), and the provider scope-proof demands a key it rejects


rodaddy:
First live firing of capture-gate.ts (PR #642, ledger item 24) on the controller's merge of PR #645 exposed the two gaps that PR's own report named as PROPOSED (fixture-proven only):

1. **Receipt probe endpoint does not exist on live Open Brain.** The gate's refusal text shows the service answering `Cannot POST /too...` (HTML error page) — the receipt query URL is wrong against the real server, so the gate can NEVER see a real receipt; every merge with the service up is refused as a skip.
2. **Provider scope-proof is self-contradictory from a head session.** `openbrain-memory` capture fails with "session_start result did not prove exact Open Brain scope: agent, channel_id, server_id, source" — but including `source` in scope fails with "scope contains unsupported keys: source". It demands a key it refuses to accept, so a head session cannot produce the receipt the gate requires by any spelling.
3. **Drain path degraded in the hook environment:** `runtime unavailable: No module named 'openbrain_memory'` — the gate's drain-first step cannot import the client package from the hook's interpreter.

Combined effect: the gate structurally blocks ALL merges including the merge of its own fix — an enforcement wedge. Operator ruling pending on interim disposition (unregister-and-fix-forward vs hold-all-merges). Fix requirements, red-first: a done-means clause that queries the REAL service for a receipt written through the REAL provider path (the seeded-spool drain-delivers clause PR #642 already named), plus a provider fix or documented scope spelling a head session can satisfy, plus the hook interpreter gaining the client package. Related: #451 (remains open for exactly these live gaps).

rodaddy:
Fixed and MERGED (PR #650, controller-verified via verify-lane receipt at 9b068e0 against the LIVE service). Root cause was two defects: server/tools/session-lifecycle.ts returned existing lanes without exact-scope establishment (2,011 lanes on the dogfood DB could never accept a capture — they now heal lazily on next capture; bulk repair remains an operator option), and the provider reported mismatches in response vocabulary, telling operators to send 'source' which the request validator rejects (now reports the request vocabulary: 'platform'). The third finding from the original report — capture-gate's hook interpreter lacking openbrain_memory — is moot: that gate is retired (ledger item 25, PR #649). Live proof: capture into a real head lane returned durable:true status:saved. Service redeploy to merged main is the remaining operational step, happening now.

PR #650, opened 2026-08-08
fix(memory): let an existing lane establish exact scope, and stop naming a key requests may not carry (#646)

rodaddy:
## Summary

- A head session could not capture to Open Brain **by any spelling**. Capture failed with `session_start result did not prove exact Open Brain scope: agent, channel_id, server_id, source`, but adding `source` — the key that error names — failed with `scope contains unsupported keys: source`. Two independent defects produced that contradiction; both are fixed here.
- **The capture failure (server).** The live `session_start` (`server/tools/session-lifecycle.ts`) returned an existing lane verbatim, so a lane opened without the full exact-scope predicate could never afterwards prove it — every capture into it was lost, permanently. **2011 such lanes existed on the dogfood database** when measured. This ports the exact-scope establishment `src/tools/session-start.ts:34-90` already had into the handler the running service actually serves. `src/` was the tree everyone reads; `server/main.ts` is the tree that runs.
- **The misleading error (client).** The server stores `platform` under the lane column `source`, so the comparison correctly uses `source` — but `validate_exact_fields` also *reported* it, telling operators to send a key `_SCOPE_KEYS` rejects. It now takes `report_names` so an error names only keys a request may carry (`platform`), per `docs/agent-context-pack-contract.md:99-105`.
- The backfill is one-way and is **not** a scope widening: `COALESCE` only fills NULLs, and a lane whose established scope disagrees is refused by name rather than silently re-pointed. The contract requires exactly that (`docs/agent-context-pack-contract.md:537`, "different agent is denied").

## Verification

- Done-means: scripts/done-means/646-provider-scope.sh

**RED** (pre-fix service `552cf48`, corrected fixture) — clauses (a), (b), (c) fail, control (d) passes:

```
(a) contradiction absent       : FAIL — CONTRADICTION PRESENT — demand: [session_start result did not prove exact Open Brain scope: channel_id, server_id, source] / reject: [scope contains unsupported keys: source]
(b) capture succeeds end-to-end: FAIL — status=lost durable=false rows=0
(c) errors name accepted keys  : FAIL — scope-proof error names 'source', which no request may carry
(d) control, healthy stays ok  : PASS — fresh-lane capture still healthy: status=saved durable=true rows=1
DONE-MEANS 646: FAIL
```

**GREEN** (fixed service `9b068e0`, same fixture, live against 127.0.0.1:3100):

```
(a) contradiction absent       : PASS — no demand for 'source'; rejecting the unknown key remains correct
(b) capture succeeds end-to-end: PASS — status=saved durable=true, ob_session_events rows=1 in the partial-scope lane
(c) errors name accepted keys  : PASS — failure surfaced without the misleading scope-proof wording
(d) control, healthy stays ok  : PASS — fresh-lane capture still healthy: status=saved durable=true rows=1
DONE-MEANS 646: PASS
```

The gate discriminates by revision, not by luck: it was run against **both** deployed revisions (`552cf48` -> FAIL, `9b068e0` -> PASS), with the control green in both, using `local-clone-deploy.sh`'s revision proof each time.

- Server regression tests are red-first: with the fix stashed, the two new `#646` tests fail while the other four pass and 23 assertions still execute (proving harness health, not a collection error). With the fix: `6 pass, 0 fail, 30 expect() calls`.
- Python regression is mutation-proven: removing `report_names={"source": "platform"}` turns the new test RED (2 failed), restoring it turns it GREEN.
- Full suites on the rebased branch: `bun run test:isolated` -> **3720 pass, 0 fail** across 242 files; `uv run pytest -q` -> **574 passed, 14 skipped**; `mypy` and `ruff` clean; `bunx tsc --noEmit` clean.
- **Real-world proof, not just fixtures:** a capture into an actual head-created lane (`09ea2a7d...`, `agent=openbrain-capture` with NULL scope columns) returned `durable:true`/`status:saved` and backfilled the lane to `claudex|verify-646|local`. That row was removed afterwards.

## Critical Self-Review

- Highest-risk behavior: the `UPDATE` in `establishExactStartScope`. It writes to a table row that another session may own. Mitigated by the guard clause — every field is `COALESCE`d (NULL-only fill) and the `WHERE` refuses the update if any already-set coordinate disagrees, so a foreign lane is left byte-identical and reported. The conflict test asserts exactly that: after a refusal, `agent` and `source` are unchanged.
- Assumptions that could be wrong: that `agent` matching is the right predicate for backfill. The contract says a different agent is denied (`:537`), so I followed it rather than my first fixture, which assumed the opposite and was wrong. If the intended semantics are "the lane's creator may be superseded", clause (b)'s fixture and the guard both need an operator ruling — flagged, not silently chosen.
- Missing/weak tests: no test covers `thread_id` backfill on an existing lane (the `thread_id` arm of the guard is simplified from `src/`'s more elaborate version — I kept it strictly narrower, never wider). No concurrency test for two sessions racing to establish scope on the same lane; Postgres row locking makes the `UPDATE` atomic, but last-writer-wins between two *different* complete scopes is untested.
- Security/permission risk: the scope predicate is an isolation boundary, so a wrong change here leaks one session's memory into another's context. The change cannot widen it: it can only turn NULL into a value, and only when the request already matches every non-NULL coordinate. `auth.namespace` (not a caller-supplied namespace) remains the outer predicate, unchanged.
- Migration/deploy risk: none — no schema change, no migration. Deploy is the ordinary `local-clone-deploy.sh` path; rollback is `--rollback` or redeploying the prior ref, both exercised this session (`552cf48` <-> `9b068e0`, revision proof passing each way).
- Downstream client/runtime risk: **not applicable by the doc's own trigger.** `docs/downstream-rollout.md:12` scopes rollout to MCP tool names, input schemas, output shapes, annotations, auth, and namespace behavior. None change: `inputSchema` and `registerTool` are untouched (verified by diffing those lines), the success envelope keeps its `lane`/`events`/`events_returned`/`is_new` shape, and the new refusal reuses the existing `errorResult` path already present in `src/`. The only client-visible difference is that a previously-unprovable lane now succeeds.
- Rollback/cleanup concern: the done-means check seeds and removes its own lanes under randomized `session_key`s carrying a run id, and fails hard if it cannot reach the database, so a PASS can never come from a run that could not clean up. All probe rows and lanes from this lane's investigation were removed; the 5 remaining rows matching `%646%` are pre-existing real memories that merely contain the digits, deliberately left alone.
- Fixes made before PR: (1) the done-means fixture seeded a *conflicting* agent and read the server's correct refusal as a failure — corrected to the real-world shape and RED re-proven afterwards; (2) the shared test `pool` was closed in the first suite's `afterAll`, making the new suite fail with `Cannot use a pool after calling end`, which reads exactly like a real red — teardown moved to the last suite.
- Known residual risk: the 2011 pre-existing partial lanes are *repairable* but not repaired — each one heals on its next capture, and none is backfilled proactively. That is deliberate (no bulk data migration in a bug fix), but it means the fleet converges lazily rather than at deploy.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the two reusable lessons here are operating knowledge for lanes (a done-means fixture that encoded the wrong real-world shape; a shared pool closed by an earlier suite faking a red), which `docs/lane-contract.md` Tightenings owns. They are in the lane report for the controller's mandatory harvest.

## Review Gate

Live evidence is the RED/GREEN transcripts above, both taken against the running dogfood service on `127.0.0.1:3100` with `local-clone-deploy.sh` revision proofs (`pid 25323 -> 28052 -> 96387`).

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface changed. Tool name, `inputSchema`, annotations, and the success envelope are untouched; this fixes handler behavior for an already-published shape, and the parity fixtures encode the shape, not the branch taken.

## Downstream Rollout

Notes/evidence:

- All four are **not applicable**, on the doc's stated trigger (`docs/downstream-rollout.md:12`): no MCP tool name, input schema, output shape, annotation, auth, or namespace behavior changes. Downstream clients cache *schemas*; the schema is byte-identical, so there is nothing to refresh. Verified by diffing the `inputSchema`/`registerTool` lines in `server/tools/session-lifecycle.ts` — zero changed lines.

### DIFF
--- a/python/openbrain-memory/src/openbrain_memory/_runtime_validation.py
+++ b/python/openbrain-memory/src/openbrain_memory/_runtime_validation.py
@@ -92,7 +92,19 @@ def validate_started_lane(
         candidate["server_id"] = metadata["server_id"]
     expected = exact_scope_fields(namespace, scope)
     expected["source"] = expected.pop("platform")
-    validate_exact_fields(candidate, expected, "session_start result")
+    # The server stores/returns `platform` under the lane column `source`
+    # (src/tools/session-start.ts:9,45,84), so the COMPARISON must use the
+    # response spelling. The REPORT must not: `source` is not in the contract's
+    # scope vocabulary (docs/agent-context-pack-contract.md:99-105, restated as
+    # a closed set at :671-674) and `_SCOPE_KEYS` (runtime.py) rejects it on
+    # input. Naming it told operators to add a key the provider then refused —
+    # the #646 contradiction, where no spelling could succeed.
+    validate_exact_fields(
+        candidate,
+        expected,
+        "session_start result",
+        report_names={"source": "platform"},
+    )
 
 
 def validate_context_pack_scope(
@@ -677,10 +689,19 @@ def validate_exact_fields(
     candidate: Mapping[str, Any],
     expected: Mapping[str, Any],
     label: str,
+    report_names: Mapping[str, str] | None = None,
 ) -> None:
-    """Require every expected field to be present with its exact value."""
+    """Require every expected field to be present with its exact value.
+
+    ``report_names`` maps a comparison key to the name the CALLER may actually
+    supply, for the cases where the server's response spelling differs from the
+    request vocabulary. An error must only ever name keys a request can carry;
+    naming a response-only key is an instruction the validator itself rejects
+    (#646).
+    """
+    renames = report_names or {}
     mismatches = sorted(
-        name
+        renames.get(name, name)
         for name, value in expected.items()
         if name not in candidate or candidate[name] != value
     )
--- a/python/openbrain-memory/tests/test_cli_request_diagnostics.py
+++ b/python/openbrain-memory/tests/test_cli_request_diagnostics.py
@@ -133,6 +133,44 @@ def test_operator_capture_rejects_unproven_start_scope(
     assert client.appended is None
 
 
+@pytest.mark.parametrize("client", [SparseStartClient(), NullScopeStartClient()])
+def test_scope_proof_error_names_only_request_vocabulary(
+    client: SparseStartClient,
+) -> None:
+    """The scope-proof error must never name a key a request may not carry.
+
+    Regression for #646. The server stores `platform` under the lane column
+    `source`, and the validator compares using that response spelling -- but it
+    also REPORTED it, so an operator was told to add `source` to their scope
+    while `_SCOPE_KEYS` rejected exactly that key ("scope contains unsupported
+    keys: source"). The provider demanded a key it refused, so a head session
+    could not capture by any spelling. The contract vocabulary is `platform`
+    (docs/agent-context-pack-contract.md:99-105).
+    """
+    output = execute_json(
+        {
+            "config": _CONFIG,
+            "operation": "capture",
+            "content": "Scope proof error vocabulary",
+            "distilled": True,
+            "event_type": "fact",
+            "scope": _FULL_SCOPE,
+        },
+        client=client,
+    )
+
+    error = str(output["receipt"]["error"])
+    named = error.split("did not prove exact Open Brain scope:", 1)[1]
+    reported = {part.strip() for part in named.split(",") if part.strip()}
+
+    assert reported, f"no scope keys were named: {error}"
+    # Every named key must be one a caller may actually send.
+    assert reported <= set(_FULL_SCOPE) | {"namespace"}, error
+    # The specific false instruction that made #646 unfixable from the outside.
+    assert "source" not in reported, error
+    assert "platform" in reported, error
+
+
 def test_operator_capture_rejects_sparse_lane_before_v2_dispatch() -> None:
     """A server below session_start v2 cannot supply handshake scope proof."""
     client = UnderVersionedSparseStartClient()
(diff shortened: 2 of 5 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/646-provider-scope.sh exit=0 sha=9b068e030315f398b057fc94b695e2d8e47c3bb4 at=2026-08-08T18:54:17.579Z

```
=== done-means #646 — provider scope contradiction (run 77f49841cd65) ===
  (a) contradiction absent       : PASS — no demand for 'source'; rejecting the unknown key remains correct: [scope contains unsupported keys: source]
  (b) capture succeeds end-to-end: PASS — status=saved durable=true, ob_session_events rows=1 in the partial-scope lane
  (c) errors name accepted keys  : PASS — failure surfaced without the misleading scope-proof wording: [Open Brain tool returned an error context=call_tool:session_start body=Existing lane exact scope does not match session_start request]
  (d) control, healthy stays ok  : PASS — fresh-lane capture still healthy: status=saved durable=true rows=1
DONE-MEANS 646: PASS
```

</details>

rodaddy:
harvested: c72a753 — round-12 Tightenings on wip/2026-08-07.

PR #648, opened 2026-08-08
feat(647): make a silent capture lane loud, per role, in /health

rodaddy:
## Summary

- Closes #647 (ledger item 25). The capture merge-gate was retired because gating a human's PR is the wrong instrument for a background pipeline. This is the replacement: **liveness on the automatic lane**, the #625 pattern applied to the second background lane.
- **This is not a new design — it is an existing one made executable.** `docs/decisions/capture-never-drops-a-turn.md` has specified the correct capture health check since 2026-07-28 (`:182-200`) and records at `:280` that "the check it asks for was never run". This PR runs it.
- New `openbrain.apps.capture.liveness`: a **pure function over counts**. It opens no file, holds no lock, and takes no clock of its own. That shape is deliberate — `outage.py`'s module docstring records what happened the last time a capture-path component held a lock another waited on: a **healthy** hook stalled 31.4s, past the 5s Stop deadline, and the turn delivered zero batches ("the turn was DROPPED by its own telemetry"). A liveness reader that could do that would be strictly worse than the blindness it removes.
- Three faults, all count-derived: watermark advancing **zero bytes while sessions run**; spool **accumulating with no outage announced** (`OutageLatch.note_spooled` returns `None` for three distinct reasons, so a latch that never fires is indistinguishable from a healthy one at the call site — `outage.py:411`); and any **role** delivering nothing.
- **Per role is load-bearing, not decoration.** `capture-never-drops-a-turn.md:188-200` specifies it and explains why with a measurement: the operator's numbers stayed healthy for six days while the assistant side sat at zero (#447). On 2026-08-02 that was 365 user rows against 13 assistant rows — a lane-total counter reads 365 turns and reports busy traffic. Summing across roles rebuilds the exact blind spot this exists to remove.
- `server/transport/health.ts` composes it exactly as `producerHealth` is composed (health.ts:129-134), **including the absence rule**: a process composing no capture lane publishes no block and stays green. Absent means "not my job"; `stale: true` means "my job and I am not doing it".
- **No wall-clock assertion anywhere.** `silence_seconds` is reported and never compared (lane-contract Tightenings round 5).

## Verification

- Done-means: scripts/done-means/647-capture-liveness.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

[...]

Issue #647, opened 2026-08-08
Automatic-capture liveness check: prove the raw capture lane delivered for recent sessions, loud on silence (replaces the retired capture merge-gate)


rodaddy:
Ledger item 25 (docs/issue-graph.md): the capture merge-gate tier is retired — raw capture is automatic (Stop hooks → watermark → durable spool) and distillation is the DREAM pipeline's designed job (docs/dream-design.md). The enforcement that replaces it is LIVENESS on the automatic lane, the #625 pattern: a wedged or silently-dead capture pipeline must be loudly distinguishable from a healthy one.

Shape (delta on existing designs, not an invention): a deterministic check that recent sessions' capture events actually reached the server (or are sitting in the spool with the outage latch announced) — asserting on event counts/watermark advancement, NEVER wall-clock assertions (lane-contract round 5). Loud on: watermark not advancing while sessions run; spool accumulating with no outage latch; zero capture events across N recent active sessions. Control clause required: healthy path stays healthy (#624 clause-z / round-8 rule). Related: #646 (provider scope contradiction), #451 (hydration stamp + recall telemetry remain), #625 (the pattern's precedent).

PR #652, opened 2026-08-08
feat(652): compose the capture-liveness reader into live /health

rodaddy:
## Summary

- Closes #652. PR #648 (issue #647) merged the capture-liveness reader and the `/health` plumbing that accepts it — and **its own report named what was left undone**: "no process composes `captureHealth` yet, so no live `/health` reports capture today." An optional input that no composition root supplies is dead code with a docstring. This composes it.
- The gap was verifiable, not a matter of opinion. At `origin/main`, `captureHealth` appeared in exactly two places, both inside `server/transport/health.ts` (the input declaration at `:138` and the read at `:194`). The only composition root — `createShadowApplication` (`server/application/index.ts:114`), which `server/main.ts:280` calls to build the serving process — supplied `producerHealth` (`:159`) and nothing at all for capture.
- **`server/capture/liveness-observer.ts`** gathers the counts from the only vantage point a SERVER has: arrivals in `ob_raw_turns`, grouped **per role**, exactly as `docs/decisions/capture-never-drops-a-turn.md:193-197` publishes the check. The per-role grouping is load-bearing rather than stylistic — `:291` records 365 user rows beside 13 assistant rows over the six days #447 hid behind a healthy-looking lane total.
- **Expected roles are seeded at zero before the rows fold in.** A role that delivered nothing has no rows and therefore no group, so a fold over returned rows alone reports a busy lane. Without the seed the silent-role fault is structurally unreachable — the dead speaker is precisely the one with no rows to group. `server/capture/liveness-observer.test.ts` pins it, and the test was proven to fail with the seed removed.
- **`server/application/index.ts` places the reading LATE-BOUND**, in the same shape and for the same stated reason as `producerHealth` immediately above it: the closure is invoked per request, so `/health` publishes what the capture lane looks like now rather than at boot — which is the stale-but-confident answer the feature exists to replace.
- **Honest about what this vantage point cannot see, announced rather than silent** (AGENTS.md). The watermark store and outage latch are strictly single-key with no enumeration (`watermark.py:196-245`, `outage.py:375-440`) and are per-hook CLIENT-side files a server cannot read at all — which is the blocker #648's residual-risk field named. So the spool fault stays unobservable from here and is reported absent rather than fabricated, and delivered turns stand in for bytes advanced. That substitution is deliberate: a literal `0` there IS the wedged-watermark fault, so a gatherer that passed zero because it cannot measure bytes would degrade every healthy deployment on earth. The substitution preserves the property the field is about rather than its units (round-10 lesson on picking a neutral value that keeps the property under test intact).
- **Absence stays absence** (round-13's rule, honored as briefed): a deployment that legitimately runs no capture lane composes no observer, publishes no block, and stays green. A composed observer that has seen nothing at all returns `undefined` and is treated identically — a fresh process and a quiet night are not evidence of a dead lane.
- **No wall-clock verdict anywhere.** `silence_seconds` is computed and carried for an operator to read; nothing compares it to a threshold. A test asserts a full day of reported silence still reads healthy.
- Two type exports were added to the transport barrel (`TransportCaptureHealth`, and `TransportProducerHealth` alongside it) because neither crossed the boundary before and a composition root now needs one.

## Verification

- Done-means: scripts/done-means/652-capture-health-composed.sh

**RED (against the `origin/main` composition, restored by file-copy — no `git stash`, per round 13):** 5 of 8 clauses failed.

```
FAIL  (a) the serving composition supplied NO capture input — /health published no capture block
FAIL  (b) live GET /health over HTTP names-capture=false (status 200)
FAIL  (c) silent lane (365 user turns beside 0 assistant) -> HTTP 200, status=healthy, stale=undefined,
          silent_roles=null — silent-but-green reproduced (the #652 defect)
FAIL  (h) same composed app, observation changed: first=healthy then=healthy — reading is late-bound
PASS  (g) simulated 1800s of capture silence in 0.021s of wall clock — clock is injected, not slept
FAIL  (d) control: delivering lane -> HTTP 200, block published=false
PASS  (e) control: no capture lane composed -> HTTP 200, status=healthy, capture block absent=true
PASS  (f) control: observer composed but observation unavailable -> HTTP 200, capture block absent=true
```

The three controls (e, f, g) pass **pre-fix**, which is what distinguishes a discriminating check from one that fails everywhere. Clause (c) reproduces the defect verbatim: a half-dead lane reports `healthy`.

This RED was **re-proven in the check's final form** after the driver was edited for a typecheck fix — a repaired clause has never failed in its current form (round 8/11).

**GREEN (after):** 8/8.

```
PASS  (a) the serving composition supplied a capture input — /health published a capture block with 9 field(s)
PASS  (b) live GET /health over HTTP names-capture=true (status 503)
PASS  (c) silent lane (365 user turns beside 0 assistant) -> HTTP 503, status=degraded, stale=true,
          silent_roles=["assistant"]
PASS  (h) same composed app, observation changed: first=degraded then=healthy (stale=false)
          — reading is late-bound, not captured at boot
PASS  (g) simulated 1800s of capture silence in 0.021s of wall clock (carried as silence_seconds=1800,
          never compared) — clock is injected, not slept
PASS  (d) control: delivering lane -> HTTP 200, status=healthy, block published=true, stale=false
PASS  (e) control: no capture lane composed -> HTTP 200, status=healthy, capture block absent=true
PASS  (f) control: observer composed but observation unavailable -> HTTP 200, capture block absent=true
```

**Suites:** `bun run test:isolated` → **3728 pass, 0 fail** (243 files, db `ob_isolated_1313_mskvh8sm7p`, dropped on exit). `bunx tsc --noEmit` clean.

**Mutation-tested** (round-9: a green clause is not evidence until seen to fail). Four mutations, each caught by its own clause rather than by blanket failure:

| Mutation | Caught by |
|---|---|
| `stale: true` always | (d) control, and (h) |
| sum across roles instead of per-role (rebuilds #447) | (c) *only* |
| absent observation returns a fabricated healthy reading | (e) and (f) *only* |
| reading captured at boot instead of per request | (h) *only* |

The unit tests were separately proven able to fail: removing the expected-role seed turns the no-rows silent-speaker test red (7 pass / 1 fail), and restoring it returns 8/8.

## Critical Self-Review

- Highest-risk behavior: a **false-positive `stale` degrading a healthy worker**, since `/health` gates load balancing across core01's two workers. Three structural mitigations, each with a clause: absence and `undefined` both stay green (e, f); a lane below `MIN_SESSIONS_FOR_SILENCE` is never called dead; and the healthy control (d) fails loudly if anything degrades a delivering lane. The residual exposure is real and named below.
- Assumptions that could be wrong: (1) **that `user` and `assistant` are the roles a healthy lane must deliver** — `EXPECTED_ROLES` is a constant here, while `ingest-raw-turn.ts:26` admits a third role, `tool`. Seeding `tool` would make every non-tool-using session report a silent speaker, so it is deliberately excluded; if a deployment genuinely expects tool turns, this constant is wrong for it and nothing detects that. (2) That `session_ref` is a sound session denominator — `count(DISTINCT session_ref)` counts NULLs as none, so sessions ingested without a ref undercount the denominator and make silence *harder* to declare, which fails safe but silently. (3) That a 60-minute window and a 60s refresh suit every deployment; both are defaults with no operator override wired.
- Missing/weak tests: the gather SQL is exercised against a **fake pool**, so the query's actual grouping behavior against Postgres is unproven here — the fold is tested, the SQL is not. No test drives the `setInterval` refresh cadence (it is `autoStart: false` in every test). The done-means check covers composition and the endpoint; neither covers a real database round trip.
- Security/permission risk: the query is parameterized and namespace-scoped (`WHERE namespace = $1`), matching the repo's namespace-isolation rule. The reading is content-free by construction — counts, booleans, and role names only; no transcript text, path, or endpoint reaches the health body. `/health` is unauthenticated, so the block is public: it exposes turn and session COUNTS for the configured namespace, which is a volumetric signal and not content. Worth a reviewer's explicit opinion.
- Migration/deploy risk: none. No schema change, no migration. `captureObserver` is optional, so every existing caller compiles and behaves identically — proven by 3728 tests unchanged and a clean `tsc`.
- Downstream client/runtime risk: no MCP tool, schema, or protocol surface changed. `/health` gains a `capture` block only on a process that composes an observer; `SingleWorkerHealth.capture` was already an optional field merged in #648.
- Rollback/cleanup concern: revert is a clean three-file revert plus deleting the new module and its test. No state is written and no migration runs, so nothing needs undoing.
- Fixes made before PR: (1) **`tsc` caught that `TransportCaptureHealth` was never exported from the transport barrel** — #648 added the type to `health.ts` but not to `index.ts`, so the first composition to need it could not import it cleanly; added it and `TransportProducerHealth` together. (2) Removed an unused injected-clock knob from the composition input after realizing the observation already carries `silenceSeconds` — shipping an unused parameter would have been a second opinion about where time comes from. (3) Fixed three implicit-`any` parameters in the driver, then **re-proved RED** because the check had been edited.
- Known residual risk: **the composition is WRITTEN, not RUNNING.** This PR makes the reading reachable and proves it over a real HTTP listener from the real composition, but `server/main.ts` does not yet construct a `createCaptureLivenessObserver` — the observer is a port the composition accepts, and wiring it into the production entrypoint means choosing a namespace, a window, and a refresh cadence as operator-visible configuration rather than as constants I pick silently. Composing it here with hardcoded defaults would be exactly the "adjusted silently" failure the repo standard forbids. Stated plainly rather than implied away: **after this merges, `/health` on the local clone still publishes no capture block until that config decision is made.** Flagged for the decisions pass.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the lessons here are operating lessons for lanes (the seed-the-expected-roles trap, the substitute-don't-zero rule for an unobservable count), not review knowledge for a reviewer lane — they belong in `docs/lane-contract.md` Tightenings, and are listed in the lane report for the controller's harvest.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: nothing constructs the observer in `server/main.ts` yet, so there is no live surface to smoke. The composition is WRITTEN, not RUNNING, and this PR says so rather than implying deployment.

## Contract Parity

- Contract parity: [x] runtime-specific because: the `capture` block is an optional, additive field on the TypeScript `/health` response that #648 already introduced; this PR only supplies it from a composition root. No contract fixture asserts its presence and nothing crosses the MCP wire.

## Downstream Rollout

Notes/evidence:

- Not applicable across every row, for one reason: **no MCP tool, schema, or protocol surface changed.** The change is a composition-root wiring plus a server-internal observer module. `/health` is the only surface touched, its `capture` field was already declared optional by #648, and it remains absent on every process until the entrypoint wiring named in the residual risk lands. Nothing downstream can observe this PR today; the rollout classification is worth re-running against a surface that actually exists once it does.

### DIFF
--- a/scripts/done-means/652-capture-health-composed.driver.ts
+++ b/scripts/done-means/652-capture-health-composed.driver.ts
@@ -1,0 +1,373 @@
+/**
+ * DONE-MEANS driver for #652 — the capture reading reaches a LIVE /health.
+ *
+ * Subject: the real serving composition. `createShadowApplication`
+ * (server/application/index.ts:114) is what `server/main.ts:280` calls to build
+ * the process that serves traffic, and this driver binds its Express app to an
+ * ephemeral port and makes real HTTP requests against it. Nothing here
+ * re-implements the composition — a check that rebuilds its subject proves the
+ * rebuild works (docs/lane-contract.md, #624 harvest).
+ *
+ * The distinction from #647's driver is the whole point of this issue: that one
+ * called `getSingleWorkerHealth(...)` directly with a hand-supplied
+ * `captureHealth`, which proves the FUNCTION honors the input. It cannot prove
+ * any process ever SUPPLIES one, and #648's own report recorded that none does.
+ * Here the input must arrive from the composition itself.
+ *
+ * Clock: injected. `silence_seconds` is carried through and never compared
+ * against a threshold; every verdict is a count or a boolean
+ * (docs/lane-contract.md Tightenings round 5 — wall-clock assertions are flake
+ * generators, #632/#634).
+ */
+import express from "express";
+import type { AddressInfo } from "node:net";
+import type { Server } from "node:http";
+import { createShadowApplication } from "../../server/application/index.ts";
+import type { ShadowApplication } from "../../server/application/index.ts";
+import { parseServerConfig } from "../../server/config.ts";
+import type { ServerConfig } from "../../server/config.ts";
+import type { Database, DatabaseHealth } from "../../server/db/pool.ts";
+import { silentLogger } from "../../server/transport/testing/silent-logger.ts";
+
+const driverStartedAt = Date.now();
+
+/**
+ * Simulated capture silence, in seconds. Orders of magnitude beyond this
+ * driver's own runtime — clause (g) asserts the gap, which is what proves the
+ * clock is assigned rather than slept through.
+ */
+const SIMULATED_SILENCE_SECONDS = 1_800;
+
+const results: Array<{ clause: string; ok: boolean; detail: string }> = [];
+
+function clause(name: string, ok: boolean, detail: string): void {
+  results.push({ clause: name, ok, detail });
+  console.log(`${ok ? "PASS" : "FAIL"}  (${name}) ${detail}`);
+}
+
+const CONNECTED: DatabaseHealth = { connected: true, total: 2, idle: 2, waiting: 0 };
+
+function testConfig(overrides: Record<string, string> = {}): ServerConfig {
+  const result = parseServerConfig({
+    DB_HOST: "db.internal",
+    DB_NAME: "open_brain_test",
+    DB_USER: "open_brain",
+    LOG_FILE: "logs/open-brain.log",
+    OPEN_BRAIN_SERVER_IP: "192.0.2.21",
+    ...overrides,
+  });
+  if (!result.ok) {
+    throw new Error(`invalid driver configuration: ${JSON.stringify(result.issues)}`);
+  }
+  return result.config;
+}
+
+function fakeDatabase(): Database {
+  return {
+    pool: {} as Database["pool"],
+    close: async () => {},
+    health: async () => CONNECTED,
+  } as Database;
+}
+
+interface LiveApp {
+  readonly base: string;
+  close(): Promise<void>;
+}
+
+/**
+ * Compose the REAL application and bind it to a kernel-assigned ephemeral port.
+ *
+ * `overrides` is spread as `object` for the same reason #647's driver did it:
+ * on the pre-fix tree the capture input is not a field of the composition's
+ * input type, so a typed literal would be a COMPILE error and the check would
+ * fail to start. A check that cannot run does not distinguish "not fixed yet"
+ * from "check is broken" (#625 precedent), and the RED must be a runtime
+ * observation.
+ */
+async function listen(overrides: Record<string, unknown> = {}): Promise<LiveApp> {
+  const application: ShadowApplication = createShadowApplication({
+    config: testConfig(),
+    logger: silentLogger(),
+    database: fakeDatabase(),
+    authenticate: ((
+      _request: unknown,
+      _response: unknown,
+      next: () => void,
+    ) => next()) as never,
+    parseRequestBody: express.json(),
+    serverFactory: () => ({ connect: async () => {} }) as never,
+    fetch: (async () => new Response("{}", { status: 200 })) as unknown as typeof fetch,
+    ...(overrides as object),
+  } as never);
+
+  const server: Server = await new Promise((resolve) => {
+    const listener = application.app.listen(0, "127.0.0.1", () => resolve(listener));
+  });
+  const { port } = server.address() as AddressInfo;
+  return {
+    base: `http://127.0.0.1:${port}`,
+    close: async () => {
+      await new Promise<void>((resolve) => server.close(() => resolve()));
+      await application.close();
+    },
+  };
+}
+
+interface HealthResponse {
+  readonly status: number;
+  readonly body: Record<string, unknown>;
+  readonly raw: string;
+}
+
+async function getHealth(base: string): Promise<HealthResponse> {
+  const response = await fetch(`${base}/health`);
+  const raw = await response.text();
+  return {
+    status: response.status,
+    body: JSON.parse(raw) as Record<string, unknown>,
+    raw,
+  };
+}
+
+/**
+ * A capture observation as the gatherer produces one.
+ *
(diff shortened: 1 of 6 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/652-capture-health-composed.sh exit=0 sha=df2585d19a117a26d3c6b4382300144abc297e6f at=2026-08-08T21:22:02.694Z

```
INFO  repo:   /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-652-11849mskvq02t
INFO  driver: scripts/done-means/652-capture-health-composed.driver.ts (injected clock, ephemeral listener, no DB)

PASS  (a) the serving composition supplied a capture input — /health published a capture block with 9 field(s)
PASS  (b) live GET /health over HTTP names-capture=true (status 503)
PASS  (c) silent lane (365 user turns beside 0 assistant) -> HTTP 503, status=degraded, stale=true, silent_roles=["assistant"]
PASS  (h) same composed app, observation changed: first=degraded then=healthy (stale=false) — reading is late-bound, not captured at boot
PASS  (g) simulated 1800s of capture silence in 0.033s of wall clock (carried as silence_seconds=1800, never compared) — clock is injected, not slept
PASS  (d) control: delivering lane -> HTTP 200, status=healthy, block published=true, stale=false
PASS  (e) control: no capture lane composed -> HTTP 200, status=healthy, capture block absent=true
PASS  (f) control: observer composed but observation unavailable -> HTTP 200, status=healthy, capture block absent=true

clauses: 8 run, 0 failed

DONE-MEANS #652: PASS — the serving composition reports capture state on live /health.
```

</details>

rodaddy:
harvested: dae5a4b — round-14 Tightenings on wip/2026-08-07.

Issue #654, opened 2026-08-08
provider session_start fails exact-scope proof on 'namespace' against live main — blocks capture/checkpoint E2E scenarios


rodaddy:
Found by the #578 E2E gate's first composed live run (PR #653, lane report 2026-08-08). RUNNING evidence: against the live dogfood clone at merged main, provider session_start fails with "session_start result did not prove exact Open Brain scope: namespace" — on the default rico namespace, reproduced, not a stale-deploy artifact. Localized (PROPOSED, not isolated) to python/openbrain-memory/src/openbrain_memory/_runtime_validation.py:78-92 and :688-711. Note relationship to #646 (fixed the demand-source/reject-source contradiction and server-side scope establishment, PR #650): this is a DIFFERENT key (namespace) surviving that fix. Blocks the capture and checkpoint scenarios in the E2E gate (scripts/done-means/578-e2e-gate.sh clauses a/e). Red exists already: the gate's live run IS the failing check.

rodaddy:
Fixed by PR #657 (merged 120ad41) and independently verified against the live dogfood service before merge: verify-lane receipt bound to 3f21578, done-means scripts/done-means/654-namespace-scope-proof.sh exit 0, all five clauses — capture durable in the requested namespace (lane_namespace matched the throwaway e2e654 namespace, read from the database), weak-token delegation refused with nothing written, refusal names the cause (403 'Role not permitted to delegate namespace'), and the default path stays healthy (lane_namespace=rico). Root cause was client-side: runtime.py hardcoded delegate_namespace=False since #294, so X-Namespace was never sent and delegated-intent sessions silently landed in namespace 'admin'. Fix adds OPENBRAIN_DELEGATE_NAMESPACE (required-explicit, default OFF); no server redeploy needed (lane's claim, consistent with the live verify passing against the unmodified service). Downstream-rollout: additive opt-in, Hermes canary becomes required at first flag-on deployment (recorded on the PR). Lessons harvested: lane-contract round 15 (d84206e).

PR #294, opened 2026-07-17
feat(memory): add first-class exact-scope runtime

rodaddy:
## Summary

- add a package-owned first-class memory runtime for recall, distilled capture, checkpoint, and wrap across Claude, Codex, and Pi adapters
- route direct authenticated Streamable HTTP MCP first, optional fixed-command `mcp2cli` second, and atomic ordered JSONL spool last
- add opt-in bounded `durable_lane_context` with exact seven-coordinate server-side scope enforcement
- atomically establish missing legacy-lane coordinates without overwriting asserted scope
- advance the public contract to `2026-07-17.memory-tools.v22`, `agent_context_pack` v2, and `append_session_event` v8
- require v22 capability semantics for `openbrain-memory` 0.1.8 first-class runtime operations

## Changes

- introduce `FirstClassMemoryRuntime`, bounded JSON CLI operations, truthful receipts, fixed fallback routing, and group-safe spool retention/replay
- make capture/checkpoint/wrap establish every exact nullable coordinate and make checkpoint/wrap summaries immediately recallable
- enforce auth-derived namespace plus agent, platform, server, channel, thread, and session scope for durable lane context
- enforce context latency budgets without detached DB work, total-order event ties, and truthfully report omitted events
- harden spool full-write/fsync durability, malformed-group retention, fallback write evidence, direct-to-fallback lane verification, bounded child output, and nullable-thread checks
- update the shared NATS context-pack caller to use the same authoritative builder; NATS remains runtime-disabled unless separately rolled out
- add security, concurrency, exact-scope, redaction, fallback, spool, contract, real-Postgres, and process-boundary regressions
- split new oversized runtime/context/test owners below the repository's 750-line warning
- capture PR #294 MEDIUM+ review patterns in `docs/sme/`

## Validation

- [x] TypeScript: `bunx tsc --noEmit`
- [x] Server suite with disposable PostgreSQL 18 databases: 1,364 passed, 24 skipped, 0 failed
- [x] Skipped migration/scratch suites rerun independently: 28 passed, 0 failed
- [x] DB anti-skip guard: 28 live-Postgres cases; all six required suites present; zero required skips/failures/errors
- [x] Focused lifecycle/context/contract/NATS suite: 210 passed, 0 failed
- [x] Python Ruff format/check, mypy, and combined Pyright: zero errors
- [x] Python package suite: 340 passed, 5 credential-gated live skips
- [x] `uv build`; installed wheel imports runtime modules and passes malformed/oversized CLI boundary smokes
[...]

PR #657, opened 2026-08-08
fix(memory): request the configured namespace over the wire, or refuse by name (#654)

rodaddy:
## Summary

- **The defect.** `FirstClassMemoryRuntime` constructed its direct client with a hardcoded `delegate_namespace=False` (`python/openbrain-memory/src/openbrain_memory/runtime.py:520`, unchanged since #294), so the `X-Namespace` header was **never sent**. The server then fell back to the token's own namespace (`server/tools/memory-helpers.ts:125`) while the client validated the response against the **configured** namespace (`runtime.py:1165`). The two disagreed, `validate_started_lane` correctly refused to prove the scope, and every capture was lost with `session_start result did not prove exact Open Brain scope: namespace`.
- **Why #650 did not cover it.** This is #646's disease one key over. PR #650 fixed the demand/reject contradiction on `source` and taught the serving tree to establish exact scope on an existing lane. `namespace` survived that fix because it is not a lane column the server backfills — it is the predicate the server derives from *identity*, so client and server can silently disagree about it.
- **RUNNING evidence of real harm, not just a failed check.** Against the live dogfood service (revision `212f916`) with `OPENBRAIN_NAMESPACE=e2e654-<run>` and a delegation-capable token, the lane landed in namespace **`admin`** — the operator asked for one namespace and the write went to another. That is precisely the failure `server/auth/middleware.ts:9-13` names as the dangerous variant: *"Silently ignoring the header is the dangerous variant: the caller believes it wrote somewhere it did not."*
- **The fix, at the owning boundary — two parts.**
  1. `delegate_namespace` becomes a real config key (`OPENBRAIN_DELEGATE_NAMESPACE`), plumbed through to the client that **already** knew how to send the header (`client.py:1542-1543`). No new mechanism was invented.
  2. A namespace mismatch is diagnosed **by name** before the generic comparison. It names what the server bound, what was configured, and how to authorize the difference.
- **Default stays OFF, deliberately.** The server gates `X-Namespace` on **role, not on value**: an `agent`-role token presenting `X-Namespace: rico` is a 403 even though `rico` is already its own namespace (observed live, 403 vs 200 for admin/ob-admin). Sending it unconditionally would 403 every non-admin caller in the fleet, all of whom are correctly served today.
- **The validator was NOT weakened.** Accepting whatever namespace came back is the silent mis-scope itself. Namespace isolation is a security boundary, so clause (c) exists to prove the boundary still holds.

## Verification

- Done-means: scripts/done-means/654-namespace-scope-proof.sh

**RED** (pre-fix, unmodified `origin/main`, live against `127.0.0.1:3100` revision `212f916`):

```
(a) namespace scope proof         : FAIL — RED ANCHOR PRESENT — [session_start result did not prove exact Open Brain scope: namespace] status=lost durable=false lane_namespace=admin
(b) capture durable in requested ns: FAIL — status=lost durable=false lane_namespace=admin expected_namespace=e2e654-ed7589d7a264
(c) isolation held for weak token : PASS — refused (durable=false); nothing written to e2e654-ed7589d7a264 (lane_namespace=rico)
(d) refusal names the cause       : PASS — error names namespace: [session_start result did not prove exact Open Brain scope: namespace]
(e) control, default path healthy : PASS — status=saved durable=true lane_namespace=rico

=== DONE-MEANS 654: FAIL (2 failing clause(s)) ===
```

Note `lane_namespace=admin` on the RED run: the clause failed *and* demonstrated the mis-scope, which is stronger than the issue's original report.

**GREEN** (this branch, same service, no redeploy — the change is entirely client-side):

```
(a) namespace scope proof         : PASS — no namespace scope-proof failure (status=saved durable=true)
(b) capture durable in requested ns: PASS — status=saved durable=true lane_namespace=e2e654-b9036450de3e
(c) isolation held for weak token : PASS — refused (durable=false); nothing written to e2e654-b9036450de3e (lane_namespace=<none>)
(d) refusal names the cause       : PASS — error names namespace: [Open Brain HTTP error status=403 context=initialize body={"error": "Role not permitted to delegate namespace"}]
(e) control, default path healthy : PASS — status=saved durable=true lane_namespace=rico

=== DONE-MEANS 654: PASS ===
```

The gate refuses to run at all unless its token can actually delegate (a `200` pre-check against `/mcp` with `X-Namespace`). Without that, a blanket 403 would make clause (a) look green because nothing was ever attempted — a gate that examined nothing is not a pass (#583).

**Regression tests are red-first, proven by mutation, not asserted:**

- Reverting the plumbing to `delegate_namespace=False` → `test_delegate_namespace_config_reaches_the_direct_client` fails with `assert False is True`; restoring it passes.
- Removing the named-diagnostic block → `test_namespace_mismatch_names_the_cause_and_the_remedy` fails with `assert 'token-namespace' in 'session_start result did not prove exact Open Brain scope: namespace'` — the old message, verbatim; restoring it passes.

**Python bars on the branch:** `uv run pytest -q` → **577 passed, 14 skipped**; `uv run mypy src/openbrain_memory` → clean (17 files); `uv run ruff check src tests` → clean.

**`bun run test:isolated` not run, and here is the honest reason:** this diff touches **zero** TypeScript files (`git diff --name-only | rg '\.ts$'` → 0). The change is confined to the Python client and one shell check. Running the TS suite would prove nothing about this diff.

## Critical Self-Review

- Highest-risk behavior: the new `X-Namespace` header being sent at all. If it were sent by default it would 403 the entire non-admin fleet, converting a lost-capture bug into a total outage. Mitigated by defaulting to off and by clause (e), which drives the ordinary no-override path and asserts it still lands in the token's namespace.
- Assumptions that could be wrong: that opt-in is the right ergonomic. A caller who *should* delegate but does not set the flag now gets a clear refusal rather than a silent mis-scope — better, but still a manual step. I considered auto-detecting by comparing the configured namespace to the token's, and rejected it: there is no identity-introspection endpoint (`/whoami` and `/auth/whoami` both 404), so the client cannot know the token's namespace without first making a write and reading where it landed. Inferring it from a failed write would mean writing to the wrong namespace to discover that it was wrong.
- Missing/weak tests: no test covers a *delegated* namespace flowing through the fallback/spool path — the spool already stamps a parked namespace and refuses cross-namespace drain (`runtime.py:1055-1067`), but that interaction with delegation is untested here. No test covers `X-Namespace` for `checkpoint`/`wrap` specifically; they share the same client construction, so the coverage is structural rather than per-operation.
- Security/permission risk: this is a namespace-isolation change, so the risk is real and it is why clause (c) is adversarial rather than cosmetic: the weak token **requests** delegation (`run_capture ... 1`) and must still be refused. Post-fix the header genuinely goes out, so (c) tests the server's role gate rather than the client politely declining. Verified independently at the wire: admin/ob-admin → `200`, observation token → `403`, including when the requested namespace equals its own. Nothing in this diff can widen a predicate; the only new capability requires a token that already had delegation authority.
- Migration/deploy risk: none. No schema change, no migration, no server change, and therefore no redeploy — the GREEN run above is against the already-running `212f916`. Rollback is reverting the commit; the default-off flag means an un-upgraded client behaves exactly as before.
- Downstream client/runtime risk: the gate **applies** — `docs/downstream-rollout.md:14` names `python/openbrain-memory` client behavior explicitly. See the Downstream Rollout section; the honest summary is that behavior is additive and default-off, but I did not run the downstream canaries and am not claiming them.
- Rollback/cleanup concern: the check tears down its own throwaway namespace and all three lanes by exact `RUN_ID`, verified `0` rows remaining after the final run. Three lanes from my manual reproduction (`654-repro-*`, `654-ns-*`) were also removed; confirmed `0`.
- Fixes made before PR: (1) the first draft of the gate read `AUTH_TOKEN_ADMIN` from the repo `.env`, which returns `401` — the live service is served from `/Volumes/ThunderBolt/open-brain-local/app` and takes tokens from the launchd environment, so the gate would have hard-failed on a healthy system; it now reads the running process's environment first. (2) The gate initially referenced `PROBE_NS` in its delegation pre-check before `RUN_ID` was defined; caught and reordered. (3) The delegation pre-check itself was added after I realised a blanket 403 would produce a false green on clause (a).
- Known residual risk: the 403 for a non-delegating token surfaces as an `initialize`-stage transport error (`Open Brain HTTP error status=403 ... Role not permitted to delegate namespace`) rather than a purpose-built message. It is accurate, names the cause, and satisfies clause (d), but it is a raw HTTP error where a shaped one would read better. Left alone deliberately: shaping it means adding an error-translation path this issue did not ask for.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the reusable lesson here is operating knowledge for lanes (a live-service gate must read the *serving* process's credentials, and must prove its own precondition or risk a vacuous green), which `docs/lane-contract.md` Tightenings owns. It is in the lane report for the controller's mandatory harvest.

## Review Gate

Live evidence is the RED/GREEN transcripts above, both taken against the running dogfood service on `127.0.0.1:3100` at revision `212f916` (unchanged across both — the fix is client-side, so the same server produced both results).

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface changed. No MCP tool name, `inputSchema`, annotation, or response shape is touched — zero TypeScript files are in this diff. `X-Namespace` is a pre-existing, already-specified header (`server/auth/middleware.ts:9-13`) that this client simply never sent; the parity fixtures encode the wire shape, which is unchanged.

## Downstream Rollout

Notes/evidence:

- The gate **applies**: `docs/downstream-rollout.md:14` names `python/openbrain-memory` client behavior, and this changes it. I am not marking it away.
- **rtech-mcps / mcp2cli: not applicable.** Both consume the MCP tool surface, which is byte-identical here — no tool name, schema, annotation, or response shape changed, and no `.ts` file is in the diff. They cache schemas; there is no schema delta to refresh.
- **rtech-hermes and the live canaries are left UNCHECKED deliberately, because I did not run them.** Checking them would be a claim I cannot support. What I can say precisely: the change is additive and default-off, so a Hermes runtime that does not set `OPENBRAIN_DELEGATE_NAMESPACE` constructs its client exactly as before (`delegate_namespace=False`) and its behavior is unchanged — the 577-test suite passing with the default off is the evidence for that. Any Hermes agent that today *intends* to write to a namespace its token does not grant is already silently mis-scoping and would need the new flag; identifying whether such an agent exists is a downstream survey this lane did not perform. Flagging for the controller rather than closing it out.

### DIFF
--- a/python/openbrain-memory/src/openbrain_memory/_runtime_validation.py
+++ b/python/openbrain-memory/src/openbrain_memory/_runtime_validation.py
@@ -90,6 +90,31 @@ def validate_started_lane(
     }
     if isinstance(metadata, Mapping) and "server_id" in metadata:
         candidate["server_id"] = metadata["server_id"]
+    # A namespace mismatch is diagnosed separately, BEFORE the generic
+    # comparison, because it is the one scope key an operator cannot fix by
+    # editing the request (#654). `namespace` is env-carried and never a
+    # request key (the provider's own `--help` says so), so the generic
+    # "did not prove exact Open Brain scope: namespace" pointed at a key there
+    # is no way to send — the same shape of dead end as the #646 `source`
+    # contradiction, one key over.
+    #
+    # What it actually means is that the server bound the lane to the TOKEN's
+    # namespace rather than the configured one, because the request carried no
+    # `X-Namespace` delegation header. Naming the cause and the remedy turns an
+    # opaque refusal into an actionable one, and it must stay a REFUSAL: writing
+    # into whatever namespace came back is the silent mis-scope this issue is
+    # about (server/auth/middleware.ts:9-13).
+    served = lane.get("namespace")
+    if "namespace" in lane and served != namespace:
+        raise ValueError(
+            "session_start bound the lane to namespace "
+            f"'{served}', not the configured '{namespace}'. The namespace is "
+            "carried by OPENBRAIN_NAMESPACE and is never a request key; to bind "
+            "a namespace your token does not already grant, set "
+            "OPENBRAIN_DELEGATE_NAMESPACE=1 with an admin/ob-admin token so the "
+            "request sends the X-Namespace header. Nothing was written to "
+            f"'{namespace}'."
+        )
     expected = exact_scope_fields(namespace, scope)
     expected["source"] = expected.pop("platform")
     # The server stores/returns `platform` under the lane column `source`
--- a/python/openbrain-memory/src/openbrain_memory/cli.py
+++ b/python/openbrain-memory/src/openbrain_memory/cli.py
@@ -130,6 +130,13 @@ _HELP_ENVIRONMENT = {
         "OPENBRAIN_NAMESPACE -- namespace for every request; NOT a request key",
         "OPENBRAIN_TOKEN -- the bearer value sent as the Authorization header",
     ],
+    "optional": [
+        "OPENBRAIN_DELEGATE_NAMESPACE -- 1 to send OPENBRAIN_NAMESPACE as the "
+        "X-Namespace header, binding a namespace the token does not already "
+        "grant; requires an admin/ob-admin token and is refused for other "
+        "roles. Off by default: without it the server uses the token's own "
+        "namespace",
+    ],
     "note": (
         "The example is a request body and carries no identity. base_url, "
         "token, and namespace are read from these variables, or from an "
--- a/python/openbrain-memory/src/openbrain_memory/runtime.py
+++ b/python/openbrain-memory/src/openbrain_memory/runtime.py
@@ -98,6 +98,7 @@ MAX_DISTILLED_CONTENT_BYTES = _runtime_validation.MAX_DISTILLED_CONTENT_BYTES
 _CONFIG_KEYS = {
     "allow_insecure_http",
     "base_url",
+    "delegate_namespace",
     "fallback_enabled",
     "namespace",
     "project",
@@ -307,6 +308,22 @@ class RuntimeConfig:
     timeout: float = 30.0
     fallback_enabled: bool = False
     spool_path: Path | None = None
+    # Send the configured namespace as `X-Namespace` so the server binds THIS
+    # namespace instead of the token's own (#654).
+    #
+    # Off by default, and that default is not timidity — the server gates the
+    # header on ROLE, not on the value: an `agent`-role token presenting
+    # `X-Namespace: rico` is a 403 even though `rico` is already its own
+    # namespace (observed live 2026-08-08 against revision 212f916). Sending it
+    # unconditionally would therefore 403 every non-admin caller in the fleet,
+    # all of whom are correctly served by their token's namespace today.
+    #
+    # It must nonetheless be REACHABLE, because without it a caller configured
+    # for a namespace its token does not grant is silently written to the
+    # token's namespace instead. That is the exact failure
+    # `server/auth/middleware.ts:9-13` names as the dangerous variant: "the
+    # caller believes it wrote somewhere it did not."
+    delegate_namespace: bool = False
 
     def __post_init__(self) -> None:
         object.__setattr__(self, "base_url", _require_text(self.base_url, "base_url"))
@@ -374,6 +391,12 @@ class RuntimeConfig:
                 "fallback_enabled",
                 env,
                 "OPENBRAIN_MCP2CLI_FALLBACK",
+            ),
+            delegate_namespace=_source_bool(
+                values,
+                "delegate_namespace",
+                env,
+                "OPENBRAIN_DELEGATE_NAMESPACE",
             ),
             spool_path=Path(str(spool_value)) if spool_value is not None else None,
         )
@@ -517,7 +540,7 @@ class FirstClassMemoryRuntime:
                     timeout=config.timeout,
                     transport=transport,
                     allow_insecure_http=config.allow_insecure_http,
-                    delegate_namespace=False,
+                    delegate_namespace=config.delegate_namespace,
                 )
             except Exception as error:
                 # Threaded into the router, which surfaces it on the first call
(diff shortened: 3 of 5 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/654-namespace-scope-proof.sh exit=0 sha=3f2157820e06d2eedd3fc598edb9cd0cbe5adf64 at=2026-08-08T22:10:46.028Z

```
run marker : 46c38d6a34be
probe ns   : e2e654-46c38d6a34be (throwaway, removed at exit)
token ns   : rico
service    : http://127.0.0.1:3100

(a) namespace scope proof         : PASS — no namespace scope-proof failure (status=saved durable=true)
(b) capture durable in requested ns: PASS — status=saved durable=true lane_namespace=e2e654-46c38d6a34be
(c) isolation held for weak token : PASS — refused (durable=false); nothing written to e2e654-46c38d6a34be (lane_namespace=<none>)
(d) refusal names the cause       : PASS — error names namespace: [Open Brain HTTP error status=403 context=initialize body={"error": "Role not permitted to delegate namespace"}]
(e) control, default path healthy : PASS — status=saved durable=true lane_namespace=rico

=== DONE-MEANS 654: PASS — the configured namespace is requested over the wire and proven in the response, an unauthorized namespace bind is still refused by name, and the default path is unchanged. ===
```

</details>

rodaddy:
harvested: d84206ecb62b1975a65f4c9d9463fb107172299d — docs/lane-contract.md round 15 (#654 lane + verify-run lessons: serving-process credentials, unexercised refusal branches, silent-default identity as tenant mis-scope, dead-end-error class, head-bound red anchors).

rodaddy:
Downstream-rollout classification (docs/downstream-rollout.md): Python-client contract change, additive and opt-in — new config key OPENBRAIN_DELEGATE_NAMESPACE defaults OFF, so every existing consumer (mcp2cli, rtech-hermes runtime, generated skills) keeps byte-identical request behavior until a deployment explicitly enables it. Hermes live canary: N/A at this merge; it becomes REQUIRED at the first deployment that turns the flag on, and the unchecked canary boxes in the PR body record that honestly (PROPOSED, not claimed). Server side needs no redeploy per the lane, and the verify run neither proved nor needed that claim.

PR #645, opened 2026-08-08
Neutrality scrub: environment-specific values move to env lookups and placeholders

rodaddy:
## Summary

- Makes the repo neutral for outside users (#636): environment-specific hosts, volumes, and paths move to env lookups or `192.0.2.x` / `example.com` placeholders across shipped code, client-facing docs, CI, and ~44 test/fixture files. Internal ops runbooks (`docs/core01-*`, `_plans/`, `_DOCS/`, `specs/`) keep real values via an explicit allowlist, per the issue's scope ruling.
- Adds `scripts/done-means/636-neutrality.sh`: an rg-based gate that fails on any matching value in a tracked file outside the allowlist.
- **This lane inherited an in-progress scrub from a prior lane that died mid-run. That work is PROPOSED until verified, and verifying it found eight defects** — six edits that renamed something real, plus two gate bugs. Each is corrected in its own commit with the reason. Details under Critical Self-Review.

The governing distinction, learned the hard way here: a literal path is only safe to neutralise when **nothing compares it to reality**. Where a value is matched against the filesystem, replayed byte-for-byte against an external runtime, or pasted by an operator, a neutral placeholder does not neutralise it — it silently disables the thing the value was for.

## Verification

- Done-means: scripts/done-means/636-neutrality.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [ ] Live Open Brain smoke passed or is not applicable — N/A, no runtime/serving behavior changes

**RED** (committed check against the pre-scrub tree, `git archive HEAD` into a clean checkout):

```
VIOLATION src/tools/__tests__/upsert-entity.test.ts:175:          metadata: { ip: "10.71.20.35" },
VIOLATION src/transport.ts:22:// was 603 ms. The case that matters is core01 with 60-70 clients: every one of
FAIL: neutrality check found environment-specific values in 127 tracked file(s) outside the explicit internal-ops allowlist.
exit 1   (583 violation lines across 127 files)
```

**GREEN** (branch head):

```
PASS: tracked files are neutral outside the explicit internal-ops allowlist.
exit 0
```

**Negative control** — the exceptions are narrow, not blanket. Injecting `10.71.1.21` into `scripts/deploy-ref-gate.ts`, a file that HAS a line exception:

```
VIOLATION scripts/deploy-ref-gate.ts:251:// canary probe: 10.71.1.21
FAIL: ... 1 tracked file(s) outside the explicit internal-ops allowlist.
```

[...]

PR #658, opened 2026-08-08
fix(655): tear down the eval namespace the scenario runner creates

### LINKED ISSUE #655: E2E scenario runner leaks one namespace per run (eval-live-recall-scenario-*) — teardown missing
Found by the #578 E2E gate (PR #653): each scenario run leaves eval-live-recall-scenario-* / eval-live-recall-redprobe namespaces in the dogfood DB — clause (e) of scripts/done-means/578-e2e-gate.sh fails on it. Leaked rows reported, not cleaned (delete is operator-gated). Fix: scenario runner tears down its own prefix-guarded namespaces per ledger item 20's narrow auto-drop exception (self-created + prefix-guarded + session-scoped), with the leak clause staying RED-provable.

rodaddy:
## Summary

- Closes #655. The E2E scenario runner tore down its RECORDS but never its NAMESPACE, leaving one `eval-live-recall-scenario-*` namespace behind per run. This adds a prefix-guarded namespace purge that runs AFTER the existing record teardown.
- **The record teardown was not broken — it cannot do this job.** Two mechanisms, both measured in the dogfood database on 2026-08-08 rather than reasoned about:
  - `archive_entry` is a SOFT delete (`src/tools/archive-entry.ts:54`): it sets `archived_at` and the row stays. A namespace in this schema has no registry row (there is no namespaces table) — it exists exactly as long as some row carries it — so an archived row keeps its namespace alive permanently. **24** `eval-live-recall-*` namespaces hold nothing but archived rows: teardown reported success and the namespace is still there.
  - When the archive call THROWS, the tally counts `failed` and the row is left fully LIVE. **6** `eval-live-recall-scenario-*` namespaces are in exactly that state (`archived_at IS NULL`), from a receipt reading `attempted=1 archived=0 already_absent=0 failed=1`.
- `eval/open-brain/live/namespace-purge.ts` (new) hard-deletes rows for ONE namespace after `teardownRecords`, under `docs/issue-graph.md` ledger item 20's narrow auto-drop exception. All three conditions are structural, not promised: **self-created** (the namespace derives from a per-invocation crypto nonce in `makeRunId`), **prefix-guarded** (`assertPurgeableNamespace` refuses anything outside `eval-live-recall-` before issuing a single statement, and the prefix is a module constant, not a parameter a caller could dial open), **session-scoped throwaway** (fixture text seeded seconds earlier).
- `LiveScenarioTransport.cleanup` composes the purge after the record pass. A purge failure increments the teardown `failed` tally rather than throwing, so the #578 gate's clause (e) still reads it and the record tally that explains what DID get cleaned is not discarded.
- **Scope note:** this fixes the leak going forward. The rows already in the dogfood database are pre-existing data and are NOT touched — see "Pre-existing leaked rows" below for the count and the exact operator-gated cleanup SQL.

## Verification

- Done-means: scripts/done-means/655-eval-teardown.sh

**RED** (pre-fix reproduction via `DONE_MEANS_655_SKIP_PURGE=1`, which skips only the new purge and leaves everything else intact):

```
INFO  gate passed=true scenarios=3 teardown={"attempted":6,"archived":4,"already_absent":2,"failed":0} purged_rows=0
PASS  (c) the run really executed 3 scenario(s) — not a silent empty run
PASS  (c) scenario verification read its own seeded record BEFORE teardown ran
PASS  (c) the run's own assertions still pass with teardown in place
PASS  (a) record teardown reported failed=0
FAIL  (a) 2 eval-live-recall-% row(s) left behind — the run leaked its namespace
thoughts|eval-live-recall-scenario-donemeans655-c2bf5739ccbe|1
PASS  (b) the guard refused all 6 non-eval name(s) it was pointed at
PASS  (b) a refused purge deleted ZERO rows — it refuses BEFORE it mutates
PASS  (b) the guard still ALLOWS a genuine per-run eval namespace — it discriminates
=== RESULT: FAIL (1 failing clause(s)) ===
```

That RED is the production defect exactly: 3/3 scenarios pass, teardown self-reports `failed=0`, and the namespace leaks anyway. A teardown that reports success while leaving residue is why this needed a row-count gate rather than a tally assertion.

**GREEN**:

```
INFO  gate passed=true scenarios=3 teardown={"attempted":6,"archived":4,"already_absent":2,"failed":0} purged_rows=2
PASS  (c) the run really executed 3 scenario(s) — not a silent empty run
PASS  (c) scenario verification read its own seeded record BEFORE teardown ran
PASS  (c) the run's own assertions still pass with teardown in place
PASS  (a) record teardown reported failed=0
PASS  (a) 0 eval-live-recall-% rows remain — the run removed the namespace it created
PASS  (b) the guard refused all 6 non-eval name(s) it was pointed at
PASS  (b) a refused purge deleted ZERO rows — it refuses BEFORE it mutates
PASS  (b) the guard still ALLOWS a genuine per-run eval namespace — it discriminates
=== RESULT: PASS ===
```

`purged_rows=2` is exactly the 2 rows that leaked in RED.

Other checks, all RUNNING this session:

- `bun run test:isolated` — **3741 pass, 35 skip, 0 fail** across 244 files, fresh database `ob_isolated_46330_mskx6quux8`, dropped on exit.
- `bunx tsc --noEmit` — clean, no output.
- `bun test eval/open-brain/live/__tests__/` — 204 pass, 0 fail.
- **Guard mutation proof.** Swapping `startsWith` for `includes` in `assertPurgeableNamespace` turns the unit suite from 13 pass / 0 fail into 12 pass / **1 fail**; restoring returns it to 13 pass. The guard test is proven able to fail, so its green is worth something.
- `scripts/done-means/636-neutrality.sh` — FAILs with exactly 2 violations, both in `server/application/index.ts:203` and `server/transport/health.ts:136`, both present on `origin/main` (`git show origin/main:server/transport/health.ts` line 136), neither in a file this PR touches. Pre-existing, reported not fixed. **Labelled from a same-tree comparison, not a single-file run.**

## Critical Self-Review

- Highest-risk behavior: a hard `DELETE` running automatically. The mitigation is that the prefix is a module constant rather than a parameter, so no caller can widen it, and clause (b) plants a canary row under each refused name and verifies it SURVIVES — proving the guard refuses before mutating rather than throwing after the deletes have run.
- Assumptions that could be wrong: `PURGE_TABLES` is hand-enumerated. If a future migration adds a namespaced table the eval suites seed, the purge will silently miss it and the leak returns for that table. I chose the explicit list over discovering from `information_schema` deliberately — a discovered list would quietly grow to cover tables this module was never reviewed against — but the tradeoff is real and is the most likely way this regresses.
- Missing/weak tests: the done-means driver uses a local `ScenarioTransport` against a real fresh database rather than `LiveScenarioTransport` against a live deployment. `runScenarioGate`, `teardownRecords`, `purgeNamespace` and the guard are all the shipped code and the residue measured is real, but the composition inside `LiveScenarioTransport.cleanup` itself is **WRITTEN, not RUNNING** — see residual risk.
- Security/permission risk: the purge bypasses the MCP auth layer and talks to Postgres directly, so no namespace predicate is derived from a token. That is why the prefix guard is the boundary, and why it is tested against `rico` and `shared-kb` specifically. The refusal message deliberately carries no namespace content (a refusal can be pointed at an operator's real namespace) and there is a test asserting the offending value does not appear in it.
- Migration/deploy risk: none. No schema change, no migration; eval-only code path.
- Downstream client/runtime risk: none. `docs/downstream-rollout.md` classifies this as not applicable — no MCP tool, schema, transport, or Python-client surface changes. `eval/` is not imported by the server or the Python package.
- Rollback/cleanup concern: revert is a clean `git revert`; nothing persists.
- Fixes made before PR: (1) the check originally wrote its receipt to the repo root and claimed it was git-ignored — it was not, and the check would have left an untracked file behind, the same class of residue it exists to catch. Moved to the temp-workspace scratch bucket and the false claim removed. (2) First RED attempt died on a module-not-found import and never measured the leak at all; replaced with `DONE_MEANS_655_SKIP_PURGE=1`, which reproduces the pre-fix world with everything else intact and keeps RED regenerable without deleting the fix. (3) Two schema mismatches in the driver (`created_by` NOT NULL; `ob_session_lanes` has no `platform`/`server_id` columns) surfaced as `scenario_transport_error` — announced in the code comment rather than silently dropped, with the fixture's `platform`/`server_id` folded into `source`/`project`.
- Known residual risk: `LiveScenarioTransport.cleanup` is **WRITTEN, not RUNNING** — it has not been executed against a live deployment. Running it needs an admin/ob-admin token with X-Namespace delegation authority, and since the neutrality scrub (PR #645) the repo `.env` carries empty placeholders (`AUTH_TOKEN_ADMIN` has length 0); the values live only in the serving process's environment, and harvesting them from a running process is what the #578 gate header forbids. The next `bash scripts/done-means/578-e2e-gate.sh` run with real coordinates is the RUNNING proof, and its clause (e) reads the tally this PR feeds.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the operating lesson here (a teardown can report success and still leave residue, because soft-delete is not removal) is a LANE lesson, harvested into `docs/lane-contract.md` Tightenings by the controller's merge pass per ledger item 19, not a reviewer-facing code-review pattern.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no admin/ob-admin token with X-Namespace delegation is available in this lane (see residual risk); the #578 gate is the live proof and is the controller's to run.

## Contract Parity

- Contract parity: [x] runtime-specific because: eval-harness teardown only. No MCP tool, schema, or wire contract changes, so there is no parity fixture to update.

## Downstream Rollout

Notes/evidence:

- Not applicable on every downstream line: the change is confined to `eval/` and `scripts/done-means/`. No tool schema, transport, or client-facing surface is touched, so no downstream consumer can observe it.

### Pre-existing leaked rows — REPORTED, NOT DELETED

Counted in the dogfood database (`open_brain_local_20260724`) this session. **Nothing here was deleted**: pre-existing data is operator-gated, so this is the count plus the exact SQL for the operator to run by hand.

The six namespaces this issue names, all with `archived_at IS NULL` (the thrown-archive case), 1 `thoughts` row each:

```
eval-live-recall-scenario-3be56b18bc82-0be200df59a1
eval-live-recall-scenario-3be56b18bc82-3de7dae8061c
eval-live-recall-scenario-3be56b18bc82-76708f80140a
eval-live-recall-scenario-3be56b18bc82-a422aa9f8559
eval-live-recall-scenario-3be56b18bc82-ba69018f3ed7
eval-live-recall-scenario-3be56b18bc82-f16d2f1a7932
```

The wider `eval-live-recall-*` family is **24 further namespaces holding 94 rows**, every one of them already archived — the soft-delete case, where teardown genuinely succeeded and the namespace survived anyway.

Counted this session against `open_brain_local_20260724`: **100 rows across 30 `eval-live-recall-*` namespaces** — 69 `thoughts`, 30 `decisions`, 1 `ob_session_lanes`, 0 `sessions`. (An earlier draft of this body said 105 rows / 28 namespaces; that was wrong and is corrected here against the query output.)

Operator cleanup SQL, scoped by the same prefix the code guard uses. Run the SELECT first and read the counts:

```sql
-- 1. Review before removing anything.
SELECT 'thoughts' AS tbl, namespace, count(*) FROM thoughts
  WHERE namespace LIKE 'eval-live-recall-%' GROUP BY 1, 2
UNION ALL SELECT 'decisions', namespace, count(*) FROM decisions
  WHERE namespace LIKE 'eval-live-recall-%' GROUP BY 1, 2
UNION ALL SELECT 'ob_session_lanes', namespace, count(*) FROM ob_session_lanes
  WHERE namespace LIKE 'eval-live-recall-%' GROUP BY 1, 2
ORDER BY 2, 1;

-- 2. Remove, children before parents. Wrap in a transaction and check the
--    row counts before COMMIT.
BEGIN;
DELETE FROM ob_session_events e USING ob_session_lanes l
  WHERE e.lane_id = l.id AND l.namespace LIKE 'eval-live-recall-%';
DELETE FROM ob_session_lanes WHERE namespace LIKE 'eval-live-recall-%';
DELETE FROM thoughts        WHERE namespace LIKE 'eval-live-recall-%';
DELETE FROM decisions       WHERE namespace LIKE 'eval-live-recall-%';
DELETE FROM sessions        WHERE namespace LIKE 'eval-live-recall-%';
-- COMMIT;   -- uncomment after reading the counts above
```

One namespace outside the prefix, `eval-kwdiag-1889c167` (1 row), is from a different eval family and is deliberately excluded from both the guard and this SQL — it is not this issue's residue and removing it is a separate operator decision.

### DIFF
--- a/eval/open-brain/live/__tests__/namespace-purge.test.ts
+++ b/eval/open-brain/live/__tests__/namespace-purge.test.ts
@@ -1,0 +1,159 @@
+import { describe, expect, test } from "bun:test";
+import {
+  EVAL_NAMESPACE_PREFIX,
+  NamespacePurgeRefused,
+  assertPurgeableNamespace,
+  purgeNamespace,
+} from "../namespace-purge.ts";
+import { makeRunId, runNamespaces } from "../config.ts";
+
+/**
+ * The prefix guard is the whole authority for auto-removal here
+ * (`docs/issue-graph.md` ledger item 20, condition 2: "prefix-guarded so it
+ * structurally cannot name anything it did not create"). These tests exist so
+ * that claim is EXERCISED rather than asserted, and so they run with no
+ * database at all — a guard reachable only behind a live connection is a guard
+ * nobody fires, and an unfired guard is the one that turns out to be wrong.
+ *
+ * Following `server/tools/search-read-scope.test.ts`, the purge tests assert on
+ * the SQL and bound parameters the code actually sends, not merely on its
+ * return value: a namespace predicate is only a boundary if it is in the
+ * statement.
+ */
+describe("assertPurgeableNamespace", () => {
+  test("accepts the namespaces the eval config actually generates", () => {
+    for (const label of ["scenario-abc123", "live578", "epic336kw"]) {
+      const { primary, negative } = runNamespaces(
+        makeRunId({ prefix: label, randomHex: "0123456789ab", env: {} }),
+      );
+      expect(() => assertPurgeableNamespace(primary)).not.toThrow();
+      expect(() => assertPurgeableNamespace(negative)).not.toThrow();
+    }
+  });
+
+  test.each([
+    ["an operator namespace", "rico"],
+    ["the shared knowledge base", "shared-kb"],
+    ["a bare empty string", ""],
+    ["one character off the prefix", "eval-live-recal-abc123"],
+    ["a different eval family", "eval-kwdiag-1889c167"],
+  ])("refuses %s", (_label, namespace) => {
+    expect(() => assertPurgeableNamespace(namespace)).toThrow(
+      NamespacePurgeRefused,
+    );
+  });
+
+  test("refuses the bare prefix, which no run ever produces", () => {
+    // `runNamespaces` always appends a nonce, so the bare prefix names nothing
+    // this process created. Accepting it would mean accepting a name that could
+    // only have come from somewhere else.
+    expect(() => assertPurgeableNamespace(EVAL_NAMESPACE_PREFIX)).toThrow(
+      NamespacePurgeRefused,
+    );
+  });
+
+  test("refuses a name that merely CONTAINS the prefix", () => {
+    // The specific defect a `.includes()` guard ships with.
+    expect(() =>
+      assertPurgeableNamespace(`not-${EVAL_NAMESPACE_PREFIX}abc123`),
+    ).toThrow(NamespacePurgeRefused);
+    expect(() =>
+      assertPurgeableNamespace(`rico-${EVAL_NAMESPACE_PREFIX}abc123`),
+    ).toThrow(NamespacePurgeRefused);
+  });
+
+  test("the refusal message carries no namespace content", () => {
+    // A refusal can be pointed at an operator's real namespace, so the message
+    // must not echo it into a log or a receipt.
+    const secret = "rico-private-namespace";
+    let message = "";
+    try {
+      assertPurgeableNamespace(secret);
+    } catch (error) {
+      message = error instanceof Error ? error.message : String(error);
+    }
+    expect(message).not.toContain(secret);
+    expect(message).toContain("REFUSED");
+  });
+});
+
+describe("purgeNamespace", () => {
+  test("refuses BEFORE issuing any statement", async () => {
+    // The mutation proof, in unit form: a pool that throws on ANY query. If the
+    // guard ran after the first DELETE, this would surface as the pool's error
+    // rather than the refusal.
+    const exploding = {
+      query: () => {
+        throw new Error("purgeNamespace issued a statement on a refused name");
+      },
+    } as never;
+
+    await expect(purgeNamespace(exploding, "rico")).rejects.toThrow(
+      NamespacePurgeRefused,
+    );
+  });
+
+  test("every statement is a namespace-parameterized DELETE", async () => {
+    const seen: Array<{ sql: string; params: unknown[] }> = [];
+    const recording = {
+      query: (sql: string, params: unknown[]) => {
+        seen.push({ sql, params });
+        return Promise.resolve({ rowCount: 1 });
+      },
+    } as never;
+
+    const namespace = runNamespaces(
+      makeRunId({ prefix: "scenario-unit", randomHex: "0123456789ab", env: {} }),
+    ).primary;
+    const result = await purgeNamespace(recording, namespace);
+
+    expect(seen.length).toBeGreaterThan(0);
+    expect(result.deleted).toBe(seen.length);
+    for (const call of seen) {
+      // No statement may reach a row by any predicate other than this exact
+      // namespace, and the namespace is bound, never interpolated.
+      expect(call.sql).toContain("DELETE FROM");
+      expect(call.sql).toContain("WHERE namespace = $1");
+      expect(call.params).toEqual([namespace]);
+    }
+  });
+
+  test("records a failing table instead of aborting the rest", async () => {
+    // A partially-purged namespace still needs its tally reported; swallowing
+    // the remainder is the defect #655 is about.
+    let calls = 0;
+    const flaky = {
+      query: () => {
+        calls += 1;
+        if (calls === 1) return Promise.reject(new TypeError("boom"));
+        return Promise.resolve({ rowCount: 0 });
+      },
+    } as never;
+
+    const namespace = runNamespaces(
+      makeRunId({ prefix: "scenario-unit", randomHex: "0123456789ab", env: {} }),
+    ).primary;
+    const result = await purgeNamespace(flaky, namespace);
+
+    expect(Object.keys(result.failed_tables)).toHaveLength(1);
+    // The error CLASS, never the driver's message, which can echo row content.
+    expect(Object.values(result.failed_tables)).toEqual(["TypeError"]);
+    expect(calls).toBeGreaterThan(1);
+  });
+
+  test("reports zero deletions without inventing a table entry", async () => {
+    const empty = {
+      query: () => Promise.resolve({ rowCount: 0 }),
+    } as never;
+
+    const namespace = runNamespaces(
+      makeRunId({ prefix: "scenario-unit", randomHex: "0123456789ab", env: {} }),
+    ).primary;
+    const result = await purgeNamespace(empty, namespace);
+
+    expect(result.deleted).toBe(0);
+    expect(result.deleted_by_table).toEqual({});
+    expect(result.failed_tables).toEqual({});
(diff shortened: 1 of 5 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/655-eval-teardown.sh exit=0 sha=5bb07c2e4b5c6ae7b856d3fdcc7a308a65829fed at=2026-08-08T22:14:23.645Z

```
=== done-means #655: does the scenario runner remove the namespace it created? ===
throwaway database: done_means_655_69561_1786227262

PASS  created done_means_655_69561_1786227262
PASS  migrations applied
PASS  precondition: 0 eval-live-recall-% rows before the run

--- driving the real gate against the fresh database ---
INFO  receipt: /Volumes/ThunderBolt/_tmp/open-brain/_scratch/655-eval-teardown/driver-done_means_655_69561_1786227262.json
INFO  driver namespace: eval-live-recall-scenario-donemeans655-f3ffc7ad14c1
INFO  gate passed=true scenarios=3 teardown={"attempted":6,"archived":4,"already_absent":2,"failed":0} purged_rows=2
INFO  scenario capture-round-trip: passed=true
INFO  scenario durable-memory-pack-shape: passed=true
INFO  scenario checkpoint-wrap-round-trip: passed=true
INFO  guard: 6/6 refused, rows_touched_on_refusal=0, allows_own=true

--- (c) control: the run examined something, and teardown did not eat it ---
PASS  (c) the run really executed 3 scenario(s) — not a silent empty run
PASS  (c) scenario verification read its own seeded record BEFORE teardown ran
PASS  (c) the run's own assertions still pass with teardown in place

--- (a) the leak: nothing eval-prefixed is left behind ---
PASS  (a) record teardown reported failed=0
PASS  (a) 0 eval-live-recall-% rows remain — the run removed the namespace it created

--- (b) control: the prefix guard REFUSES a non-eval name, without mutating ---
PASS  (b) the guard refused all 6 non-eval name(s) it was pointed at
PASS  (b) a refused purge deleted ZERO rows — it refuses BEFORE it mutates
PASS  (b) the guard still ALLOWS a genuine per-run eval namespace — it discriminates

=== RESULT: PASS — the scenario run removes the namespace it created, the guard refuses everything else without mutating, and the evidence survives long enough to be verified ===
INFO  teardown: dropped done_means_655_69561_1786227262
```

</details>

rodaddy:
harvested: 60ae314b54dfe86f45815c94101493384e4c6735 — docs/lane-contract.md round 16 (#655 lane lessons: tally-is-not-proof teardown, namespaces as emergent row properties, SKIP-flag RED over broken-import RED, guard canary vs exception, session-scoped lookup window).

PR #656, opened 2026-08-08
feat(656): wire the capture-liveness observer into the serving process

rodaddy:
## Summary

Fills the composition port that PR #652 left open. #652 (MERGED) landed the
capture-liveness gatherer (`server/capture/liveness-observer.ts`) and the
`captureObserver` port on the application root (`server/application/index.ts:109`,
read per-request at `:204`), but nothing filled the port: `server/main.ts` passed
no `captureObserver`, so the serving process published no capture block on
`/health`. #652's own report named the blocker as an open config question, and
the operator ruling of 2026-08-08 (ledger item 28, `docs/issue-graph.md`) settled
it. State of this branch: WRITTEN and locally verified; no live process has been
observed publishing a capture block.

**1. `server/capture/liveness-observer.ts`** — new exported
`createCaptureHealthRuntime({ env, pool, logger })`.
`OPENBRAIN_CAPTURE_HEALTH_NAMESPACE` is REQUIRED config with NO fallback: unset
or empty builds NO observer and emits a `logger.warn` of
`capture_health_namespace_missing` carrying `config_key`. Window defaults to 360
minutes (6h) via `OPENBRAIN_CAPTURE_HEALTH_WINDOW_MINUTES`; refresh defaults to
60000 ms via `OPENBRAIN_CAPTURE_HEALTH_REFRESH_MS`. Applied defaults are
announced in a `capture_health_defaults` info line carrying `window_minutes`,
`refresh_interval_ms`, `window_source`, and `refresh_source`; an unparseable or
non-positive override is announced at warn (`capture_health_override_invalid`)
with the rejected text and the applied fallback — nothing is adjusted silently.
`observation()` was added to `CaptureLivenessObserver` so the runtime hands the
application the RAW COUNTS. The alternative — deriving an observation back out of
a finished reading — would have to invent a per-role split from a total, and the
per-role split is the entire point of #447.

**2. `server/main.ts`** — constructs the runtime before composition and registers
the observer as a `BackgroundRuntime` named `capture-health-observer`, so the
refresh timer stops on the application's EXISTING ordered-shutdown path
(`server/application/index.ts` owns it via `backgroundRuntimes`) rather than a
second hand-rolled teardown. A design lookup (qmd) surfaced that port and its
"a runtime running but absent from `backgroundRuntimes` is the abandoned-lease
bug" reasoning; the delta here is one runtime entry, not a new mechanism.

**3. Operator-owned live config edit — announced explicitly.** This branch
[...]

Issue #612, opened 2026-08-07
Component child-logger lines never reach the clone log files


rodaddy:
## Problem

No \`"component":"..."\` line from a pino child logger reaches the local clone's log files. Found by the #384 lane while proving the maintenance sweep producer live: \`maintenance_sweep_complete\` (child logger, \`component: "maintenance"\`) never appears in the clone logs, while module-logger lines (\`graph_derivation_enqueue_sweep\`) do.

## Why it matters

- Pre-existing routing quirk of the server path — not introduced by PR #609; durable \`maintenance_jobs\` rows prove the producer runs regardless.
- The same unrouted stream would carry the sweep's warn-level **bound-reached** lines that #384's acceptance criteria require, so this defect blocks the final criterion of #384.

## Evidence

- PR #609 self-review: "no \`"component":"..."\` line does [reach the clone's log files]".
- Reproduce: trigger a sweep on the local clone (\`server/main.ts\` entrypoint), compare child-logger vs module-logger emissions against the clone log files.

## Done means

An executable check (scripts/done-means/) that greps the clone log file for a child-logger line with a \`component\` field emitted during the run, red on current behavior.

Blocks: #384 (final logging criterion).

PR #660, opened 2026-08-08
fix(local-clone): pass capture-health config to the server child and announce dropped keys

### LINKED ISSUE #659: local-clone launcher env allowlist silently drops OPENBRAIN_CAPTURE_HEALTH_* — merged observer wiring cannot go RUNNING
## Observed (RUNNING, this session)

PR #656 merged the capture-health observer wiring in server/main.ts, ledger item 28 supplied the config ruling, and OPENBRAIN_CAPTURE_HEALTH_NAMESPACE=rico sits in /Volumes/ThunderBolt/open-brain-local/local-clone.env (line 65, byte-clean, mtime before boot). The clone was redeployed at 1d2e848 with revision proof PASSED — and live /health publishes NO capture block.

Evidence chain:
- Boot log (autostart.out.log 22:21:35Z): `capture_health_namespace_missing ... capture_health_enabled:false` — the observer's loud-on-absence path, working as designed.
- Launcher process 3067 (`bun run scripts/local-clone.ts --start`): HAS the var in its env (the autostart wrapper sources local-clone.env correctly).
- Server child 3463 (`bun run server/main.ts`): does NOT have it.

## Root cause

`scripts/local-clone.ts` `buildChildEnvironment()` passes the child an ALLOWLIST (`CHILD_ENV_KEYS`) plus two prefix families (`AUTH_TOKEN_USER_*`, `OPENBRAIN_TRACING_*`). `OPENBRAIN_CAPTURE_HEALTH_*` is in neither, so the launcher drops it. This is the third instance of the deploy-coupling class the code itself documents at the #530 tracing comment ("must reach the server child or createTracingRuntime silently stays off").

Two defects here, not one:
1. The three OPENBRAIN_CAPTURE_HEALTH_* keys don't reach the server child.
2. The allowlist violates "nothing is adjusted silently" (repo coding standard, operator ruling 2026-08-08): it drops configured keys without announcing which. Every future server config key re-buys this bug until the drop is announced.

## Done-means design (red-first)

RED against current main: a unit-level check on buildChildEnvironment with OPENBRAIN_CAPTURE_HEALTH_NAMESPACE set in the input env asserts the key in the output — fails today. Plus an announce clause: with a known-unlisted key set, the launcher's output names the dropped key (fails today, silent). Control: DB_*/AUTH_TOKEN_* still pass through; a deliberately-unlisted junk key still does NOT pass (the allowlist stays an allowlist — announced, not abolished). Live GREEN after redeploy: /health publishes the capture block (that step is the controller's, post-merge).

Also worth a clause: done-means 656 clause (b) drove createShadowApplication directly, so this whole launcher seam was outside its vantage — the check measured the composition, not the deployment chain. The lane should say whether a cheap chain-level clause (spawn the launcher's buildChildEnvironment output into the shadow app) is feasible.

rodaddy:
## Summary

- Closes #659. `scripts/local-clone.ts` `buildChildEnvironment()` passes the server child an allowlist (`CHILD_ENV_KEYS`) plus two prefix families. The three `OPENBRAIN_CAPTURE_HEALTH_*` keys were in neither, so the launcher dropped them — and dropped them **silently**. `OPENBRAIN_CAPTURE_HEALTH_NAMESPACE=rico` sat in the deployed clone's env file, the redeploy's revision proof PASSED, and live `/health` still published no capture block. The launcher process had the variable; the child it spawned did not. The merged #656 observer could not go RUNNING.
- **Defect 1 — the keys.** The three capture-health keys are now in `CHILD_ENV_KEYS`, listed **by name** rather than swept in by an `OPENBRAIN_CAPTURE_HEALTH_` prefix, so adding a fourth stays a deliberate act with a reviewer attached. This is the third instance of the deploy-coupling class the file already documents at its own #530 tracing comment ("must reach the server child or `createTracingRuntime` silently stays off").
- **Defect 2 — the silence.** New `describeDroppedChildEnvironment()` reports every configured key the child allowlist drops, and the start path announces it **before** spawning. Fixing only defect 1 closes the issue and leaves the mechanism that produced it: every future server config key would re-buy this same bug. This is the class already recorded in `docs/sme/gotcha-agent.md` from #464 — "An allowlist that drops unrecognized keys is accept-and-ignore, not tolerance" — whose first review question is exactly *does the drop path name what it dropped?*
- **The allowlist stays an allowlist.** Reporting is not honoring; the ruling is announced, not abolished. A key named in the report is one a human must either add to `CHILD_ENV_KEYS` or delete from the env file. Clause (d) fails any "fix" that passes the whole environment through, and the junk key it uses carries the `OPENBRAIN_` prefix so a lazy whole-family passthrough fails too.
- **Two scoping decisions, both asserted rather than assumed.** The report is scoped to Open Brain server-config prefixes because `scripts/local-clone-autostart.sh:56-59` does `set -a; source local-clone.env`, making the launcher's environment the env file unioned with launchd's — an unscoped report would name `PATH` and `HOME` on every boot, noise an operator learns to skip, which is silence with extra steps. And it returns **names only, never values**: the dropped set contains secrets in the general case (an unlisted `AUTH_TOKEN_*` spelling lands there) and this line goes to a log file the deploy runbook tells an operator to read.

**Found while fixing, wider than the issue as filed (PROPOSED, not addressed here):** the deployed env file configures **seven** keys this allowlist drops, not three. Besides the capture-health key: `EMBEDDING_WATCHDOG_RESTART_SCRIPT`, `LOG_LEVEL`, `LOG_MAX_BYTES`, `LOG_MAX_FILES`, `OPENBRAIN_MCP_AUDIT_ENABLED`, `SERVICE_NAME` — every one read by the serving tree, and `LOG_LEVEL`/`SERVICE_NAME` read by `server/config.ts` itself, where they silently default (`:124`, `:126`). So the operator sets `LOG_LEVEL` and the server runs at `info` regardless. Those are left for an operator ruling rather than swept into the allowlist here: honoring a key changes deployed behavior and is not this issue's scope. **The announcement now makes all six visible on the next boot instead of leaving them invisible.**

## Verification

- Done-means: scripts/done-means/659-launcher-env-passthrough.sh

**RED** (at `origin/main`, before the fix — real nonzero exit):

```
FAIL  (a) capture-health keys DROPPED by the launcher: OPENBRAIN_CAPTURE_HEALTH_NAMESPACE, OPENBRAIN_CAPTURE_HEALTH_WINDOW_MINUTES, OPENBRAIN_CAPTURE_HEALTH_REFRESH_MS
FAIL  (b) scripts/local-clone.ts exports no describeDroppedChildEnvironment — the launcher has no way to report what it dropped
FAIL  (b:no-values) (same cause)
PASS  (c) control: DB_*, AUTH_TOKEN_*, AUTH_TOKEN_USER_* and OPENBRAIN_TRACING_* all still reach the child
PASS  (d) control: the allowlist holds — junk key and ambient host vars still excluded
FAIL  (e) (same cause)
FAIL  (f) chain: launcher exit=0, announced 0 line(s), names the dropped key=false, capture namespace reached the spawned child=false
clauses: 7 run, 5 failed          EXIT=1
```

Controls (c) and (d) **pass pre-fix by design** — a check that fails everywhere proves only that it fails. Clause (f) is the live symptom exactly: the launcher ran to completion, announced nothing, and the namespace never reached the child.

**GREEN** (after the fix): `clauses: 7 run, 0 failed`, exit 0. Also re-run from a **clean clone** of the pushed branch (7/7) — the only thing that catches a lane file silently dropped by `.gitignore`.

- `bun run test:isolated` — **3747 pass, 35 skip, 0 fail**, 15298 expect() calls, 244 files; isolated DB dropped cleanly.
- `bunx tsc --noEmit` — clean. It caught the real boundary: adding `announce` to `LocalCloneLauncherDependencies` broke the test fixture, which is the type system doing its job.

**Mutation proofs** (every negative-match clause seen to fail):

1. *Allowlist abolished* (pass whole env) → (d) FAILS naming `PATH, HOME, SSH_AUTH_SOCK` as leaked; (b)(e)(f) also catch it.
2. *Scope filter removed* (announce everything) → (e) FAILS with `ambient host vars announced=PATH,HOME,SSH_AUTH_SOCK`. **Only (e) catches this** — every other clause accepts an announce-everything filter, which is why the scope is asserted from both sides in one clause.
3. *One capture key renamed* → (a) FAILS naming just that key, proving it is not a substring superset (round 17).
4. *Key removed from allowlist* → the new unit test `passes the capture-health observer configuration to the server child` FAILS. New tests proven able to fail before being trusted.

## Critical Self-Review

- Highest-risk behavior: the `SERVER_CONFIG_PREFIXES` scope. Too narrow and a genuinely-dropped key stays invisible (the #659 defect, quieter); too wide and the boot log becomes noise an operator skips. Asserted from both directions by clause (e) and mutation 2, but it is a heuristic over key *names* and a future key outside these eight prefixes would not be reported.
- Assumptions that could be wrong: that announcing on **stderr** reaches the operator. Verified the autostart wrapper does not redirect the streams separately, so both land in the same log file — but I did not observe the deployed log after a real restart, because redeploy is the controller's post-merge step. If some launchd plist splits the streams, the line lands in a file nobody reads.
- Missing/weak tests: no test asserts the announcement's exact wording or that it precedes the spawn — clause (f) proves ordering only implicitly (the announce happens before `deps.child.spawn`, but nothing fails if that order is swapped). The `announce`-before-spawn intent is enforced by code position and a comment, not by a check.
- Security/permission risk: the report could log a secret. Actively defended: `b:no-values` asserts values never appear, and mutation-checked by construction (the junk key's value is a distinctive string the clause greps for). Names of secret-bearing keys can still appear, which is intended and is not itself a leak.
- Migration/deploy risk: none in-repo. Deploy-side, this changes what the server child receives — that is the point. The capture-health observer will construct on the next redeploy where the namespace is set, so `/health` gains a `capture` block. Per #656's design absence stays green, so a clone without the namespace is unaffected.
- Downstream client/runtime risk: none. `scripts/local-clone.ts` is the local dogfood launcher only; it is explicitly not the core01 deploy path (`com.rico.open-brain`) and no client, MCP tool, schema, or Python surface is touched.
- Rollback/cleanup concern: revert the commit. The `announce` dependency is a new **required** member of `LocalCloneLauncherDependencies`, so any out-of-tree construction of that object would need updating — there are none in this repo beyond `productionDependencies` and the test fixture.
- Fixes made before PR: (1) my first RED was a **false RED** — a static import of the not-yet-existing export killed the driver at module resolution, so clauses (a)/(c)/(d) measured nothing; switched to a dynamic import so the pre-fix tree gives a clean per-clause RED, and re-proved RED afterward. (2) Clause (f) initially failed on three successive *fixture* defects (WAL path, clone-root existence, `QMD_PATH`), not the defect under test; corrected so (f) fails for the right reason. (3) The clone-root fixture was moved to a repo-relative gitignored `_scratch/` path rather than an absolute machine path, which died with EACCES on the Linux CI runner once before (#612).
- Known residual risk: **the live half is unproven from here.** Every claim in this PR is at the code/check level. That the deployed clone actually publishes a capture block requires a redeploy plus a live `/health` read, which is the controller's post-merge step — and #659 exists precisely because a passing revision proof was mistaken for that. Also: the six other silently-dropped keys are named above but not honored; `LOG_LEVEL` and `SERVICE_NAME` still silently default in `server/config.ts` regardless of the env file.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the class is already the active entry "An allowlist that drops unrecognized keys is accept-and-ignore, not tolerance" (`docs/sme/gotcha-agent.md`, from #464). This PR is a third instance of that recorded pattern, not a new one — its review questions are what shaped the fix. A new entry would duplicate it; the lane report proposes adding #659 as provenance to the existing entry instead.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: the launcher is proven through its injected boundaries with no DB and no network; the live proof is the post-merge redeploy plus a `/health` read, which belongs to the controller and is named as residual risk above.

## Contract Parity

- Contract parity: [x] runtime-specific because: this is the local dogfood launcher's process-spawn boundary. No MCP tool, schema, transport, or client contract is touched.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: the change is confined to `scripts/local-clone.ts` and its test, plus a new done-means check. No MCP tool, schema, protocol, or client-facing surface changes, so no downstream runtime consumes anything different. The one deploy-side consequence — the server child receiving three more env keys — lands on the local dogfood clone only, not core01.

### DIFF
--- a/scripts/done-means/659-launcher-env-passthrough.driver.ts
+++ b/scripts/done-means/659-launcher-env-passthrough.driver.ts
@@ -1,0 +1,465 @@
+/**
+ * DONE-MEANS driver for #659 — the local-clone launcher's child-environment
+ * allowlist carries the capture-health keys, and ANNOUNCES every configured key
+ * it drops instead of discarding it in silence.
+ *
+ * ---------------------------------------------------------------------------
+ * THE DEFECT, AS OBSERVED
+ * ---------------------------------------------------------------------------
+ * PR #656 merged the capture-health observer wiring into `server/main.ts` and
+ * ledger item 28 supplied the config ruling. `OPENBRAIN_CAPTURE_HEALTH_NAMESPACE`
+ * was then set in the deployed clone's env file, the clone was redeployed with a
+ * passing revision proof — and live `/health` still published no capture block,
+ * because `buildChildEnvironment()` never handed the key to the server child.
+ * The launcher process HAD the variable; the child it spawned did not.
+ *
+ * That is two defects, and this check asserts both:
+ *
+ *   1. The three `OPENBRAIN_CAPTURE_HEALTH_*` keys do not reach the child.
+ *   2. The allowlist drops CONFIGURED keys with no signal at all, so every
+ *      future server config key re-buys this same bug. That violates the repo
+ *      coding standard "Nothing is adjusted silently" (operator ruling
+ *      2026-08-08): a tool that transforms its input to satisfy a constraint
+ *      announces the adjustment in its normal output.
+ *
+ * The SME KB already carries this exact class — `docs/sme/gotcha-agent.md`,
+ * "An allowlist that drops unrecognized keys is accept-and-ignore, not
+ * tolerance" (#464). Its first review question is the one this check enforces:
+ * *does the drop path name what it dropped?* A count is unactionable at the
+ * moment it matters; names cost one list. Its second is why clause (a) is a bug
+ * in both directions: a dropped key that appears in the project's own constant
+ * set is not an unknown future field, it is a supported concept the reader
+ * forgot.
+ *
+ * ---------------------------------------------------------------------------
+ * WHY THE ALLOWLIST STAYS AN ALLOWLIST
+ * ---------------------------------------------------------------------------
+ * The launcher exists to hand the server child a tightly scoped environment —
+ * `scripts/local-clone.test.ts` asserts `PATH`, `SSH_AUTH_SOCK` and a host
+ * identity var never reach it. Passing the whole environment would fix #659 by
+ * deleting the property the launcher was built for. Clause (d) is the control
+ * that fails any such "fix": an unlisted junk key must still be dropped.
+ *
+ * ---------------------------------------------------------------------------
+ * WHY THE ANNOUNCEMENT IS SCOPED, AND WHY THAT SCOPE IS ITSELF ASSERTED
+ * ---------------------------------------------------------------------------
+ * `scripts/local-clone-autostart.sh:56-59` does `set -a; source local-clone.env`,
+ * so the launcher's own environment is the env file UNIONED with launchd's
+ * ambient environment. Announcing every dropped variable would name `PATH`,
+ * `HOME`, `SSH_AUTH_SOCK` and dozens of others on every boot — noise that
+ * teaches an operator to ignore the line, which is silence with extra steps.
+ *
+ * So the announcement is scoped to keys that look like Open Brain SERVER
+ * configuration, and clause (e) asserts that scope from BOTH sides: an ambient
+ * host variable is NOT announced, and a configured-looking one IS. A scope
+ * rule with only its positive half tested passes just as well when the filter
+ * is "announce nothing".
+ *
+ * No database, no network, no child process: `buildChildEnvironment` and
+ * `describeDroppedChildEnvironment` are pure functions over a record. Clause
+ * (f) drives the REAL `runLocalCloneLauncher` start path through its injected
+ * boundaries so the announcement is proven to reach the launcher's actual
+ * output, not merely to be computable in isolation.
+ */
+import {
+  buildChildEnvironment,
+  runLocalCloneLauncher,
+  type LocalCloneLauncherDependencies,
+} from "../local-clone.ts";
+import {
+  CAPTURE_HEALTH_NAMESPACE_ENV,
+  CAPTURE_HEALTH_REFRESH_ENV,
+  CAPTURE_HEALTH_WINDOW_ENV,
+} from "../../server/capture/liveness-observer.ts";
+import { EventEmitter } from "node:events";
+import { mkdirSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import type { ChildProcess } from "node:child_process";
+
+/**
+ * A real directory for the clone-root fixture.
+ *
+ * `src/local-clone-mode.ts` requires OPENBRAIN_LOCAL_CLONE_ROOT to EXIST, and
+ * clause (f) drives that validation for real. Repo-relative and gitignored
+ * (`.gitignore:119`) rather than an absolute machine path — a hardcoded
+ * `/Volumes/...` default died with EACCES on the Linux CI runner once already
+ * (docs/lane-contract.md, #612 harvest).
+ */
+const CLONE_ROOT = fileURLToPath(
+  new URL("../../_scratch/done-means-659-clone-root", import.meta.url),
+);
+
+const results: Array<{ clause: string; ok: boolean; detail: string }> = [];
+
+function clause(name: string, ok: boolean, detail: string): void {
+  results.push({ clause: name, ok, detail });
+  console.log(`${ok ? "PASS" : "FAIL"}  (${name}) ${detail}`);
+}
+
+/**
+ * The env the deployed clone actually runs with: the env file's keys unioned
+ * with a handful of ambient launchd/shell variables. Values are placeholders —
+ * this driver never connects to anything — but the KEY SET is the real one,
+ * read from `/Volumes/ThunderBolt/open-brain-local/local-clone.env` on
+ * 2026-08-08. A fixture whose key set is invented would prove nothing about the
+ * deployment this issue is about (round 12: query the real distribution before
+ * inventing a fixture shape).
+ */
+function deployedEnv(): Record<string, string | undefined> {
+  return {
+    // --- allowlisted today, must keep passing (clause c) ---
+    ALLOWED_ORIGINS: "http://127.0.0.1:7100",
+    AUTH_TOKEN_ADMIN: "placeholder-admin",
+    AUTH_TOKEN_AGENT: "placeholder-agent",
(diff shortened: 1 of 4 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/659-launcher-env-passthrough.sh exit=0 sha=89ff7067a6fd2f0e8b3d4b74a63e6df7965a1c2e at=2026-08-08T22:42:07.160Z

```
INFO  repo:   /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-660-85165mskykzot
INFO  driver: scripts/done-means/659-launcher-env-passthrough.driver.ts (pure env projection + injected launcher boundaries; no DB, no network)

PASS  (a) all three capture-health keys reach the server child with their configured values (namespace=rico)
PASS  (b) a configured-but-unlisted key is named in the launcher's drop report: OPENBRAIN_DEFINITELY_NOT_A_REAL_SERVER_KEY
PASS  (b:no-values) the drop report names keys without echoing their values
PASS  (c) control: DB_*, AUTH_TOKEN_*, AUTH_TOKEN_USER_* and OPENBRAIN_TRACING_* all still reach the child
PASS  (d) control: the allowlist holds — OPENBRAIN_DEFINITELY_NOT_A_REAL_SERVER_KEY and the ambient host vars are still excluded
PASS  (e) scope, both directions: ambient host vars announced=no, configured unlisted key announced=true
PASS  (f) chain: launcher exit=0, announced 1 line(s), names the dropped key=true, quiet on ambient=true, capture namespace reached the spawned child=true

clauses: 7 run, 0 failed

DONE-MEANS #659: PASS — the capture-health keys reach the server child, and configured keys the allowlist drops are announced by name.
```

</details>

rodaddy:
harvested: ae37e4f (wip/2026-08-07) — docs/lane-contract.md round 18 (#659 lane + controller lessons: revision proof ≠ feature-live proof; ask which seam the passing check could not see; allowlist announce-on-drop; dynamic import for new-export checks; both-halves-one-clause scope rules; read WHY each RED clause failed; own-formatting before known gate defect; self-reported rm reflex).

Issue #676, opened 2026-08-09
[cutover-note] DB_NAME silently defaults to open_brain — identity-selecting config with a fallback (ledger 28/31)


rodaddy:
Found by the pre-cutover config audit (2026-08-09). Identity-config-loud-on-absence review (ledger 28/31).

## The one that matters
src/db/pool.ts:22 — DB_NAME defaults to \`open_brain\` if unset. This SELECTS WHICH BRAIN you connect to: a core01 deploy missing DB_NAME from its env connects to a database named open_brain and, if one exists, silently serves it. Same class as the capture-health namespace ruling (identity-selecting), but with a silent fallback instead of a loud refusal. On a cutover this is how you serve the wrong database and not know.

## Lesser, same family
- OPEN_BRAIN_SERVER_IP / OPEN_BRAIN_BIND_HOST unset → LAN detection then literal "unknown" (CONFIG_REFERENCE.md:151-166): /health advertises a guessed identity. Bounded to private ranges; low severity but a guess presented as fact.
- OPEN_BRAIN_RUN_MIGRATIONS defaults ON (src/index.ts:320, !== "0") — fail-open for schema mutation; interacts with the unversioned-plist blocker (#675 B2): if the plist omits it for a non-primary worker, both workers migrate.
- SERVICE_NAME / OPEN_BRAIN_WORKER_NAME default/unset — log attribution only, note.

## Also
docs/CONFIG_REFERENCE.md self-admits it was measured against src/ on 2026-07-30, predates server/, and documents none of the capture-health keys. Worth a refresh pass so the cutover has an accurate config map.

## Wanted (operator ruling per ledger 28/31 — required config, loud on absence)
Make DB_NAME required-no-fallback at minimum; decide the rest. Then a done-means check that a missing identity key fails loudly rather than defaulting. Not a lane's call to change identity-config semantics silently.

Issue #643, opened 2026-08-08
flaky: ob-backfill quadratic-blowup test asserts an absolute wall-clock threshold (1005ms vs 1000ms)


rodaddy:
`scripts/ob-backfill.test.ts:259` asserts a wall-clock duration:

```ts
expect(performance.now() - startedAt).toBeLessThan(1_000);
```

Observed failing in CI at **1005.5852320000013 ms** — missing by 5.6 ms — on run 31267844990 (`db-integration`, `Test (JUnit reporter for the anti-skip guard)`). A re-run of the **same SHA** (30efb2a, run 31267860992) passed with the identical tree, so this is timing, not behavior. Suite result on the failing run: 3711 pass / 34 skip / **1 fail**.

This is the flake class already recorded in `docs/lane-contract.md` Tightenings (2026-08-08 round 5): "Wall-clock assertions (`toBeLessThan(1000)` ms) are CI flake generators" — previously filed as #632 and #634 for other instances. A shared runner under load makes any absolute millisecond threshold a coin flip near the boundary.

**What the test is actually defending:** that `sanitize()` is not quadratic in token count. That property is better expressed as a RATIO between two input sizes (e.g. 12k tokens takes < ~4x the time of 3k tokens) than as an absolute ceiling, because the ratio is robust to a slow or contended runner while still failing loudly on real quadratic blowup.

Found by the #563 lane (PR #639), which does not touch this file — reported rather than absorbed, per the lane contract.

rodaddy:
Two more datapoints from tonight's cutover-blocker lanes (2026-08-09), both on branches that touch neither `scripts/ob-backfill.ts` nor its test:

- **PR #690** (lane/677, scheduled backup) — `check` FAILED at SHA `adb766bf`: `Expected: < 1000` / `Received: 1112.08`. 3760 pass / 35 skip / **1 fail** across 3796 tests; the one fail is this test.
- **fix/679-register-dev-safety-gates** — the SAME test failed on `pull_request` at SHA `585a9772`, and **the identical SHA passed on its `push` run**. That is the two-runs-same-SHA flake signal, on a SHA neither run's branch shares with #690.

Differential evidence that it is not branch-owned, for #690 specifically:
- `git diff origin/main -- scripts/ob-backfill.test.ts scripts/ob-backfill.ts` is EMPTY — neither the test nor its subject is touched.
[...]

PR #690, opened 2026-08-09
feat(677): schedule backup.ts on launchd, prove one restore, guard against deploy overlap

### LINKED ISSUE #677: [cutover-blocker] No scheduled backup on core01 — tooling built & CI-drilled but never run; only backup is 16 days old, 15 migrations behind
Found by the pre-cutover durability audit (2026-08-09, read-only, live dogfood DB queried). The single highest-value finding of the pre-cutover sweep.

## The gap (RUNNING — verified absent this session)
The backup substrate is real and good: scripts/backup.ts, backup-verify.ts, restore.ts, content-free manifest, and a LIVE restore drill in CI (scripts/__tests__/backup-restore-live.test.ts, db-integration job with OPENBRAIN_BACKUP_DRILL=1). What does not exist is any INVOCATION:
- docs/backup-restore.md:149-156: stale-backup alerting is DOCUMENTED-ONLY; launchd/cron wiring "deliberately left to the operator runbook and NOT shipped in this repo."
- `rg -rn -i backup deploy/` → 0. `fd -e plist` → 0 schedulers.
- Documented backups root /Volumes/ThunderBolt/open-brain/backups/ does NOT exist.
- Only real set on disk: /Volumes/ThunderBolt/open-brain-local/backups/core01-20260724/ — created 2026-07-24, manifest head 031. Repo head is 046. 15 migrations of schema drift.
- Stated RPO is 24h (docs/backup-restore.md:17); actual observed RPO is 16 days and unbounded.

## Data-loss scenario
core01 disk fails after cutover. Restore possible only from the 07-24 set. Everything since is gone — live count: 74,904 raw turns, 26,453 thoughts, 11,268 session events, 6,441 candidate memories, 3,480 lanes. Restore WOULD work (verify passes restorable_with_migrations, 031→046 run forward), yielding a functioning Open Brain containing July's memory.

Process crash (vs disk loss) is fully handled and loses nothing — watermark model + Postgres transactions. This blocker is disk-loss / catastrophic only, but that's the case backups exist for.

## Wanted
1. Schedule backup.ts on core01 (launchd plist — the one-command fix; the script is already tested and CI-drilled).
2. Run ONE restore end-to-end before cutover to prove the current-schema path, not just the 07-24 drill.
3. Encode in the schedule that a backup job must NOT overlap a deploy (docs/backup-restore.md:200: mid-migration dumps capture a half-applied schema verify can't detect from outside).
This single action moves the durability surface from blocker to sound.

Truth grammar: backup/restore tooling MERGED + TESTED-in-CI; scheduled backup on core01 PROPOSED; a current restorable backup NOT RUNNING.

rodaddy:
## Summary

- Closes #677 (cutover blocker B4). The backup substrate was already real, tested, and CI-drilled; what did not exist was any **invocation**. The only set on disk was 16 days and 15 migrations stale against a stated 24h RPO.
- **The schedule:** a launchd plist template (03:00 daily, LowPriorityIO) plus an installer that renders, lints, and bootstraps it, following the existing install-qmd-sync-launchagent.sh pattern exactly — staged write, XML metacharacter validation before anything is created, plutil lint. Adds a render-only mode so the done-means check asserts what the **installer** produces rather than hand-substituting tokens itself.
- **The runner** (scripts/backup-scheduled-run.sh) is what launchd invokes: dated set, explicit env load (launchd inherits almost no environment — the classic reason a scheduled job works by hand and fails at 03:00), then backup, then verify the set it just wrote, then a root-level staleness check. It never removes anything; retention stays operator-run per the existing doc. DB_NAME is **required**, not defaulted (#676: it selects which brain gets backed up, and a job quietly backing up the wrong database would look healthy indefinitely).
- **The overlap guard** (scripts/deploy-lock.ts): docs/backup-restore.md already recorded that a dump taken mid-migration captures a half-applied schema that verify **cannot detect from the outside**. That artifact is worse than a missing backup because it passes verification, so scheduling without a guard would have made this blocker subtler rather than smaller. The mechanism is a session-scoped Postgres advisory lock on the database both operations touch — the same primitive server/db/migrations.ts already uses to serialize migrations, with a distinct key. Deliberately asymmetric: the deploy waits (bounded, attended); the backup refuses immediately with exit 4, writes nothing, and names the deploy as the cause.
- **The restore proof:** one full backup-to-restore round trip run locally end to end at migration head 046, comparing row counts across all 23 tables and an information_schema-derived schema hash between source and restored databases, computed from **outside** both scripts.
- **Not executed by this lane:** the core01 installation step is WRITTEN into the cutover runbook section of docs/backup-restore.md. core01 was never contacted.

## Verification

- Done-means: scripts/done-means/677-scheduled-backup.sh

RED (pre-change tree, final version of the check): 4 passed / 7 failed, exit 1. Failing: (a) no plist template, (b) no installer, (d1) backup SUCCEEDS while the deploy lock is held, (d2) no refusal to judge, (f) deploy never takes a lock, (g) x2 runbook still claims the wiring is not shipped.

GREEN (post-change): 12 passed / 0 failed, exit 0.

Clauses (c) and (e) PASS pre-fix **by design** and are labeled as controls in the check header: (c) is the restore proof whose value is the receipt it produces, not a red-to-green transition, and (e) proves the check does not simply fail everywhere. RED was re-proven against the pre-change tree with the FINAL version of the check after every repair, by file-copy rather than git stash.

Also: bunx tsc --noEmit clean. backup.test.ts + backup-lib.test.ts = 65 pass / 0 fail via bun run test:isolated. Pre-push hook ran the full suite (3891 tests across 244 files) and passed.

Not applicable, stated rather than skipped: no Python package surface is touched, and no live Open Brain smoke applies because nothing here serves traffic — this is operator tooling plus a scheduler, and its proof is the restore round trip.

## Critical Self-Review

- Highest-risk behavior: the guard sits in the path of every backup, so a defect in it does not corrupt data but **prevents backups** — the blocker restated in a subtler form. This is exactly why clause (e) exists as a control, and why the lock is taken and released rather than held for the dump's duration.
- Assumptions that could be wrong: that take-and-release at the START of the backup is sufficient. It closes the window where a deploy is already running; the other direction (a deploy starting mid-dump) is closed by the deploy's own bounded WAIT rather than by holding the lock across the whole dump. Stated openly in the module's own comments. Holding it for the full dump would be stronger, but coupling a long-lived lock to the backup pool's lifetime risks a backup failure leaving the lock held for the pool idle timeout — trading a rare corrupt dump for a recurring blocked deploy. If the operator prefers the stronger form, that is a one-line change and a ruling.
- Missing/weak tests: the launchd job has never been **bootstrapped** — the check renders and lints the plist, it does not prove launchctl accepts and fires it. That proof only exists on a machine where the agent is installed, which is the core01 runbook step. Called out as WRITTEN-not-RUNNING below rather than implied away.
- Security/permission risk: the installer interpolates values into XML and validates every one against metacharacters before writing; clause (b) proves the refusal AND that a refused run leaves no plist behind. Plist is written 0600, runner installed 0700, matching the qmd-sync precedent. No credential is read, written, or logged by any new file.
- Migration/deploy risk: core01-deploy-local.sh gains a lock acquisition before staging and migrate. If the lock helper cannot connect, the deploy FAILS CLOSED with a named error rather than proceeding unguarded. A deploy killed at any point releases the lock automatically because it is session-scoped, so there is no stale-lock file and no cleanup path to get wrong.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, contract, or client-facing surface is touched — the change is operator tooling and a scheduler.
- Rollback/cleanup concern: rollback is removing the LaunchAgent (operator, on core01) and reverting the commit. Nothing here mutates existing data, and the runner deliberately never deletes a backup set.
- Fixes made before PR: the installer's placeholder guard matched the template's own explanatory comment and refused its own correct output — scoped to string values. The identical over-broad match existed in the checker, so it was one mistake made twice, in the checker and in the subject. Also: (d2) was a bare substring match that PASSED pre-fix because the successful backup's receipt embeds a path this check itself chose containing the word deploy — a false green, repaired by gating the clause on the refusal having actually occurred. Clause (f) matched the deploy script's own comments rather than its commands, reporting the lock as taken after migrate when it is taken before. The seed used column and table names that do not exist, and had a silent fallback that would have hidden exactly the schema drift the clause detects.
- Known residual risk: the schedule is WRITTEN, not RUNNING. No backup has ever fired on core01 from this job, and the backups root does not exist there yet. The blocker is not closed by merging this — it is closed by the operator running the install step and confirming a set lands. Second residual: the staleness alert exits non-zero and writes to the launchd log; there is no push notification, so it is only seen by something that reads that log.
- SME review-memory update: [ ] docs/sme/ updated or [x] not applicable because: the lessons here are lane-operating lessons (false-green shapes, cluster-wide pg_locks readings, killing a client does not release a session lock) rather than reviewer-facing code patterns, and they are supplied in the lane report for the lane-contract Tightenings harvest.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: nothing in this change serves traffic; the proof is the restore round trip, run locally against isolated databases

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, schema, or tool surface is touched — this is operator tooling and a launchd schedule

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool, schema, protocol, transport, or client-facing behavior changes. No downstream consumer can observe this change; it adds a scheduler, a guard between two operator scripts, and a doc section.
- One measured finding worth carrying: killing a psql client does NOT release its advisory lock — the backend stays inside its query and holds it. Release is server-side. This cost a full false failure that read exactly like an implementation defect.

### DIFF
--- a/docs/backup-restore.md
+++ b/docs/backup-restore.md
@@ -146,14 +146,92 @@ Fail-closed, no partial success:
    `err.message`. A failed `pg_dump` also removes its partial dump file so a
    retry is not blocked behind `--force`.
 
-### Stale-backup alerting — DOCUMENTED-ONLY (mechanism TESTED, scheduling deferred)
+### Stale-backup alerting — mechanism TESTED, now WIRED into the scheduled job
 
 `backup-verify --dir <backups-root> --max-age-hours N` exits 3 with a `stale`
 status when the newest VALID backup is older than N hours (a corrupt set does
-not count as a valid backup — TESTED in `scripts/backup-lib.test.ts`). Wire
-it into cron/launchd on core01 and alert on nonzero exit; the actual launchd
-plist/notification wiring is deliberately left to the operator runbook and is
-NOT shipped in this repo (scripts stay operator-run, no service wiring).
+not count as a valid backup — TESTED in `scripts/backup-lib.test.ts`).
+
+This is no longer left to the operator to wire. `scripts/backup-scheduled-run.sh`
+runs it as the last step of every scheduled backup, at `--max-age-hours 26`
+(one daily interval plus two hours of slack), across the whole backups root.
+That is deliberately a DIFFERENT question from "did tonight's run succeed":
+a job that has been failing silently for a week passes the per-set check on
+the night it finally works and still fails this one.
+
+### Scheduled backup on core01 (#677) — WRITTEN, not yet RUNNING
+
+**Status: the wiring SHIPS in this repo and is proven locally; it has NOT been
+installed on core01.** Everything below is the cutover-runbook step, to be run
+BY THE OPERATOR ON CORE01. Nothing in this section has been executed against
+core01 by any agent.
+
+Issue #677 (cutover blocker B4) measured the gap this closes: the backup
+substrate was real, tested, and CI-drilled, and had **never been invoked** —
+the only set on disk was 16 days and 15 migrations stale, against a stated
+24h RPO.
+
+What ships:
+
+| File | Role |
+|---|---|
+| `docs/deploy/com.rico.open-brain-backup.plist.template` | launchd job, 03:00 daily, `LowPriorityIO` |
+| `scripts/install-backup-launchagent.sh` | renders + lints + bootstraps it (same staged-write/validation pattern as `install-qmd-sync-launchagent.sh`) |
+| `scripts/backup-scheduled-run.sh` | what launchd runs: dated set, env load, backup, verify, staleness alert |
+| `scripts/deploy-lock.ts` | the deploy-overlap guard shared by backup and deploy |
+
+**The core01 install step (operator, on core01):**
+
+```bash
+cd /Volumes/ThunderBolt/open-brain/app
+bash scripts/install-backup-launchagent.sh
+# Prove the wiring immediately — do not wait for 03:00:
+launchctl kickstart -k gui/$(id -u)/com.rico.open-brain-backup
+# Then confirm a set landed and verifies:
+bun run scripts/backup-verify.ts --dir /Volumes/ThunderBolt/open-brain/backups --max-age-hours 26
+```
+
+Note that `/Volumes/ThunderBolt/open-brain/backups/` does not exist on core01
+today (#677 confirmed its absence); the runner creates it on first run.
+
+**`DB_NAME` is required, not defaulted.** The runner refuses to start without
+it. `DB_NAME` selects *which brain* gets backed up, and #676 records that it
+otherwise silently defaults to `open_brain` — a scheduled job quietly backing
+up the wrong database would look healthy indefinitely.
+
+**Retention stays operator-run**, exactly as the retention row above specifies.
+The scheduled job creates and verifies sets; it never removes one.
+
+#### The deploy-overlap guard
+
+A dump taken mid-migration can capture a half-applied schema that
+`backup-verify` **cannot detect from the outside** (see the upgrade matrix
+below). That is worse than a missing backup, because it passes verification.
+Scheduling backups without a guard would have made blocker B4 subtler rather
+than smaller.
+
+The guard is a session-scoped Postgres advisory lock on the database both
+operations touch — the same primitive `server/db/migrations.ts` already uses to
+serialize migrations, with a distinct key (`:openbrain-deploy`). It is
+deliberately asymmetric:
+
+- **`scripts/core01-deploy-local.sh` takes and HOLDS the lock** for the whole
+  deploy, acquired after the env file is sourced and before staging, migrate,
+  and the runtime swap. It waits (bounded) rather than refusing: a deploy is
+  attended, and queueing behind a running backup is correct.
+- **`scripts/backup.ts` REFUSES immediately** if the lock is held, exits `4`,
+  and writes nothing. It is an unattended job; skipping to the next clean run
+  beats holding a connection open against a database being migrated. The
+  refusal names the deploy as the cause.
+
+Because the lock is session-scoped, a deploy that is killed at any point
+releases it automatically. There is no stale-lock file and no cleanup path to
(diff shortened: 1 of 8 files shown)


rodaddy:
**CI: `check` is red on a known main-owned flake (#643), not on this branch's change.**

The single failure across 3796 tests is `ob-backfill sanitize > scans token-dense transcripts without quadratic blowup` — `Expected: < 1000`, `Received: 1112.08`. Evidence it is not branch-owned, gathered rather than asserted:

1. `git diff origin/main -- scripts/ob-backfill.test.ts scripts/ob-backfill.ts` is **empty**. Neither the test nor its subject is touched by this PR, and nothing in the diff references `ob-backfill` — this change is a launchd schedule, a deploy/backup advisory lock, and a doc section.
2. Clean `origin/main`, separate worktree, separate fresh database (the #609 differential standard): the file ran 3 consecutive times, **22 pass / 0 fail** every time.
3. The same test failed tonight on `fix/679-register-dev-safety-gates` at SHA `585a9772`, and **that identical SHA passed on its `push` run** — the two-runs-same-SHA flake signal.
4. The assertion is a bare `performance.now()` delta against a fixed 1000ms ceiling on a pure-CPU function (`scripts/ob-backfill.test.ts:251-260`), so it measures runner contention. Prior recorded observation on #643 was 1005ms; this one is 1112ms.

Evidence added to #643 as a comment. Not absorbed, not retried into green.

Every other check is green: `validate`, `db-integration`, `python-package`, `python-provider`, `python-capture`, `contract-parity`, GitGuardian, and the PR Body workflow.

**This PR is NOT merged and should not be merged by me** — the controller re-runs `scripts/done-means/677-scheduled-backup.sh` independently first, per the lane contract.

Issue #512, opened 2026-08-03
Codex adapter: the same Python hook structure for Codex/Claudex lanes


rodaddy:
## Destination

Codex/Claudex sessions carry the same Open Brain client structure proven on Claude: canon at session start, capture at lifecycle events, enforcement gates on tool use — from the same installed `python/openbrain` + `python/openbrain-provider` packages, against the **local dogfood brain**. State: PROPOSED.

## Scope

- Map Codex's hook/lifecycle surface to the wrapper-invoked console scripts (same commands now live in `~/.claude/settings.json` and the sol profile after the 2026-08-02 swap — the sol profile swap already covers proxied-Claudex sessions; this issue covers the native Codex runtime).
- Respect the existing Codex memory contract (`docs/memory-contract.md`) — the mcp2cli MCP route stays Codex's durable path for remote/uninstalled boxes; this adapter is additive on boxes where the Python stack is installed, not a replacement.
- Prove the same triple: canon injection, capture row, sane gate verdicts.

## Constraints

- **Dogfood-scoped.** No change for any consumer still on core01 until swap-back (operator rule, 2026-08-02).
- Sequenced after 0.9.0 soak and after the Pi lane proves the structure transfers across harnesses.

Refs: #443 (map), docs/memory-contract.md, PR #500 (fleet rollout plan), sibling Pi adapter issue.

rodaddy:
Sequencing decision, operator, 2026-08-04: the Codex adapter is to be done pretty soon if not next; then Pi, which is expected to be easy because it is the same shape and fully operator-controlled.

Recorded for ordering only — this is NOT yet a go-ahead to start.

rodaddy:
Status after PR #703 (squash 971ae13 on wip/2026-08-07, verify-lane VERIFIED, CI clean): **canon side is RUNNING on Codex** — live `codex exec` proof answered from injected canon, 7 Python-provider hooks in ~/.codex/hooks.json (.bak beside it), and the pre-existing Codex parser in bulk/formats.py fixed (time_to_first_token_ms optional; 27196 lines/430 turns/0 errors, was full-ingest abort on one line).

[...]

PR #703, opened 2026-08-09
fix(codex): make task_complete telemetry optional so Codex rollouts parse (#512)

rodaddy:
## Summary

Codex sessions could not carry the Python Open Brain provider structure, and the
reason turned out to be two separate things — one researched, one measured.

**The code defect.** A Codex transcript reader already existed for the bulk
ingester (`openbrain.apps.bulk.formats.codex_raw_turn_from_line`). It was broken
on the installed Codex CLI 0.147.0: `CodexTaskCompletePayload.time_to_first_token_ms`
was a required int, and Codex omits that field when a turn produced no tokens.
`ingest.py` re-raises rather than skipping, so ONE such line aborted an entire
ingest with a file:line.

Measured across every rollout written 2026-08-09: **144** `task_complete` records
carry the field, **1** does not — and that one came from `codex exec`, which is
exactly the non-interactive path automation and Workflow lanes use. Rare input,
total failure.

The field is telemetry; no turn content depends on it. `turn_id`,
`last_agent_message`, and the timestamps stay REQUIRED, so malformed records are
still quarantined loudly.

After the fix, over all of that day's real rollouts:

```
lines=27196 turns=430 errors=0     (before: aborted on the first tokenless turn)
```

**The research deliverable** (`docs/codex-adapter.md`) is the part #512 actually
asked for — what Codex's native surface can and cannot express:

- Codex 0.147.0 HAS a native hook engine (`features.hooks = true`,
  `~/.codex/hooks.json`), with Claude-shaped events and **identical snake_case
  payload fields**. `SessionStartHook`/`StopHook` are all-optional with
  `extra="ignore"`, so the canon read path needs no payload change.
- **Codex hash-pins hooks, and untrusted hooks fail silently.** A new or edited
  entry does not run, while Codex still prints `Completed` for the entries that
  did. Measured both ways: an untrusted probe and blocker never fired; with
  `--dangerously-bypass-hook-trust` the same blocker returned
  `UserPromptSubmit Blocked`. Corroborated by an unrelated trace hook registered
  2026-07-30 that has never written a single line.
- Gaps are documented rather than faked: no `Notification` event, no
  non-interactive trust grant, an unmeasured `additionalContext` ceiling, the
  capture hook not yet routed by transcript shape, and subagent capture
  unverified.

The `mcp2cli` route stays for remote/uninstalled boxes — #512 is additive, and
the done-means pins that so it cannot be dropped silently.

Refs #512

## Verification

- Done-means: scripts/done-means/512-codex-adapter.sh
- RED before the fix: clauses (a), (b), (d), (f) FAILED. Controls (c) and (e)
  PASSED in the same RED run, which is what proves the check discriminates
  rather than failing everything.
- PASS after: all six clauses.
- `uv run mypy src/openbrain` — Success, 50 source files.
- `uv run ruff check src tests` — All checks passed.
- `uv run pytest -q` — 610 passed, 1 skipped, **1 pre-existing failure**
  (`test_receipts_gate_crosslang.py::test_python_capture_receipt_clears_the_gates_capture_block`).
  Confirmed pre-existing by stashing this branch's changes and re-running on the
  untouched base, where it fails identically. NOT introduced here and NOT fixed
  here.

## Critical Self-Review

- Highest-risk behavior: relaxing a required field on a validating payload model.
  Contained by keeping `turn_id`, `last_agent_message`, and the timestamps
  required, and by done-means clause (c), which feeds a `task_complete` missing
  `turn_id` and asserts it STILL raises `MalformedCodexRecordError`.
- Assumptions that could be wrong: that `time_to_first_token_ms` is purely
  telemetry. Checked against the adapter — no turn content, role, or identity is
  derived from it. If a future consumer wants latency data it now has to handle
  `None`, which is the honest shape of the source.
- Missing/weak tests: the done-means covers the parser end-to-end, but capture is
  NOT wired to the Codex reader by this PR, so nothing here proves a Codex hook
  writes a row. That is stated as a gap in the doc rather than implied to work.
  There is no automated test for the hook trust behaviour — it is a property of
  the Codex binary, evidenced by measurement in the doc, not by CI.
- Security/permission risk: none introduced. No auth, namespace, or SQL path is
  touched; namespace stays token-derived server-side. The doc explicitly forbids
  hardcoding a host and points at `$OPENBRAIN_BASE_URL`/`$OPENBRAIN_TOKEN`.
- Migration/deploy risk: none. Making a field optional is backward compatible —
  every record that parsed before still parses, and the fixtures prove it.
- Downstream client/runtime risk: Codex is a downstream runtime, so
  `docs/downstream-rollout.md` applies. This PR changes no MCP tool, schema, or
  transport contract; it fixes a client-side parser and adds documentation. The
  `mcp2cli` route for Codex is unchanged and done-means clause (e) enforces that.
- Rollback/cleanup concern: revert is a single-field revert. The probe hooks used
  to measure trust behaviour were removed and `~/.codex/hooks.json` was restored
  from a timestamped backup; verified no probe references remain.
- Fixes made before PR: two. (1) The first done-means fixture invented a
  `user_message` shape without the attachment fields — measuring 391 real records
  showed all of them carry `images`/`local_images`/`audio`/`local_audio`/
  `text_elements`, so the FIXTURE was wrong, not the code, and it was corrected to
  the observed shape. (2) The check originally targeted
  `capture.records.raw_turn_from_line`; searching the repo first surfaced the
  existing bulk Codex adapter, so the gate was retargeted to reuse it instead of
  adding a second parser.
- Known residual risk: capture is not routed to the Codex reader by transcript
  shape yet, so a Codex session still does not write turns. The parser is now
  correct and proven over 27k real lines; the routing is follow-on work, named as
  gap 4 in the adapter doc.
- SME review-memory update: [x] not applicable because: the findings here are runtime surface facts and a schema-drift fix, captured in docs/codex-adapter.md where the next agent wiring Codex will read them; no review-lane pattern changed.

## Review Gate

- Live Open Brain checks: [x] not applicable because: this PR changes a transcript parser and adds documentation; it wires no hook and performs no Open Brain read or write. The live measurements that back it are Codex-side and are recorded in docs/codex-adapter.md.

### DIFF
--- a/docs/codex-adapter.md
+++ b/docs/codex-adapter.md
@@ -1,0 +1,180 @@
+# Codex Adapter — the direct Python Open Brain provider on Codex
+
+State: **WRITTEN**, with one RUNNING measurement set (below). The repo-side
+parser fix and this document are merged code and prose; the machine-side hook
+registration is described here but is NOT claimed as running on any box other
+than where its receipt says so.
+
+Issue: #512. Sibling of the Claude swap (2026-08-02) and the Pi adapter.
+
+## Why this exists
+
+Claude sessions on this box reach Open Brain through wrapper-invoked console
+scripts from the installed `openbrain` / `openbrain-provider` packages:
+
+```
+sh /Users/rico/.local/share/openbrain-memory/env/openbrain-hook-env <console-script>
+```
+
+Codex's route, per `docs/memory-contract.md`, is daemon-mode
+`mcp2cli open-brain`. That route **stays** — it is the correct answer for remote
+and uninstalled boxes, and #512 is explicitly additive. This document covers the
+native Codex runtime on a box where the Python stack IS installed.
+
+## Codex's real hook surface (measured, Codex CLI 0.147.0)
+
+Codex has a native hook engine. It is enabled by `features.hooks = true` in
+`~/.codex/config.toml` and configured in **`~/.codex/hooks.json`** — a separate
+file from `config.toml`, in Claude's `{"hooks": {EventName: [{matcher, hooks:
+[{type, command, timeout}]}]}}` shape.
+
+Event vocabulary, read from the binary's own strings (`hooks/src/events/*.rs`):
+
+| Codex event | Claude counterpart | OB use |
+| --- | --- | --- |
+| `SessionStart` | `SessionStart` | canon injection |
+| `UserPromptSubmit` | `UserPromptSubmit` | gates |
+| `PreToolUse` | `PreToolUse` | `ob-guard`, gates |
+| `PostToolUse` | `PostToolUse` | — |
+| `PermissionRequest` | (none) | — |
+| `PreCompact` / `PostCompact` | same | checkpoint |
+| `SessionEnd` | `SessionEnd` | wrap |
+| `SubagentStart` / `SubagentStop` | `SubagentStop` | subagent capture |
+| `Stop` | `Stop` | capture |
+
+Codex emits **snake_case payload fields identical to Claude's**. Captured live
+from a real `codex exec` run:
+
+```json
+{"session_id":"...","transcript_path":"/Users/rico/.codex/sessions/.../rollout-....jsonl",
+ "cwd":"...","hook_event_name":"SessionStart","model":"...",
+ "permission_mode":"bypassPermissions","source":"startup"}
+```
+
+`UserPromptSubmit` adds `turn_id` and `prompt`; `SessionEnd` adds `reason`.
+Because `SessionStartHook` and `StopHook` are all-optional with
+`extra="ignore"`, **the canon read path needs no payload change at all.**
+
+## THE TRUST GATE — read this before registering anything
+
+Codex **hash-pins every hook entry**. A new or edited entry is untrusted until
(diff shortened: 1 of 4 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/512-codex-adapter.sh exit=0 sha=4368e446d5fae1a7b4c484a7c1b36c2073f0388e at=2026-08-09T18:48:20.757Z

```
=== done-means #512 — Codex adapter (direct Python provider) ===
  (a) real rollout parses           : PASS — real-shaped rollout parsed to 2 turns with no exception
  (b) both roles, human attributed  : PASS — roles=[assistant,user] human_prompts=1
  (c) control, strictness preserved : PASS — a task_complete missing turn_id still raises MalformedCodexRecordError
  (d) control, silence is not a turn: PASS — 2 turns from 6 records — non-speech records correctly declined
  (e) mcp2cli route preserved       : PASS — memory-contract still documents the mcp2cli route (additive change; remote boxes unaffected)
  (f) trust gate documented         : PASS — docs/codex-adapter.md documents the hook trust gate and its bypass
DONE-MEANS 512: PASS
```

</details>

PR #713, opened 2026-08-10
fix(gates): pr-body-gate feeds the validator the PR's real tree and head ref (#709)

rodaddy:
## Summary

- Fixes #709, and closes the first of the two defects reported in #641. `.claude/hooks/pr-body-gate.ts` is the only caller of `scripts/validate-pr-body.ts` that runs at the boundary, and it never delivered the input the validator needs. #706 shipped a correct three-tier `Done-means` resolution and a hook that fed it neither the right tree nor the head ref, so half of #706 was unreachable in production.
- Two independent defects, both pointing the gate at the base branch. First, `input.cwd` is the HARNESS payload's cwd — the SESSION's directory. A lane opens its PR as `cd <worktree> && gh pr create` in one Bash call, and that `cd` does not move the session, so the "tree under review" was the primary checkout: the exact tree #706 existed to stop consulting. Second, `PR_HEAD_REF` was never set and `--head` was never parsed, so `git cat-file -e <ref>:<path>` — the tier #706 asked for by name — was dead code from the only live caller.
- The hook now resolves both in a fixed precedence and ANNOUNCES which source answered: `--head`/`-H` on the intercepted command, else the branch checked out in the directory the command `cd`s into, else the payload cwd. `PR_REPO_DIR` follows the same target directory, keeping the tree and ref tiers complementary rather than alternatives.
- Inputs are printed on the REFUSAL path too. #709 was diagnosed from a refusal that named the tree it searched but not where that tree came from, so it read as "your path does not exist" when the truth was "the gate looked in the wrong place".
- Harvest: `docs/lane-contract.md` round 29 plus one SME entry. Round 28's own first bullet — a seam added to make a gate testable is not the path that runs — recurred in the very next lane, and round 29 records why the paragraph did not prevent it.
- Found live and FILED, not absorbed: #711 (`core.hooksPath` absolute, so every worktree runs the primary checkout's hooks) and #712 (`bun test` aborts with `WriteFailed` when its stdout is git's pipe, failing pushes while the suite is green).

## Verification

- Done-means: scripts/done-means/pr-body-gate-fires.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

### Bootstrap ordering, declared — not routed around

The `Done-means` line above names `pr-body-gate-fires.sh`, NOT this lane's new `709-hook-feeds-head-ref.sh`, and that is a deliberate declaration rather than a convenience.

[...]

Issue #709, opened 2026-08-10
pr-body-gate never sets PR_HEAD_REF, so #706's branch-resolution tier is unreachable from the only caller that runs


rodaddy:
## What happens

The #706 fix landed in two halves and only one of them is reachable in
production.

`scripts/validate-pr-body.ts` resolves a `Done-means` path in three tiers —
the tree under review (`PR_REPO_DIR`), the validator's own tree, and
`git cat-file -e <PR_HEAD_REF>:<path>` against the head branch
(`validate-pr-body.ts:331-337`). The third tier is the one #706 asked for by
name: it is what lets a lane cite a check that exists only on its own branch.

`.claude/hooks/pr-body-gate.ts` — the only caller that runs at the boundary —
sets `PR_REPO_DIR` (`pr-body-gate.ts:356`) and **never sets `PR_HEAD_REF`**.
It also parses no `--head` flag from the `gh pr create` command line
(`rg -n 'head' .claude/hooks/pr-body-gate.ts` matches only prose in comments).
So the branch tier is unreachable from the hook, and the fix reduces to
"resolve against `input.cwd`".

That would still work if `input.cwd` were the lane worktree. It is not.
`input.cwd` is the harness payload's cwd — the SESSION's directory — and a
`cd /path/to/worktree && gh pr create ...` inside a single Bash call does not
change it. The hook's own comment states the intended source correctly ("the
tree the command was actually run from — the lane worktree"), but the payload
does not carry that.

## Measured 2026-08-10, lane `lane/issue-artifacts-outcomes` (rebased at `134b0b8`)

Creating the PR from the lane worktree, with the check committed and pushed on
the branch, is refused — and the refusal names the primary checkout as the only
tree it looked in:

```
BLOCKED by pr-body-gate: this PR body fails scripts/validate-pr-body.ts.
- Verification field 'Done-means' must name an existing repo-relative path;
  not found: scripts/done-means/issue-resolution-artifacts.sh
  (looked in: /Volumes/ThunderBolt/Development/open-brain)
```

The same body, same validator, same primary checkout, with `PR_HEAD_REF`
supplied by hand, passes — and announces the branch tier exactly as designed:

```
PR_BODY="$(cat prbody.md)" PR_TITLE="..." \
  PR_HEAD_REF="lane/issue-artifacts-outcomes" bun scripts/validate-pr-body.ts
# => Done-means resolved in branch lane/issue-artifacts-outcomes
[...]

Issue #704, opened 2026-08-09
test_receipts_gate_crosslang executes the out-of-repo _ob context-budget-gate.ts, which has drifted — fails locally, silently skipped in CI


rodaddy:
## What

`python/openbrain/tests/test_receipts_gate_crosslang.py::test_python_capture_receipt_clears_the_gates_capture_block` fails on the untouched base tree and on the primary checkout. It executes the **out-of-repo** `/Volumes/ThunderBolt/Development/_ob/scripts/context-budget-gate.ts`, whose behavior has drifted from what the test pins. CI never sees the failure because the `_ob` file is absent on runners, so the test skips — CI stays green while every local push is blocked by pre-push.

State per LAW 0: RUNNING (reproduced by the #512 lane, 2026-08-09, on base `4368e44`-era tree and primary checkout).

## Why it matters

- A test whose CI verdict and local verdict permanently disagree trains people to bypass it — the #512 lane had to use the test's own documented `OPENBRAIN_CONTEXT_BUDGET_GATE` override to push (no `--no-verify`).
- Silent-skip-in-CI + fail-locally is the exact fail-open shape B7/#679 is about, on the test side.

## Done-means

The cross-language receipt test either (a) pins a gate version/contract it vendors or resolves hermetically, or (b) skips loudly with a reason naming the missing/drifted dependency in BOTH environments — never green-in-CI/red-locally for the same tree. Regression check proves the old split verdict is impossible.

Found by the #512 lane (PR #703); reported not fixed, per lane scope.

Part of the test/CI hygiene cluster (#683 adjacent).

PR #718, opened 2026-08-10
fix(pr-body-gate): resolve a --head branch name to a ref the tree can read (#714)

### LINKED ISSUE #714: pr-body-gate never sets PR_HEAD_REF, so the validator's branch fallback is dead in the gh pr create path it was written for
## What happens

`.claude/hooks/pr-body-gate.ts` never sets `PR_HEAD_REF`, so
`scripts/validate-pr-body.ts`'s third resolution fallback — "the file may still
be committed on the branch being merged" — is **dead in the exact path it was
written for**.

The validator supports it (`scripts/validate-pr-body.ts:336`,
`headRef: process.env.PR_HEAD_REF?.trim() || undefined`), and its own comment
names the case: *"the `gh pr create` case the hook hits, where the lane has
pushed but no local checkout carries the file."* The hook sets `PR_BODY`,
`PR_TITLE`, and `PR_REPO_DIR` — but not `PR_HEAD_REF` (`pr-body-gate.ts:350-358`).

## Why the #709 fix does not cover this

#709 made the hook pass the payload's `cwd` as the tree under review, which
fixes the case where the agent's shell is IN the lane worktree. But the hook
reads `input.cwd` from the harness payload, which is the **session** cwd. An
agent whose tool layer resets cwd between calls (or which `cd`s inside the
command string) reports the PRIMARY checkout, sitting on the base branch, where
a lane's brand-new done-means check does not exist.

Measured on `lane/712-pre-push-writefailed`, body identical in all three runs:

```
# 1. from the lane worktree, direct  -> PASSES
$ PR_BODY=... bun scripts/validate-pr-body.ts
Done-means resolved in the tree under review:
  /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/lane-712-pre-push-writefailed/scripts/done-means/712-pre-push-pipe-safe.sh
PR body validation passed.

# 2. through the hook at `gh pr create` -> REFUSED
BLOCKED by pr-body-gate: this PR body fails scripts/validate-pr-body.ts.
- Verification field 'Done-means' must name an existing repo-relative path;
  not found: scripts/done-means/712-pre-push-pipe-safe.sh
  (looked in: /Volumes/ThunderBolt/Development/open-brain)

# 3. from the PRIMARY, with the head ref supplied -> PASSES
$ PR_HEAD_REF=origin/lane/712-pre-push-writefailed bun scripts/validate-pr-body.ts
Done-means resolved in branch origin/lane/712-pre-push-writefailed
  (/Volumes/ThunderBolt/Development/open-brain), not on disk in any available tree
PR body validation passed.
```

Note the refusal text ends `(looked in: <primary>)` with no `; and in ref ...`
clause — the validator's own message proves no ref was ever offered.

## Why it matters

This is the #706 failure with one more layer: a valid PR is refused because the
gate asked the wrong tree, and every cheap way past is a false receipt — name a
different pre-existing check that never judged this lane, or claim
`not applicable`. That is the reflex ledger item 17 exists to prevent, and the
gate teaching it is worse than the gate not existing.

It also lands hardest on exactly the lanes that are doing the right thing: only
a lane that wrote its OWN done-means check (the RLVR requirement) can hit it.

Have the hook derive the head ref and pass it, so the validator's existing
fallback is reachable:

- parse `--head` from the `gh pr create` argv when present;
- else `git -C <reviewRoot> symbolic-ref --quiet --short HEAD`;
- prefer the pushed `origin/<branch>` when it resolves, since the PR is created
  from the remote ref.

Keep the resolution ORDER unchanged (tree under review first, own tree, then
ref) — this only makes the third step reachable. The containment guard on
`existsInRef` is unchanged, so this widens WHERE a path may resolve, never WHAT
may be named.

A done-means clause should drive the REAL hook (the `gh pr create` shape) from a
cwd that is NOT the lane worktree, since that is the only shape that reproduces
it — `docs/lane-contract.md` round 28.

## State

WRITTEN — measured above on `lane/712-pre-push-writefailed` at `b3d1c3f`; no fix
in code. Found by the #712 lane, which was blocked from opening its own PR by
this and did NOT route around it (no `--no-verify`, no weakened body, no false
`not applicable`).

Related: #706 and #709 (the two prior instances of a gate judging from the wrong
tree), #711 (absolute `core.hooksPath`, same family).

rodaddy:
## Summary

- Closes #714 — but **not as filed**. The issue's premise is stale and the lane says so with a transcript instead of implementing it: PR #713 (issue #709) already made `pr-body-gate.ts` parse `--head`/`-H`, derive a branch from the `cd` target, and set `PR_HEAD_REF`. `scripts/done-means/709-hook-feeds-head-ref.sh` is 5/5 GREEN holding that, so "never sets PR_HEAD_REF" does not reproduce.
- Re-verifying the premise (canon: an issue older than the current code re-verifies before implementation) found a **narrower residue that does reproduce**: a branch NAME is not a ref until something can resolve it. `gh pr create --head <branch>` names a branch on the REMOTE, and run from a root checkout that has fetched but never created a local branch of that name — no `cd`, no worktree on disk — the bare name resolves to nothing. The tier #709 made reachable was being handed a dead string.
- `resolvableRef()` now resolves the branch name to a ref the tree being asked can actually read: the name **as given** first (so a correct `origin/lane/x`, a SHA, a tag, or a local-only checkout with no remote is never mangled), then `<remote>/<name>` for each configured remote with origin first. Each candidate is asserted on **positively** via `rev-parse --verify`, so a name that resolves nowhere skips the branch tier rather than feeding it garbage.
- Second-order fix: the old refusal printed `; and in ref lane/x` for a ref that resolves to nothing, which reads as *"your check is not on the branch either"* when the truth is *"that name means nothing here."* A refusal that misdirects is worse than one that admits it looked nowhere — the cheapest way past it is a false receipt, which is the exact reflex this gate exists to prevent. The head-ref line now has a third state naming the unresolvable name and what to do about it.
- `existsInRef`'s containment guard on the PATH is untouched. This widens WHERE a check may resolve, never WHAT may be named.

### Reproduction, against the unfixed hook at `17ada37`

Fixture is a genuine remote — bare `upstream.git`, a seed repo that pushes `lane/714-fixture` carrying the check, and a `primary` clone that has fetched the branch but holds no local branch of that name and no worktree:

```
[pr-body-gate] tree under review: the hook payload's cwd (.../primary)
[pr-body-gate] head ref: lane/714-fixture (source: explicit --head on the gh command)
- Verification field 'Done-means' must name an existing repo-relative path;
  not found: scripts/done-means/lane-only-check.sh
  (looked in: .../primary, /Volumes/ThunderBolt/Development/open-brain;
   and in ref lane/714-fixture)                                        EXIT=2
```

The head ref **is** supplied — that is what #709 fixed. The mechanism of the residue:

```
$ git -C primary cat-file -e lane/714-fixture:...lane-only-check.sh
fatal: invalid object name 'lane/714-fixture'.      exit=128
$ git -C primary cat-file -e origin/lane/714-fixture:...lane-only-check.sh
                                                    exit=0
```

## Verification

- Done-means: scripts/done-means/714-head-ref-resolves-remote.sh

6 clauses, every one driving the REAL hook with a synthetic PreToolUse payload against a REAL remote. RED 3/6 before the fix, GREEN 6/6 after:

```
CLAUSE 1 (a --head naming a REMOTE-ONLY branch resolves): FAIL -> PASS
CLAUSE 2 (an explicit origin/<branch> is not mangled): PASS (control)
CLAUSE 3 (MUTANT CONTROL: a path on no ref is still REFUSED): PASS (control)
CLAUSE 4 (MUTANT CONTROL: a --head naming nothing is not claimed as consulted): FAIL -> PASS
CLAUSE 5 (the ref the hook settled on is announced, never silent): FAIL -> PASS
CLAUSE 6 (REGRESSION: a local branch still resolves): PASS (control)
```

Clauses 2/3/6 being green before the fix is deliberate: they fail a BAD fix (blind `origin/` prepending, a blanket pass, a remote-dependent regression), not the defect.

**Round 28's false-green trap, called out in the check's header:** `709-hook-feeds-head-ref.sh` clause 3 already drives `--head` through the real hook and passes throughout — because its fixture holds the lane branch LOCALLY, so the bare name resolves. It proves `--head` is parsed, never that the parsed value is resolvable. The defect is invisible in any fixture where the lane branch also exists locally, which is why this check builds a genuine bare remote instead of faking the ref layout.

Regression surface, all GREEN after the fix: `709-hook-feeds-head-ref.sh` (5/5), `pr-body-gate-fires.sh` (7/7), `706-done-means-resolves-pr-head.sh` (5/5), `pr-template-passes-validator.sh` (3/3), `done-means-field-required.sh` (5/5), `bunx tsc --noEmit` (0).

## Critical Self-Review

- Highest-risk behavior: `resolvableRef()` choosing a DIFFERENT ref than the caller named. Mitigated by ordering — the name as given is always tried first, so a remote-prefixed ref, a SHA, or a local branch wins before any prefixing is attempted, and clause 2 fails the double-prefix mistake.
- Assumptions that could be wrong: that a stale remote-tracking ref cannot silently answer for a newer local branch. If both resolve, the as-given (local) name wins by ordering, so the stale ref is only consulted when nothing local exists — but an `origin/<branch>` that is behind the pushed branch could in principle resolve a path the newest commit deleted. That is an ACCEPT, not a proof: the tier only ever adds a resolution for a path that exists somewhere real, and the mutant controls keep it from becoming a blanket pass.
- Missing/weak tests: no clause covers two remotes both carrying the branch name (origin-first ordering is asserted only by construction, not driven). No clause drives a detached-HEAD checkout with an explicit `--head`.
- Security/permission risk: none new. The widening is on WHERE a ref is read; `existsInRef`'s absolute/`..` containment guard on the PATH is unchanged, so no path outside the repo becomes nameable. All added git calls are read-only (`rev-parse`, `remote`) via `spawnSync` with an argv array — no shell, no user string interpolated into a command line.
- Migration/deploy risk: none. Repo-local hook only, no schema, no service, no config.
- Downstream client/runtime risk: none. `docs/downstream-rollout.md` classified: this touches no MCP tool, schema, transport, or Python client surface — it is a `.claude/` PreToolUse hook in this repo. Not applicable.
- Rollback/cleanup concern: revert the single commit; the hook returns to #709 behavior and the branch tier goes back to being fed an unresolvable name. The check's fixture is archived with `mv` by its own teardown, never deleted.
- Fixes made before PR: the "no head ref" announcement initially reported an unresolvable name as "not on a named branch", which sends the reader after the wrong problem; it now has a distinct third state naming the unresolvable value and the remedy.
- Known residual risk: the two untested shapes above (multi-remote, detached HEAD + `--head`). Neither is on the path #714 reports; both fail CLOSED (no ref resolved -> tier skipped -> refusal that says so) rather than open.

Per the RLVR harvest model the lane returns lessons to the controller rather than writing `docs/sme/` or `docs/lane-contract.md` itself; the tracking-scribe owns those files at root. Findings returned: the resolvable-ref family (a gate fed a value it cannot resolve is the same dead tier by another route), and the fixture-locality trap (a check whose fixture makes the defect impossible passes forever).

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: this changes a repo-local PreToolUse hook only; no Open Brain service, tool, schema, or namespace path is touched.

## Contract Parity

- Contract parity: [x] runtime-specific because: the change is a Claude Code hook in `.claude/hooks/`, which has no cross-runtime contract fixture — no client/contract path in `contracts/parity-paths.txt` is touched.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool, schema, transport, or Python client surface changes. The diff is one `.claude/` hook, one new done-means check, one worklog.
- Pre-push blocker declared: `test_receipts_gate_crosslang.py::test_python_capture_receipt_clears_the_gates_capture_block` fails on this lane AND identically on the untouched wip head `17ada37` (proven in a detached probe worktree carrying none of this lane's files). That is **#704**, already filed, not this lane's. Pushed using the test's own documented `OPENBRAIN_CONTEXT_BUDGET_GATE` override — the route #704 records the #512 lane using. `--no-verify` was NOT used; the rest of the suite passed (`pre-push: all checks passed ✓`).

### DIFF
--- a/.claude/hooks/pr-body-gate.ts
+++ b/.claude/hooks/pr-body-gate.ts
@@ -309,6 +309,78 @@ function branchOf(dir: string): string | null {
   return branch ? branch : null;
 }
 
+/** Does `ref` name something `git` can resolve inside `dir`? */
+function refResolves(dir: string, ref: string): boolean {
+  const probe = spawnSync(
+    "git",
+    ["-C", dir, "rev-parse", "--verify", "--quiet", `${ref}^{commit}`],
+    { encoding: "utf8" },
+  );
+  return probe.status === 0;
+}
+
+/**
+ * The ref that actually CONTAINS the PR's commits, given a branch NAME.
+ *
+ * ISSUE #714. A branch name is not a ref until some repository can resolve it,
+ * and `gh pr create --head <branch>` names a branch on the REMOTE. Run from a
+ * root checkout that fetched but never created a local branch of that name —
+ * which is exactly where a lane opens a PR with `--head` and no `cd` — the bare
+ * name resolves to nothing:
+ *
+ *   $ git cat-file -e lane/x:scripts/done-means/check.sh
+ *   fatal: invalid object name 'lane/x'.
+ *   $ git cat-file -e origin/lane/x:scripts/done-means/check.sh   # exit 0
+ *
+ * #709 made the hook FEED the branch tier; this makes what it feeds RESOLVABLE.
+ * Without it the tier is fed a dead string and the refusal prints
+ * `; and in ref lane/x`, which reads as "your check is not on the branch" when
+ * the truth is "that name means nothing here" — a refusal that misdirects is
+ * worse than one that admits it looked nowhere, because the cheapest way past a
+ * misdirecting refusal is a false receipt (#706's lesson, one layer out).
+ *
+ * Candidates are tried in order and each is ASSERTED ON POSITIVELY — round 28:
+ * a lookup whose failure is indistinguishable from its legitimate empty case
+ * must be asserted on positively, never by observing that nothing broke. So a
+ * name that resolves nowhere returns null and the branch tier is skipped
+ * entirely rather than fed a string that will fail for the wrong reason.
+ *
+ *   1. the name as given — a local branch, a SHA, a tag, or a ref the caller
+ *      already spelled correctly (`--head origin/lane/x`). Trying this FIRST is
+ *      what stops a correct ref being mangled, and keeps a local-only checkout
+ *      working with no remote at all.
+ *   2. `<remote>/<name>` for each configured remote, origin first. This is the
+ *      #714 case. Remotes are ENUMERATED rather than assuming `origin` exists,
+ *      because a checkout with a differently-named remote is not an error.
+ *
+ * This only ever changes WHICH ref is consulted; `existsInRef`'s containment
+ * guard on the PATH is untouched, so it widens where a check may live and never
+ * what may be named.
+ */
+function resolvableRef(dir: string, name: string): string | null {
+  if (refResolves(dir, name)) return name;
+
+  const remotesProbe = spawnSync("git", ["-C", dir, "remote"], {
+    encoding: "utf8",
+  });
+  if (remotesProbe.status !== 0) return null;
+  const remotes = (remotesProbe.stdout ?? "")
+    .split("\n")
+    .map((line) => line.trim())
+    .filter(Boolean)
+    .sort((a, b) => Number(b === "origin") - Number(a === "origin"));
+
+  for (const remote of remotes) {
+    // A name that already carries this remote's prefix was covered by the
+    // as-given attempt above; re-prefixing it would only build origin/origin/x.
+    if (name.startsWith(`${remote}/`)) continue;
+    const candidate = `${remote}/${name}`;
+    if (refResolves(dir, candidate)) return candidate;
+  }
+
+  return null;
+}
+
 function refuse(lines: string[]): never {
   console.error(lines.join("\n"));
   process.exit(2);
@@ -477,6 +549,20 @@ if (!existsSync(VALIDATOR)) {
  * checkout this process can see. Every step degrades to the pre-#709 behaviour
  * rather than guessing, and `scripts/done-means/709-hook-feeds-head-ref.sh`
  * drives THIS file with real payloads to hold it there.
+ *
+ * ---------------------------------------------------------------------------
+ * ISSUE #714 — a branch NAME is not a ref until something can resolve it.
+ * ---------------------------------------------------------------------------
+ *
+ * The precedence above produces a branch NAME. `gh pr create --head <branch>`
+ * names a branch on the REMOTE, and in a root checkout that fetched it but
+ * never created a local branch of that name, the bare name resolves to nothing
+ * — so the tier #709 made reachable was still handed a dead string. The name is
+ * therefore run through `resolvableRef()` (as-given, then `<remote>/<name>`),
+ * each candidate asserted on positively, and the SETTLED ref is what reaches
+ * `PR_HEAD_REF`. `scripts/done-means/714-head-ref-resolves-remote.sh` holds it
+ * with a genuine bare-remote fixture, because the defect is invisible in any
+ * fixture where the lane branch also exists locally.
  */
 const payloadCwd = input.cwd?.trim() || process.cwd();
 const cdTarget = cdTargetBefore(commands, targetIndex, payloadCwd);
@@ -486,18 +572,20 @@ const reviewRootSource = cdTarget
   : `the hook payload's cwd (${payloadCwd})`;
 const reviewRoot = cdTarget ?? payloadCwd;
 
-let headRef: string | null = null;
-let headRefSource = "";
-if (target.head?.trim()) {
-  headRef = target.head.trim();
-  headRefSource = `explicit --head on the gh command`;
-} else {
-  const derived = branchOf(reviewRoot);
-  if (derived) {
-    headRef = derived;
-    headRefSource = `branch checked out in ${reviewRoot}`;
-  }
-}
+// The branch NAME the PR is opened from, and where that name came from.
+const headName = target.head?.trim() || branchOf(reviewRoot);
+const headNameSource = target.head?.trim()
+  ? `explicit --head on the gh command`
+  : `branch checked out in ${reviewRoot}`;
+
+// ...and the ref that can actually be READ in the tree being asked (issue
+// #714). A name that resolves nowhere yields null, so the branch tier is
+// skipped rather than fed a string that fails for a reason nobody can act on.
+const headRef = headName ? resolvableRef(reviewRoot, headName) : null;
+const headRefSource =
+  headRef && headRef !== headName
+    ? `${headNameSource}; '${headName}' does not resolve in ${reviewRoot}, so the remote-tracking ref was used`
+    : headNameSource;
 
 const result = spawnSync("bun", [VALIDATOR], {
   encoding: "utf8",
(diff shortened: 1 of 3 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/714-head-ref-resolves-remote.sh exit=0 sha=6f9603f11acd15cf5fbee74354b0d205b19b352f at=2026-08-10T14:10:15.505Z

```
CLAUSE 1 (a --head naming a REMOTE-ONLY branch resolves): PASS — --head naming a remote-only branch resolved via the branch tier and announced it (exit=0)
CLAUSE 2 (an explicit origin/<branch> is not mangled): PASS — an explicit origin/<branch> resolves untouched, no double-prefixing (exit=0)
CLAUSE 3 (MUTANT CONTROL: a path on no ref is still REFUSED): PASS — nonexistent Done-means path still REFUSED, naming the path (exit=2)
CLAUSE 4 (MUTANT CONTROL: a --head naming nothing is not claimed as consulted): PASS — a --head naming no existing ref is REFUSED without claiming to have consulted it (exit=2)
CLAUSE 5 (the ref the hook settled on is announced, never silent): PASS — the hook announces the ref it settled on, naming both what --head said and what resolved
CLAUSE 6 (REGRESSION: a local branch still resolves): PASS — a checkout carrying the branch locally still resolves (exit=0), no remote dependency introduced
```

</details>

rodaddy:
harvested: returned to controller for tracking-scribe (RLVR harvest model, docs/sop-rlvr-lanes.md step 5 — the lane does not write docs/lane-contract.md, docs/sme/, or the ledger at root).

Lessons returned:

1. RESOLVABLE-REF FAMILY. #709 made the hook FEED the branch tier; this lane found the fed value was not resolvable in the tree being asked. A gate handed a value it cannot resolve is the same dead tier by another route — 'is it supplied?' and 'does it resolve?' are two different questions and only the second one is the gate working. Sibling of round 28's positive-assertion rule: assert the ref resolves, never infer it from the absence of an error.

2. A REFUSAL THAT MISDIRECTS IS WORSE THAN ONE THAT ADMITS IT LOOKED NOWHERE. The old message printed '; and in ref lane/x' for a ref resolving to nothing, reading as 'your check is not on the branch' when the truth was 'that name means nothing here'. The cheapest way past a misdirecting refusal is a false receipt — the exact reflex ledger item 17 exists to prevent.

3. FIXTURE-LOCALITY TRAP (new spelling of round 28's false-green family). 709-hook-feeds-head-ref.sh clause 3 drives --head through the REAL hook and passed throughout, because its fixture holds the lane branch LOCALLY so the bare name resolves. Round 28 says the invocation shape must be real; this says the fixture's ENVIRONMENT must also be able to express the defect. A check whose fixture makes the defect impossible passes forever.

4. STALE-PREMISE RE-VERIFICATION PAYS EVEN WHEN IT DOES NOT CLOSE THE ISSUE. #714 as filed did not reproduce; re-verifying rather than implementing found a real, narrower residue. Implementing the filed premise would have added a second PR_HEAD_REF assignment next to the working one.

Issue #719, opened 2026-08-10
pre-commit gate is silently absent: repo core.hooksPath displaces the global gitleaks + protected-branch hook, and nothing runs in its place


rodaddy:
## Family classification

**Family B** — a gate letting its ENVIRONMENT decide the verdict instead of the code.
Sibling of #705 (inherited env) and #712 (pipe writability). Root enabler #711 fixed
`core.hooksPath` to the relative `_githooks`, but nothing asserts that `_githooks/`
CONTAINS the hooks the value displaced.

## Defect

`core.hooksPath = _githooks` (repo-local) overrides the operator's global
`core.hooksPath = /Users/rico/.config/git/hooks`. Git consults **exactly one**
hooks directory — it does not chain. `_githooks/` ships only `pre-push` and
`install.sh`, so this repo has **no pre-commit hook at all**.

The global hook that is being displaced performs two controls:

- `gitleaks git --pre-commit --staged --redact=100` — local secret scanning
- LAW #8 — blocks commits on `main`/`master`/`develop`/`production`/`staging`

Both are **silently OFF for every commit in this repo**. AGENTS.md requires "No
secrets in git, logs, reports, fixtures, PRs, issues, or screenshots"; the
mechanism enforcing that requirement is not running.

There is a `.git/hooks/pre-commit` shim whose own docstring says it exists to
restore "the user's global gitleaks pre-commit, which this repo's local
core.hooksPath override was silently bypassing". It is **unreachable**: git
ignores `.git/hooks` entirely when `core.hooksPath` is set. The repair was
written, and it has never once run.

## Reproduction

```
$ git config --get core.hooksPath
_githooks

$ git config --global --get core.hooksPath
/Users/rico/.config/git/hooks

$ ls _githooks/
install.sh
pre-push

# What git ACTUALLY runs for pre-commit in this repo:
$ git hook run pre-commit
error: cannot find a hook named pre-commit

# POSITIVE CONTROL — the same command, same tree, global path. The scan is real
# and it works; only the path is wrong:
$ git -c core.hooksPath=/Users/rico/.config/git/hooks hook run pre-commit
INF 0 commits scanned.
INF scanned ~0 bytes (0) in 20.6ms
INF no leaks found
```

The control is the whole finding: gitleaks is installed and functional, and this
repo's config is what stops it.

## Why it survived

#711 asserted the hooksPath VALUE (`_githooks`, matching the installer) and
[...]

Issue #711, opened 2026-08-10
core.hooksPath is absolute in .git/config, so every lane worktree runs the primary checkout's hooks


rodaddy:
## What happens

`core.hooksPath` in this clone's `.git/config` is the ABSOLUTE path
`/Volumes/ThunderBolt/Development/open-brain/_githooks`. Because
`.git/config` is shared by every linked worktree, a lane worktree under
`{temp_workspace}/open-brain/_worktrees/` runs the PRIMARY CHECKOUT's copy of
`_githooks/pre-push` — the copy sitting on the base branch, not the one the
lane is changing.

Measured this session (lane for #709):

```
$ git config --show-origin --get core.hooksPath
file:.git/config	/Volumes/ThunderBolt/Development/open-brain/_githooks

$ cd <lane worktree> && git config --get core.hooksPath
/Volumes/ThunderBolt/Development/open-brain/_githooks
```

## Why this is a defect and not a preference

`_githooks/install.sh` already decided the opposite, in a comment written
against this exact trap:

```
# Resolve to a repo-root-relative value, not an absolute one: an absolute path
# is a clone-specific string that would be wrong in every other checkout, and
# core.hooksPath is per-repo config that only this clone reads anyway.
```

It sets `core.hooksPath=_githooks`. Git resolves a relative `core.hooksPath`
against the worktree root, so the installer's value gives each worktree its
OWN hooks. The live config does not match what the installer writes, so the
divergence happened after (or instead of) an install run. Nothing detects it:
the installer is idempotent and reports "already set" only for an exact
match, and no check asserts the value is relative.

## Consequence

`docs/lane-contract.md` round 28 named this as the third instance of one
family — a gate resolving its base or its tree from something other than the
change under review (the other two were #706 and #705). The specific bite:

- a lane fixing a git hook structurally **cannot exercise its own fix** on
  push; it exercises the base branch's copy.
- a lane whose pre-push is broken ON THE BASE BRANCH pays for it even when its
  own branch fixes it, and vice versa: a lane can break `_githooks/pre-push`
  and every local push still passes.
- proving a hook fix live requires `-c core.hooksPath=<lane worktree>/_githooks`
  by hand (what PR #708 had to do), which is a workaround a lane must know to
  reach for.

## Suggested direction (not implemented here)

1. Set `core.hooksPath` to the relative `_githooks` — i.e. what
   `_githooks/install.sh` already writes — so each worktree resolves its own.
[...]

PR #727, opened 2026-08-18
fix(hooks): restore the gitleaks + LAW #8 controls core.hooksPath displaced (#719)

rodaddy:
Refs #719

`core.hooksPath = _githooks` REPLACES the operator's global hooks directory. Git
consults exactly one directory; it never chains and never overlays. `_githooks/`
shipped only `pre-push` and `install.sh`, so this repo had NO pre-commit hook at
all, and the two controls the displaced global hook carried were silently OFF
for every commit made here: the gitleaks staged secret scan, and the LAW #8
protected-branch commit block.

Sweeping the WHOLE displaced set — not just the hook the issue named — found a
SECOND live gap: the global `pre-push` also carried a gitleaks range scan and a
default-branch push guard, and this repo's own `pre-push` replaced it carrying
neither. Secret scanning was absent from BOTH ends of the commit path.

Changes:

- `_githooks/pre-commit` (new): LAW #8 branch block, then a gitleaks
  `--staged --redact=100` scan. Fails CLOSED when gitleaks is absent, per
  `_DOCS/STANDARDS-git.md`. Announces every stage, because #719's failure mode is
  that an ABSENT gate and a PASSED gate look identical.
- `_githooks/pre-push`: restores the displaced range scan and default-branch
  guard (keeping the global hook's `ALLOW_DEFAULT_PUSH` escape hatch), placed
  before typecheck and the suite; and extends the #711 drift assertion so
  "hooksPath value correct, directory INCOMPLETE" warns and names the missing
  hook. It warns rather than refuses: refusing a push over a defect the push
  cannot cause is what trains `--no-verify`.
- `_githooks/displaced-hooks-allowlist.txt` (new): dropping a displaced control
  becomes a written decision with a reason.
- `scripts/done-means/719-hooks-dir-complete.sh` (new): 10 clauses driving REAL
  `git commit` and `git push` in fixtures configured exactly like this repo,
  with a fake canary (Stripe's published test value; never a real credential)
  and a paired control for every refusal.

The controls are reimplemented in tracked files rather than delegated to the
global hook: that path is absolute and machine-specific (#711's root enabler),
untracked, and absent in CI and every fresh clone, so a delegating hook would
degrade to no gate exactly where oversight is thinnest.

## Verification

- Done-means: scripts/done-means/719-hooks-dir-complete.sh
- Result: RED on the pre-fix tree (`8fd5630`, the rebase parent) with 9 of 10
  clauses FAIL and a nonzero exit; GREEN 10/10 on this branch. Both runs
  observed this session (RUNNING).
- Red excerpt (pre-fix tree): `CLAUSE b ... FAIL — git commit of a staged fake
  credential SUCCEEDED — the secret scan is not on the path git runs`;
  `CLAUSE d ... FAIL — a commit on 'main' SUCCEEDED; the protected-branch block
  does not run`; `RESULT: FAIL`.
- Green excerpt (this branch): `CLAUSE b ... PASS — git commit of a staged fake
  credential was REFUSED (exit=1) and no commit object was created`;
  `RESULT: PASS (10 clause(s) judged)`.
- Regression: `scripts/done-means/711-hookspath-relative.sh` — clauses a-d PASS,
  clause e ENV-SKIP (no local `core.hooksPath` in this throwaway clone).
- `bunx tsc --noEmit` clean after `bun install --frozen-lockfile`.
- Live receipt: pushing this branch ran the restored `pre-push` gitleaks range
  scan and printed `INF no leaks found` (RUNNING).
- Six mutants were applied and reverted one at a time in the originating lane
  (recorded in `_plans/worklog/fix-719-2026-08-10.md`); the sharpest, M2, leaves
  the hook printing a green tick while scanning nothing, and only clause (b)
  catches it. Those mutant runs are MERGED-journal evidence from that session,
  not re-executed here.
- No TypeScript, Python, or runtime code changed, so the bun suite and the
  Python package gates are unaffected by this diff.

## Critical Self-Review

- Highest-risk behavior: `_githooks/pre-commit` now refuses commits. A false
  positive from gitleaks, or a bug in the branch block, blocks all local work in
  this repo; clauses (c) and (e) are the paired controls proving the refusals are
  specific and that clean commits on feature branches still succeed.
- Assumptions that could be wrong: that the operator's displaced hook set is
  exactly `pre-commit` and `pre-push`. Clause (a) computes the displaced set from
  the global directory at run time instead of a hardcoded list, so a hook added
  there later goes RED rather than being silently dropped — but on a machine with
  no global `core.hooksPath` that clause has nothing to compare against.
- Missing/weak tests: the check drives real git in fixtures, but it cannot prove
  the hook behaves identically under every gitleaks version; it probes the canary
  at startup and HARNESS-ERRORs if the installed scanner does not flag it, which
  converts that ambiguity into a stop rather than a verdict. Nothing tests the
  `ALLOW_DEFAULT_PUSH` escape hatch.
- Security/permission risk: this restores a secret-scanning control that was off;
  the direction is strictly toward more scanning. The canary is Stripe's
  published test value, never a real credential. The hook fails CLOSED when
  gitleaks is missing, which is the correct failure direction and was proven by
  mutant M5.
- Migration/deploy risk: none to the service. Effective only in clones that have
  run `./_githooks/install.sh`; a clone that has not is in the same state as
  before this PR, not a worse one.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python
  client surface changes, so `docs/downstream-rollout.md` does not apply.
- Rollback/cleanup concern: revert the commit and re-run `install.sh`; nothing
  persists outside `_githooks/` and `scripts/done-means/`. Fixtures are created
  under the temp workspace and are self-contained.
- Fixes made before PR: four fixture defects were found and fixed in the
  originating lane, each of which had produced a RED that was NOT the subject
  failing — an undetectable canary, a false PASS on clause (f) where the fixture
  died before the scan, untracked scaffolding, and a defective control fixture
  for clause (j). Detail in the lane journal.
- Known residual risk: the hook's cost is added to every commit here, and the
  push-side scan is added ahead of typecheck. This restores what the override
  displaced; it is deliberately NOT the full seven-check gate
  `docs/CI_CD_REQUIREMENTS.md` specifies as the eventual required state.
- SME review-memory update: [ ] `docs/sme/` updated [x] not applicable because: this lane produced no new MEDIUM+ review finding — the defect and its lessons are recorded in `_plans/worklog/fix-719-2026-08-10.md`, and the enforcement itself is the artifact.

## Review Gate

- Live Open Brain checks: [ ] linked below [x] not applicable because: this change touches only git hooks and a done-means shell check, with no Open Brain service, schema, or memory-path surface involved.

### DIFF
--- a/_githooks/displaced-hooks-allowlist.txt
+++ b/_githooks/displaced-hooks-allowlist.txt
@@ -1,0 +1,24 @@
+# Displaced-hook allowlist (issue #719)
+#
+# `core.hooksPath = _githooks` REPLACES whatever hooks directory was configured
+# before it. Git consults exactly one directory; it never chains and never
+# overlays. So every hook in the displaced directory that `_githooks/` does not
+# provide is a control that SILENTLY STOPPED RUNNING in this repo — which is
+# what #719 was: the displaced `pre-commit` carried gitleaks secret scanning and
+# the LAW #8 protected-branch block, and neither ran here for as long as the
+# override existed.
+#
+# `scripts/done-means/719-hooks-dir-complete.sh` clause (a) therefore reads the
+# displaced directory and requires each executable hook in it to be either
+# provided by `_githooks/` or listed HERE with a reason. The point of the file
+# is that dropping a control becomes a DECISION SOMEONE WROTE DOWN rather than
+# a side effect nobody noticed.
+#
+# Format, one per line:   <hook-name>: <reason it is deliberately not provided>
+#
+# A reason must say why the control is not needed HERE, not merely that nobody
+# has written it yet. "Not implemented" is a gap to file, not an allowlist
+# entry — put the issue number in and the next reader can find its status.
+#
+# Currently empty: both hooks in the displaced directory (`pre-commit`,
+# `pre-push`) are provided by `_githooks/`, so nothing is dropped.
--- a/_githooks/install.sh
+++ b/_githooks/install.sh
@@ -63,6 +63,11 @@ printf '\nHooks git will now run from %s/:\n' "$target"
 for hook in "$repo_root/$target"/*; do
   name="${hook##*/}"
   [[ "$name" == "install.sh" ]] && continue
+  # Not hooks: the installer itself, and the #719 allowlist that records which
+  # displaced hooks are deliberately not provided. Skipped explicitly, because
+  # the "NOT executable" branch below would otherwise report a data file as a
+  # broken hook and send the reader chmod'ing something git never runs.
+  [[ "$name" == "displaced-hooks-allowlist.txt" ]] && continue
   if [[ -x "$hook" ]]; then
     printf '  %s\n' "$name"
   else
--- a/_githooks/pre-commit
+++ b/_githooks/pre-commit
@@ -1,0 +1,130 @@
+#!/usr/bin/env bash
+# Open Brain pre-commit gate (issue #719).
+#
+# ---------------------------------------------------------------------------
+# WHY THIS FILE EXISTS AT ALL
+# ---------------------------------------------------------------------------
+# `core.hooksPath` selects EXACTLY ONE hooks directory. Git does not chain it
+# and does not overlay it: the value REPLACES whatever was configured before.
+# This repo sets it to `_githooks` (per `_githooks/install.sh`), which displaces
+# the operator's global `~/.config/git/hooks`. That global directory carried a
+# pre-commit hook enforcing two controls, and `_githooks/` shipped no
+# pre-commit at all — so for every commit made in this repo, both were OFF:
+#
+#   1. `gitleaks` staged secret scanning
+#   2. LAW #8 — no commits on a protected branch
+#
+# Measured on the pre-fix tree (issue #719, reproduced again in this lane's
+# worktree before this file was written):
+#
+#     $ git hook run pre-commit
+#     error: cannot find a hook named pre-commit
+#
+#     # positive control — the SAME command, the SAME tree, the displaced path:
+#     $ git -c core.hooksPath=~/.config/git/hooks hook run pre-commit
+#     INF no leaks found
+#
+# The control is the finding: gitleaks was installed and working the whole
+# time, and this repo's own config was the only thing stopping it. AGENTS.md
+# requires "No secrets in git, logs, reports, fixtures, PRs, issues, or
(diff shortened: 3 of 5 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/719-hooks-dir-complete.sh exit=0 sha=5ef61720db77e1a5c54cdd1ccfc59402ac736fa9 at=2026-08-18T02:09:35.946Z

```
CLAUSE a (every displaced hook is provided by _githooks/ or allowlisted): PASS — displaced set from /Users/rico/.config/git/hooks fully accounted for — provided: pre-commit pre-push allowlisted: none
CLAUSE b (REAL git commit REFUSES a staged fake credential): PASS — git commit of a staged fake credential was REFUSED (exit=1) and no commit object was created
CLAUSE c (control: a clean commit still succeeds (not blanket-refusing)): PASS — control — a clean commit on a feature branch SUCCEEDED (exit=0, 1 commit), so the refusal in (b) is specific and not blanket
CLAUSE d (LAW #8: a commit on main is REFUSED, naming the branch): PASS — LAW #8 — a clean commit on 'main' was REFUSED (exit=1), no commit object created, and the refusal names the branch
CLAUSE e (control: the same commit on a feature branch succeeds): PASS — control — the SAME clean commit on 'feat/allowed' SUCCEEDED, so (d) refused the branch and not the content
CLAUSE f (REAL git push REFUSES a commit containing a fake credential): PASS — git push of a commit containing the fake credential was REFUSED (exit=1), nothing reached the remote, and the refusal attributes itself to the secret scan
CLAUSE g (control: a clean push still succeeds): PASS — control — a clean push SUCCEEDED (exit=0, 1 commit on the remote), so (f) refused the secret and not every push
CLAUSE h (the hook ANNOUNCES each control it ran): PASS — the pre-commit run names the hook and the secret-scan stage, so an absent gate cannot be mistaken for a passed one
CLAUSE i (the #711 drift assertion also catches an INCOMPLETE directory): PASS — with a displaced hook missing from _githooks/, the pre-push drift assertion WARNS and names the missing hook — the state #711 could not see
CLAUSE j (control: that assertion is silent when the directory is complete): PASS — control — with the displaced hook provided, the drift assertion is silent, so (i) reflects the directory's contents and not an unconditional warning

RESULT: PASS (10 clause(s) judged)
```

</details>

rodaddy:

Issue #724, opened 2026-08-17
P0 next session: nats-worker vanished Aug 14 -- 3 days of false-receipt captures; interim repair RUNNING, backlog re-embed + silent-gap alarms needed


rodaddy:
**FIRST TASK, NEXT OPEN-BRAIN SESSION (Rico ruling 2026-08-17).** The interim repair is RUNNING but three residuals need this repo's owner: verify the backlog re-embed, decide the worker's supervision hardening, and find what removed the plist.

## What happened (verified on the mini, 2026-08-17)

Memory recall froze at **Aug 14** while every capture kept receipting `status: saved` — a false-receipt failure that survived three days undetected.

1. **Root cause:** `~/Library/LaunchAgents/com.rico.open-brain-local-nats-worker.plist` — the third service of the "Local Mac Dogfood NATS Lane" (`docs/core01-nats-worker-runbook.md`, stood up 2026-08-04) — **vanished on 2026-08-14 at 10:53** (its logs were zeroed at that timestamp and never written again; `mdfind` finds no copy of the plist anywhere). That worker also runs the maintenance pipeline (embedding/promotion), so new rows stored but never became retrievable.
2. **Compounding event:** Aug 16 ~15:00 local, the Postgres 18 relocation bounced the DB — `57P03` storm in `autostart.err.log`, then a bare `Local clone embedding preflight failed` line. The whole stack (nats-server, mlx embedding server on 8791, local-clone launcher) restarted at ~15:56 — **without** the worker, which had no plist to restart from.
3. **Why nothing alarmed:** `3100/health` reported healthy the entire time — correctly per the runbook ("What /health does and does NOT show"): the HTTP process cannot observe the separate worker. `capture` showed "delivering" because raw turns were in fact landing. The maintenance **producer** ticked happily with no consumer.
4. **Collateral symptom that led to discovery:** two Development sessions' manual `openbrain-memory` wrap calls failed with `Existing lane exact scope does not match session_start request`. Separately, receipts show contract skew: real sessions on `2026-07-23.memory-tools.v23`, plus 06:56/07:18Z entries on `2026-08-09.memory-tools.v24` with placeholder identity (`opaque-session_abc123`).

## Interim repair (done 2026-08-17, RUNNING with receipts)

[...]

Issue #613, opened 2026-08-07
Test suites leave parity-source-registry-* fixtures behind; any global producer makes them live


rodaddy:
## Problem

Several suites leave \`parity-source-registry-*\` rows in \`ob_sources\` (approved/active) without cleanup. On main these sat inert because nothing produced from them. PR #609's maintenance sweep is a cross-namespace producer by design (\`selectSourcesNeedingDerivation\` with \`undefined\` writable namespaces), so those leftovers became real queued \`graph.derive\` jobs mid-suite and broke an unrelated test (\`026 maintenance queue\` concurrent-claim race: Expected 1, Received 2).

## What happened in #609

The lane worked around it in its OWN test (cleanup by \`job_kind\` across all namespaces, runtime stopped before delete) — proven by full-suite runs: main 3701/0, branch-before 3701/1, branch-after 3702/0. The owning suites' missing cleanup was out of that lane's scope and remains.

## Ask

Fix cleanup in the suites that create \`parity-source-registry-*\` fixtures so they tear down what they seed. The next global producer added to the codebase will hit the same rows.

## SME

Pattern captured in \`docs/sme/correctness.md\` → \`review.producer_activation_promotes_inert_fixtures\` (2026-08-07).

PR #729, opened 2026-08-18
fix(embeddings): give the repair primitive the bulk caller it never had (#724)

rodaddy:
## What this is

Issue #724 item 1 — rows captured 2026-08-14..2026-08-17 stored but not
embedded/indexed. This PR carries the red-first done-means gate **and the fix**:
`scripts/repair-embeddings.ts`, the bulk caller the embedding-repair primitive
never had.

The window backlog is drained: eligible-but-unembedded `ob_session_lanes` went
**549 → 0**, and the gate moved from exit 1 to exit 0.

## The premise, as measured

The lane brief located the backlog in `ob_session_events`. Measured against the
live dogfood database (RUNNING, observed this session), it is not there:

- `ob_session_events`, window, namespace `rico`: **55 rows, 55 embedded**, zero
  NULL embeddings, `embedded_at - created_at` max lag −0.0008s — embedding is
  synchronous on that write path.
- The window gap was in `ob_session_lanes`: **2 of 2 eligible rows unembedded**
  (549 all-time eligible-but-unembedded across 12 namespaces).

So the brief's shape is **killed**; the invariant it reached for is what the gate
asserts and what the fix restores.

## Root cause — a design boundary, not a broken worker

Measured before touching anything, because the lane asked whether the restored
worker drains on its own. **It does not:**

| Observation | Result |
| --- | --- |
| Eligible-but-unembedded `ob_session_lanes` at T0 | 549 |
| Same count 4 minutes later | 549 |
| `maintenance_jobs` rows updated in the preceding 10 min | 0 |
| `embedding.repair` jobs ever recorded | 0 |

`docs/embedding-repair.md` ("Server-owned runtime and enqueue boundary") states
the boundary outright: `src/maintenance-bootstrap.ts` starts the
`embedding.repair` runner, but the bootstrap **"invents no namespace and enqueues
no job"** — enqueue is deliberately an explicit, auth-scoped CALLER boundary.

Nothing occupied that caller position for a bulk historical backlog:
`src/maintenance-sweep.ts` enqueues only `memory.distill` and graph-derivation,
and `scripts/backfill.ts` covers only the five original domain tables, which
excludes `ob_session_lanes`. The runner was healthy and correctly idle.

## The fix — smallest thing that rides the existing path

`scripts/repair-embeddings.ts` adds no pipeline, no table, and no SQL. It loops
the existing bulk primitive whose own docstring says it is for
"scripts/backfill-style bulk runs":

- `src/embedding-repair.ts:682` `repairStaleBatch(...)` — the bulk primitive.
- `src/embedding-targets.ts:326` `EMBEDDING_TARGET_NAMES` — the registry, read at
  runtime; a second hand-maintained copy is the #433 defect class.
- `src/embedding.ts` `generateEmbeddingWithMetadata`, `src/db/pool.ts`
  `createPool()` — the same provider and pool the server uses.

Every safety property is the primitive's, inherited unchanged: namespace-bound
reads *and* guarded writes, embeddings generated outside locks, idempotent
convergence, retryable-vs-permanent classification, content-free logging. Scope
is mandatory — `--namespace <ns>` or the separately named `--global`.

Two behaviors are measurement-driven, not preference:

**Selection asks for `missing` alone.** `buildSelection`
(`src/embedding-repair.ts:250`) ORs every reason into one *unordered*
`SELECT ... LIMIT n`, and the `source_drift` arm matches rows that already have
an embedding. Observed: all-reasons `limit=500` returned 100 mixed candidates
while 549 rows sat unembedded, and with no `ORDER BY` repeated batches need never
converge. Restricting to `missing` makes every selected row repairable, which is
what makes `repaired === 0` a truthful stop condition. Drift repair stays the
queue handler's job.

**A per-table failure no longer aborts the run.** The first `--global` run died on
a leaked test fixture live on the dogfood database: a `BEFORE UPDATE` trigger
`dream_bulk_release_trigger` on `thoughts` that RAISEs on every update, from
`src/tools/__tests__/bulk-set-tier.pg.test.ts` (#613 fixture-leakage class). This
PR does **not** drop it — destructive DDL is out of lane scope — it only refuses
to let one table hide the others' backlogs, and reflects it in the exit code.

## Run result

`bun run scripts/repair-embeddings.ts --global` — **1,627 rows repaired**:
decisions 1064, `ob_session_lanes` 489, `ob_session_events` 62, `ob_entities` 12;
relationships/projects already clean. `thoughts` and `sessions` aborted on the
two pre-existing defects noted below.

## Verification

- Done-means: scripts/done-means/724-backlog-recallable.sh

**RED**, before the run: **exit 1** — clause 1 FAIL (`ob_session_lanes` 2 of 2
eligible window rows unembedded), clause 2 PASS (keyword=1 semantic=16), clause 3
control PASS (keyword=1 semantic=25). A passing control with a failing clause 1 is
what makes the red valid. Exit 3 separately confirmed against an unreachable
database.

**GREEN**, after the run: **exit 0** — all three clauses PASS, `unembedded total:
0` across 101 eligible window rows, `ob_session_lanes: 0 of 2 ... clean`.

Also: `bunx tsc --noEmit` clean; `bun run test:isolated
src/embedding-repair.test.ts src/embedding-targets.test.ts` — 76 pass, 0 fail.

## Critical Self-Review

- Highest-risk behavior: a bulk loop writing embeddings across every table under `--global`. Mitigated by delegating every write to the existing guarded primitive (no SQL is written here), by a mandatory explicit scope with no unscoped default, by a dry-run mode proven to mutate nothing (549 unchanged after a dry run), and by a per-table batch bound so a non-converging table cannot spin.
- Assumptions that could be wrong: that `missing` is the right reason set for a bulk drain. If a future backlog is drift-shaped rather than missing-shaped, this CLI will not see it and the queue handler must; `--reason` is exposed for that, but the default deliberately does not chase drift.
- Missing/weak tests: the CLI itself has no unit test — it is evidenced by an observed red-to-green transition on the real database plus the primitive's own 76 passing tests. `repairAll` and `parseArgs` are exported specifically so a test can drive them with a fake `db`/`embedFn`; that test is not written here.
- Security/permission risk: namespace scope is mandatory and validated before any query, `--global` and `--namespace` are mutually exclusive so a scope cannot be silently widened, and `--table` is checked against the registry allowlist. All isolation predicates are the primitive's, bound on both read and guarded write.
- Migration/deploy risk: none. No schema change, no migration, no service or launchd change. The CLI is run by hand and nothing schedules it.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, REST surface, or Python client is touched, so `docs/downstream-rollout.md` classification is not-applicable.
- Rollback/cleanup concern: the embedding writes are not reversible by removing the file, but they are convergent and idempotent — re-running with unchanged source yields the identical vector/hash/model. Rows that had no embedding now have one; nothing was overwritten, because `missing` selects only NULL-embedding rows.
- Fixes made before PR: the reason-set narrowing and the per-table fault isolation, both after observing real failures (a non-converging all-reasons batch, and a whole `--global` run killed by one poisoned table).
- Known residual risk: two tables did not drain and are NOT fixed here — the leaked `dream_bulk_release_trigger` blocks 1,297 `thoughts` rows, and `idx_sessions_content_hash` collisions block 11,166 `sessions` rows. Both need their own issues. Nothing schedules this CLI, so an empty enqueue boundary can produce the same backlog again. The gate counts rows, so a row carrying a wrong vector still passes clause 1. Attribution of the backlog to the nats-worker outage remains unverified.
- SME review-memory update: [ ] docs/sme/ updated [x] not applicable because: the reusable findings here are recorded in the commit message, the PR body, and `_plans/worklog/lane-724-backlog-reembed.md` for the lane-contract harvest; the two blocked tables are being filed as their own issues rather than as reviewer-lane entries.

## Review Gate

- Live Open Brain checks: [x] linked below [ ] not applicable because: n/a
  - The `--global` run against the live dogfood database repaired 1,627 rows across five tables (counts quoted above), and the done-means gate's own read-only probe of that database moved from exit 1 to exit 0 with all three clauses passing.

Refs #724

### DIFF
--- a/_plans/worklog/lane-724-backlog-reembed.md
+++ b/_plans/worklog/lane-724-backlog-reembed.md
@@ -1,0 +1,135 @@
+# Lane A — issue #724 item 1: Aug 14-17 backlog re-embed
+
+Status of every claim below: **RUNNING** where it says "observed", meaning it was
+measured against the live dogfood database (`open_brain_local_20260724` on
+`127.0.0.1`) during this session on 2026-08-17/18. The code change is **WRITTEN**
+and, once its PR lands, MERGED — nothing here is deployed to core01.
+
+## 1. Does the restored worker drain the backlog on its own?
+
+**No. CONFIRMED by measurement, not inference.**
+
+The lane brief asked this first, so it was measured rather than assumed:
+
+| Observation | Result |
+| --- | --- |
+| Eligible-but-unembedded `ob_session_lanes` at T0 (21:51:20 -04) | 549 |
+| Same count at T1, 4 minutes later (21:55:20 -04) | 549 |
+| `maintenance_jobs` rows updated in the preceding 10 minutes | 0 |
+| `embedding.repair` jobs ever recorded in `maintenance_jobs` | 0 |
+
+`maintenance_jobs` has only ever held `memory.distill` (3, last updated
+2026-08-08) and `system.facts` (30, last updated 2026-07-26). The queue has been
+idle for over a week.
+
+### Root cause — it is a design boundary, not a broken worker
+
+`docs/embedding-repair.md` ("Server-owned runtime and enqueue boundary") states
+it outright: `src/maintenance-bootstrap.ts` starts the `embedding.repair` runner,
+but the bootstrap **"invents no namespace and enqueues no job."** Enqueue is
+deliberately an explicit, auth-scoped CALLER boundary, because a namespace cannot
+be fabricated by a background process.
+
+Nothing in the repo occupies that caller position for a bulk historical backlog:
+
+- `src/maintenance-sweep.ts` enqueues `memory.distill` and graph-derivation jobs
+  only — it never enqueues `embedding.repair`.
+- `scripts/backfill.ts` covers only the five original domain tables
+  (`BACKFILL_TABLES`: thoughts, decisions, relationships, projects, sessions).
+  `ob_session_lanes` — where the window gap actually was — is **not** in that
+  list.
+
+So the runner was healthy and correctly idle. The backlog had no caller to push
+it. Attribution of the backlog's *origin* to the nats-worker outage remains
+**unverified**; this lane observed rows, not causes.
+
+## 2. The fix — smallest thing that rides the existing path
+
+`scripts/repair-embeddings.ts`: a CLI that occupies the missing caller position.
+It adds no pipeline, no table, and no SQL. It loops the existing bulk primitive
+whose own docstring (`src/embedding-repair.ts:682`) says it "is for
+scripts/backfill-style bulk runs":
+
+- `src/embedding-repair.ts:682` `repairStaleBatch(...)` — the bulk primitive.
+- `src/embedding-targets.ts:326` `EMBEDDING_TARGET_NAMES` — the table registry,
+  read at runtime. A second hand-maintained copy of that list is the #433 defect
+  class, so there is none here.
+- `src/embedding.ts` `generateEmbeddingWithMetadata` and `src/db/pool.ts`
+  `createPool()` — the same provider and pool the server uses.
+
+Every safety property is the primitive's, inherited unchanged: namespace-bound
+reads *and* guarded writes, embeddings generated outside locks, idempotent
+convergence on replay, retryable-vs-permanent failure classification,
+content-free logging. Scope is mandatory — `--namespace <ns>` (repeatable) or the
+separately named `--global`; there is no unscoped default.
+
+### Two things measurement forced into the design
+
+**(a) `reasons: ["missing"]` is required for a bulk drain to converge.**
+`buildSelection` (`src/embedding-repair.ts:250`) ORs all staleness reasons into
+one *unordered* `SELECT ... LIMIT n`, and the `source_drift` arm is
+`embedding IS NOT NULL AND content_hash IS NOT NULL` — rows that already *have*
+an embedding — with the real hash comparison done in JS afterward. Observed: with
+all reasons enabled, `limit=500` on `ob_session_lanes` returned 100 candidates of
+mixed reason while 549 rows sat unembedded. With no `ORDER BY`, repeated batches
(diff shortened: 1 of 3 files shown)


rodaddy:

rodaddy:
verify-lane receipt: check=scripts/done-means/724-backlog-recallable.sh exit=0 sha=df49c13ce09ec9fa876912e3acb647769621fd0a at=2026-08-18T02:14:32.563Z

```

=== DONE-MEANS #724 item 1 (Aug 14-17 backlog is embedded AND recallable) ===
window          : 2026-08-14 .. 2026-08-18 (exclusive)
namespace probed: rico (discovered, not hardcoded)
rows examined   : 101 eligible window rows across the whole EMBEDDING_TARGETS registry
unembedded total: 0

CLAUSE 1 no unembedded backlog in window : PASS
    thoughts: 0 of 0 eligible window rows unembedded — clean
    decisions: 0 of 0 eligible window rows unembedded — clean
    relationships: 0 of 0 eligible window rows unembedded — clean
    projects: 0 of 0 eligible window rows unembedded — clean
    sessions: 0 of 44 eligible window rows unembedded — clean
    ob_session_lanes: 0 of 2 eligible window rows unembedded — clean
    ob_session_events: 0 of 55 eligible window rows unembedded — clean
    ob_entities: 0 of 0 eligible window rows unembedded — clean

CLAUSE 2 recall reaches the window       : PASS
  window content is reachable in BOTH arms: keyword=1 semantic=16 rows dated inside 2026-08-14..2026-08-18

CLAUSE 3 control (pre-window recallable) : PASS
  known-good pre-2026-08-14 row (namespace rico, newest embedded 2026-08-13) recalled through the real path: keyword=1 semantic=25 in-range hits. The retrieval path is live, so a clause-2 miss means the WINDOW is unreachable, not the service.

RESULT: PASS — every eligible row captured in the window carries an embedding, and window content returns from the real recall path in both the lexical and the vector arm.
```

</details>

Issue #722, opened 2026-08-10
Both pre-push done-means checks (705, 712) are RED on untouched main: unpinned fixture core.hooksPath trips #711's assertion, misreported as substantive regressions


rodaddy:
## Family classification

**Family B** — environment deciding a verdict. A cross-defect INTERACTION: the
#711 fix (an assertion added to the hook) breaks the #705 check's fixtures,
because those fixtures never isolated the environment they inherit.

## Defect

`scripts/done-means/705-pre-push-base-selection.sh` is **RED (exit 1), 5 of 6
clauses failing, on the untouched primary checkout at `183d066`** — with a
clean tree and no local modifications.

The #705 fix itself is NOT broken. The check's fixtures are.

Line 146 builds each throwaway repo with a bare init:

```sh
git -C "$FIXTURE" init -q -b main
```

which sets no `core.hooksPath`, so the fixture inherits the operator's GLOBAL
value (`/Users/rico/.config/git/hooks`). #711 later added an assertion to
`_githooks/pre-push` that refuses any `core.hooksPath` other than `_githooks`.
That assertion runs BEFORE the `--explain` early exit, so every fixture
invocation now dies at the hooksPath check and prints the #711 refusal instead
of the base-selection output the clauses parse:

```
CLAUSE a: FAIL — decision fields missing from the explain output (memory='' openbrain=''):
  ✗ core.hooksPath does not match what _githooks/install.sh writes.
      configured: /Users/rico/.config/git/hooks
      expected:   _githooks
CLAUSE c: FAIL — the hook did not name the base ref it compared against: <same refusal>
CLAUSE d: FAIL — an upstreamless branch made the hook fail rather than fall back (exit=1)
CLAUSE f: FAIL — the real push path produced no readable package decision
```

Clause b's message is actively misleading — "a real python/openbrain edit was
NOT detected", "the packages are no longer separately gated" — describing a
#705 regression that has not happened. The empty strings it reports are parse
failures against a refusal, not classification failures.

## Proof the hook is sound

Same shipped `_githooks/pre-push`, same fixture shape, with the one line the
fixture is missing (`git config core.hooksPath _githooks`):

```
$ (cd $FIXTURE && ./_githooks/pre-push --explain)
  base: main — fallback (no configured upstream, no origin/main)
  changed: python/openbrain-memory=no python/openbrain=no parity=no
  base_ref=main
  base_source=fallback (no configured upstream, no origin/main)
```

[...]

Issue #705, opened 2026-08-10
pre-push gate diffs against origin/main, so every wip-branch lane inherits 80 commits of Python attribution


rodaddy:
## What happens

`_githooks/pre-push` decides whether a push touched the Python packages by
diffing against `origin/main`. Every lane in the current flow branches from
`wip/2026-08-07`, which is **80 commits ahead of `origin/main`**. So the hook
attributes all 80 commits' Python changes to whatever lane is pushing.

Measured on lane `lane/issue-artifacts-outcomes` (2026-08-09), which changes
zero Python files:

```
git diff --name-only $(git merge-base HEAD origin/main) HEAD -- python/openbrain | wc -l
# => 1     (what the hook believes the push changed)

git diff --name-only origin/wip/2026-08-07 HEAD -- python/openbrain | wc -l
# => 0     (what the push actually changed)
```

The hook then runs `uv run pytest -q` for `python/openbrain` and fails the push
on `test_receipts_gate_crosslang.py::test_python_capture_receipt_clears_the_gates_capture_block`
— which is **#704**, a known main-owned failure that reproduces on the
unmodified base commit `a45e7d9` in a separate checkout.

Net effect: a lane whose diff contains no Python is blocked by a Python test it
did not touch, inherited from five commits already on the wip branch.

## Where

Both base-selection paths hardcode `origin/main`:

- `_githooks/pre-push:32-37` — the stdin-range path, taken when the branch is
  new (`remote_sha` is the zero SHA), which is every first lane push.
- `_githooks/pre-push:59-65` — the manual-invocation path.

## Why it matters beyond the annoyance

The gate stops discriminating. Once every lane is told it changed the Python
package, a push that genuinely breaks Python looks exactly like one that
touched only markdown, and the honest response to the failure ("not mine")
becomes the habitual one. That is how a gate turns into noise that gets
routed around with `--no-verify` — which the lane contract forbids, correctly.

## Suggested direction (not implemented here)

Compare against the branch's actual integration target rather than a hardcoded
`origin/main`: the configured upstream (`@{upstream}`) when one exists, else
the merge-base with the branch the PR targets. Whatever is chosen should be
announced in the hook's output — "comparing against X" — so a future lane can
see which base produced the verdict instead of inferring it from a failure.

Filed from the artifact-resolution lane, which hit it on a zero-Python diff.
State: WRITTEN (defect confirmed by the two measurements above); no fix
proposed in code yet.

rodaddy:
[...]

PR #731, opened 2026-08-18
fix(done-means): pin the fixture hook environment so 705/712 judge the code (#722)

### LINKED ISSUE #712: pre-push fails every push: bun test aborts with WriteFailed when its stdout is git's pipe (tests are green)
## What happens

Every `git push` in this repo fails. `_githooks/pre-push` runs `bun test`
(line 229) with its stdout attached to git's pipe, and bun aborts partway
through printing the coverage table:

```
 src/validation-errors.ts       |  100.00 |   90.
error: An internal error occurred (WriteFailed)
error: failed to push some refs to 'github.com:rodaddy/open-brain.git'
```

The tests are not failing. The same suite, same tree, redirected to a FILE
instead of a pipe, is green:

```
$ bun test > out.txt 2>&1 ; echo $?
0
$ tail -5 out.txt
 3372 pass
 525 skip
 0 fail
 11819 expect() calls
Ran 3897 tests across 244 files. [30.84s]
```

## Measured 2026-08-10 — main-owned, not branch-owned

Reproduced on the UNTOUCHED primary checkout at `e3917ab` (`wip/2026-08-07`),
clean tree, no local changes:

```
$ git log --oneline -1
e3917ab fix(gates): resolve Done-means against the PR head and pre-push against
        the integration target (#706, #705) (#708)
$ git status --short          # empty
$ ( bun test 2>&1; echo "BUN_EXIT=$?" ) | tail -1
BUN_EXIT=1
$ bun test > /dev/null 2>&1; echo $?
0
```

Pipe → exit 1 with `WriteFailed`. Not-a-pipe → exit 0. The only difference is
whether stdout is a pipe.

It is output-SIZE dependent, so it is not simply "bun cannot write to a pipe":

```
$ ( bun test ./src/audit-log.test.ts 2>&1; echo "BUN_EXIT=$?" ) | tail -1
BUN_EXIT=0          # one file, small output, piped — fine
```

`--coverage=false` does not avoid it either (`BUN_EXIT=1`), so disabling the
coverage table is not on its own the fix.

Environment: `bun 1.3.14 (0d9b296a)`, macOS (Darwin 27.0.0).

## Why this is urgent

The push path is the enforcement path. `_githooks/pre-push` is where #705's
base-selection fix lives and where typecheck/test/parity gating happens, and
right now it fails on EVERY push for a reason that has nothing to do with the
change being pushed. That is the precise condition under which `--no-verify`
becomes habitual — the failure mode #705's own commit message warns about:
"once every lane inherits the attribution ... 'not mine' becomes habitual,
which is how a gate becomes noise routed around with --no-verify."

A gate that fails identically for a green push and a broken one has stopped
carrying information.

1. Do not let the hook's own reporting decide the verdict: run the suite with
   its output going somewhere that is not a bounded pipe (a log file under
   `{temp_workspace}/open-brain/_scratch/`, tee'd back), and gate on bun's
   EXIT CODE. The hook already owns the summary it prints.
2. Whatever the shape, the hook must distinguish "tests failed" from "the
   runner could not write its output" and SAY WHICH (AGENTS.md, nothing is
   adjusted silently). Today both present as a bare `WriteFailed` and a
   refused push.
3. A done-means clause should drive the hook through a real pipe with
   full-suite-sized output, since that is the only shape that reproduces it —
   `docs/lane-contract.md` round 28: a seam that avoids the real invocation
   shape proves nothing.

## State

WRITTEN — reproduced above on an untouched `e3917ab` primary checkout; no fix
in code. Blocking: the #709 lane (`lane/709-pr-head-ref`, committed at
`20d5aaf`, done-means 5/5 GREEN, typecheck clean) cannot push. Reported rather
than routed around: `--no-verify` was NOT used, per `docs/sop-rlvr-lanes.md`
"a broken process fails HARD".

Related: #705 (the pre-push base-selection fix), #711 (core.hooksPath absolute,
so worktrees run the primary's copy of this same hook).

rodaddy:
## Summary

`scripts/done-means/705-pre-push-base-selection.sh` and
`scripts/done-means/712-pre-push-pipe-safe.sh` built their throwaway git
fixtures with a bare `git init`, which sets no `core.hooksPath`. The fixture
therefore INHERITED the operator's global value, and #711's in-hook assertion —
which runs before the `--explain` early exit — refused every fixture
invocation. Every clause then parsed a refusal instead of the hook's decision,
and both checks reported that as a substantive regression of #705/#712 in
confident, specific, FALSE text.

Family B (the environment deciding a verdict), reaching a new surface: the
gates' own checks.

Measured on the untouched tree at `origin/wip/2026-08-07`, under the operator's
real global `core.hooksPath=/Users/rico/.config/git/hooks`:

- `705`: 5 of 6 clauses FAIL, exit 1. Only clause (e), which reads the hook's
  source structurally and never invokes it, passes. Clause (b)'s text —
  "the packages are no longer separately gated" — is about a hook that never ran.
- `712`: 6 of 6 clauses FAIL. Clause (f)'s "redirecting stdout alone does not
  fix #712" is a specific verdict on the FIX's shape, emitted about a hook whose
  redirect code never executed.

### The fix

1. **Pin the fixture environment.** Both fixtures set
   `core.hooksPath=_githooks` immediately after `git init`, so the fixture is
   hermetic against the operator's global config and the hook under test reaches
   its own subject code. The hook IS these checks' subject, so the pin is the
   real value — contrast `709`/`714`, which point at an empty hooks directory
   because the hook is NOT their subject.
2. **A HARNESS-ERROR guard, not a FAIL.** Before any clause runs, each check
   asserts the fixture's effective `core.hooksPath`. A mismatch exits 3, because
   a blinded fixture is not a statement about the subject in either direction.
3. **A parse failure is not a regression.** `blinded_by` is anchored on the
   ABSENCE of every marker the hook owns for a successful decision, not on a
   list of known refusal texts, so an invariant added tomorrow still degrades to
   "produced no decision" rather than to a false regression.

### Second defect, found by finishing the salvage (not in the original checkpoint)

With the pin applied, `712` still failed clause (d) with the hook FULLY REACHED
— a real verdict, not a blind (`3372 pass 0 fail`, `DM712_HOOK_EXIT=0`). So the
pin alone was not the whole defect.

Clause (d) required the announced log path to sit under the fixture's own
`OPENBRAIN_TEMP_WORKSPACE` (`$SCRATCH/ws`), which `run_hook_through_pipe` sets
on the hook invocation. **The fixture cannot win that assertion**, for the same
reason #722 exists — the ambient environment decides the verdict:

```
env OPENBRAIN_TEMP_WORKSPACE=/probe/ws zsh  -c 'echo $OPENBRAIN_TEMP_WORKSPACE'
  -> /Volumes/ThunderBolt/_tmp      # the caller's value is discarded
env OPENBRAIN_TEMP_WORKSPACE=/probe/ws bash -c 'echo $OPENBRAIN_TEMP_WORKSPACE'
  -> /probe/ws                      # bash honours it
```

The hook's shebang is `#!/usr/bin/env zsh`; `~/.zshenv` is read by EVERY zsh —
including a non-interactive script — and sources
`~/Library/Application Support/rtech-infra/env-roots.zsh`, whose line 6 is an
unconditional `export OPENBRAIN_TEMP_WORKSPACE=/Volumes/ThunderBolt/_tmp`. An
unconditional export in a startup file overwrites what the caller passed, so the
hook always logs to the operator's real temp workspace. Measured with a probe
copy of the check: `has_buntest=1`, `has_ws=0`, announced path
`.../open-brain/_scratch/pre-push/bun-test-47592-1787017419.log`.

So the old assertion tested the OPERATOR'S SHELL PROFILE, not the hook. The
clause's stated subject is that the redirect is ANNOUNCED and the log NAMED;
it now checks the announcement plus a concrete `bun-test-<pid>-<epoch>.log`
filename. Pinning the DIRECTORY is not this check's business and is not
something the fixture is able to do; pinning the SHAPE is, and it still fails if
the hook stops naming its log.

## Verification

- Done-means: scripts/done-means/722-fixture-hookspath-pinned.sh

RED, on the untouched tree at `origin/wip/2026-08-07`, with the operator's real
global `core.hooksPath` in force (and reproduced identically with a simulated
global via `GIT_CONFIG_GLOBAL` pointed at a fixture config, so the real global
was never touched):

```
705 RED EXIT=1
CLAUSE a: FAIL — decision fields missing ... ✗ core.hooksPath does not match ... (issue #711)
CLAUSE b: FAIL — the packages are no longer separately gated
CLAUSE c: FAIL — the hook did not name the base ref it compared against ... (issue #711)
CLAUSE d: FAIL — an upstreamless branch made the hook fail rather than fall back (exit=1)
CLAUSE e: PASS   (structural — never invokes the hook)
CLAUSE f: FAIL — the real push path produced no readable package decision

712: 6 of 6 clauses FAIL
CLAUSE f: FAIL — redirecting stdout alone does not fix #712
```

GREEN, on this branch, under the operator's REAL environment — the same one
where the checks were RED:

```
705-pre-push-base-selection REAL-ENV EXIT=0     (6/6 PASS)
712-pre-push-pipe-safe      REAL-ENV EXIT=0     (6/6 PASS)
722-fixture-hookspath-pinned REAL-ENV EXIT=0    (4/4 PASS)
```

`722`'s own clauses, which prove the pinning rather than assuming it:

```
CLAUSE a: PASS — 705 exit=0 (6 clauses) and 712 exit=0 (6 clauses) on this tree
CLAUSE b: PASS — DELIBERATE RE-BLIND: with the pin removed (705: 1 line, 712: 1 line;
          ambient core.hooksPath='/Users/rico/.config/git/hooks') both checks fail,
          both name core.hooksPath, and NEITHER emits its original regression text
CLAUSE c: PASS — a blinded run exits 3 (HARNESS-ERROR), never 1
CLAUSE d: PASS — all 5 done-means checks that build a git fixture pin
          core.hooksPath explicitly
```

Clause (b) is the negative control: it removes the pin, re-blinds each check,
and asserts both that the check FAILS and that the false regression text is
GONE. Without it, the pin could be deleted and everything would still look green.

Mutation test on the new clause (d) assertion: replacing the
`bun-test-<pid>-<epoch>.log` pattern with a sentinel the hook never emits makes
clause (d) FAIL, so the clause is not vacuously green.

`bunx tsc --noEmit`: clean (after `bun install --frozen-lockfile`; the change is
shell-only and adds no TypeScript).

Not run: the full `bun run test:isolated` suite. This diff touches only
`scripts/done-means/*.sh` and no runtime code, and the three affected checks
each drive the real shipped `_githooks/pre-push` end to end.

## Critical Self-Review

- Highest-risk behavior: a fixture pinned to `core.hooksPath=_githooks` could
  mask a REAL #711 drift in the operator's config. It cannot: the pin is
  fixture-local (`git -C "$FIXTURE" config`), the operator's global is never
  read or written, and #711's assertion still runs in the real hook on every
  real push. These checks are about #705/#712, not about the operator's config.
- Assumptions that could be wrong: that `_githooks` is the correct pin value.
  Verified against `_githooks/pre-push` line 83 (`EXPECTED_HOOKS_PATH="_githooks"`)
  and against `_githooks/install.sh`, which is what writes it — not assumed.
- Missing/weak tests: clause (d)'s new assertion pins the log's SHAPE, not its
  DIRECTORY, so a hook that logged to a wrong-but-plausibly-named path would
  still pass. That is deliberate and documented in the file: the fixture
  provably cannot pin the directory (zsh startup overwrites the variable), and
  an assertion the fixture cannot satisfy is exactly the #722 defect. Pinning
  the directory needs `env-roots.zsh` to stop unconditionally exporting, which
  is out of this lane's scope and is NOT fixed here.
- Security/permission risk: none. No auth, namespace, network, or DB surface is
  touched. No secrets in the diff; the pre-commit secret scan reported
  `no leaks found`.
- Migration/deploy risk: none. No migration, no schema, no runtime code.
- Downstream client/runtime risk: none. `scripts/done-means/*.sh` are
  repo-local verification scripts; no MCP tool, transport, schema, or Python
  client surface changes, so `docs/downstream-rollout.md` does not apply.
- Rollback/cleanup concern: revert the two commits; nothing is stateful. The
  lane clone and the baseline worktree are archived at teardown (below), and
  the probe copy of `712` was moved to `_archive/` rather than deleted.
- Fixes made before PR: the clause (d) defect above — found by actually running
  the salvaged checkpoint instead of trusting it. The checkpoint was committed
  UNTESTED and was NOT sufficient on its own; it left `712` at 5 of 6.
- Known residual risk: `env-roots.zsh` unconditionally exporting
  `OPENBRAIN_TEMP_WORKSPACE` means NO zsh-shebanged script in this repo can
  have its temp workspace overridden by a caller. This PR works around that in
  one clause and documents it; it does not fix the root cause, and other
  callers that believe they can set that variable for a zsh child are still
  wrong. Filing that separately is the controller's call.
- SME review-memory update: [x] not applicable because: the lane's operating
  lesson (a fixture must pin every environment input its subject reads, and a
  check that cannot express the defect is a harness error rather than a
  verdict) belongs in `docs/lane-contract.md` Tightenings, which the merge pass
  harvests from this report; `docs/sme/` covers reviewer knowledge, and no new
  reviewer-facing code pattern was found.

## Review Gate

- Live Open Brain checks: [x] not applicable because: this change touches only
  `scripts/done-means/*.sh` verification scripts — no MCP tool, transport,
  schema, namespace, or Open Brain runtime surface is involved, so there is
  nothing for a live Open Brain canary to exercise.

Refs #722

### DIFF
--- a/scripts/done-means/705-pre-push-base-selection.sh
+++ b/scripts/done-means/705-pre-push-base-selection.sh
@@ -144,6 +144,18 @@ git_f() { git -C "$FIXTURE" -c user.name="done-means-705" -c user.email="done-me
 
 setup_fixture() {
   git -C "$FIXTURE" init -q -b main 2>/dev/null || return 1
+
+  # PIN THE FIXTURE'S HOOKS DIRECTORY (issue #722). A bare `git init` sets no
+  # `core.hooksPath`, so the fixture INHERITS the operator's global value --
+  # this fixture is not hermetic without this line. #711 added an assertion to
+  # the shipped hook that refuses any value other than `_githooks`, and it runs
+  # BEFORE the `--explain` early exit, so an inherited value made every
+  # invocation below die at that assertion and every clause parse a refusal
+  # instead of a base decision. The hook IS this check's subject, so the pin is
+  # the real value (contrast 709/714, which neutralise hooks entirely because
+  # the hook is not their subject).
+  git -C "$FIXTURE" config core.hooksPath _githooks || return 1  # DM722-PIN
+
   mkdir -p "$FIXTURE/python/openbrain/src" \
            "$FIXTURE/python/openbrain-memory/src" \
            "$FIXTURE/contracts" \
@@ -211,6 +223,34 @@ SPEC
 
 setup_fixture || fail_hard "could not build the fixture repository (see $FIXTURE)"
 
+# ---------------------------------------------------------------------------
+# FIXTURE-ENVIRONMENT GUARD (issue #722)
+# ---------------------------------------------------------------------------
+# Assert the pin actually took EFFECT before any clause runs. `git config
+# --get` here reads exactly what the hook's own assertion reads, so this guard
+# and the subject agree on the value by construction.
+#
+# WHY THIS IS EXIT 3 AND NOT A CLAUSE FAILURE. A blinded fixture says NOTHING
+# about base selection: the hook refuses before reaching a single line of its
+# subject code. Reporting that as FAIL would be a verdict about #705 that the
+# run did not earn -- and that is precisely the defect #722 filed, where five
+# clauses reported confident, specific, FALSE regressions ("the packages are no
+# longer separately gated") about a hook they never reached. A harness error is
+# not a failure of the thing under test.
+FIXTURE_HOOKS_PATH="$(git -C "$FIXTURE" config --get core.hooksPath 2>/dev/null || true)"
+if [ "$FIXTURE_HOOKS_PATH" != "_githooks" ]; then
+  fail_hard "$(printf '%s' "\
+BLIND FIXTURE -- this run cannot say anything about #705.
+    fixture core.hooksPath: '${FIXTURE_HOOKS_PATH:-<unset>}'
+    required:               '_githooks'
+  The fixture is inheriting an ambient core.hooksPath instead of pinning its
+  own, so the shipped hook will refuse at its #711 assertion before reaching
+  any base-selection code. Every clause below would then be parsing a REFUSAL,
+  not a decision. This is a defect in this CHECK's fixture (#722), not in the
+  hook. Fix: restore the 'git -C \"\$FIXTURE\" config core.hooksPath _githooks'
+  line in setup_fixture.")"
+fi
+
 # A lane off wip, tracking wip — the exact shape every lane in this flow has.
 new_lane_branch() {
   local name="$1"
@@ -223,6 +263,42 @@ new_lane_branch() {
 explain() {
   EXPLAIN_OUT="$(cd "$FIXTURE" && "$FIXTURE/_githooks/pre-push" --explain 2>&1 </dev/null)"
   EXPLAIN_EXIT=$?
+}
+
+# ---------------------------------------------------------------------------
+# REFUSAL DETECTION (issue #722) -- "I could not reach the subject" is not "the
+# subject regressed".
+# ---------------------------------------------------------------------------
+# The fixture guard above catches the ONE blind that has actually happened. This
+# is the general form, and it is the part that survives the next unrelated
+# invariant someone adds to the hook: the hook has a growing set of preconditions
+# that refuse BEFORE any base-selection code runs, and each one turns the
+# clauses' parse into a refusal.
+#
+# `blinded_by` returns the refusal's first meaningful line when the output
+# carries no base-selection decision at all, and empty otherwise. Clauses use
+# it to say WHAT BLINDED THEM instead of reporting empty decision fields as a
+# classification defect -- a parse failure and a wrong answer are different
+# defects with different owners (lane-contract round 29).
+#
+# It is deliberately anchored on the ABSENCE of every marker the hook owns for a
+# successful decision, not on a list of known refusal texts: a refusal added
+# tomorrow is unknown to this file, but "produced no decision" is checkable
+# today and stays true.
+blinded_by() {
+  local out="$1"
+  # Any of these means the hook REACHED its subject and produced a decision.
+  printf '%s\n' "$out" | rg -q '^[[:space:]]*base_ref=' && return 0
+  printf '%s\n' "$out" | rg -q '^[[:space:]]*changed_' && return 0
+  printf '%s\n' "$out" | rg -qF 'Python package changed' && return 0
+  printf '%s\n' "$out" | rg -qF 'Python package unchanged' && return 0
+  printf '%s\n' "$out" | rg -qF 'openbrain package changed' && return 0
+  printf '%s\n' "$out" | rg -qF 'openbrain package unchanged' && return 0
+  # No decision anywhere -> blinded. Report the first non-blank line, which is
+  # the refusal's own headline.
+  local first
+  first="$(printf '%s\n' "$out" | rg -v '^[[:space:]]*$' | head -1)"
+  printf 'the hook REFUSED before reaching its base-selection subject; first line: %s' "${first:-<no output at all>}"
 }
 
 # Read one decision field out of the hook's explain output.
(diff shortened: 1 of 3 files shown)


rodaddy:

rodaddy:
verify-lane receipt: check=scripts/done-means/722-fixture-hookspath-pinned.sh exit=0 sha=93015cc285eb82705d8ffa7d5946c05b97e8270b at=2026-08-18T02:12:37.288Z

```
CLAUSE a (both pre-push done-means checks are GREEN on this tree): PASS — 705 exit=0 (6 clauses) and 712 exit=0 (6 clauses) on this tree
CLAUSE b (MUTANT: a re-blinded check fails, names the BLIND, and drops the false regression text): PASS — with the pin removed (705: 1 line(s), 712: 1 line(s); ambient core.hooksPath='/Users/rico/.config/git/hooks') both checks fail, both name core.hooksPath, and neither emits its original regression text
CLAUSE c (a blinded run exits 3 (HARNESS-ERROR), never 1): PASS — a blinded run exits 3 (HARNESS-ERROR) in both checks -- the blind is classified as a harness problem, not as a failure of the subject
CLAUSE d (no done-means git fixture inherits core.hooksPath): PASS — all 5 done-means checks that build a git fixture pin core.hooksPath explicitly (pinned to _githooks where the hook IS the subject; neutralised where it is not)
```

</details>

PR #732, opened 2026-08-18
fix(runtime): manual wrap adopts an existing lane's scope (#724)

rodaddy:
Refs #724 (item 4 — manual wrap adopts the hook-owned lane's scope).

A manual `openbrain-memory --event wrap` (and `--event checkpoint`) against a
lane the capture hook opened failed with `Existing lane exact scope does not
match session_start request`. Two Development sessions lost their wraps to it,
and it is the collateral symptom that led to discovering the nats-worker
outage.

The capture hook creates the base lane with `agent` set and every other
exact-scope coordinate NULL. The client made `session_start` claim the CALLER's
own full exact scope, which trips the server's #646 one-way fill, and
`session_wrap` was never reached at all.

**The server does not change.** `server/` and `src/` carry zero modifications in
this PR — the exact-scope predicate is an isolation boundary. Operator ruling
2026-08-17: the client stops asserting a scope it does not own.

## What changed

Claim first, adopt only on that one refusal:

1. `session_start` claims the caller's own exact scope, exactly as before. On
   success nothing else happens — same single call, same arguments — so the
   ordinary path is byte-for-byte unchanged and adoption costs nothing until it
   is actually needed.
2. Only on the specific scope refusal, `session_start` is re-issued claiming
   nothing but `session_key`. That is the server's `!hasCompleteExactScope`
   branch (`server/tools/session-lifecycle.ts:67,155`), which returns the
   existing lane verbatim rather than attempting a fill.
3. `session_wrap` then carries the coordinates that came back ON THAT LANE, not
   the requester's, because the write verb must still satisfy the server's
   lane-scope predicate. Adoption is not merely "send fewer fields on start".

Discovering unconditionally (an unscoped `session_start` before every wrap) was
tried first and is wrong twice over: on a lane that does not exist yet it
CREATES a bare NULL-scoped lane and permanently strands it, and it adds a wire
call to the sequence the cross-runtime contract fixture pins. Both were
observed as real test failures before the approach was changed.

`validate_adopted_lane` binds the adopting start on `session_key` + `namespace`
— the coordinates the caller genuinely owns — because there is no requested
exact scope for the response to prove. The namespace refusals (#654/#662) are
extracted into `_require_proven_namespace` and enforced unchanged on BOTH
paths, rather than copied into a second block that could drift.

Nothing else is widened: capture (`append_session_event`) and the context-pack
exact-scope proof are untouched, and the adopting flag is scoped to the single
discovery call so a spooled `session_start` replay still proves its own parked
scope.

Two fidelity gaps in the test fakes are corrected, both toward the real server:
`LaneAwareTransport` now mirrors the `hasCompleteExactScope` gate (it carried
only the refusal half of that server branch, so it refused a request the real
server accepts), and `StartThenFailClient` reads exact-scope coordinates with
`.get()`, matching the server schema where all of them are `.optional()`.

## RESOLVED BY RULING

Operator ruling (Rico, 2026-08-17): `test_checkpoint_first_lane_denies_hostile_same_namespace_claim` is RETIRED (commit `d7f5c06`) because the client provably cannot distinguish the capture hook lane from another same-namespace session lane — the wire sequences are identical — same namespace is the same tenant, the server namespace predicate is the real boundary, and AGENTS.md Coding Standards already state that client-side convenience checks are not security controls. A comment at the former test site records the retirement, and `docs/memory-contract.md` now states that wrap/checkpoint adoption is namespace-scoped by design.

## Downstream rollout (`docs/downstream-rollout.md`)

This is a **client behavior change** in `python/openbrain-memory/`, with no MCP
tool, schema, or protocol change.

- **rtech-mcps / mcp2cli** — N/A. No tool, schema, or transport contract moved;
  the server is byte-identical and `mcp2cli` speaks to it, not to this client.
- **rtech-hermes Python runtime/plugin** — N/A for correctness, but it is a
  consumer of this package. Its wraps only reach the new code path when a
  `session_start` is refused for scope, which today means its wrap was failing
  outright; the change strictly converts that failure into a write.
- **Live Hermes agent canary** — the isolation question is settled by the ruling
  above; the real canary is the one the issue
  actually asks for: a manual `openbrain-memory --event wrap` from a
  Development session against a capture-hook-owned lane, which is the exact
  call that has never produced a wrap receipt (`receipts.json` shows zero wrap
  receipts ever). That is a RUNNING check and has not been performed — this PR
  is verified against fake transports only.
- **Package version bump / republish** — not done in this PR; the controller
  should decide whether this rides an existing release lane.

## Verification

- Done-means: scripts/done-means/724-wrap-lane-adoption.sh
- RED (commit `dd5de5a`, observed live this session):
  `cd python/openbrain-memory && uv run pytest tests/test_runtime_wrap_lane_adoption.py -q`
  -> `6 failed, 1 passed in 0.09s`, headline failure
  `AssertionError: Open Brain tool returned an error context=call_tool:session_start body=existing lane exact scope does not match session_start request`
- GREEN (this commit): same command -> `7 passed in 0.06s`
- Full package (commit `d7f5c06`): `uv run pytest -q` -> `585 passed, 14 skipped in 7.11s`,
  zero failures. Baseline on `dd5de5a` was `6 failed, 580 passed, 14 skipped`.
- `uv run mypy src/openbrain_memory` -> `Success: no issues found in 17 source files` (re-run on `d7f5c06`)
- `uv run ruff check src tests` -> `All checks passed!` (re-run on `d7f5c06`)
- `uv run ruff format --check src tests` -> 9 files would be reformatted, which
  is the UNCHANGED pre-existing baseline on `dd5de5a` (measured by stashing);
  the one file this PR touched that drifted was formatted back.
- TypeScript: not run to completion in this clone — `bunx tsc --noEmit` stops on
  `Cannot find type definition file for 'bun-types'` because the fresh clone has
  no `bun install`. This PR modifies zero TypeScript files, so this is stated as
  a gap rather than a pass.

## Critical Self-Review

- Highest-risk behavior: the adopting `session_start`/`session_wrap` pair, which
  deliberately writes into a lane using coordinates the caller did not supply.
  If the adopted scope were read from the wrong object the wrap would land on a
  lane the session never intended to touch.
- Assumptions that could be wrong: that the server's `!hasCompleteExactScope`
  branch returns an existing lane verbatim with no fill. Read directly in
  `server/tools/session-lifecycle.ts:67,155` this session, but never exercised
  against the RUNNING service — every check here is against fake transports.
  Also assumed: the refusal string is stable, since `_is_lane_scope_refusal`
  matches it as a substring; a server-side reword would silently disable
  adoption and restore the original failure.
- Missing/weak tests: no test covers the fallback branch where a wrap runs on a
  lane opened earlier in the same process (`conversation_key` already set) — it
  reuses the requester's scope and is reasoned about, not proven. No test
  exercises adoption through the spool/replay drain. No live-service test at
  all.
- Security/permission risk: this is the PR's central risk and it is the
  RESOLVED BY RULING section above, stated in full rather than buried. The
  namespace boundary is untouched and still fails closed (#654/#662 preserved
  verbatim via `_require_proven_namespace`), but the intra-namespace exact-scope
  assertion is genuinely weakened for the wrap/checkpoint verbs. That was not
  resolved on implementer authority: the operator ruled on 2026-08-17 that
  same-namespace is same-tenant and the server namespace predicate is the
  boundary.
- Migration/deploy risk: low. No schema, no migration, no server change. A
  client older than this change keeps its current behavior, which is the
  failure being fixed, not a new break.
- Downstream client/runtime risk: classified in the Downstream rollout section.
  Nothing was verified against a live downstream runtime.
- Rollback/cleanup concern: revert is the single commit; there is no persisted
  state to unwind. Worth noting that a wrap which already adopted has written
  durable content under the lane's scope, which is correct and is not undone by
  a revert.
- Fixes made before PR: two approaches were built and discarded on evidence
  rather than shipped — unconditional discovery (stranded fresh lanes with NULL
  scope; broke the pinned contract fixture) and a COALESCE-fill fake (unneeded
  once claim-first landed, and reverted to keep the fake diff minimal). The
  ruff import-order error and the one formatting drift were fixed.
- Known residual risk: nothing in this PR has been observed against the RUNNING
  service, so the end-to-end claim the issue actually cares about — a real
  manual wrap producing a real wrap receipt — is UNVERIFIED. Every green result
  above is a fake-transport result.
- SME review-memory update: [x] not applicable because: the lesson is now a recorded operator ruling captured in docs/memory-contract.md and at the retired test site, not a review-swarm pattern.

## Review Gate

- Live Open Brain checks: [x] not applicable because: this lane was verified against fake transports only; the live manual-wrap canary is named above as an outstanding RUNNING check and is not performed in this PR.

## Contract Parity

- Contract parity: [x] runtime-specific because: this PR changes only the Python lifecycle runtime's manual wrap/checkpoint path (the only manual-wrap surface currently in use); clients/ts is deliberately unchanged and its port of the ruled adoption behavior is filed as #736 — no shared contract fixtures encode the old client-side refusal, so none required updating here.

### DIFF
--- a/docs/memory-contract.md
+++ b/docs/memory-contract.md
@@ -77,6 +77,39 @@ through another authorized durable path.
 3. Call `session_wrap` to checkpoint the summary to durable storage
 4. Context compacts. Lane persists.
 
+#### Wrap/checkpoint against a lane the session does not own
+
+Status: WRITTEN 2026-08-17 (#724 item 4), client-side only — the server's
+exact-scope predicate is unchanged.
+
+A manual `wrap`/`checkpoint` is frequently issued against a lane some OTHER
+process opened. The capture hook is the common case: it creates the base lane
+with `agent` set and every other exact-scope coordinate NULL, so a session that
+claims its own full scope trips the server's one-way fill (#646) and the write
+is refused with `Existing lane exact scope does not match session_start
+request`. Two Development sessions lost their wraps to exactly this.
+
+The `openbrain-memory` client now **claims first and adopts on refusal**:
+
+1. `session_start` claims the caller's own exact scope, exactly as before. On
+   success nothing else happens — the ordinary path is unchanged, same single
+   call with the same arguments.
+2. **Only** on that specific scope refusal, the client re-issues
+   `session_start` claiming nothing but `session_key`. That is the server's
+   `!hasCompleteExactScope` branch, which returns the existing lane verbatim
+   instead of attempting a fill.
+3. `session_wrap` then carries the coordinates that came back ON THAT LANE, not
+   the requester's, because the write verb must still satisfy the server's
+   lane-scope predicate.
+
+The adopting `session_start` has no requested exact scope for the response to
+prove, so it is validated on `session_key` + `namespace` — the coordinates the
+caller genuinely owns. The namespace refusals (#654/#662) are enforced
+unchanged on this path. Adoption never re-points the lane, and it is scoped to
+the wrap/checkpoint verbs: `append_session_event` (capture) and
+`agent_context_pack` still prove full exact scope, and no other validation is
+widened.
+
 ### Resuming After Compaction
 1. Call `session_start` → lane + events are immediately available
 2. Search OB for related past work if needed
--- a/docs/memory-contract.md
+++ b/docs/memory-contract.md
@@ -108,7 +108,10 @@ caller genuinely owns. The namespace refusals (#654/#662) are enforced
 unchanged on this path. Adoption never re-points the lane, and it is scoped to
 the wrap/checkpoint verbs: `append_session_event` (capture) and
 `agent_context_pack` still prove full exact scope, and no other validation is
-widened.
+widened. Adoption is therefore NAMESPACE-SCOPED by design: a same-namespace lane
+is the same tenant, the client cannot observe which same-namespace session owns
+a lane, and the boundary that matters is the server's namespace predicate — not
+a client-side check (operator ruling, 2026-08-17).
 
 ### Resuming After Compaction
 1. Call `session_start` → lane + events are immediately available
--- a/python/openbrain-memory/src/openbrain_memory/_runtime_validation.py
+++ b/python/openbrain-memory/src/openbrain_memory/_runtime_validation.py
@@ -64,32 +64,24 @@ def distilled_content(
     return value
 
 
-def validate_started_lane(
-    result: Any,
-    namespace: str,
-    scope: RuntimeScopeCoordinates,
-) -> None:
-    """Require a session-start result to prove the requested exact scope."""
+def _started_lane(result: Any) -> Mapping[str, Any]:
+    """Return the lane object from a session-start result, or refuse."""
     if not isinstance(result, Mapping):
         raise ValueError("session_start result missing lane object")
     lane = result.get("lane")
     if not isinstance(lane, Mapping):
         raise ValueError("session_start result missing lane object")
-    metadata = lane.get("metadata")
-    candidate = {
-        name: lane[name]
-        for name in (
-            "namespace",
-            "session_key",
-            "agent",
-            "source",
-            "channel_id",
-            "thread_id",
-        )
-        if name in lane
-    }
-    if isinstance(metadata, Mapping) and "server_id" in metadata:
-        candidate["server_id"] = metadata["server_id"]
+    return lane
+
+
+def _require_proven_namespace(lane: Mapping[str, Any], namespace: str) -> None:
+    """Refuse a lane whose namespace is absent or is not the configured one.
+
+    Extracted verbatim from ``validate_started_lane`` so the adopted-lane
+    validator (#724 item 4) enforces the SAME refusals rather than a second
+    copy that could drift. The two branches and their reasoning below are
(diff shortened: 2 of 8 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/724-wrap-lane-adoption.sh exit=0 sha=d5046ec8c2e03d59c7e4b7d3da6034f7712f45b8 at=2026-08-18T03:31:43.019Z

```

=== DONE-MEANS #724: wrap-lane adoption in the Python runtime ===

acceptance file: python/openbrain-memory/tests/test_runtime_wrap_lane_adoption.py
runner:          uv run pytest tests/test_runtime_wrap_lane_adoption.py -q
tally:           7 passed, 0 failed, 0 error(s) (runner exit 0)

PASS — 7 tests executed, 0 failures, exit 0

transcript: /Volumes/ThunderBolt/_tmp/open-brain/_scratch/done-means-724-wrap-f877a2156a9b.log.done.done

VERDICT: PASS
```

</details>

rodaddy:
harvested: a5adda8 docs/lane-contract.md round 31 (this PR's lessons — the identical-wire-sequence proof and the pr-body-gate path rule — are in that round; the retire-test ruling is ledger item 33, commit 230cee1)

Issue #733, opened 2026-08-18
verify-lane.ts runs every done-means check with bash — a .test.ts done-means is executed as a shell script and dies exit 2


rodaddy:
Found during the #724 wave merge pass (2026-08-17). PR #728's Done-means line points at `scripts/run-nats-worker-embed-watermark.test.ts` (a bun test, the natural shape for a code lane). `scripts/verify-lane.ts:473` unconditionally runs `spawnSync("bash", [checkPath])`, so bash parsed TypeScript, errored line-by-line, and exited 2 — correctly posting no receipt, but for the wrong reason: harness error, not a failed check (exit grammar says 2/3-class harness errors are not verdicts).

Fix direction (owner's call): dispatch by extension — `.sh` via bash, `.ts`/`.test.ts` via `bun run test:isolated` (or refuse .ts paths loudly with the wrapper recipe named). Until then every code lane must ship a .sh wrapper, which is what #728 is doing to merge.

Refs #724.

PR #728, opened 2026-08-18
feat(nats-worker): surface embed watermark liveness on worker /health (#724)

rodaddy:
## Summary

- The maintenance producer ticked for three days with no consumer draining the embed queue. Raw rows kept arriving, embedded rows stopped, and nothing anywhere COMPARED the two — so every liveness surface stayed green while the corpus quietly stopped being searchable. This adds the comparison to a health surface.
- Adds an optional `embed_watermark` block to the NATS worker's `/health` on 3110, composed by `startHealthServer` (`scripts/run-nats-worker.ts:206-250`). `stale: true` flips `status` to `degraded` with HTTP 503 **even when `nats.availability` is `available`** — a healthy bridge must not be able to hold the endpoint green while the embed lane is days behind, which is exactly the shape of the outage.
- **Shape matched, not invented.** The block follows the convention `maintenance_producer` (#625, `server/transport/health.ts:14-33`) and `capture` (#647, `server/transport/health.ts:35-79`) already established for a background-lane liveness block: a snake_case named block, a boolean `stale` that is THE verdict, raw counters beside it, an explicit `*_threshold_*` naming the bound the verdict used, and a content-free `reason`. `*_age_seconds` is this repo's spelling for an age (`src/operator-doctor.ts:111` `oldest_undistilled_age_seconds`); `quiet_ms` on the producer block is a duration, not an age, so it is not the precedent. Injection matches the optional-input pattern at `server/transport/health.ts:118-138`, and degradation matches `server/transport/health.ts:185-196` where `producer?.stale === true` and `capture?.stale === true` each flip status.
- **Absence is not staleness.** A worker composing no observer emits no `embed_watermark` key and cannot be degraded by one — the ordinary case for a worker that does not own the embed lane, exactly as those two blocks are optional at `server/transport/health.ts:93-105`. Absent means "not my job"; `stale: true` means "my job and I am not doing it".
- **`stale` additionally requires `raw_rows_recent > 0`.** A quiet week produces an old embedded row too, and alarming on an idle corpus is how a check stops being read — the same argument `server/capture/liveness-observer.ts` MIN_SESSIONS_FOR_SILENCE makes for capture.
[...]

Issue #674, opened 2026-08-09
[cutover-blocker] core01 serves src/index.ts, not server/main.ts — capture-health chain (#656/#659/#661) absent from the production tree


rodaddy:
Found by the pre-cutover config audit (2026-08-09, read-only, source at wip head f673299). CUTOVER-BLOCKER.

## The gap
- scripts/local-clone.ts:31 pins the dogfood clone to server/main.ts.
- package.json:8-9: core01 runs `bun run src/index.ts` (start) / scripts/run-two-worker.ts (start:two-worker), and run-two-worker.ts:50 spawns `bun run src/index.ts` per worker.
- `rg -c "createCaptureHealthRuntime|CAPTURE_HEALTH" src/index.ts` → 0. server/main.ts → 2.

So the entire capture-health chain merged tonight — observer, /health capture block, the required-namespace WARN at server/main.ts:273, and by extension the value of #659/#661's launcher key fixes — is absent from the tree core01 actually runs. Setting OPENBRAIN_CAPTURE_HEALTH_* on core01 reads nothing.

## Failure scenario
Set OPENBRAIN_CAPTURE_HEALTH_NAMESPACE on core01, deploy, curl /health, see no capture block, misdiagnose as an allowlist drop (#659's shape) and hunt an allowlist that isn't on this path. The code simply is not there.

## This is a known repo pattern
docs/sme/correctness.md:1822 — "TWO serving trees... a cross-cutting install done on only one leaves the other tree's call sites permanently inert." PR #599 shipped this exact class once.

## Wanted (operator decision required — do NOT pick silently)
Either (a) cut core01 over to server/main.ts as its entrypoint, or (b) port the capture-health runtime composition into src/index.ts. This is an architecture decision about which tree is canonical, not a lane's to choose. Whichever: a done-means check must assert the capture block is live on the core01 serving tree (feature-signal, not revision), and the two-serving-trees SME entry gets a regression note. Truth grammar: WRITTEN in server/, NOT in the core01 serving path.

rodaddy:
[...]

PR #737, opened 2026-08-18
test(nats-worker): red-first proof the LIVE worker composes no watermark observer (#724)

rodaddy:
## What this closes

Issue #724 item 3, second half. PR #728 landed the entire `embed_watermark`
health surface — the block shape, the configurable threshold, degraded/503
composition in the worker's `/health`, the runbook section — and proved every
one of those behaviors by **injecting** an `embedWatermarkHealth` option.
Nothing constructed a real observer. `startNatsWorkerProcess` only forwarded
`options.embedWatermarkHealth`, and the live entrypoint calls it with
`{ env: process.env }` and nothing else, so the deployed worker logged
`embed_watermark_observed: false`, published no block, and **could never
alarm**. The surface existed in the tree and was absent from the serving
process — the #674 class, which this repo had already paid for once at #656
(capture observer wired).

`startNatsWorkerProcess` now default-constructs an observer against its own
pool when the caller supplies none. The option remains an override, so #728's
fixture test (`scripts/run-nats-worker-embed-watermark.test.ts`) is green
unchanged, and the live entrypoint needs no change — that was the point of the
gate.

## It rides the existing registry, not a new table list

The watermark queries every `EMBEDDING_TARGETS` entry whose
`provenance.hasEmbeddedAt` is true — the same registry lane A's embedding
repair drives off:

- `EMBEDDING_TARGETS` — `src/embedding-targets.ts:179`
- `FULL_PROVENANCE` (`hasEmbeddedAt: true`) — `src/embedding-targets.ts:146-150`
- `ENTITY_PROVENANCE` (`hasEmbeddedAt: false`) — `src/embedding-targets.ts:158-162`
- `EMBEDDING_TARGET_NAMES` — `src/embedding-targets.ts:326`
- `thoughts` timestamp columns — `src/db/migrations/001_init.sql:7-19`

`ob_entities` is skipped on its declared provenance rather than queried for an
`embedded_at` that does not physically exist. Table names reaching the SQL come
only from that allowlist — the same rule `getEmbeddingTarget` enforces for
repair. Adding a target to the registry extends the watermark with no change to
the observer.

The observer plugs into the **existing seam**: `startHealthServer` already
accepted `embedWatermarkHealth` (`scripts/run-nats-worker.ts`), and the
construction happens in `startNatsWorkerProcess` after `createDbPool()` and
before the `startHealthServer` call, using the worker's existing pool.

## Cheap and non-blocking: a 30s TTL cache

`startHealthServer`'s fetch handler is **synchronous**, so a probe must not be
able to block on a database round trip or issue one query per embedding target
per request. The accessor returned by `createEmbedWatermarkObserver` therefore
**never queries**: it returns the cached reading and, once that reading is older
than `EMBED_WATERMARK_CACHE_TTL_MS` (**30 seconds**), schedules a background
refresh whose result serves the next probe. One refresh is in flight at a time.
The observer is primed once during startup so the first probe answers with
numbers rather than absence.

Thirty seconds against a default bound of one hour: a reading that stale cannot
move the `stale` verdict in any realistic case, while the endpoint's cost
collapses to at most one aggregate query per 30s regardless of scrape rate.

**Nothing is adjusted silently.** The startup summary now carries
`embed_watermark_cache_ttl_seconds` and `embed_watermark_observer_source`
(`default_pool_observer` or `injected`) alongside the threshold fields #728
already announced, so a reader can tell which observer is in force and how old
the block may be without reading code.

## The failure shape I chose, and why

**A DB query failure composes `stale: false` with a `reason` naming the read
failure**, carrying the numeric fields from the last successful reading, or `-1`
(explicitly "not measured", never an age) when there has never been one. Every
failure is logged as `Open Brain embed watermark read failed`. The refresh
catches its own errors in the background, so a rejected query is never thrown
into the request path.

Three things that shape rules out at once:

- **It does not flip `stale: true`.** A database the observer cannot reach is
  evidence about the *observer*, not about the embed lane. Alarming on it makes
  a transient blip indistinguishable from the three-day outage this block exists
  to catch, and a 503 that fires for unrelated reasons is how a check stops
  being read — the same argument `raw_rows_recent` already carries for an idle
  corpus.
- **It does not go absent, and it does not hold a silent green.** The runbook's
  absence-is-not-staleness contract says a missing block means "not my job"; a
  failed read is *not* "not my job", and publishing nothing would be
  indistinguishable from a pre-#724 worker. The block stays published with a
  reason saying the read failed and `-1` rather than a plausible-looking zero,
  so "measured and fine" is distinguishable from "could not measure" — which a
  silently-omitted block cannot express. This is the runbook contract extended,
  not overridden: absence still cannot degrade, and the deployed worker now
  never goes absent.
- **It cannot crash `/health`.** The synchronous accessor only reads state.

**Empty corpus** (a case the gate deliberately left open): no rows in any
target reports `-1` ages, `raw_rows_recent: 0`, `stale: false`. An empty corpus
is not a stalled lane — there is nothing the lane failed to do — and `stale`
already requires `raw_rows_recent > 0`, so the guard that keeps an idle week
quiet keeps an empty database quiet on the same argument.

## Verification

- Done-means: scripts/done-means/724-watermark-live-observer.sh

The wrapper is bash-shaped because `scripts/verify-lane.ts` runs every Done-means check with `bash` (#733); it invokes `bun run test:isolated` on `scripts/run-nats-worker-live-watermark.pg.test.ts` and exits 0 green / 1 red / 3 harness error.

RED (observed by the author lane on commit `b2f0a83`, before this commit existed): `bash scripts/done-means/724-watermark-live-observer.sh` -> 0 pass, 4 fail, wrapper exit 1, isolated DB `ob_isolated_88305_msy45rtn3u`. Every failure was the missing wiring: `startNatsWorkerProcess must construct a real observer when the caller supplies none` (`Expected: true, Received: false`) and `the DEFAULT worker composition published no embed_watermark block` (`Received: undefined`) on the other three.

GREEN (RUNNING, observed this session on commit `fcd0339`, isolated DB `ob_isolated_74162_msy4erbv6p`, created and dropped by the harness):

```
bash scripts/done-means/724-watermark-live-observer.sh
tally:   4 pass, 0 fail, 0 skip (runner exit 0)
VERDICT: PASS
```

LOAD-BEARING (RUNNING, this session): with the default construction neutered to `if (false)` and nothing else changed, the same runner returns **0 pass, 4 fail**. The wiring is what makes it green, not the test's own harness. The file was restored from a copy held outside the repo; `git diff --stat` confirmed the restored tree.

NO REGRESSION in the touched area (RUNNING, this session):

```
bun run test:isolated scripts/run-nats-worker*.test.ts src/embedding-targets.test.ts
 38 pass
 0 fail
Ran 38 tests across 4 files.
  tests exited 0
```

`bunx tsc --noEmit` exit 0 on the committed tree (re-run after the pre-commit prettier reformat, so the formatted tree is the one that is green). Per `AGENTS.md`, no bare full-suite `bun test` count is cited.

## Critical Self-Review

- Highest-risk behavior: the observer serves a CACHED reading, so `/health` can report a verdict up to 30s stale. Against a one-hour default bound that cannot change the verdict; against a threshold configured below ~60s it could, and the TTL is a module constant rather than env-tunable today. Second-order: the deployed worker can now return 503 for an embed stall where before it structurally could not, because it published no block at all — that is the intended fix, but it is a live behavior change for anything alerting on 3110.
- Assumptions that could be wrong: that every `hasEmbeddedAt` target also has a `created_at` column. True for all seven `FULL_PROVENANCE` targets in the current schema, but it is an implicit coupling the registry does not declare, so a future target with `embedded_at` and no `created_at` would break the SQL at runtime rather than at typecheck. Also PROPOSED: that the newest-across-targets aggregate is the right roll-up — one busy table can currently mask a second table whose own lane is stalled, because `max()` is taken across the UNION rather than per target.
- Missing/weak tests: only `thoughts` carries seeded rows, so the multi-target UNION is exercised structurally but not with rows in a second table; the masking case named above is therefore unasserted. The TTL expiry path and the read-failure composition have no dedicated case (the `now` and `log` seams exist for them). The empty-corpus branch is reasoned in code and docs, not asserted.
- Security/permission risk: none new. Table names are interpolated only from `EMBEDDING_TARGET_NAMES`, the same static allowlist `getEmbeddingTarget` enforces for repair; the threshold is a bound parameter. The block is aggregate counts and ages — no content, no identifiers, no namespace values. The query is deliberately NOT namespace-scoped: it is an operator liveness reading on the worker's own 3110 port, not a tenant read, and scoping it would make the reading blind to exactly the namespaces most likely to stall.
- Migration/deploy risk: no schema change, no migration. Deploy effect is that the worker's `/health` starts publishing a block it did not before and CAN now return 503 where it previously could not. A supervisor configured to restart on 503 would restart the worker for an embed stall, which restarting does not fix; the runbook names the new meaning, wiring a restart policy to it is out of scope here.
- Downstream client/runtime risk: `docs/downstream-rollout.md` classification — NOT APPLICABLE. No MCP tool, schema, protocol, or client-facing contract is touched. 3110 is the worker's own operator endpoint; no rtech-mcps, mcp2cli, or rtech-hermes surface reads it.
- Rollback/cleanup concern: revert is one self-contained commit — one new module, one wiring block, one runbook section. The `embedWatermarkHealth` injection override is unchanged, so reverting cannot break #728's fixture test. Both isolated test databases were dropped by the harness, confirmed by its own teardown lines.
- Fixes made before PR: initial `LoggerLike` shape failed typecheck against the repo logger's `Record<string, unknown>` signature; narrowed. The threshold read was moved ABOVE the observer construction so the bound the verdict is taken against is provably the same value announced in the startup summary, rather than two independent reads of the same env key.
- Known residual risk: the 30s TTL interacts with a sub-minute configured threshold as described, and no test pins that interaction. The per-target masking question is unresolved. This lane makes the deployed worker able to alarm; whether the alarm is wired to anything that acts on it is not established here and should not be assumed from this merge.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no MEDIUM+ review finding surfaced in this lane; the reusable lesson (a health surface proved only by injection is not wired, the #656/#674 class) is already carried by the gate test's own header and belongs in `docs/lane-contract.md` Tightenings, which is the controller's merge-pass harvest step.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: this lane is code, tests, and docs only with no deploy — no running instance carries this change, so there is nothing to smoke and claiming otherwise would be a false receipt.

## Contract Parity

- Contract parity: [x] runtime-specific because: the change is confined to the TypeScript NATS worker's own `/health` payload on 3110 and the registry it reads. There is no cross-runtime contract fixture for that surface — the Python client does not read it, and no MCP tool schema describes it.

## Downstream Rollout

- [x] mcp2cli handoff is complete or not applicable
- [x] rtech-hermes handoff is complete or not applicable

Closes part of #724 (item 3).

### DIFF
--- a/docs/core01-nats-worker-runbook.md
+++ b/docs/core01-nats-worker-runbook.md
@@ -375,7 +375,9 @@ matches their shape (`server/transport/health.ts:14-33`, `:35-79`).
 - **Absence is not staleness.** A worker that composes no embed observer emits
   NO `embed_watermark` key and cannot be degraded by one. Missing block means
   "not my job"; `stale: true` means "my job and I am not doing it". Do not read
-  an absent block as a failure.
+  an absent block as a failure. In the DEPLOYED worker the block is always
+  present (see "Who builds the observer" below), so an absent block there means
+  the process is older than #724 item 3 — not that the lane is idle.
 - `lag_threshold_seconds` reports the bound the verdict was actually taken
   against, so a reading is interpretable without knowing the deployed config.
 
@@ -387,6 +389,47 @@ an hour). Nothing is adjusted silently: the worker's startup log line carries
 `invalid_env_default`), plus `embed_watermark_observed`, so an unset key is
 visible as a default rather than looking configured, and an unusable value
 announces the original it replaced.
+
+**Who builds the observer, and what it reads.** `startNatsWorkerProcess`
+constructs one by default against the worker's own pool
+(`src/embed-watermark-observer.ts`); the `embedWatermarkHealth` option remains
+an override for tests, but its absence no longer means "no observer". The live
+entrypoint passes only `{ env: process.env }` and gets a real one. The startup
+log line reports `embed_watermark_observer_source` as `default_pool_observer`
+or `injected`, so which one is in force is visible without reading code.
+
+The watermark is registry-driven, not a hard-coded table: it queries every
+`EMBEDDING_TARGETS` entry whose `provenance.hasEmbeddedAt` is true
+(`src/embedding-targets.ts`), which is the same registry the embedding repair
+path drives off. `ob_entities` is skipped because it declares
+`ENTITY_PROVENANCE` — it has an `embedding` column and no `embedded_at`. Adding
+a target to the registry extends the watermark with no change here.
+
+**It is cached, and `/health` never waits on Postgres.** The health handler is
+synchronous, so the accessor returns a cached reading and schedules a
+background refresh once that reading is older than
+`EMBED_WATERMARK_CACHE_TTL_MS` — **30 seconds**, announced at startup as
+`embed_watermark_cache_ttl_seconds` beside the threshold. A reading up to 30s
+old cannot move a verdict taken against a one-hour bound, and the cost of the
+endpoint is at most one aggregate query per 30s no matter how often it is
+scraped. The observer is primed once during startup so the first probe answers
+with numbers.
+
+**When the query itself fails.** The block is still published, with
+`stale: false` and a `reason` naming the read failure; the numeric fields carry
+the last successful reading, or `-1` (explicitly "not measured", never an age)
+if there has never been one. A database the observer cannot reach is evidence
+about the OBSERVER, not the embed lane — flipping to `degraded` on it would
+make a transient blip indistinguishable from the three-day outage, and a 503
+that fires for unrelated reasons is how a check stops being read. It equally
+does not go silently absent or hold a stale green: the reason says the read
+failed and every failure is logged as `Open Brain embed watermark read failed`.
+The refresh catches its own errors, so a failing query can never crash
+`/health`.
+
+**Empty corpus.** No rows in any target reports `-1` ages,
+`raw_rows_recent: 0`, and `stale: false`. An empty corpus is not a stalled lane;
+this is the same guard that keeps an idle week quiet.
 
 ```zsh
 curl -fsS http://127.0.0.1:3110/health | jq .embed_watermark
--- a/scripts/done-means/724-watermark-live-observer.sh
+++ b/scripts/done-means/724-watermark-live-observer.sh
@@ -1,0 +1,163 @@
+#!/usr/bin/env bash
+# DONE-MEANS check for issue #724, item 3, SECOND HALF — the acceptance gate,
+# not the fix.
+#
+#   bash scripts/done-means/724-watermark-live-observer.sh
+#
+# WHAT THIS GATES, AND WHY IT IS SEPARATE FROM 724-embed-watermark-health.sh.
+#
+# PR #728 landed the embed-watermark health SURFACE on the NATS worker: the
+# block, the threshold config, degraded/503 composition, the runbook section,
+# all proven by `scripts/run-nats-worker-embed-watermark.test.ts` — which
+# INJECTS an `embedWatermarkHealth` option and asserts what gets composed from
+# it. Correct test, correct scope.
+#
+# It is silent about who constructs the observer in production. The live
+# entrypoint calls `startNatsWorkerProcess({ env: process.env })` and
+# `startNatsWorkerProcess` only FORWARDS `options.embedWatermarkHealth`
+# (scripts/run-nats-worker.ts:333-335), so the deployed worker logs
+# `embed_watermark_observed: false` (:345), publishes no `embed_watermark`
+# block, and can never alarm. The surface is in the tree and absent from the
+# serving process — the #674 class, and the same shape #656 already caught once
+# for the capture observer.
+#
+# The executable acceptance for the DEFAULT composition is a bun test file:
+#
+#   scripts/run-nats-worker-live-watermark.pg.test.ts
+#
+# It starts the worker with NO watermark option (the way the deployed process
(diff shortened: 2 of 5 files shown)


rodaddy:
harvested: c0e4d30 docs/lane-contract.md round 32

rodaddy:
verify-lane receipt: check=scripts/done-means/724-watermark-live-observer.sh exit=0 sha=fcd0339b683e77bdcf209ea4e4290dab9f637e4d at=2026-08-18T03:51:04.504Z

```

=== DONE-MEANS #724 item 3 (second half): the LIVE worker composes a real observer ===

acceptance file: scripts/run-nats-worker-live-watermark.pg.test.ts
runner:          bun run test:isolated scripts/run-nats-worker-live-watermark.pg.test.ts
isolated db:     ob_isolated_9406_msy4kzmnf9
tally:           4 pass, 0 fail, 0 skip (runner exit 0)

PASS — 4 tests executed against ob_isolated_9406_msy4kzmnf9, 0 failures, exit 0

transcript: /Volumes/ThunderBolt/_tmp/open-brain/_scratch/done-means-724-live-observer-cdfa1985c777.log.done

VERDICT: PASS
```

</details>

Issue #757, opened 2026-08-25
Embedding: move the local clone from the MLX qwen server to OMLX (as on the Air) and offload embedding to the 5060 Ti


rodaddy:
Operator ruling relayed from the 2026-08-25 head session (Rico, in chat). Recorded as scope; not started.

## What is RUNNING now on the Mini (checked 2026-08-25 09:55)

```
PID 3444  56.7% CPU  /Users/rico/Library/Application Support/OpenBrainLocal/mlx-embedding-server/.venv/bin/python \
  -m uvicorn qwen_embedding_server.main:app --host 127.0.0.1 --port 8791
```

That is the local clone's embedding provider (`EMBEDDING_BASE_URL` → 127.0.0.1:8791, AGENTS.md "Stack"). It is the process taking the GPU. The python3.14 processes on the box are `codex-openai-shim.py`, idle.

## Ruling

1. The local clone should be on OMLX, not the MLX qwen server. OMLX was already found to be better; the Air is already running it, so its setup is the reference.
2. Embedding should be offloaded to the 5060 Ti box rather than run on the Mini at all.

## Open before anyone builds

- What exactly the Air runs (binary, model, port, launchd unit) — read it from the Air, do not reconstruct.
- Which host carries the 5060 Ti and how the Mini reaches it (hostmap, not a guess).
- `EMBEDDING_WATCHDOG_RESTART_SCRIPT` and the launchd unit for the current server need retiring in the same change, or the watchdog restarts what was just replaced.
- The embedding model must stay `embeddinggemma-300m` at halfvec 768 or every stored vector is invalid; a model change is a re-embed, which is its own lane.

Out of #750's scope. Filed so the ruling is not only in chat.

rodaddy:
Two clarifications from the head session, 2026-08-25.

**Cutover shape (operator, in chat):** switch, then prove it, then keep it. Not a hard switch. Minimum proof before the MLX unit is retired: point a copy of the clone env at TN01, embed a fixed probe set through both endpoints and compare cosine distance (MLX 8-bit vs GGUF Q8_0 are the same weights, not the same bits), run `bun run test:isolated`, then a recall smoke against the dogfood database with the real query path. Only then retire the launchd unit and the watchdog script.

**Which logic model the dream cycle needs resident on TN01 (asked from the rtech-infra side):** none today.
- Light: no model, hard requirement (`docs/dream-design.md:83`).
[...]

Issue #634, opened 2026-08-08
flaky: wall-clock timing assertions fail non-deterministically under CI load (3 different tests in 3 runs)


rodaddy:
## Symptom

CI's `check` job fails intermittently, with a **different** test each run, on branches whose diffs cannot plausibly affect them. Three consecutive runs on PR #629 (a diff touching only `.claude/hooks/`, `scripts/`, and `.claude/settings.json` — zero `src/`, zero `server/`):

| Run | Failure |
|---|---|
| 1 | `0 fail`, `1 error` — unhandled `proc.kill is not a function` at `src/tools/search-all.ts:110` (filed as #632) |
| 2 | `1 fail` — `session lifecycle event bounds (#515)`, `a beforeEach/afterEach hook timed out`, exactly `5000.98ms` |
| 3 | `1 fail` — `ob-backfill sanitize > scans token-dense transcripts without quadratic blowup`, `1274.00ms` |

## Root cause: wall-clock assertions on shared CI hardware

`scripts/ob-backfill.test.ts:259`:

```ts
expect(performance.now() - startedAt).toBeLessThan(1_000);
```

This asserts a **wall-clock duration**, so it measures the runner's current load as much as the code's complexity. It failed at 1274ms — 27% over a threshold that has nothing to do with correctness. The `session-lifecycle.pg` failure is the same shape at the hook level: a 5000ms `beforeEach`/`afterEach` bound, hit exactly, i.e. timed out rather than errored.

## Differential proof (per the #609 standard)

Both failing tests were run against **PR #629's exact head content** in a fresh worktree with a fresh migrated database:

```
server/tools/session-lifecycle.pg.test.ts   4 pass, 0 fail
scripts/ob-backfill.test.ts                22 pass, 0 fail
```

So neither failure is reproducible from the code under test. `main`'s own recent CI runs are green, which is why these are NOT being labelled "pre-existing" — the honest label is **load-sensitive and non-deterministic**.

## Why this is worth fixing rather than re-running

A suite that fails on a different unrelated test each run trains every lane to re-run CI until it is green, which is indistinguishable from the habit that lets a *real* failure through. It also makes red CI uninformative exactly when a controller is deciding whether to merge.

Suggested direction: assert on the property the test actually cares about — complexity/behaviour (e.g. operation counts, or a ratio between two input sizes) rather than absolute elapsed milliseconds on shared hardware.

Issue #632, opened 2026-08-08
flaky: search-all.ts proc.kill() races the qmd subprocess, erroring between tests with 0 fail


rodaddy:
## Symptom

`bun test` in CI intermittently reports `0 fail` **and** `1 error`, exiting 1. The error is not a failed assertion — it is an unhandled rejection between tests:

```
# Unhandled error between tests
-------------------------------
110 |             proc.kill();
                       ^
TypeError: proc.kill is not a function. (In 'proc.kill()', 'proc.kill' is undefined)
      at <anonymous> .../src/tools/search-all.ts:110:18
```

## Evidence it is a flake, not a defect in the triggering PR

Observed on PR #629, whose diff touches **zero** `src/` files (two `.claude/hooks/` gates, one shared parser module, a done-means check, and `.claude/settings.json`).

- Same job, **re-run with no code change**: `check` went from `fail` to `pass`. That is the definition of a race.
- Full suite on clean `origin/main` (9daf96d) locally: `3318 pass, 0 fail, 0 error` — but 520 skipped without `OPENBRAIN_TEST_DATABASE_URL`, so a local green does not reproduce CI's conditions and is NOT offered as the proof. The re-run is.

## Root cause hypothesis (PROPOSED, not verified)

`src/tools/search-all.ts` races a `qmd` subprocess against a `setTimeout`. The timeout callback calls `proc.kill()` unconditionally, but `proc` is undefined at that moment on the path where spawning failed or the handle was never assigned. When the subprocess loses the race under full-suite load, the timer fires against a missing handle and the rejection escapes as an unhandled error between tests.

Suggested fix direction: guard the timeout callback (`proc?.kill()`), and clear the timer when the subprocess settles first so it cannot fire after the race is already decided.

## Why this matters beyond the noise

A suite that exits non-zero with `0 fail` teaches readers to discount red CI, which is how a real failure gets waved through later. It also makes every lane pay a re-run tax on an unrelated diff.

rodaddy:
Recurrence on PR #648 (lane for #647), with the two-runs-disagree evidence the round-8 rule asks for:

- Both runs are the SAME head SHA `be0f6b2ed543c9815dd0b919f8fe3421f446bc14`.
- run 31272342608 → `check` **SUCCESS** (49s)
- run 31272430693 → `check` **FAILURE** (1m11s)

Failing run reports `3674 pass / 35 skip / 0 fail / 1 error` — the signature in this issue's title. The error verbatim:

```
[...]

Issue #608, opened 2026-08-07
search-all: proc.kill TypeError in the qmd-search timeout race — intermittent db-integration failures


rodaddy:
Found by the #592 CI-repair lane, 2026-08-07, while proving its own failures were unrelated.

`src/tools/search-all.ts:110` — when the timeout wins the `Promise.race`, the loser path calls `proc.kill` and throws `TypeError: proc.kill is not a function`. Fires only on timeout, hence intermittent: green on main's last 6 db-integration runs, red on one run of the #592 push, passed on re-run. The branch that observed it was byte-identical to main in that file.

Not fixed in #592 (smallest-correct-change scope: unrelated main-owned code). Suggested done-means: a test that forces the timeout branch and asserts the process handle is terminated without throwing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #759, opened 2026-08-25
feat(local-clone): fleet embedding host opt-in and launchd-safe autostart (#757)

rodaddy:
Closes part of #757.

Two commits, both scoped to local-clone mode. Neither touches the hosted service or any MCP contract.

## What changed

**`b2643ed` — clone-mode opt-in `OPEN_BRAIN_EMBEDDING_HOST_ALLOW`.** Local clone mode refuses a non-loopback `EMBEDDING_BASE_URL` by default, which is correct: a dogfood clone silently embedding against an arbitrary host is the failure the guard exists to prevent. This adds a single named-host escape hatch. The operator sets `OPEN_BRAIN_EMBEDDING_HOST_ALLOW=<host>` and only that exact host is accepted; everything else still refuses as before. The key is added to `CHILD_ENV_KEYS` so the launcher actually passes it through to the spawned process, and `docs/local-clone-dogfood.md` gains the one line describing it.

**`7c3eb1b` — the autostart script runs the launcher as a child instead of `exec`ing it.** Under launchd, `scripts/local-clone.ts --verify` reached the fleet embedding host and passed, then `--start` died every 30 seconds on "Local clone embedding preflight failed". Root cause: macOS Local Network privacy attributes the grant to the launchd job's **main process**. When bash `exec`s bun, bun *becomes* the main process and its outbound request to a LAN address is refused with a misleading `Was there a typo in the url or port?`. Keeping bash as the main process and running bun as its child inherits the grant and the same fetch returns 200. The script now spawns the launcher, forwards `SIGTERM`/`SIGINT` to it, and exits with the child's status so launchd still sees an accurate exit code.

## Why this host

Measured on #757 against 1000 stored rows: `embed-gemma-dense` on the fleet host produces cosine p50 **0.974** against the existing stored vectors — the same embedding space, so no re-embedding is required. `embed-qwen3-4b` scores cosine **~0** (a different space, 2560-wide), which would mean re-embedding all 178,282 vectors plus altering nine `halfvec(768)` columns and their indexes. Gemma-dense now; Qwen is its own issue if anyone wants it.

## Verification

- Done-means: src/local-clone-mode.test.ts
- `bunx tsc --noEmit` → clean, exit 0.
- `bunx oxlint --deny-warnings src/local-clone-mode.ts src/local-clone-mode.test.ts scripts/local-clone.ts` → clean, exit 0.
- `/opt/homebrew/bin/bash -n scripts/local-clone-autostart.sh` → clean, exit 0.
- `bun test src/local-clone-mode.test.ts scripts/local-clone.test.ts` → 69 pass, 1 skip, 0 fail, 105 assertions.
- Pre-push hook ran the full suite on these two commits: 3407 pass, 531 skip, 0 fail across 248 files; gitleaks clean.
- RUNNING, not merely merged: the local clone currently serves this exact code as `ad3b59c` (cherry-picked here as `7c3eb1b`), reports `embedding.connected=true` with embedder `embed-gemma-dense`, and the fleet host's llama-swap log shows `POST /v1/embeddings` arriving from the Mini at ~20 ms. Process tree confirms bash is the launchd main process with bun as its child.

## Critical Self-Review

- Highest-risk behavior: the launchd Local Network attribution. The fix depends on macOS granting the Local Network entitlement to the job's main process (bash) and that grant being inherited by the bun child. This was proven by two throwaway probe plists — `exec`'d bun fails the fetch, bun-as-child returns 200 — but it is an OS privacy behavior, not a documented API contract, so it can change under a macOS update.
- Assumptions that could be wrong: that TCC continues to grant Local Network access to bash. If Apple later gates bash the same way it gates bun, the clone will start failing the embedding preflight again under launchd with the identical misleading error. That is the first thing to check if this regresses; the probe plists reproduce it in about ten seconds.
- Missing/weak tests: the allowlist logic has RED-first cases in `src/local-clone-mode.test.ts` (exact-host accept, non-allowlisted host still refused, unset variable preserves the old refusal). Signal forwarding in the autostart script is proven with a dummy child rather than against launchd itself — there is no unit-testable seam for the TCC behavior, so that half rests on the live probes and the running deployment, not on an automated test.
- Security/permission risk: this widens a deliberate network guard, which is the whole point of the change, so it is opt-in and single-valued. The variable names exactly one host; it is not a list, not a CIDR, and not a wildcard, and an unset variable leaves the previous loopback-only behavior untouched. It applies only in local clone mode and has no effect on the hosted service.
- Migration/deploy risk: no schema change and no re-embedding — `embed-gemma-dense` lands in the same vector space as the stored data (cos p50 0.974), so existing rows stay valid. The env change is the deploy. Already RUNNING on the local clone as `ad3b59c`, with `EMBEDDING_BASE_URL` and `EMBEDDING_MODEL` switched and backups written beside each env file.
- Downstream client/runtime risk: none identified. No MCP tool, schema, transport, or Python client path is touched, so `docs/downstream-rollout.md` does not apply — this is a local-clone launcher and a clone-mode env guard only.
- Rollback/cleanup concern: restore `local-clone.env` from `local-clone.env.bak-20260825-1103` and redeploy `8f1f2a8`. That reverts both the host and the launcher in one step. The local MLX embedding unit is still loaded and running, so the previous provider is available immediately without a reinstall.
- Fixes made before PR: the autostart commit itself is the fix found after the first deploy attempt — the allowlist alone passed `--verify` and still could not `--start` under launchd, which is what exposed the `exec` problem. `OPEN_BRAIN_EMBEDDING_HOST_ALLOW` was also added to `CHILD_ENV_KEYS` after the first run showed the launcher was not forwarding it to the child.
- Known residual risk: the fleet host is now a hard dependency of local embedding. If it is down, local tests that embed fail with network errors rather than skipping, and the clone's preflight refuses to start. Separately, the fleet host carries an ungrouped `embed-qwen3-4b` test entry whose implicit exclusive group would evict the pinned concurrent group if anyone requested it; harmless today, but it should be grouped or removed before anything relies on it.
- SME review-memory update: [x] not applicable because: no review finding was produced — this is an author-side gate on a two-commit local-clone change, and the launchd/TCC lesson is recorded in `_DOCS/_handoff/2026-08-25-embedding-tn01-cutover.md` rather than as a reviewer lane pattern.

## Review Gate

- Live Open Brain checks: [x] not applicable because: no Open Brain tool, schema, or namespace behavior changes here; the live evidence that matters is the clone's own `/health` embedding status, cited under Verification.

### DIFF
--- a/docs/local-clone-dogfood.md
+++ b/docs/local-clone-dogfood.md
@@ -141,6 +141,8 @@ DB_POOL_MAX=10
 OPEN_BRAIN_RUN_MIGRATIONS=0
 
 EMBEDDING_BASE_URL=http://127.0.0.1:<local-embedding-port>/v1
+# Optional (#757): name ONE non-loopback embedding host to allow, exactly.
+# OPEN_BRAIN_EMBEDDING_HOST_ALLOW=10.71.1.11
 EMBEDDING_API_KEY=<local-only-key-if-required>
 EMBEDDING_MODEL=embeddinggemma-300m-8bit
 EMBEDDING_DIMENSIONS=768
--- a/scripts/local-clone-autostart.sh
+++ b/scripts/local-clone-autostart.sh
@@ -95,4 +95,28 @@ if ! "${BUN_BIN}" run scripts/local-clone.ts --verify; then
 fi
 
 log "preflight verified, starting launcher"
-exec "${BUN_BIN}" run scripts/local-clone.ts --start
+
+# Hand off to the launcher as a CHILD of this shell, not via exec.
+#
+# Why (measured 2026-08-25, #757): macOS Local Network privacy attributes the
+# LAN grant to the launchd job's MAIN process. With `exec`, bun became that
+# process and its fetch to the fleet embedder (10.71.1.11) was refused with
+# "Was there a typo in the url or port?", while the same fetch from a bun child
+# of this bash succeeded. That is exactly why the --verify preflight above
+# passed under launchd and --start then died on "embedding preflight failed"
+# every 30 seconds. Keeping bash as the main process keeps the grant. TERM and
+# INT are forwarded so `launchctl kickstart -k` still stops the launcher, which
+# in turn stops its server child.
+"${BUN_BIN}" run scripts/local-clone.ts --start &
+launcher_pid=$!
+forward_signal() { kill -s "$1" "${launcher_pid}" 2>/dev/null || true; }
+trap 'forward_signal TERM' TERM
+trap 'forward_signal INT' INT
+status=0
+wait "${launcher_pid}" || status=$?
+# A trapped signal returns from `wait` before the child has exited; keep
+# waiting so the exit status below is the launcher's, not the signal's.
+while kill -0 "${launcher_pid}" 2>/dev/null; do
+  wait "${launcher_pid}" || status=$?
+done
+exit "${status}"
--- a/scripts/local-clone.ts
+++ b/scripts/local-clone.ts
@@ -78,6 +78,9 @@ const CHILD_ENV_KEYS = [
   "OPENBRAIN_CAPTURE_HEALTH_WINDOW_MINUTES",
   "OPENBRAIN_RECOVERY_WAL_PATH",
   "OPEN_BRAIN_BIND_HOST",
+  // #757: the fleet-embedding opt-in read by `validateLocalCloneMode`; without
+  // it here the key is dropped before the child sees it (the #755 trap).
+  "OPEN_BRAIN_EMBEDDING_HOST_ALLOW",
   "OPEN_BRAIN_MAINTENANCE_ENABLED",
   "OPEN_BRAIN_RUN_MIGRATIONS",
   "OPENBRAIN_TRANSPORT",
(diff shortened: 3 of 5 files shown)


rodaddy:
**CI red is environmental, not this diff.** Three separate causes, none in files #759 touches:

1. Run 32865230602 (`check`, ct108): the Test step ran `bun test v1.4.0` although setup-bun restored 1.3.13 → 61 failures, all `src/observability/observability.test.ts` + `server/observability/langfuse-tracing.test.ts`. Filed as a new issue (bun 1.4.0 runner leak).
2. Run 32864979708 (`check`, ct106, bun 1.3.13): **3804 pass / 0 fail**, exit 1 from the known `src/tools/search-all.ts:110` `proc.kill` race between tests (#608, #632).
3. Main `6065390` re-run today (ct106, bun 1.3.13): 3796 pass / 1 fail, the #634 `slow audit writes` timing flake.

Local: `bun run test:isolated` on `7c3eb1b` → 3806 pass / 35 skip / 0 fail (RUNNING).

Failed jobs on both PR runs re-requested for a fresh sample.

rodaddy:
Re-runs green: both `check` jobs (ct107, ct108) resolved `bun test v1.3.13`, 3806 pass / 0 fail. PR is CLEAN / MERGEABLE. The 1.4.0 leak is tracked in #760.

PR #771, opened 2026-08-26
chore(standards): land the config-guarded pre-commit lint hook and oxlint dependency

rodaddy:
Refs #750.

`.oxlintrc.json` landed on this branch in `2a89cf2` and enforces nothing on
`main`, because the thing that reads it lives somewhere else. The hook's oxlint
step is config-guarded — it prints `SKIPPED: no .oxlintrc.json` when the file is
absent and lints staged content when it is present — and that hook exists only
on `sprint/standards-fmt`. So `main` currently has the rule set and no reader
for it: armed and disconnected.

This connects the two.

**The hook is carried unchanged.** `_githooks/pre-commit` is byte-identical at
`b2d4252` (where it landed) and `1e1e006` (the sprint branch tip) — `git diff
b2d4252 1e1e006 -- _githooks/pre-commit` is empty — so what is here is the
reviewed file, not a re-derivation of it.

**oxlint becomes a real dependency.** `package.json` gains `lint` and
`lint:fix`, and `oxlint@^1.76.0` moves into `devDependencies` so CI's three
`bun install --frozen-lockfile` jobs resolve it from the lock instead of
letting `bunx` fetch an unpinned version at hook time.

**Deliberately not included.** No prettier, no `.prettierrc.json`, no `format`
scripts: that step is guarded the same way and correctly reports `SKIPPED`,
which is the honest state until a formatting lane lands its own config. No
`.github/workflows/ci.yml` change — this arms the local pre-commit gate only,
and moving CI is a separate decision with its own blast radius.
`.oxlintrc.json` is untouched.

**The hooksPath defect this surfaced.** `core.hooksPath` selects exactly one
directory; it does not chain. An earlier probe on this work committed clean and
read as a passing gate — because hooksPath still resolved to the operator's
global hooks directory, so the hook under test was never the one git executed.
The done-means check now asserts that path before trusting any result, since a
gate git does not run cannot refuse anything.

## Verification

- Done-means: scripts/done-means/750-precommit-lint-gate-fires.sh
- RED, and it is the point of the check. The commit was REFUSED, exit 1, by the
  oxlint step naming each rule:

      [pre-commit] oxlint (staged content)
        ...__done_means_750_probe__.ts:1:10: error typescript(no-explicit-any)
        ...__done_means_750_probe__.ts:2:1: error eslint(no-console)
        ...__done_means_750_probe__.ts:4:8: error eslint(max-lines-per-function):
          has too many lines (125). Maximum allowed is 100.
[...]

PR #776, opened 2026-08-26
fix(verify-lane): install the PR head's dependencies before running the check

### LINKED ISSUE #775: verify-lane installs deps at origin/main, then verifies the PR head: any PR that adds a dependency cannot get a receipt
## Claim state

MEASURED (twice, 2026-08-25) against PR #771 at head `2d67702`. WRITTEN fix proposed; nothing merged.

## Defect

`scripts/verify-lane.ts` stands its verification worktree up in two moves that
disagree about which commit's dependencies are installed:

1. Step 3 (`scripts/verify-lane.ts:365-378`) shells out to
   `scripts/lane-bootstrap.ts`, which cuts the worktree from `origin/main` and
   runs `bun install --frozen-lockfile` there
   (`scripts/lane-bootstrap.ts:452`). The installed `node_modules` therefore
   matches **origin/main's** `package.json` / `bun.lock`.
2. Step 3's tail (`scripts/verify-lane.ts:~410`) does `git fetch origin <sha>`
   followed by `git reset --hard <headSha>`, moving the worktree's *source* to
   the PR head — but `node_modules` is never reinstalled.
3. Step 4 runs the done-means check against that mismatched tree.

Any PR that ADDS a dependency therefore cannot get a receipt: the check runs
with origin/main's `node_modules`, the new tool is absent, and verify-lane
throws before posting.

## Measurement

PR #771 adds `oxlint` as a devDependency. Its done-means check fails under
verify-lane with:

```
MISSING TOOL: .../node_modules/.bin/oxlint
```

and no receipt is posted. Running a second `bun install --frozen-lockfile` in
that same verification worktree installs the 2 missing packages, after which
the identical check passes. Full log:
`{temp_workspace}/open-brain/_scratch/verify-771-run2.log` (operator machine).

## Why it matters

The merge gate (`.claude/hooks/merge-gate.ts`) requires a verify-lane receipt.
A whole class of PR — any that adds or bumps a dependency — is structurally
unable to produce one, and the failure presents as the lane's check being
broken rather than as a verify-lane environment defect.

## Proposed fix

After the hard reset to the PR head and before `run-check`, compare
`package.json` and `bun.lock` between the bootstrap ref (`origin/main`) and the
head. If they differ, run `bun install --frozen-lockfile` in the worktree with
the script's existing fail-loud shape. Print a `deps-at-head` step line on
BOTH paths — reinstalled or unchanged — per the repo's "nothing is adjusted
silently" rule.

rodaddy:
Closes #775.

`scripts/verify-lane.ts` stood its verification worktree up in two moves that
disagreed about which commit's dependencies were installed. Step 3 shells out
to `scripts/lane-bootstrap.ts`, which cuts the worktree from `origin/main` and
runs `bun install --frozen-lockfile` there, so `node_modules` describes
origin/main. The same step's tail then does `git fetch origin <sha>` plus
`git reset --hard <headSha>`, moving the SOURCE onto the PR head — and never
reinstalls. Step 4 ran the done-means check against that mismatched tree.

Any PR that adds or bumps a dependency was therefore structurally unable to
earn a receipt, and the failure presented as the lane's check being broken
rather than as a verify-lane environment defect. That matters because
`.claude/hooks/merge-gate.ts` requires a verify-lane receipt to merge.

**Measured twice** (2026-08-25) against PR #771 at head `2d67702`, which adds
`oxlint` as a devDependency: the check died on `MISSING TOOL:
.../node_modules/.bin/oxlint` and NO receipt was posted, while a second
`bun install --frozen-lockfile` in that same verification worktree installed
the 2 missing packages and the identical check passed.

**The fix.** After the hard reset and before `run-check`, compare
`package.json` and `bun.lock` between `origin/main` and the head, and reinstall
when they differ. `git diff --quiet` answers 0 for "same" and 1 for "differs",
so it cannot go through `capture()`, which throws on any non-zero status; every
OTHER exit code (a bad ref, a missing object) fails loudly rather than being
read as "unchanged", which would reintroduce the original defect whenever git
could not compare.

Both paths print a `deps-at-head` line — `reinstalled: lockfile differs from
origin/main` or `unchanged: lockfile matches origin/main` — per AGENTS.md
"Nothing is adjusted silently": a no-op that prints nothing is
indistinguishable from a step that never ran.

`main()` now sits behind an `import.meta.main` guard, the convention already
used by 10 other scripts in `scripts/`, so the decision function is importable
by a test without the script trying to verify a PR on import.

The change is confined to `scripts/verify-lane.ts`. `lane-bootstrap.ts` needed
no hook: it correctly installs deps for the ref it cuts, and the mismatch is
created entirely by verify-lane's own later reset, so the reconciliation
belongs at that boundary.

## Verification

- Done-means: scripts/done-means/775-verify-lane-installs-head-deps.sh
- **RED first, against the unfixed script.** `git checkout origin/main --
  scripts/verify-lane.ts` then running the check exits 1 with clauses 1, 2 and
  4 FAIL (`deps-at-head:none`; the unit suite exits 1; no fail-closed guard).
  Clause 3 stays PASS by design — it exercises git's own discrimination on a
  fixture repository, not the fix. Restoring the fix returns exit 0 with all
  four PASS.
- `bun test scripts` -> 384 pass, 35 skip, 0 fail (419 tests across 32 files).
- `bun test scripts/verify-lane.test.ts` -> 6 pass, 0 fail.
- `bash scripts/done-means/verifier-agent-grounded.sh` -> exit 0, still green.
- `bash -n scripts/done-means/775-verify-lane-installs-head-deps.sh` -> clean.
- Second commit re-verified after hardening: check exits 0 on the fix and 1 on
  the unfixed script, `bun test scripts` unchanged at 384 pass / 0 fail, and
  the worktree HEAD is unchanged across a full run of both fixtures.

The new check's four clauses assert the fix's load-bearing properties rather
than its prose: that the step is ordered strictly between the reset and
run-check AND actually invokes `bun install` (position is the whole fix — a
reinstall before the reset would install origin/main's deps twice and change
nothing); that both branches announce themselves; that the comparison really
fires on a dependency-adding head and really skips an unrelated one; and that
an unusable comparison fails closed.

## Critical Self-Review

- Highest-risk behavior: an unconditional reinstall would slow every
  verification run and could mask a genuinely broken lockfile. Mitigated by
  gating on a real manifest diff and by keeping `--frozen-lockfile`, which
  still refuses a lockfile that does not match `package.json`.
- Assumptions that could be wrong: that `origin/main` is always the ref
  lane-bootstrap cut from. It is, at `scripts/lane-bootstrap.ts`, but the two
  now name it independently. Named once here as `BOOTSTRAP_REF` with a comment
  saying why; if lane-bootstrap ever gains a configurable base ref, this
  constant is the single place that must follow, and a wrong base ref produces
  a spurious reinstall (slow, safe) rather than a skipped one.
- Missing/weak tests: the unit suite covers the decision function and the git
  signal feeding it, not a full end-to-end verify-lane run against a live PR —
  that needs `gh`, a real PR, and a bootstrap, which is not something a test
  should stand up. The done-means check closes the gap by asserting the call
  site's ordering and the presence of the install itself.
- Security/permission risk: none. No namespace, auth, or predicate logic
  touched. `bun install --frozen-lockfile` cannot resolve a package outside the
  committed lockfile, which is the property that makes running it against
  untrusted PR head content acceptable here — the same install lane-bootstrap
  already runs.
- Migration/deploy risk: none. No schema, no migration, no runtime code path.
- Downstream client/runtime risk: none. `scripts/verify-lane.ts` is a
  controller-side tool; it ships in no client and no server surface.
- Rollback/cleanup concern: revert is a single commit. The done-means fixture
  repositories are written under the repo-relative, gitignored `_scratch/verify-lane-deps/` (the convention of `src/operator-doctor.test.ts:29`; the earlier temp-workspace default was Mac-only and failed CI with EACCES on `/Volumes`) and left in place —
  teardown is printed, never executed, matching ledger item 15.
- Fixes made before PR: **the fixtures could escape into the surrounding
  repository, and did.** Both the unit test and the done-means check built a
  throwaway git repo and then ran git against it with a bare `cwd`/`-C`. When
  the fixture is not a repository, git walks UP to the first enclosing one and
  operates there. One `mkdtempSync` fixture came back without a `.git` — `git
  init` did not take, and nothing checked — so its `base` and `docs only`
  commits were authored into the surrounding worktree by "verify-lane test
  <test@example.invalid>", deleting every tracked file in it. The branch ref
  was moved back to the real commit and the tree recovered intact with no work
  lost, and the remote never carried those commits. Closed in the second
  commit rather than merely noted: every git call in both fixtures now pins
  `--git-dir` and `--work-tree` (which makes the upward walk impossible), and
  `git init` — which predates the repo and so cannot use the pinned helper —
  now has its success asserted on both the exit code and the resulting `.git`
  before any other git call runs. Reviewers should weigh this: a check that can
  mutate the repository it is verifying is a worse defect than the one it
  gates, and it is the reason clause 3 builds its fixture the way it does.
- Fixes made before PR: the check originally used `mktemp -d`, which AGENTS.md
  forbids outright because it resolves to a sandbox-local `$TMPDIR` that no
  other process can see. Replaced with an explicit pid+epoch path under the repo's own
  gitignored `_scratch/`, git-dir and work-tree pinned on every call. The test fixture also had to commit on a branch named
  `fixture` rather than `main`, because the global protected-branch hook
  refuses commits to `main` even in a throwaway repository.
- Known residual risk: the fix reinstalls only when `package.json` or
  `bun.lock` differ. A PR that changes neither but still needs a different
  dependency tree — a patch applied by a postinstall script, say — would still
  verify against origin/main's `node_modules`. No such case is known in this
  repo, and widening the trigger to "always reinstall" trades a real slowdown
  on every run for a hypothetical.
- SME review-memory update: [x] not applicable because: this is a fix to
  controller tooling found by running it, not a review-swarm finding or a
  missed review pattern.

## Review Gate

- Live Open Brain checks: [x] not applicable because: this changes only the
  controller-side verification script; no server, tool schema, namespace, or
  client-facing surface is touched, and no Open Brain endpoint is involved.

### DIFF
--- a/docs/lane-contract.md
+++ b/docs/lane-contract.md
@@ -87,6 +87,57 @@ Every lane, no exceptions:
 
 Newest first. Every entry: what changed, and the observation that forced it.
 
+### 2026-08-26 (round 34) — harvest of the verify-lane deps-at-head lane (PR #776, #775)
+
+- **verify-lane installed dependencies at `origin/main` and then verified the
+  PR head, so the first dependency-adding PR could not get a receipt.**
+  `lane-bootstrap.ts` cuts the verification worktree from `origin/main` and
+  runs `bun install --frozen-lockfile` THERE; verify-lane then hard-resets the
+  same worktree onto the PR head and never reinstalled. The SOURCE moved and
+  `node_modules` did not. Measured twice on #771, which adds `oxlint`: the
+  check died on `MISSING TOOL: .../node_modules/.bin/oxlint` and NO receipt was
+  posted, while a second install in that same worktree made the identical check
+  pass. Fixed by reinstalling when `package.json`/`bun.lock` differ between the
+  bootstrap ref and the head. `git diff --quiet` answers 0 for "same" and 1 for
+  "differs", so it cannot go through `capture()` (which throws on any non-zero
+  status); every OTHER status fails closed rather than being read as
+  "unchanged", because reading a broken comparison as "unchanged" reintroduces
+  the original defect exactly when git cannot compare.
+- **A test that shells out to git MUST pin `--git-dir` and `--work-tree` on
+  every call, and MUST assert `git init` succeeded before any other git call.**
+  A bare `-C <fixture>` does not fail when the fixture is not a repository — it
+  walks UP to the first enclosing one and operates there. One `mkdtempSync`
+  fixture came back without a `.git` (init did not take, nothing checked), so
+  the fixture's own `base` and `docs only` commits were authored into the
+  surrounding worktree as `verify-lane test <test@example.invalid>`, deleting
+  every tracked file in it, and the fixture's `git config` calls wrote
+  `core.bare=true`, the test identity, and `commit.gpgsign=false` into the
+  CLONE's `.git/config` (the controller restored it). Proven directly rather
+  than argued: against a non-repo directory, `git -C <d> rev-parse HEAD`
+  returns the enclosing worktree's HEAD while the pinned form returns
+  `fatal: not a git repository`. Recovery used `git update-ref` plus a mixed
+  reset, because the guard correctly refuses `git reset --hard` — the guard was
+  right and the workaround is the supported path, not a bypass.
+- **Test scratch is repo-relative `_scratch/`, never a hardcoded Mac path.**
+  Both the test and the done-means check defaulted to
+  `/Volumes/ThunderBolt/_tmp` when `OPENBRAIN_TEMP_WORKSPACE`/`DEV_TMP` were
+  unset. That is the operator's machine, so CI failed in both `check` and
+  `db-integration` with `EACCES: permission denied, mkdir '/Volumes'` on the
+  Linux runner. The repo already had the right shape
+  (`src/operator-doctor.test.ts:29`): a `_scratch/<name>/` directory resolved
+  from the repo root, already excluded by `.gitignore:119`. A fixture that
+  requires one machine's volume layout gates nothing in CI, which is where the
+  check most needs to run. Keeping the fixture inside the repo also bounds the
+  escape above — the worst a stray git call can reach is a directory the repo
+  already ignores.
+- **A check that can mutate the repository it verifies is a worse defect than
+  the one it gates.** Proving such a check is not "it exited 0": it is
+  `git config --list --local` and `git status --short` of the ENCLOSING repo,
+  before and after, plus confirming HEAD is unchanged and no fixture path shows
+  up as untracked. The escape above passed its own clauses green while
+  corrupting the branch it ran on; only inspecting the enclosing repo caught
+  it.
+
 ### 2026-08-26 (round 33) — harvest of the plans lane (PR #772) and the L1 lint-gate lane (PR #771)
 
 - **A docs-only PR still carries an executable done-means.** The merge gate
--- a/docs/sme/entries/2026-08-26-a-test-fixture-that-shells-out-to-git-escapes-into-the-enclosing-repository.md
+++ b/docs/sme/entries/2026-08-26-a-test-fixture-that-shells-out-to-git-escapes-into-the-enclosing-repository.md
@@ -1,0 +1,75 @@
+---
+lane: gotcha-agent
+order: 75
+---
+## [2026-08-26] A test fixture that shells out to git escapes into the enclosing repository
+
+**Provenance:** PR #776, issue #775. Severity: HIGH. Status: open.
+**Scope:** `scripts/*.test.ts`, `scripts/done-means/*.sh`, any test or check
+that builds a throwaway git repository.
+
+A done-means check and its unit test both built a scratch git repository and
+then drove it with `spawnSync("git", args, { cwd: fixture })` and
+`git -C "$FIXTURE"`. Both spellings look scoped and neither is: when the target
+directory is NOT a repository, git does not fail — it walks UP the directory
+tree to the first enclosing `.git` and operates on THAT.
+
+One `mkdtempSync` fixture came back without a `.git` directory. `git init` had
+not taken and nothing checked its exit code, so every subsequent call in the
+fixture's setup resolved to the surrounding worktree. The fixture's own commits
+— `base` and `docs only` — were authored into the branch under development as
+`verify-lane test <test@example.invalid>`, and because the fixture commits a
+tree containing only `package.json`, `bun.lock`, and `README.md`, the commit
+DELETED every other tracked file in the worktree. The setup's `git config`
+calls landed in the enclosing repository too, writing `core.bare=true`, the
+test identity, and `commit.gpgsign=false` into the clone's `.git/config`.
+
+The check reported all clauses PASS while doing this. Its assertions were about
+git's diff behavior on the fixture, which remained true; nothing in the check
+looked at the repository it was running inside. The corruption was found only
+because the branch was pushed and the diff was inspected.
+
+Two properties make this dangerous rather than merely annoying. First, the
+failure is silent and delayed: a fixture that fails to initialize produces no
+error, and the damage surfaces later as "why did my branch delete everything".
+Second, moving fixtures into the repo's own `_scratch/` (correct for CI
+portability — a hardcoded `/Volumes/...` default failed the Linux runner with
+`EACCES: permission denied, mkdir '/Volumes'`) puts the fixture strictly INSIDE
+the repository, which makes the upward walk shorter and guaranteed to find a
+target.
+
+The fix is two mechanical rules, both fail-closed. Pin `--git-dir <fixture>/.git`
+and `--work-tree <fixture>` on every call, which makes the walk impossible —
+git errors instead of finding a parent. And assert `git init` succeeded on BOTH
+its exit code and the resulting `.git` before any other git call runs, since an
+unchecked init is what enables the escape at all. Verified directly against a
+non-repo directory nested in the worktree:
+
+```
+git --git-dir <d>/.git --work-tree <d> rev-parse HEAD
+  -> fatal: not a git repository
+git -C <d> rev-parse HEAD
+  -> 365ae45          # the ENCLOSING worktree's HEAD
+```
+
+### Review Questions
+
+- **Does any test or check run git with a bare `cwd` or `-C` against a
+  directory it created?** That is the escape. Require `--git-dir` and
+  `--work-tree` on every invocation; `-C` alone scopes nothing when the target
+  is not a repository.
+- **Is `git init`'s success asserted, or assumed?** An init that silently did
+  not take converts every following call into an operation on the enclosing
+  repo. Check the exit code AND that `.git` exists before proceeding.
+- **Does the check inspect the repository it is running INSIDE?** A check that
+  can mutate its host must prove it did not: `git status --short`,
+  `git rev-parse HEAD`, and `git config --list --local` of the enclosing repo,
+  before and after. Green clauses are not evidence here — this one was green
+  throughout.
+- **Where does the fixture root come from?** Repo-relative `_scratch/`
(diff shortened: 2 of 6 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/775-verify-lane-installs-head-deps.sh exit=0 sha=57d4792c10ad709f65dbf746fb885c2df26d8c80 at=2026-08-26T09:17:02.835Z

```

#775 — verify-lane installs the PR head's dependencies

  CLAUSE 1 (step ordered between reset and check): PASS
    deps-at-head at byte 22492, strictly between checkout-head (20486) and run-check (23431), and it runs bun install --frozen-lockfile through capture() (fail-loud)
  CLAUSE 2 (both paths announce themselves):       PASS
    scripts/verify-lane.test.ts green (6 pass) — covers the reinstalled path, the unchanged path, and the head SHA appearing on both
  CLAUSE 3 (comparison fires on a real head):      PASS
    dependency-adding head -> git diff --quiet exit 1 (reinstall), unrelated-change head -> exit 0 (skip); fixture at /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-776-20507mt9vqzfa/_scratch/verify-lane-deps/fixture-20665-1787735822
  CLAUSE 4 (fail-closed on an unusable compare):   PASS
    verify-lane rejects any git diff status outside {0,1} rather than reading it as 'unchanged'

  Fixture repositories are LEFT IN PLACE under /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-776-20507mt9vqzfa/_scratch/verify-lane-deps —
  teardown is printed, never executed. Remove them yourself when done.

PASS — verify-lane reinstalls at the PR head and says which path it took.
```

</details>

rodaddy:
Review round 1 (Light tier, one read-only lane, Codex via codex-rescue) bound to head 365ae45. Verdict: **BLOCK**. P2 :92 (Mac-only fallback) and P2 :133 (env-derived scratch path) were already fixed by 0bccaf8 (repo-relative `_scratch/`). Fixer lane dispatched for the three P1s and the remaining P2s in one commit; deterministic gates re-run on the new head, no re-review (lanes run once per head).

```text
RESULTS
P1|scripts/verify-lane.ts:516 -- compares mutable `origin/main`, not the bootstrap commit -- advancing `origin/main` after bootstrap can return status 0 and skip installing dependencies absent from the bootstrapped tree.
P1|scripts/done-means/775-verify-lane-installs-head-deps.sh:157 -- unpinned `git init` honors inherited `GIT_DIR`/`GIT_WORK_TREE` -- with both set to a victim repo, it reinitializes the victim before line 159 detects the missing fixture `.git`.
P1|scripts/verify-lane.test.ts:106 -- same unpinned `git init` can reinitialize an ambient victim repository before the fixture assertion fails.
P2|scripts/verify-lane.test.ts:156 -- `diffStatus` uses bare `git diff` -- deleting `.git` or setting Git routing variables lets it read an enclosing/foreign repository.
P2|scripts/done-means/775-verify-lane-installs-head-deps.sh:133 -- environment-derived scratch paths are neither confined nor symlink-safe -- an external path or symlink redirects fixture commits outside the mandated scratch root.
P2|scripts/verify-lane.test.ts:92 -- Mac-only fallback fails on Linux when temp variables are unset -- `mkdirSync` resolves `/Volumes/ThunderBolt/_tmp/open-brain/_scratch` and cannot create it.
P2|scripts/done-means/775-verify-lane-installs-head-deps.sh:224 -- no cleanup runs on success or failure -- every created fixture remains; no `mv` cleanup exists.
P2|scripts/verify-lane.test.ts:149 -- `afterAll` deliberately leaves the scratch repository -- each unit-test run leaks a Git fixture.
P2|scripts/done-means/775-verify-lane-installs-head-deps.sh:120 -- does not run a RED pre-fix variant or invoke `verify-lane`’s `main()` -- a mutant that ignores `diff.status` can pass all four clauses while skipping installation.
verified: `git diff --name-status/stat origin/main...365ae45` -> 3 files, 536 insertions, 19 deletions; only requested production/test/check files changed; explanatory scaffolding is concentrated in check lines 7-66 and test lines 1-12.
verified: exact-head binding -> `365ae454fdb8ddbabce841fb13bf739c51c964d4` resolves; checkout advanced afterward to `57d4792`, so execution was not run against the moving checkout.
verified: `git config --list --local` and `git status --short` before/after, review checkout -> config diff empty; status diff empty.
verified: `git config --list --local` and `git status --short` before/after, `/Volumes/ThunderBolt/_tmp/open-brain/_scratch/clone-20260825` -> config diff empty; status diff empty.
verified: `bun test scripts/verify-lane.test.ts` and `bash scripts/done-means/775-verify-lane-installs-head-deps.sh` -> not run; exact-head archive creation was blocked by the read-only filesystem.
verified: temp inventory -> `mkdtempSync` only at `scripts/verify-lane.test.ts:98`, resolving under `${OPENBRAIN_TEMP_WORKSPACE}/open-brain/_scratch`; no executable `mktemp`, `/tmp`, or `os.tmpdir()` use; check line 135 is comment-only.
verified: import/caller scan -> only the unit test imports `verify-lane.ts`; hooks/verifier invoke it as a CLI; `import.meta.main` prevents import-time `main()` execution.
verified: frozen-lockfile path -> `capture("deps-at-head", "bun", ["install", "--frozen-lockfile"])` fails closed with a named error and no receipt.
verified: post-init check Git calls -> helper calls pin `--git-dir` and `--work-tree`; no `GIT_DIR` or `GIT_CEILING_DIRECTORIES` environment fence exists.
BLOCK
```

rodaddy:
verify-lane receipt: check=scripts/done-means/775-verify-lane-installs-head-deps.sh exit=0 sha=357e7003133a5e490284a97f1c6ce86d02da5a4c at=2026-08-26T09:46:25.609Z

```

#775 — verify-lane installs the PR head's dependencies

  CLAUSE 1 (mutant killed: suite fails on flipped guard): PASS
    mutant (fail-closed guard disabled) makes the unit suite exit 1 — the suite can kill it. Mutant kept at /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-776-66621mt9wsr94/_scratch/verify-lane-deps/mutation-66702-1787737584/verify-lane.ts
  CLAUSE 2 (both paths announce themselves):       PASS
    scripts/verify-lane.test.ts green (9 pass) — covers the reinstalled path, the unchanged path, and the head SHA appearing on both
  CLAUSE 3 (comparison fires on a real head):      PASS
    dependency-adding head -> git diff --quiet exit 1 (reinstall), unrelated-change head -> exit 0 (skip); fixture at /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-776-66621mt9wsr94/_scratch/verify-lane-deps/fixture-66702-1787737584
  CLAUSE 4 (fail-closed on an unusable compare):   PASS
    decideDepsAtHead() takes the raw git exit code (diffStatus: number | null) and throws on anything outside {0,1} rather than reading it as 'unchanged'
  CLAUSE 5 (negative control: victim untouched):   PASS
    victim untouched with GIT_DIR aimed at it — config sha 69820b6ef287e00657d01a6913aa14fc1c5180ac and HEAD 8efc78468b5a71b9bd70d4415bfbef7a1fd327e3 identical before/after; victim at /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-776-66621mt9wsr94/_scratch/verify-lane-deps/victim-66702-1787737585

PASS — verify-lane reinstalls at the PR head and says which path it took.

  scratch archived to: /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-776-66621mt9wsr94/_scratch/_archive/verify-lane-deps/66702-1787737584
```

</details>

Issue #764, opened 2026-08-26
python-capture: TestAgainstARealTranscript reads whatever is largest in the runner's ~/.claude, fails when that is an API-error transcript


rodaddy:
Agent-filed (open-brain session 2026-08-25, head Claude Fable 5).

## What happened

PR #763 (docs-only) run 32922625834, job `python-capture` on `open-brain-ct106`: `tests/test_capture_transcript.py::TestAgainstARealTranscript::test_operator_turns_are_found_and_tool_results_are_not` failed with `every user record was treated as an operator turn; tool results are leaking into the lane`. The push-event run of the same SHA passed.

## Why

`python/openbrain/tests/test_capture_transcript.py:690-730`: the test `rglob("*.jsonl")`s `LIVE_TRANSCRIPT_DIR` on the machine running the tests and takes the LARGEST transcript. On ct106 that file is a transcript whose every turn is `API Error: Unable to connect to API (ConnectionRefused)` (the log quotes it), so there are no tool results to distinguish and the `< user_records` assertion fails. The test's verdict depends on what happens to be on the runner's disk, not on the code under test.

## Ask

Pin the input: ship a fixture transcript (redacted) under `tests/fixtures/` and run the real-transcript variant only when an env var opts in (same pattern as the live canaries). Related flake family: #702.

Claim state: RUNNING for the log and the test source; PROPOSED for the fix.

rodaddy:
Hit again 2026-08-27 on PR #873 run 33033898375 (TS-only diff): test_operator_turns_are_found_and_tool_results_are_not, 'assert 1 < 1', the largest transcript was an API Error transcript.

rodaddy:
Hit again on PR #875 run 33034489960, same assertion (test_capture_transcript.py:725), same API-error transcript uuid 6e56dc8e.

rodaddy:
Hit again on PR #883 run 33038376440 (test-only diff), same assertion.

rodaddy:
Hit again on PR #885 run 33038453875 (test-only diff), same assertion.

rodaddy:
Hit again on PR #886 run 33038608484 (test-only diff), same assertion. Four hits this session (#873 #875 #883 #885 #886).

PR #777, opened 2026-08-26
fix(search-all): clear the qmd timeout when the process exits first

### LINKED ISSUE #608: search-all: proc.kill TypeError in the qmd-search timeout race — intermittent db-integration failures

### LINKED ISSUE #632: flaky: search-all.ts proc.kill() races the qmd subprocess, erroring between tests with 0 fail

rodaddy:
Closes #608. Closes #632.

`searchQmdInternal` races the qmd subprocess against a `setTimeout` that calls
`proc.kill()`. Nothing cancelled the timer when the subprocess won, so the
callback stayed armed and fired later, against an already-settled race --
typically after the test that started it had ended.

Under a mocked `Bun.spawn`, the handle carries only the fields the happy path
reads (`stdout`/`stderr`/`exited`) and has no `kill`, so the late callback threw
`TypeError: proc.kill is not a function`. Bun cannot attribute an error raised
between tests to any test, so it printed `# Unhandled error between tests` and
exited 1 while reporting `0 fail`.

That is the shape that made this expensive rather than merely noisy: a red run
that looks green in the test list. It read as a flake, it cost every lane a
re-run tax on diffs touching nothing in `src/`, and a suite that exits non-zero
with `0 fail` teaches readers to discount red CI.

**The fix, at the owning boundary.** The timer handle is held outside the race
and `clearTimeout`d once the race settles, so the loser cannot fire at all. The
`proc.kill` call is additionally guarded with a `typeof` check, since a partial
spawn handle legitimately may not provide one -- belt and braces, because the
cancellation is the actual fix.

**Applied to both twins.** `src/tools/search-all.ts:101` and
`server/tools/search-all.ts:156` ship byte-equivalent copies of this race, and
the server copy has no test file of its own to notice a half-fix. Both are
corrected in the same commit, and done-means clause d asserts structurally that
neither can regress alone.

**Nothing else shares this shape.** `rg -n 'setTimeout\(' src server --type ts`
over non-test files returns 23 sites; the only other `Promise.race`-with-kill is
`src/rem-terra-transport.ts:134`, which already clears its timer in a `finally`
block (line 152). It is correct and untouched.

## Verification

- Done-means: scripts/done-means/608-qmd-timeout-does-not-fire-after-exit.sh
  -> VERDICT: PASS, 5 clauses passed, 0 failed.
- RED-proven against the pre-fix tree with the regression test present: the
  done-means script exits 1 with clauses a, c and d failing. Clause a reports
  `Expected: 1 / Received: 0` on `cleared` -- the timer was never cancelled --
  and firing the captured callback throws the verbatim
  `TypeError: proc.kill is not a function. (In 'proc.kill()', 'proc.kill' is
  undefined)` from #632. Clause z passes on both sides, as a control must.
- The gate checks the run OUTPUT for `Unhandled error between tests`, not the
  exit status, because the two disagree by design in exactly this failure mode:
  on the broken code the test list can report every test passing.
- Clause c is a mutation: it removes the `clearTimeout` and requires the test
  file to go red. The mutation is killed. It is gated on a green baseline, so
  an already-red tree reports INCONCLUSIVE rather than banking a free pass.
- `bun run test:isolated src/tools/__tests__/get-stats.test.ts
  src/tools/__tests__/search-all.test.ts` -> 42 pass, 0 fail across 2 files,
  exit 0, zero `Unhandled error` lines. `get-stats.test.ts` is the file CI
  reported the leak against.
- `bun test src/tools/__tests__/search-all.test.ts` -> 38 pass, 0 fail.

## Critical Self-Review

- Highest-risk behavior: the timeout path itself. If `clearTimeout` were placed
  where the timeout branch also reached it, a genuinely hung qmd would no longer
  be killed. It sits after the race settles and before the `timedOut` branch is
  read, so the kill still runs on a real timeout; done-means clause z asserts
  both twins still bound the subprocess with `QMD_TIMEOUT_MS`, and would fail if
  a future change "fixed" the race by deleting the timeout.
- Assumptions that could be wrong: that the mocked-spawn handle is the only way
  `proc.kill` goes missing. The `typeof` guard does not depend on that being
  true -- it holds for any partial handle -- and the `clearTimeout` removes the
  late-fire path regardless of what the handle carries.
- Missing/weak tests: the server twin has no test file, so its fix is covered
  structurally (clause d greps both files) rather than behaviorally. Writing a
  server-side harness is larger than this defect warrants; the structural check
  is what stops a half-fix landing.
- Security/permission risk: none. No auth, namespace, predicate, or SQL touched.
- Downstream client/runtime risk: none. `search_all`'s tool contract, arguments,
  and result shape are unchanged; this only affects whether an already-lost
  timer fires.
- Rollback/cleanup concern: none. Single commit, revertible in isolation.
- Fixes made before PR: the regression's first form asserted
  `expect(cleared).toBeGreaterThan(0)`, which passed on the BROKEN code --
  unrelated timers in the MCP client path were being counted. Tightened to track
  only the qmd timer by its `QMD_TIMEOUT_MS` delay and assert `toBe(1)`, then
  re-RED-proven. A test that passes pre-fix proves nothing.
- Known residual risk: the regression stubs `globalThis.setTimeout` to capture
  the callback rather than waiting out the real 10s delay, and identifies the
  qmd timer by matching that delay value. If `QMD_TIMEOUT_MS` changes, the
  mirrored constant in the test must change with it; the test asserts
  `captured.length === 1`, so a drifted value fails loudly rather than silently
  measuring nothing.
- SME review-memory update: [x] not applicable because: #608 and #632 already
  record this pattern with the evidence, and the done-means check is the durable
  artifact that prevents its return.

## Review Gate

- Live Open Brain checks: [x] not applicable because: this changes only timer
  cancellation inside a subprocess race; no tool contract, transport, schema, or
  client-facing surface moves, so there is nothing for a live canary to observe.

### DIFF
--- a/scripts/done-means/608-qmd-timeout-does-not-fire-after-exit.sh
+++ b/scripts/done-means/608-qmd-timeout-does-not-fire-after-exit.sh
@@ -1,0 +1,206 @@
+#!/usr/bin/env bash
+# DONE-MEANS check for the qmd-search timeout race (Closes #608, #632).
+#
+#   bash scripts/done-means/608-qmd-timeout-does-not-fire-after-exit.sh
+#
+# WHAT THIS GATES, AND WHY IT IS NOT JUST "RUN THE TEST"
+#
+# src/tools/search-all.ts raced a qmd subprocess against a setTimeout that
+# called proc.kill(). Nothing cleared the timer when the subprocess won, so the
+# callback fired later — typically after the test that started it had already
+# ended — and called .kill() on a mocked spawn handle that has none. Bun does
+# not attribute that to any test. It reports:
+#
+#   # Unhandled error between tests
+#   TypeError: proc.kill is not a function
+#
+# and exits 1 with `0 fail`. That is the whole reason a bare `bun test` exit
+# code is insufficient evidence here: on the BROKEN code the run can report
+# every test passing. Clause b therefore greps the run output for the unhandled
+# marker rather than trusting the status, because the two disagree by design in
+# exactly the failure mode this gate exists to catch.
+#
+# The observed cost: `db-integration` and `check` red on PRs #771, #772, #776,
+# and on 2 of main's own last 6 runs, each time on a diff that touched nothing
+# in src/. Re-running the same job with no code change flipped it to green,
+# which is the definition of a race and is what taught readers to discount red.
+#
+# Clauses:
+#   a — the search-all test file is green.
+#   b — the run emitted no `Unhandled error between tests` (the real gate).
+#   c — the regression test still DISCRIMINATES: with the clearTimeout removed
+#       from src/tools/search-all.ts, the file must go red. A mutation that
+#       survives means the test is decoration.
+#   d — the server/ twin carries the same cancellation, so the fix cannot be
+#       half-applied. Both files ship the identical race; fixing one leaves the
+#       other armed.
+#   z — CONTROL: both twins still arm a timeout at all. Deleting the timeout
+#       would trivially satisfy a/b/c while removing the protection the timer
+#       exists to provide. Must PASS on both sides of the fix.
+#
+# RED PROOF (pre-fix tree, origin/main @ 1e1e006, regression test present):
+#   clause a FAILS — `expect(cleared).toBe(1)` gets `Received: 0`, i.e. the
+#   timer was never cancelled; and firing the captured callback throws
+#   `TypeError: proc.kill is not a function. (In 'proc.kill()', 'proc.kill' is
+#   undefined)` — the verbatim #632 error. Clause d FAILS (no clearTimeout in
+#   either twin). Clause z PASSES pre-fix, as a control must.
+#
+# Output is content-free: counts, names, and pass/fail states only.
+
+set -uo pipefail
+
+# Resolve the repo from THIS script's own location, so the check cannot reach
+# across trees when verify-lane runs it from a fresh worktree at the PR head.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
+cd "${REPO_ROOT}" || exit 1
+
+TEST_FILE="src/tools/__tests__/search-all.test.ts"
+SRC_TWIN="src/tools/search-all.ts"
+SERVER_TWIN="server/tools/search-all.ts"
+REGRESSION_NAME="clears the qmd timeout when the subprocess finishes first"
+UNHANDLED_MARKER="Unhandled error between tests"
+
+# Scratch is gitignored; a run leaves the working tree clean.
+WORK_DIR="${REPO_ROOT}/_scratch/608"
+mkdir -p "${WORK_DIR}"
+
+fail_count=0
+pass_count=0
+
+pass() { printf '  PASS  %s\n' "$1"; pass_count=$((pass_count + 1)); }
+fail() { printf '  FAIL  %s\n' "$1"; fail_count=$((fail_count + 1)); }
+
+printf 'done-means: #608/#632 qmd timeout does not fire after the process exits\n'
+printf '  repo: %s\n\n' "${REPO_ROOT}"
+
+for f in "${TEST_FILE}" "${SRC_TWIN}" "${SERVER_TWIN}"; do
+  if [[ ! -f "${f}" ]]; then
+    printf 'FATAL: %s not found — wrong tree or the file was renamed.\n' "${f}"
+    exit 1
+  fi
+done
+
+# ---------------------------------------------------------------------------
+# clause a — the search-all test file is green, and the regression is present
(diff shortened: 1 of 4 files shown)


rodaddy:
**`python-capture` red is an unrelated pre-existing flake, not this diff.**

The two CI runs on this same commit (699378a) disagree: `python-capture`
failed in run 32951628088 and passed in run 32951761395. Same tree, both
outcomes — a race, by definition.

Cause, from the failing log:
`tests/test_capture_transcript.py::TestAgainstARealTranscript::test_operator_turns_are_found_and_tool_results_are_not`
asserts `len(result.turns) < user_records` against whatever real transcript
happens to be on the runner's disk. It picked one holding a single record —
`API Error: Unable to connect to API (ConnectionRefused)` — so the assertion
became `assert 1 < 1`. It is a property of the runner's transcript directory,
not of the code under test.

Pre-existing on `main`: 4 of the last 8 `ci.yml` runs on `main` failed
(96978a8, 0692b63, 5ebf407, 99a293b), against 4 successes, with no Python
change between them.

This PR's diff touches zero Python files:

```
scripts/done-means/608-qmd-timeout-does-not-fire-after-exit.sh
server/tools/search-all.ts
src/tools/__tests__/search-all.test.ts
src/tools/search-all.ts
```

Worth its own issue — it is the same family as the defect this PR fixes (a
test whose green depends on ambient state), but fixing it here would take
this change outside its boundary.

`db-integration` and `check` — the two jobs #608/#632 were actually
breaking — both pass.

rodaddy:
verify-lane receipt: check=scripts/done-means/608-qmd-timeout-does-not-fire-after-exit.sh exit=0 sha=699378aebc7ebb6a32c4e739f925cd0b811d59f5 at=2026-08-26T09:16:32.238Z

```
done-means: #608/#632 qmd timeout does not fire after the process exits
  repo: /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-777-10321mt9vqbjr

  PASS  a: src/tools/__tests__/search-all.test.ts green and the #608/#632 regression is present by name
  PASS  b: no 'Unhandled error between tests' in the run output
  PASS  c: green baseline + mutation (clearTimeout removed) goes RED — it discriminates
  PASS  d: both twins clear the timer and guard proc.kill
  PASS  z: both twins still bound the qmd subprocess with QMD_TIMEOUT_MS

  passed: 5   failed: 0
  VERDICT: PASS
```

</details>

rodaddy:
No new lessons: the one mid-lane correction (a regression test that passed on the broken code because it counted unrelated timers, tightened to identify the qmd timer by its delay and re-RED-proven) is the standing red-first rule in docs/lane-contract.md, not a new tightening. The python-capture red on this head is the ambient-transcript flake already tracked in #764.

Issue #614, opened 2026-08-07
Full bun test run against the shared dogfood DB is untrustworthy as a gate


rodaddy:
## Problem

Running the full \`bun test\` suite against the shared dogfood database produces unstable failure counts on UNMODIFIED code (observed 49 → 43 → 40 across consecutive runs, 2026-08-06; same pattern as #498). Meanwhile the #609 lane's full-suite runs against a **freshly created** database were clean: \`origin/main\` 3701 pass / 0 fail.

## Implication

- A red full-suite run on the dogfood DB is not evidence a change is broken; a green one is not proof it is safe. Lanes currently compensate with differential failure-set comparison, which is labor and judgment per lane.
- The clean fresh-DB baseline shows the tests themselves are fine — the shared mutable state is the contaminant.

## Ask

Make the trustworthy path the default: test harness creates/points at an isolated database (as CI's \`db-integration\` job and the #609 lane's worktree runs already do) instead of the live dogfood DB. Documented in AGENTS.md once the mechanism lands.

Related: #498 (chunk-write flake, same class), #613 (one concrete leakage mechanism).

Issue #498, opened 2026-08-02
CI: 'check' job flakes on chunk-write live-Postgres tests (shared runner Postgres contention)


rodaddy:
## Symptom

The `check` job in `.github/workflows/ci.yml` intermittently fails on `main`
with timeouts in `src/chunk-write.pg.test.ts`, and occasionally a foreign-key
violation:

```
(fail) parent + chunk entry storage (live Postgres) > stores the parent's COMPLETE text, however long [6471.97ms]
  ^ this test timed out after 5000ms.
error: insert or update on table "thoughts" violates foreign key constraint "thoughts_parent_id_fkey"
```

## Evidence that this is infrastructure, not code

Observed during the v0.9.0 release (commit `de4429b`, run 30766200661):

- **Non-deterministic.** The first run failed two tests; the rerun of the same
  commit failed a *different* single test (`gives every chunk its own vector and
  a link to the parent`). A real defect does not move.
- **The commit contains no TypeScript.** `de4429b` changed six files: four
  version manifests, one Python test literal, and `uv.lock`. It cannot cause a
  TS chunk-storage failure, and `src/chunk-write.pg.test.ts` is untouched by it.
- **Passes locally on the exact merged tip.** `bun test src/chunk-write.pg.test.ts`
  at `de4429b` → 6 pass / 0 fail in **1.61s**, against the 5-6 second timeouts
  seen on the runner. Full suite at the same commit → 3358 pass / 35 skip /
  0 fail.
- **Not new.** The two preceding `main` commits (`2e0a2b9`, `9cdf443`) also
  failed a DB job, before the release branch existed.
- **The isolated job passes.** `db-integration` — which `ci.yml` describes as the
  "authoritative DB-backed integration gate" and which stands up its own
  ephemeral pgvector container on a unique port — passed on the same commit
  every time.

## Root cause (proposed)

`check` runs against whatever Postgres happens to be listening on
`127.0.0.1:5432` on the self-hosted runner, which `ci.yml` acknowledges in its
own comment on the `db-integration` job. `check` gives each run its own
*database* (`open_brain_test_${{ github.run_id }}_...`) but all concurrent runs
share one *server*, so parallel jobs contend for connections and shared
resources. Under contention, 5000ms assertions blow their budget, and
interleaved writes surface as FK violations.

`db-integration` does not have this problem because it owns a dedicated
container on a dynamically-chosen free port.

## Impact

Red `main` that does not mean anything, which is the expensive kind: it trains
[...]

PR #778, opened 2026-08-26
feat(config): type every server/ env read in the validated schema (L2a)

rodaddy:
Lane L2a of the `server/` hardening ladder
(`_plans/server-hardening-ladder.md`, rung L2 "Composition root"). Schema half
only -- no consumer is rewired here.

`_plans/463-server-rewrite-charter.md:108` gives `server/config/` ownership of
ALL env parsing and startup validation, "replaces 18 scattered process.env
readers with injected config", and `:119` states that domain code must not
import `process.env`. Twenty-one of the twenty-three variables non-test
`server/` code reads had no field in `environmentSchema` at all, so no consumer
could be handed a value even in principle. The rewiring L2b/L2c does was
blocked on the type not existing, which is what this fixes.

Every reader named below still reads `process.env` today, deliberately. Both
paths exist until the injection and the `no-process-env` lint rule land; that
intermediate state is the point of splitting the rung.

New groups on `ServerConfig`, each mirroring one existing reader:

| group | reader |
|---|---|
| `search` | `server/tools/search-engine.ts:148-149` |
| `fts` | `server/tools/fts-config.ts:111` |
| `qmd` | `server/tools/search-all.ts:119` |
| `recovery` | `server/tools/realtime-stores.ts:46`, `server/main.ts:153` |
| `sharedNamespaceNames` | `server/tools/shared-namespace.ts:37-52` |
| `tracing` | `server/observability/langfuse-tracing.ts:599-604` |
| `captureHealth` | `server/capture/liveness-observer.ts:404-408` |
| `http` | `server/main.ts:313,362` |

The NATS group is deliberately NOT duplicated. `server/config/nats.ts` already
parses `OPENBRAIN_TRANSPORT` and every `OPENBRAIN_NATS_*` key from the
environment object the schema's `.catchall` hands `parseNatsConfig`, so those
names already reach a consumer through validated config. Adding fields for them
would be a second opinion on a question already answered.

**The bar is start-equivalence, not tidiness.** Several of these readers
deliberately ignore an unusable value and fall back rather than failing:
`searchEmbeddingTimeoutMs` answers 3000 for `abc`, `envPositiveInteger` answers
5 for `-1`, `resolveQmdPath` treats `QMD_PATH=` as unset. Tightening any of
those into a hard rejection here would turn a deployment that boots today into
`server_configuration_invalid` at startup -- a behavior change smuggled in
under a typing change, and the same class of failure as the 2026-08-02
`EMBEDDING_API_KEY=` launchd throttle-loop this schema already carries a
preprocessor for. The fallbacks are reproduced exactly, including
`Number.parseInt`'s tolerance of a trailing suffix (`3000ms` stays 3000), which
`z.coerce.number()` would have rejected.

Fields live in `server/config/env-groups.ts` rather than inline because
`buildConfig` is already near the 100-code-lines-per-function lint limit;
`environmentSchema` composes them with one named spread.

`_plans/l2-composition-root-inventory.md` is the read-only scan the scope came
from, landed in the first commit so L2b/L2c read it from the repo rather than a
scratch directory.

## Verification

- Done-means: scripts/done-means/750-l2a-config-covers-every-env-read.sh
- RED, final script against `origin/main`'s `server/config.ts` with
  `env-groups.ts` moved away -> `FAIL: 23 env names read in server/; 21 are not
  declared in environmentSchema`, exit 1, all 21 listed by name.
- GREEN on this branch -> `PASS: all 23 env names read in server/ are declared
  in environmentSchema (server/config.ts)`, exit 0.
- RED for the tests: deleting the `OPENBRAIN_CAPTURE_HEALTH_WINDOW_MINUTES`
  field and its group read -> `(fail) extended env - captureHealth > takes an
  explicit namespace, window, and refresh`, 64 pass / 1 fail. Restored -> 65
  pass / 0 fail.
- `bun run test:isolated server/config.test.ts` -> 65 pass, 0 fail, 102
  expect() calls.
- `bun run test:isolated server/main.pg.test.ts server/application/` -> 35
  pass, 0 fail, confirming the widened `ServerConfig` broke no existing
  composer.
- `bunx oxlint --deny-warnings server/config.ts server/config/env-groups.ts
  server/config.test.ts` -> exit 0.
- The `_githooks` pre-commit gate (gitleaks + oxlint on staged content + whole
  project tsc) passed on both commits. It caught a pre-existing
  `max-nested-callbacks` violation at `server/config.test.ts:271` that the
  whole-file lint surfaces once the file is staged; fixed in place by naming
  the inner thunk.

## Critical Self-Review

- Highest-risk behavior: a parser here disagreeing with the reader it mirrors,
  so a value silently changes when L2b swaps the consumer over. Mitigated by
  reproducing each reader's exact semantics rather than a tidier version, and
  by asserting the shrug-off cases (`abc`, `0`, `-1`, blank) as fallbacks
  rather than rejections.
- Assumptions that could be wrong: that the inventory's section A/B is
  complete. Three read shapes the literal `process.env.X` scan cannot see --
  `env[CONST]` in the capture observer, injected-`env` parameters in
  fts/tracing/qmd, and `process.env[name]` in `shared-namespace.ts` -- are
  enumerated explicitly in the done-means script, so a name reached that way is
  covered but a NEW one of that shape would not be caught automatically. The
  script's header says so.
- Missing/weak tests: the group builders are covered through
  `parseServerConfig`, not called directly, so a builder reachable only from an
  unexercised branch would show as covered. Coverage reports 100% funcs/lines
  on `env-groups.ts`, which is a floor and not the argument; the RED run is.
- Security/permission risk: `tracing.secretKey` and the shared-namespace names
  are now on the config object. The secret is typed as a string like the
  existing `EMBEDDING_API_KEY`, and no logging or health emitter reads it in
  this change; nothing is added to any response payload. No namespace
  predicate, auth check, or role mapping is touched.
- Migration/deploy risk: none. No schema migration, and no variable's meaning,
  name, or default changes -- an environment that starts the server today
  produces the same values.
- Downstream client/runtime risk: none. No MCP tool, transport schema, REST
  route, or Python client shape changes; `ServerConfig` is internal to
  `server/` and gains fields rather than changing any.
- Rollback/cleanup concern: revert is two commits, one of which is a doc copy.
  No consumer depends on the new fields yet, which is what makes it cheap.
- Fixes made before PR: the done-means script initially named a variable
  `GROUPS`, which is a bash builtin holding the user's group IDs, so the
  filename expanded to a numeric group list and the check failed for the wrong
  reason. Renamed with a comment. The `max-nested-callbacks` fix above.
- Known residual risk: `environmentSchema` now spreads a second module, so the
  done-means check follows that spread. If a later change adds a SECOND spread,
  the check only follows the one named `extendedEnvironmentFields` -- it would
  fail loudly rather than pass, since those names would appear undeclared.
- SME review-memory update: [x] not applicable because: no review finding or
  missed pattern surfaced here; the lane's only new lesson is the bash-builtin
  name collision, recorded below.

## Review Gate

- Live Open Brain checks: [x] not applicable because: this change adds typed
  fields to an internal config object and rewires no consumer, so no running
  behavior differs and there is nothing for a live check to observe.

## Harvest

Tightenings candidate -- a done-means script assigning `GROUPS=` gets the
user's numeric group IDs instead, because `GROUPS` is a bash builtin. The
failure surfaces as a nonsense path in an error message ("but 20 is missing"),
which reads like a broken check rather than a name collision. Any
`scripts/done-means/*.sh` naming a variable after a common word should avoid
bash's builtin set (`GROUPS`, `SECONDS`, `LINES`, `COLUMNS`, `PWD`, `RANDOM`,
`UID`, `PIPESTATUS`).

### DIFF
--- a/_plans/l2-composition-root-inventory.md
+++ b/_plans/l2-composition-root-inventory.md
@@ -1,0 +1,58 @@
+# L2 composition-root inventory
+
+Status: WRITTEN 2026-08-26, measured by a read-only Codex lane against origin/main 96978a8 (`server/` identical on chore/oxlint-enforcement). Feeds `_plans/server-hardening-ladder.md` L2.
+
+```text
+RESULTS
+
+A.
+server/observability/langfuse-tracing.ts:599 OPENBRAIN_TRACING_ENDPOINT/PUBLIC_KEY/SECRET_KEY/ENABLED/MASKING_ENABLED -- whole-env default; trim, enabled only with flag `1` and all coordinates, masking defaults on -- NO -- `readMcpTracingConfig`
+server/tools/shared-namespace.ts:37 SHARED_NAMESPACE_CANONICAL/OPENBRAIN_SHARED_NAMESPACE/SHARED_NAMESPACE_PHYSICAL/SHARED_NAMESPACE_LEGACY/OPENBRAIN_LEGACY_SHARED_NAMESPACE -- trim; first non-empty else caller default -- SHARED_NAMESPACE_CANONICAL; others NO -- `envString` via `sharedNamespaceConfig`
+server/tools/shared-namespace.ts:45 OPENBRAIN_LEGACY_SHARED_FALLBACK -- trim; true for `1/true/yes/on`, default false -- NO -- `envBoolean` via `sharedNamespaceConfig`
+server/tools/shared-namespace.ts:52 OPENBRAIN_SHARED_FALLBACK_MIN_RESULTS -- base-10 positive integer, fallback 5 -- NO -- `envPositiveInteger` via `sharedNamespaceConfig`
+server/tools/realtime-stores.ts:46 OPENBRAIN_RECOVERY_WAL_PATH -- `?? null`, no trim/parse -- NO -- `recoveryWalStoreFor`
+server/tools/search-all.ts:119 QMD_PATH -- whole-env default; trim blank to undefined -- NO -- `resolveQmdPath`, called by `registerSearchAllTool`
+server/tools/search-engine.ts:148 OPENBRAIN_SEARCH_EMBEDDING_TIMEOUT_MS -- preferred value; base-10 integer >=1 else 3000 -- NO -- `searchEmbeddingTimeoutMs`
+server/tools/search-engine.ts:149 SEARCH_EMBEDDING_TIMEOUT_MS -- legacy fallback; same parse/default -- NO -- `searchEmbeddingTimeoutMs`
+server/tools/fts-config.ts:111 OPENBRAIN_FTS_CONFIG -- whole-env default; trim, allowlisted language, fallback english -- NO -- `corpusFtsConfig`
+server/tools/fts-config.ts:128 OPENBRAIN_FTS_CONFIG -- whole-env default; request override else corpus default -- NO -- `requestFtsConfig`
+server/tools/operator-doctor.ts:62 OPENBRAIN_TRANSPORT and OPENBRAIN_NATS_* -- whole-env default; trim/lowercase, HTTP/auth/fallback defaults and local-URL guard -- NO -- `readNatsRuntimeBoundary`
+Scan-only non-read hits: `server/application/nats.ts:19` and `server/config/nats.ts:6` are comments.
+
+B.
+main.ts:210 -- config: `loadServerConfig()` calls `parseServerConfig` at `server/config.ts:350`; skipped when `options.config` is supplied.
+main.ts:211 -- logger: `createLogger(config.logging)`.
+main.ts:218-222 -- nonempty-token guard, then pool: `createDatabase(config.database, logger)`.
+main.ts:228 -- tracing: `createTracingRuntime()`; no config passed, so it rereads tracing env before migrations/NATS.
+main.ts:232-241 -- migrations from `config.database.migrationsDirectory`, unless skipped by option.
+main.ts:243-268 -- NATS boundary, health, token map, and bridge from `config.nats`.
+main.ts:74-77,170,247,287,348 -- embedder is not constructed; imported `generateEmbedding*` functions are passed through, with module-level env state already loaded.
+main.ts:278-282 -- capture-health runtime receives raw `process.env`.
+main.ts:284-360 -- auth, REST surface, application, MCP factory, and routes are composed.
+main.ts:362-371 -- port/bind host are resolved, then the HTTP listener opens.
+Direct env departures: `:153` WAL path, `:279` capture env, `:313` `ALLOWED_ORIGINS`, `:362` `PORT`, `:364` `OPEN_BRAIN_BIND_HOST`; audit at `:159` also defaults to `src/audit-log.ts` env parsing.
+src/embedding.ts:6-10,35-36 -- import-time embedding configuration -- imported by `server/main.ts:74-77`, `src/maintenance-bootstrap.ts:41`, `server/tools/ingest-conversation-facts.ts:37`, and `src/operator-doctor.ts:7-13`.
+src/logger.ts:26,115-125,175-189,201-205,583 -- module logger/config/file-sink singleton -- imported directly by `server/observability/langfuse-tracing.ts:100` and indirectly by source modules.
+server/tools/realtime-stores.ts:26-27,34-46 -- lazy module-scoped fallback stores, not import-time; main injects stores at `main.ts:150-155`.
+
+C.
+server/config.ts:115-123 -- database fields.
+server/config.ts:124-127 -- logging/service fields.
+server/config.ts:128-135 -- transport/session fields.
+server/config.ts:136-137 -- embedding fields.
+server/config.ts:138 -- shared-namespace field.
+server/config.ts:139-144 -- auth-token fields.
+server/config.ts:145-146 -- catchall optional strings, including NATS/maintenance keys not explicitly schema fields.
+Yes, beyond main: production call is `server/config.ts:350`; direct test calls are in `server/config.test.ts`, `server/application/{shadow-application.test.ts,sdk-protocol.pg.test.ts}`, and `server/main.pg.test.ts`. `main.ts` calls `loadServerConfig`, not `parseServerConfig` directly.
+
+D.
+Import-time blockers: `src/embedding.ts` and `src/logger.ts` above.
+Runtime source blockers: `src/audit-log.ts:134-156`, `src/operator-doctor.ts:331,433-435,637,649-686`, `src/qmd-path.ts:14-22`, `src/nats-runtime.ts:493-529`, `src/shared-namespace.ts:14,21,27,44+`, `src/promotion-service.ts:235`, `src/drop-folder-collector.ts:97,145`, and `src/maintenance-bootstrap.ts:232-246`.
+Env-mutating tests needing exemption or rewiring include `server/application/sdk-protocol.pg.test.ts:395-396`, `server/config.test.ts:234-247`, `src/embedding.test.ts`, `src/operator-doctor.test.ts`, namespace-policy tests, search timeout tests, FTS tests, and promotion tests.
+
+verified: `rg -n 'process\.env' server --type ts | rg -v '\.test\.ts'` -> 22 textual hits across 11 files; 11 executable reads outside config.ts/main.ts.
+verified: `rg -n 'parseServerConfig' server src` -> 27 textual references; production definition/call only at config.ts:314/350; remaining direct calls are tests.
+verified: `rg -n 'process\.env' src --type ts | rg -v '\.test\.ts'` -> 81 legacy-source textual hits.
+verified: inventory was read-only; no edit, commit, or test command issued.
+
+```
--- a/docs/lane-contract.md
+++ b/docs/lane-contract.md
@@ -87,6 +87,45 @@ Every lane, no exceptions:
 
 Newest first. Every entry: what changed, and the observation that forced it.
 
+### 2026-08-26 (round 35) — harvest of the L2a config-schema lane (PR #778)
+
+- **A schema field that "mirrors" a reader is differenced against the reader
+  input by input, and the test calls the reader.** L2a declared 23 env names
+  with parsers that reproduced the readers' fallbacks, and every test picked
+  inputs where reader and schema agreed, so 65 green tests could not falsify
+  the PR's own start-equivalence claim. The Light-tier reviewer ran a 15-input
+  differencing set (`"3000ms"`, `"1e3"`, `"0x10"`, `"10.5"`, `""`, `"   "`,
+  `"-1"`, `"Infinity"`, ...) and found three P1 divergences: `PORT` is read
+  with `Number()` (`server/main.ts:362`) and the schema used `parseInt`, so a
+  value that crashes `listen` today silently became 3100; the capture-health
+  integers are read with `Number(raw.trim())` + `isInteger`
+  (`server/capture/liveness-observer.ts:535-550`), not `parseInt`; and the
+  shared-namespace canonical precedence was inverted against `envString`
+  (`server/tools/shared-namespace.ts:79-82`). Where the reader is exported,
+  the equivalence test calls it and compares; where the reader would crash the
+  process, the schema rejects with a named issue, never a substitute value.
+- **Cite the parse, not the constant.** The PR body cited
+  `liveness-observer.ts:404-408`, which is where the env NAMES are declared;
+  the parse is at `:535`. A citation to the name constant reads as "checked"
+  while the semantics were never opened.
+- **`??` is not "blank falls through".** `process.env.A ?? process.env.B`
+  keeps `A=""`; a blank-as-absent preprocessor makes it fall through to `B`.
+  Two readers with different blank handling cannot share one preprocessor.
+- **`GROUPS` is a bash builtin** holding the user's numeric group IDs;
+  assigning it in a done-means script silently yields that list and fails as
+  `but 20 is missing`, which reads like a broken check rather than a name
+  collision.
+- **Tooling:** the Codex git guard refuses `git checkout main` in the clone
+  for the head as well as for workers ("do not switch to main/master for
+  work"); verify-lane runs fine from a feature-branch checkout because it cuts
+  its own worktree from `origin/main`. Two CI runs on one SHA inverted their
+  flakes (`db-integration` vs `python-capture`, #614/#764); every job has a
+  green run on the SHA, which is the evidence, not a re-run to all-green.
+- **`max-lines` at 500 on a test file changes how a table-driven block is
+  added.** The fixer retired five hand-constant assertions the reader-driven
+  table subsumes so `server/config.test.ts` still passed the rule; the next
+  rung adds a sibling `*.test.ts` per group instead of removing assertions.
+
 ### 2026-08-26 (round 34) — harvest of the verify-lane deps-at-head lane (PR #776, #775)
 
 - **verify-lane installed dependencies at `origin/main` and then verified the
--- a/scripts/done-means/750-l2a-config-covers-every-env-read.sh
+++ b/scripts/done-means/750-l2a-config-covers-every-env-read.sh
@@ -1,0 +1,143 @@
+#!/usr/bin/env bash
+# DONE-MEANS check for #750 lane L2a (_plans/server-hardening-ladder.md, rung
+# L2 "Composition root").
+#
+#   bash scripts/done-means/750-l2a-config-covers-every-env-read.sh
+#
+# ---------------------------------------------------------------------------
+# What this asserts
+# ---------------------------------------------------------------------------
+# `_plans/463-server-rewrite-charter.md:108` gives `server/config/` ownership of
+# ALL env parsing and startup validation, and `:119` states that domain code
+# must not import `process.env`. The charter is a claim about the SCHEMA's
+# coverage, not about who happens to read the variable today: a consumer cannot
+# be rewired onto injected config (L2b/L2c) until the field it needs exists in
+# the validated schema.
+#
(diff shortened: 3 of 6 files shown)


rodaddy:
## CI: two runs on the same commit, each failing a job the other passed

Both runs are on `f6b71fb`, unchanged between them:

| job | run 32956141083 | run 32956441305 |
|---|---|---|
| `db-integration` | **failure** | success |
| `python-capture` | success | **failure** |
| `check`, `contract-parity`, `python-package`, `python-provider` | success | success |

The inversion is the point: the same tree passed and failed each of those two
jobs, so neither result is a property of the diff.

**`db-integration`** failed one test, in a file this branch does not touch:

```
1 tests failed:
(fail) maintenance runtime lease boundary (live Postgres) >
       dead-letters an expired lease that has consumed its attempts
       instead of reclaiming it [25.00ms]
```

That is `server/maintenance/maintenance.pg.test.ts`. Run locally against a
fresh database on this branch's tip:

```
bun run test:isolated server/maintenance/maintenance.pg.test.ts
-> 6 pass, 0 fail
```

**`python-capture`** failed only in the second run; this branch changes no
Python. `git diff --name-only origin/main HEAD` is five files:

```
_plans/l2-composition-root-inventory.md
scripts/done-means/750-l2a-config-covers-every-env-read.sh
server/config.test.ts
server/config.ts
server/config/env-groups.ts
```

This matches the shared-mutable-state flake `AGENTS.md` records for the
dogfood database (#614, same class as #498) -- a different failure count on
unmodified code across consecutive runs, clean against a fresh database.

Reported rather than re-run to green: I have not proven a fix for either job,
only that the failures do not track this diff. Both are worth a look on their
own, and `db-integration` should be re-run before merge so the gate is judged
on a run that actually exercised it.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2a-config-covers-every-env-read.sh exit=0 sha=f6b71fbb3316914b6b04b43a7218be07c023fb9b at=2026-08-26T10:11:30.998Z

```
PASS: all 23 env names read in server/ are declared in environmentSchema (server/config.ts)
```

</details>

rodaddy:
## Light-tier review — PR #778 (`feat/l2a-config-schema-env-fields`, head `f6b71fbb`)

Read-only. Bar applied is the one the PR sets for itself: **start-equivalence** — for
every input, the new schema field must answer what the CURRENT reader answers.
Divergences below were found by extracting both parse functions and differencing
them over a shared input set, not by inspection.

### Findings

**P1 `server/config/env-groups.ts:140` — `PORT` diverges from `server/main.ts:362` on six input classes, including `PORT=` (empty), which is a real deployment shape.**
The reader is `Number(process.env.PORT ?? DEFAULT_PORT)`; the field is
`fallbackInteger(1, 3100)`, which is `Number.parseInt` + a range guard. `Number`
and `parseInt` are different functions and disagree in both directions. Observed:

| `PORT` | reader (`main.ts:362`) | schema field | note |
|---|---|---|---|
| `""` / `"   "` | `0` | `3100` | `Number("")` is `0`; today this binds an **ephemeral** port |
| `"not-a-port"` | `NaN` | `3100` | |
| `"3000ms"` | `NaN` | `3000` | `parseInt` takes the trailing suffix |
| `"10.5"` | `10.5` | `10` | |
| `"1e3"` | `1000` | `1` | `parseInt` stops at `e` |
| `"0x10"` | `16` | `3100` | |
| `"-1"` / `"0"` / `"Infinity"` | passed through | `3100` | |

`PORT=` (present-but-empty) is the same shape the PR's own qmd note calls out as
having caused a live defect (`QMD_PATH=` in the dogfood env), so it is not
hypothetical. Note the module header claims "`parseInt` is NOT `Number` … Using
`z.coerce.number()` would reject it" — that reasoning is correct for the section-A
integer readers, but `PORT` is the one reader that *does* use `Number`, so the
header's justification is inverted for this field specifically.
Reproduce: `node -e 'console.log(Number(""), Number("1e3"), Number.parseInt("1e3",10))'` → `0 1000 1`.

**P1 `server/config/env-groups.ts:136-137` — `OPENBRAIN_CAPTURE_HEALTH_WINDOW_MINUTES` and `OPENBRAIN_CAPTURE_HEALTH_REFRESH_MS` diverge from `server/capture/liveness-observer.ts:535-550`.**
That reader is `readPositiveInteger`, which uses `Number(raw.trim())` and then
requires `Number.isInteger` — i.e. it deliberately **rejects** anything that is not
a clean integer. The field uses `parseInt`, which truncates and accepts suffixes.
Observed divergences for both fields:

| value | reader | schema field |
|---|---|---|
| `"3000ms"` | fallback (360 / 60000) | `3000` |
| `"10.5"` | fallback | `10` |
| `"1e3"` | `1000` | `1` |
| `"0x10"` | `16` | fallback |
| `"3,000"` | fallback | `3` |

This is a behavior change in the direction the module header says it is avoiding:
the reader announces the rejection at WARN (`capture_health_override_invalid`,
"NOTHING IS ADJUSTED SILENTLY"), and the schema would silently accept `10.5` as a
10-minute window. The PR body lists this reader at `:404-408` (the env-name
constants) rather than at `:535` (the parse), which is likely how the `Number` vs
`parseInt` difference was missed.

**P1 `server/config/env-groups.ts:247` — `sharedNamespaceGroup` inverts the canonical-name precedence in `server/tools/shared-namespace.ts:79-82`.**
The reader is `envString(["SHARED_NAMESPACE_CANONICAL", "OPENBRAIN_SHARED_NAMESPACE"], "shared-kb")`
— first non-empty **in that order**, so `SHARED_NAMESPACE_CANONICAL` wins. The new
code is `parsed.OPENBRAIN_SHARED_NAMESPACE ?? canonical`, which makes
`OPENBRAIN_SHARED_NAMESPACE` win instead. It is masked today only because
`SHARED_NAMESPACE_CANONICAL` is declared `z.literal("shared-kb").default("shared-kb")`
(`server/config.ts:167`) and so is never absent — which means the schema's
precedence is not merely untested, it is unreachable-by-the-default and would flip
the moment that literal is relaxed.

Concretely, with `OPENBRAIN_SHARED_NAMESPACE=other`: reader resolves canonical
`shared-kb`, the group resolves `other`. The same value then propagates into
`physical` (line 249), so the two halves of the canonical/physical pair — the pair
whose whole purpose per the module header is to bind isolation predicates
correctly — disagree between reader and config. `docs/sme/security.md` treats
namespace resolution as a security boundary; a divergence here is not cosmetic
even while the consumer is un-rewired.
Reproduce: set `OPENBRAIN_SHARED_NAMESPACE=other`, compare `sharedNamespaceConfig().canonicalSharedNamespace` (`shared-kb`) against `parseServerConfig({...}).config.sharedNamespaceNames.physical` (`other`).

**P2 `server/config.test.ts:552,558` — the http tests assert the schema's answer, not the reader's, so they cannot catch the P1 above.**
`it("ignores an unusable port rather than failing startup")` asserts
`PORT: "not-a-port"` → `3100`. The reader answers `NaN` for that input. The test is
not wrong about what the code does; it is asserting the new behavior as if it were
the existing one, which is exactly the gap that lets a start-equivalence claim pass
its own suite. The same shape applies to the capture-health case at `:539`
(`"soon"` / `"-5"` — both inputs happen to agree between the two implementations,
so the test passes without exercising the divergence). Every extended-env test
picks inputs on which reader and schema agree; none of the divergent inputs above
appears anywhere in the file.

Suggested (not required by me): drive the assertions from the reader itself where
it is importable — `searchEmbeddingTimeoutMs` and `readPositiveInteger` are
module-private, but `resolveQmdPath`, `parseAllowedOrigins`, `resolveFtsConfig`,
and `sharedNamespaceConfig` are exported and could be asserted equal to the group
output over a shared input table. That is the only test shape that makes
"start-equivalence" falsifiable.

**P3 `server/config/env-groups.ts:126` — `fallbackInteger(1, 5)` vs `envPositiveInteger` differ in written form but not in behavior.**
The reader guards `parsed > 0`, the field guards `parsed >= 1`. Equivalent for
integers, and `parseInt` is used on both sides, so this one is clean — noting it
only so the record shows it was differenced rather than assumed.

### Checked and clean

- **Composition safety (`server/config.ts:179`)** — no P0. Every one of the 22 new
  keys is new: `comm -12` of the pre-existing `environmentSchema` key set against
  the `extendedEnvironmentFields` key set is empty, so no existing field's parse or
  default changed. Every new field is total (each preprocess returns a value for
  `undefined`), so none can produce a new `server_configuration_invalid`. The
  `.catchall(z.string().optional())` terminator is unchanged, so NATS/maintenance
  keys still flow through as before. Startup cannot now fail where it did not.
- **Done-means is not vacuous** — `scripts/done-means/750-l2a-config-covers-every-env-read.sh`
  guards each failure mode the brief asked about: the awk block failing to locate
  the literal is an explicit `FAIL` (both for `environmentSchema` and for
  `extendedEnvironmentFields`), the `rg -o '^\s{4}([A-Z][A-Z0-9_]+):'` pattern
  requires a declaration colon at declaration indent so a comment mention does not
  satisfy it, and `(( count < 15 ))` turns an empty scan into a failure rather than
  a green. It follows the spread only into the one named sibling module. I did not
  find a way to make it pass vacuously. `OPEN_BRAIN_BIND_HOST` is in its required
  list and resolves against the pre-existing `server/config.ts:158`.
- **Tracing** (`langfuse-tracing.ts:599-604`) — `blankAsAbsent` + `?? ""` is
  equivalent to `?.trim() ?? ""` for blank, whitespace, and absent; `strictOneFlag`
  matches `=== "1"`; `strictZeroDisablesFlag` matches `!== "0"`; the
  three-coordinate `complete` conjunction matches. Secret key is not surfaced in
  `/health`. Clean.
- **Search timeout** (`search-engine.ts:148-149`) — name-before-value precedence is
  reproduced correctly, including the case the module comment calls out
  (`OPENBRAIN_…=abc` falls to 3000, not through to the legacy name). Clean.
- **qmd** (`search-all.ts:119`), **recovery** (`realtime-stores.ts:46`),
  **fts** (`fts-config.ts:111`, reuses `resolveFtsConfig` rather than a second
  allowlist), **`ALLOWED_ORIGINS`** vs `parseAllowedOrigins` (`rest.ts:115-122`,
  including empty-means-no-origin rather than wildcard) — all differenced, all
  equivalent.

### Verification

```
verified: bash scripts/done-means/750-l2a-config-covers-every-env-read.sh
  -> PASS: all 23 env names read in server/ are declared in environmentSchema (server/config.ts); exit 0
verified: bun test server/config.test.ts
  -> 65 pass, 0 fail, 102 expect() calls, 1 file, 41.00ms
verified: differencing harness over {"3000ms","10.5","1e3"," 42 ","0x10","abc","-1","0","","   ",undefined,"007","3,000","+5","Infinity"}
  -> PORT 11 divergences; CAPTURE_HEALTH_WINDOW_MINUTES 5; CAPTURE_HEALTH_REFRESH_MS 5;
     SHARED_FALLBACK_MIN_RESULTS 0
verified: comm -12 <(old environmentSchema keys) <(extendedEnvironmentFields keys)
  -> empty (no pre-existing field redefined)
```

Note the two receipts above pass while three start-equivalence divergences exist —
neither check tests reader-vs-schema equivalence, which is the PR's actual claim.

VERDICT: BLOCK

rodaddy:
## Fixer pass — all four start-equivalence findings addressed (`bae8571`)

The review's bar was the PR's own: for every input, the schema answers what the
reader answers, and where the reader would crash the process at startup, the
schema rejects with a named config issue. Three of the four divergences had one
root cause — the mirror used `Number.parseInt` where the reader uses `Number`.
Those are different functions and disagree in both directions, so a parser is
only a mirror if it uses the reader's own function.

### Fixed

- **fixed: `server/config/env-groups.ts:140` (`PORT`)** — replaced
  `fallbackInteger(1, 3100)` with a `portNumber` parser that is `Number(value)`
  exactly, mirroring `server/main.ts:362`. `""`/`"   "` → 0 (ephemeral bind),
  `"1e3"` → 1000, `"0x10"` → 16, `" 42 "` → 42, `"007"` → 7. Any result Node's
  `listen` would throw on — non-integer or outside 0..65535 (`NaN`, `10.5`,
  `-1`, `Infinity`, `"3000ms"`) — is now REJECTED with a config issue whose
  path is `PORT`. That is equivalent: the process crashes today on exactly
  those inputs, and an earlier named crash is the same outcome reached sooner.
  A silent 3100 was not — it would boot a deployment that does not boot now, on
  a port nobody configured. No `parseInt` on this field.
- **fixed: `server/config/env-groups.ts:136-137` (capture-health window/refresh)**
  — new `numberPositiveInteger` parser reproducing `readPositiveInteger`
  (`server/capture/liveness-observer.ts:535-550`) line for line:
  `Number(raw.trim())` + `Number.isFinite` + `Number.isInteger` + `> 0`, else
  the fallback. That reader is module-private, so it is reproduced rather than
  imported, with its file and line cited beside the parser. `parseInt` had been
  silently accepting `10.5` as a 10-minute window; the reader rejects it and
  announces it at WARN. The WARN stays with the reader, which owns a logger.
- **fixed: `server/config/env-groups.ts:247` (`sharedNamespaceGroup`)** — the
  canonical precedence now matches
  `envString(["SHARED_NAMESPACE_CANONICAL", "OPENBRAIN_SHARED_NAMESPACE"], "shared-kb")`
  (`server/tools/shared-namespace.ts:79-82`): first non-empty trimmed value in
  that order, so `SHARED_NAMESPACE_CANONICAL` wins. Because the schema declares
  that name as a literal with a default (`server/config.ts:167`), `parsed` can
  never express "unset" — so `buildConfig` now receives the RAW `environment`
  alongside the parsed one and passes `environment.SHARED_NAMESPACE_CANONICAL`
  in. The literal is unchanged and not weakened.
- **fixed: `server/config/env-groups.ts:110-111` + `searchGroup` (search timeout)**
  — both names are back to raw `z.string().optional()` (no `blankAsAbsent`),
  and `searchGroup` reproduces `server/tools/search-engine.ts:147-152` including
  the `!raw` guard. The reader joins the two names with `??`, which does NOT
  fall through on `""`, so `OPENBRAIN_SEARCH_EMBEDDING_TIMEOUT_MS=` with
  `SEARCH_EMBEDDING_TIMEOUT_MS=900` answers 3000 — `blankAsAbsent` would have
  answered 900.
- **fixed: `server/config.test.ts` (P2, the tests could not catch the above)** —
  added a table-driven start-equivalence block over the reviewer's input set
  `{"3000ms","10.5","1e3"," 42 ","0x10","abc","-1","0","","   ",undefined,"007","3,000","+5","Infinity"}`
  covering PORT, both capture-health fields, both search-timeout names, and the
  shared-namespace precedence. Where the reader is exported it is IMPORTED and
  called and the schema compared to it — `resolveFtsConfig`, `parseAllowedOrigins`,
  `resolveQmdPath`, `readMcpTracingConfig`. Where it is private
  (`readPositiveInteger`, `searchEmbeddingTimeoutMs`, `main.ts`'s inline
  `Number(...)`) its exact expression is written out beside its file and line.
  The pre-existing `PORT: "not-a-port"` → 3100 case asserted the schema's new
  behavior as if it were the existing one; it now asserts the named rejection.

The module header's justification is corrected too: it had claimed `parseInt`
semantics for all integer readers and that no field would reject. Both were
wrong for PORT specifically, which is the inversion the review named.

### RED evidence

The new tests were run against this branch with ONLY the source fix stashed
(`git stash push -- server/config/env-groups.ts server/config.ts`), test file
unchanged:

```
verified: git stash push -- server/config/env-groups.ts server/config.ts
       && bun run test:isolated server/config.test.ts
  -> 66 pass, 16 fail; Ran 82 tests across 1 file
```

Representative failure lines from that run, one per finding:

```
(fail) extended env — http > rejects an unbindable port by name instead of substituting a default
(fail) start-equivalence — every field answers what its reader answers > every reader agrees for "1e3"
(fail) start-equivalence — every field answers what its reader answers > every reader agrees for "10.5"
(fail) start-equivalence — every field answers what its reader answers > binds the ephemeral port for PORT= rather than substituting 3100
(fail) start-equivalence — every field answers what its reader answers > does not fall through to the legacy timeout name when the preferred is blank
(fail) start-equivalence — shared-namespace canonical precedence > resolves the physical name from {"SHARED_NAMESPACE_CANONICAL":"shared-kb","OPENBRAIN_SHARED_NAMESPACE":"other"}
```

### Verification

```
verified: bun run test:isolated server/config.test.ts
  -> 82 pass, 0 fail, Ran 82 tests across 1 file
verified: bunx tsc --noEmit -> exit 0
verified: bash scripts/done-means/750-l2a-config-covers-every-env-read.sh
  -> PASS: all 23 env names read in server/ are declared in environmentSchema; exit 0
verified: pre-commit hook (gitleaks + oxlint staged + tsc) -> all checks passed, no bypass
verified: git merge-base --is-ancestor origin/main HEAD -> true
verified: pushed -> bae85712554f805e003b6a25777a186c39beeea1
```

One note on scope, stated rather than hidden: satisfying `oxlint`'s
`max-lines: 500` on `server/config.test.ts` (code lines; the file was already at
~406 before this change) meant the new table could not simply be appended. Rather
than bypass the hook or add a file outside the agreed set, five pre-existing
assertions that the new reader-driven table strictly subsumes were removed and
replaced by a pointer comment at each site — the unparseable search timeout, the
unrecognized FTS token, the unparseable capture-health window/refresh, the blank
origin list, and the incomplete tracing coordinate. Each of those inputs is now
asserted against the reader itself rather than against a hand-written constant,
so coverage went up, not down. Nothing else in the file was touched.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2a-config-covers-every-env-read.sh exit=0 sha=b8d097aa6c4dd3805ced94e8faf0fa34015650dc at=2026-08-26T10:36:20.663Z

```
PASS: all 23 env names read in server/ are declared in environmentSchema (server/config.ts)
```

</details>

PR #779, opened 2026-08-26
feat(config): inject four tool-layer env values from validated config (L2b-1a)

rodaddy:
## Summary

- Lane **L2b-1a** of the `server/` hardening ladder (`_plans/server-hardening-ladder.md`, rung L2 "Composition root"). #778 typed every `server/` env read into a validated `ServerConfig` group but left the original readers in place, so both halves existed and only the reader was consulted. That doubled state is worse than either half alone: a group value can be correct, tested, and reported in `/health` while the tool path answers something else, and nothing fails — the config quietly becomes documentation of an intention rather than the value in force.
- **Both halves of the lane are now in this PR.** Six commits over `origin/main`:

| Commit | What it lands |
|---|---|
| `22802df` | RED: assert the typed tool readers stop reading `process.env` |
| `b7c6966` | Readers take the env values as injected parameters (1/2) |
| `407f83f` | Rescope the done-means to the four files this lane lands |
| `7b257c0` | Composition root supplies those values (2/2) |
| `fd6e294` | `server/tools/dependency-arrival.test.ts` + done-means CLAUSE 5 |
| `d0bbbd3` | The fallback recovery WAL store honors an injected path whenever it arrives |

- **Reader half (`b7c6966`).** `server/tools/fts-config.ts` takes the deployment default as a `corpusDefault` parameter, and `corpusFtsConfig` is deleted rather than kept as a wrapper — `server/config/env-groups.ts` already derives the group by calling `resolveFtsConfig`, this module's own export, so the allowlist and alias table keep exactly one owner. `server/tools/realtime-stores.ts` takes `recoveryWalPath` from dependencies. `server/tools/operator-doctor.ts` takes the NATS boundary object injected. `server/tools/types.ts` declares the new optional fields on `MemoryToolDependencies`.
- **Composition-root half (`7b257c0`).** `createServerFactory` gained a `nats: NatsPhase` input and now passes `ftsCorpusConfig: config.fts.corpusConfig`, `recoveryWalPath: config.recovery.walPath`, and `natsRuntimeBoundary: nats.boundary` into `registerMemoryTools`; the recovery WAL store in `server/main.ts` reads `config.recovery.walPath` instead of `process.env.OPENBRAIN_RECOVERY_WAL_PATH`. The boundary passed is the one the NATS phase **already built** (`natsRuntimeBoundaryFromConfig` at `server/main.ts:267`, returned as `NatsPhase.boundary`), not a second call — rebuilding it at the factory would be a second source of truth by construction, able to drift from the one the bridge and `/health` run on. `server/tools/search-brain.ts` threads `dependencies.ftsCorpusConfig` through `resolveCallerFtsConfig` into `requestFtsConfig`'s new second argument. The dead `searchEmbeddingTimeoutMs` field was removed from `server/tools/types.ts`: nothing supplied it and nothing consumed it, so leaving it declared would have advertised a wire that does not exist.
- **Arrival fix (`d0bbbd3`).** `server/tools/realtime-stores.ts` keys the fallback recovery WAL store on the `recoveryWalPath` it was built for. Callers asking for the same path keep sharing one store, which is the point of a process-lifetime fallback; a caller asking for a different path gets a store rebuilt for that path, so an injected path is honored no matter WHEN it arrives rather than whichever registration touched the fallback first. `workingSetStoreFor` is untouched. `server/tools/dependency-arrival.test.ts` gained an order-independence case that registers twice in one process with two different paths and requires both files written, and its scratch path moved to `mkdtempSync(join(tmpdir(), "ob-arrival-"))` so it does not depend on a developer-machine directory.
- `operator_doctor` gains a real property it did not have: it previously read `OPENBRAIN_TRANSPORT`/`OPENBRAIN_NATS_*` in a SECOND env read, so it could report `nats` while the process served http, because nothing tied the two reads together. It now answers from the same parsed `NatsConfig` the bridge and `/health` use.
- Start-equivalence is the bar. Every field keeps its reader's own default for a caller that injects nothing — english, a RAM-only WAL, and the boundary an empty environment declares — so this changes no behavior on a deployment that boots today.

### Scope: what is deferred, and to where

`server/tools/search-all.ts` and `server/tools/search-engine.ts` are the other two readers L2a typed. They are **not** in this PR and their `process.env` reads are still in the tree (`search-all.ts:119`, `search-engine.ts:148-149`).

The reason is a collision with a different gate. `_githooks/pre-commit` lints whole **staged** files with no main-vs-branch baseline, and both files already violate `.oxlintrc.json` on unmodified `origin/main` — 5 violations in `search-all.ts`, 9 in `search-engine.ts`. So any commit touching them is refused whether or not its own change is clean, and `--no-verify` is not an approved workaround in this repo. `.oxlintrc.json` states the intended policy — the backlog is "paid per-file as work naturally touches them" — and measured, clearing those files is a behavior-risky refactor of the core search path (`executeSearch` alone has 55 references repo-wide, a 203-line handler, an 837-code-line file). Burying a start-equivalence rewiring inside that is what the lane contract forbids.

Where each deferred piece now goes:

- `server/tools/search-all.ts` — **lane 3, next**, on the #780 checklist.
- `server/tools/search-engine.ts` — **next handoff**, the L4 split.
- `searchEmbeddingTimeoutMs` wiring belongs to the search-engine lane. That lane must **re-add** the field to `MemoryToolDependencies` and wire `config.search.embeddingTimeoutMs` to it; this PR removed it precisely so nobody reads its presence as evidence the value is flowing.
- `server/tools/search-brain.ts` lint debt was cleared separately by **#786**, which is why this PR could touch it.

## Verification

- Done-means: scripts/done-means/750-l2b1-tool-readers-take-config.sh

Five clauses, all of which must pass:

1. All four target files exist (a rename would make the scan vacuous).
2. None of the four reads `process.env`.
3. Positive control — the same pattern, tool, and invocation must still find hits in `server/config.ts`, the one file in `server/` that MUST read the environment. Guards the "0 items examined, exit 0" shape.
4. Each of the four values arrives **exactly once** at the composition root. Zero means unwired; more than one means two composition paths that can disagree.
5. The injected values are observed **at their consumers**, by running `server/tools/dependency-arrival.test.ts` and requiring exit 0 with at least 3 passing tests. It now reports 5.

**RED evidence that clauses 4 and 5 are not decoration.** The absence-only version of this check (clauses 1-3) passed at `1d0b9b2` while **nothing was wired** — three of the four readers took parameters and `server/main.ts` supplied none of them, so every one was silently running on its `?? default` forever. That is exactly the pair of P1s the #779 review raised, and clause 4 was written to make that state fail. Clause 5 in turn FAILS with the test file absent (`"...does not exist — clause 4 asserts the wiring was TYPED; nothing here asserts it WORKS"`), because clause 4 is a textual assertion: a field can be spelled correctly at the root and still be dropped by the registrar, shadowed by a fallback, or read from the wrong place by the handler. The arrival test is proven able to fail — the french-variant assertion in the `ftsCorpusConfig` case fails when the injected corpus default does not reach the emitted tsvector/tsquery.

**CI RED that `d0bbbd3` answered.** The `db-integration` job on `fd6e294` (runs 32994328337 and 32994354301) FAILED the new arrival case: `expect(existsSync(walPath)).toBe(true)` received `false`. Two real causes, both fixed in `d0bbbd3` — the test wrote its WAL under a developer-machine scratch directory absent on the runner, and the fallback recovery WAL store memoized on first touch, so in the single bun process CI uses for the whole suite whichever registration touched it first decided the path for every later caller. The order-independence case was written RED-first: `server/tools/dependency-arrival.test.ts:238` `expect(existsSync(firstPath)).toBe(true)` received `false` before the fix.

GREEN on this PR head, run in the lane clone:

```
CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present
CLAUSE 2 (no process.env in any of the 4):              PASS — no process.env read in any of the 4 files
CLAUSE 3 (scanner proven to match in server/config.ts):   PASS — server/config.ts still contains 4 process.env line(s)
CLAUSE 4 (values arrive at the composition root):        PASS — all 4 values are passed exactly once from the composition root
CLAUSE 5 (injected values observed at their consumers):  PASS — 5 test(s) passed in server/tools/dependency-arrival.test.ts
EXIT=0
```

- [x] Relevant Open Brain tests/typecheck/migrations passed — `bunx tsc --noEmit` clean; oxlint 0 over the touched and adjacent tool files; `bun run test:isolated server/` → 488 pass / 0 fail across 36 files; 112 pass / 0 fail across `main.pg`, `search-read-scope`, `namespace-denial`, and `config` on the wired head; the 5 arrival tests pass through the done-means clause 5 run shown above. Isolated databases dropped on exit.
- [x] Python package checks passed or are not applicable — not applicable; no change under `python/`.
- [x] Live Open Brain smoke passed or is not applicable — not applicable; no contract, schema, transport, or client-facing surface changes. `operator_doctor`'s payload shape is unchanged and still under the `DOCTOR_CONTRACT_VERSION` lock in `src/operator-doctor.test.ts`.

## Critical Self-Review

- Highest-risk behavior: `operator_doctor`'s NATS boundary object. It changes SOURCE (a second env read → the parsed `NatsConfig` the NATS phase already built), and the two could disagree in a deployment where they already did — which is the defect being fixed, so a difference in output there is the correction, not a regression. The frozen payload SHAPE is untouched and still under the `DOCTOR_CONTRACT_VERSION` lock.
- Assumptions that could be wrong: that `registerMemoryTools` at `createServerFactory` is the ONLY path these tools are registered through. Checked — the NATS bridge takes a separate `ToolDeps` and registers none of these tools (`rg` over `server/application/nats.ts` and the bridge found no `registerMemoryTools`/`registerOperatorDoctorTool`). If a future caller registers them elsewhere without injecting, it silently gets the documented defaults rather than the deployment's config, and nothing in this PR would catch that.
- Missing/weak tests: `server/tools/dependency-arrival.test.ts` covers the three values this lane wires through a fake McpServer, but it exercises no real transport and no real process start, so it cannot catch a value that is correct in `createServerFactory` and wrong in the launch path that builds `ServerConfig`. The two deferred readers have no arrival coverage at all — by construction, since they are not wired yet. The order-independence of the fallback recovery WAL store is now covered by the second registration case added in `d0bbbd3`.
- Security/permission risk: low. No namespace predicate, auth check, or role gate is touched. `operator_doctor`'s admin/ob-admin gate and its fixed-string error path are unchanged, and the NATS URL stays out of the doctor payload. `search-brain.ts`'s non-english corpus denial for unprivileged roles is preserved — the injected corpus default flows through `resolveCallerFtsConfig`, which still applies the privileged-role rule before returning.
- Migration/deploy risk: none. No schema, no migration. Start-equivalence is the explicit bar and each field keeps its reader's default, so an environment that boots today boots identically. `createServerFactory` gained a required `nats` input; every call site is inside `server/main.ts` and tsc is clean.
- Downstream client/runtime risk: none identified. No MCP tool schema, transport protocol, or Python client surface changes; the four readers are internal to `server/`. `docs/downstream-rollout.md` classifies this as not applicable.
- Rollback/cleanup concern: revert is clean — six commits, no state, no schema. The isolated test databases were dropped by the harness. Nothing is left stashed or dirty: the composition-root half that the previous body described as parked in `stash@{0}` is now committed as `7b257c0`, so that stash is no longer the home of any unlanded work.
- Fixes made before PR: the #779 review's two P1s, both of which were that the wiring did not exist. `7b257c0` supplies all three values from the composition root and `fd6e294` proves they arrive at their consumers; the done-means grew clause 4 (arrival at the root) and clause 5 (arrival at the consumer) so the unwired state that passed at `1d0b9b2` fails loudly. Also caught that `qmdPath` must be spread conditionally rather than assigned `undefined`, because `exactOptionalPropertyTypes` distinguishes the two, and that removing `searchEmbeddingTimeoutMs` is correct only because `search-engine.ts` still parses its own — recorded in the deferral so the search-engine lane re-adds it.
- Known residual risk: rung L2 is not finished. `search-all.ts` and `search-engine.ts` still read `process.env` in the tree, so the composition root is the single source for three of the five typed values and not yet for all of them; the done-means header says so rather than letting a green read as completion. `server/tools/dependency-arrival.test.ts` exercises no real transport and no real process start, so a value that is correct in `createServerFactory` and wrong in the launch path that builds `ServerConfig` would not be caught here.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the #779 review findings were P1s against this branch's own unfinished wiring, fixed here rather than being a reusable reviewer pattern; the transferable lesson — an absence-only check passes on an unwired reader, so pair it with an arrival assertion — is captured executably in the done-means header and harvested to the lane contract's Tightenings.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no contract, transport, schema, or client-facing change; the affected surfaces are internal composition wiring with start-equivalent values.

## Contract Parity

- Contract parity: [x] runtime-specific because: the diff touches no path in `contracts/parity-paths.txt` and no contract fixture covers these values. `operator_doctor`'s payload shape is unchanged and remains locked by `src/operator-doctor.test.ts` against `DOCTOR_CONTRACT_VERSION`.

## Downstream Rollout

Notes/evidence:

- Not applicable against every trigger in the doc's "When This Applies": no MCP tool name, schema, or output shape changes; no auth or namespace semantics change; no error envelope change; no transport or session behavior change; no externally visible migration; no `python/openbrain-memory` behavior or export change; no generated `skill/` content change. The four readers are internal to `server/`, and every injected value is start-equivalent to what the reader it replaced produced, so no mcp2cli or Hermes consumer observes a difference.

### DIFF
--- a/scripts/done-means/750-l2b1-tool-readers-take-config.sh
+++ b/scripts/done-means/750-l2b1-tool-readers-take-config.sh
@@ -1,0 +1,171 @@
+#!/usr/bin/env bash
+# DONE-MEANS check for lane L2b-1 of the server/ hardening ladder
+# (`_plans/server-hardening-ladder.md`, rung L2 "Composition root").
+#
+#   bash scripts/done-means/750-l2b1-tool-readers-take-config.sh
+#
+# ---------------------------------------------------------------------------
+# The defect this gates
+# ---------------------------------------------------------------------------
+# The rewrite charter (`_plans/463-server-rewrite-charter.md:108,119`) puts ALL
+# env parsing behind `server/config/` and forbids domain code from importing
+# `process.env`. L2a (#778) delivered the SCHEMA half: every one of these reads
+# is now typed and parsed into a `ServerConfig` group. That left the repo in a
+# deliberately doubled state — the validated group AND the original scattered
+# reader both existed, and only the reader was actually consulted.
+#
+# A doubled state is worse than either half alone, because it looks finished
+# from the config side. `config.search.embeddingTimeoutMs` can be correct,
+# tested, and reported in `/health` while the search path continues to read
+# `process.env` directly and answer something else entirely. Nothing fails; the
+# two simply drift, and the config becomes documentation of an intention rather
+# than the value in force.
+#
+# This lane closes the five readers in `server/tools/` that L2a typed. The
+# assertion is therefore about ABSENCE at a named boundary: those five files
+# must contain no `process.env` at all, because every value they need now
+# arrives through a parameter or through `MemoryToolDependencies`.
+#
+# ---------------------------------------------------------------------------
+# WHY CLAUSE 3 EXISTS — the vacuous-pass problem
+# ---------------------------------------------------------------------------
+# The natural check is "rg finds no `process.env` in these five files", and that
+# assertion passes for two very different reasons: because the reads are gone,
+# or because the scan examined nothing. A renamed file, a typo in a path, an
+# `rg` that is not on PATH, a bad `--glob` — every one of them produces a silent
+# clean sweep and an exit 0 that means nothing. This repo has been burned by the
+# "0 items examined, exit 0" shape before, which is why the toolbox convention
+# is to prove the tool can still see a hit somewhere it MUST see one.
+#
+# So the check is three clauses, and all three must pass:
+#
+# CLAUSE 1 — ALL FIVE TARGET FILES EXIST.
+#   Each path is `test -f`'d individually and named in the output. A file that
+#   was renamed or deleted makes the scan vacuous, so it is a hard fail here
+#   rather than an invisible pass in clause 2.
+#
+# CLAUSE 2 — NONE OF THEM READS `process.env`.
+#   `rg -n 'process\.env' <five files>` must produce no output. Any match is
+#   printed, so a failure names the file and line instead of just refusing.
+#
+# CLAUSE 3 — THE SCANNER IS PROVEN TO MATCH.
+#   `rg -c 'process\.env' server/config.ts` must be > 0. `server/config.ts` is
+#   the composition root's own env reader: it is the one file in `server/` that
+#   MUST read the environment, because it is the boundary the charter puts the
+#   parsing behind. If the same pattern, the same tool, and the same invocation
+#   find nothing THERE, the clean result in clause 2 is a broken scan and not a
+#   clean tree. This is a positive control, deliberately pointed at a file this
+#   lane does not touch and no future lane should empty.
+#
+# Exit 0 only when all three clauses pass. Exit 3 is a harness error (missing
+# tool / wrong repo root), which is NOT a fail of the thing under test.
+set -uo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
+fail_hard() {
+  printf 'HARNESS-ERROR: %s\n' "$1" >&2
+  exit 3
+}
+
+command -v rg >/dev/null 2>&1 || fail_hard "rg (ripgrep) not on PATH"
+[ -d "$REPO_ROOT/.git" ] || [ -f "$REPO_ROOT/.git" ] || fail_hard "not a git repo at $REPO_ROOT"
+
+# The five readers L2a typed and this lane rewires. Listed literally rather
+# than globbed: a glob that stops matching is exactly the vacuous pass clause 3
+# guards against, and naming them makes a rename fail loudly in clause 1.
+TARGETS="server/tools/search-engine.ts
+server/tools/fts-config.ts
+server/tools/search-all.ts
+server/tools/realtime-stores.ts
+server/tools/operator-doctor.ts"
+
+# The positive control: the composition root's OWN env reader, which must keep
+# reading `process.env` for the boundary to exist at all.
+CONTROL="server/config.ts"
+
+CLAUSE1=FAIL; CLAUSE1_EVIDENCE=""
+CLAUSE2=FAIL; CLAUSE2_EVIDENCE=""
+CLAUSE3=FAIL; CLAUSE3_EVIDENCE=""
+
+# ---------------------------------------------------------------------------
+# CLAUSE 1 — every target file is present.
+# ---------------------------------------------------------------------------
+MISSING=""
+N_TARGETS=0
+while IFS= read -r t; do
+  [ -n "$t" ] || continue
+  N_TARGETS=$((N_TARGETS + 1))
+  [ -f "$REPO_ROOT/$t" ] || MISSING="${MISSING}${t} "
+done <<EOF
+$TARGETS
+EOF
+
+if [ "$N_TARGETS" -ne 5 ]; then
+  CLAUSE1_EVIDENCE="expected 5 target paths, the list yielded $N_TARGETS — the check itself is miswritten"
+elif [ -n "$MISSING" ]; then
+  CLAUSE1_EVIDENCE="target file(s) NOT found: ${MISSING}— a scan over missing files passes vacuously"
+else
+  CLAUSE1=PASS
+  CLAUSE1_EVIDENCE="all 5 target files present under $REPO_ROOT"
+fi
+
+# ---------------------------------------------------------------------------
+# CLAUSE 2 — none of the five reads process.env.
+# ---------------------------------------------------------------------------
+if [ "$CLAUSE1" != PASS ]; then
+  CLAUSE2_EVIDENCE="skipped — target files are not all present, so any result is vacuous"
+else
+  # Build the argument list from the same literal set clause 1 verified.
+  set --
+  while IFS= read -r t; do
+    [ -n "$t" ] || continue
+    set -- "$@" "$REPO_ROOT/$t"
+  done <<EOF
+$TARGETS
+EOF
+
+  HITS="$(cd "$REPO_ROOT" && rg -n 'process\.env' "$@" 2>/dev/null)"
+  RG_STATUS=$?
+  # rg exits 1 for "no matches" (the pass) and 2 for a real error. A 2 is a
+  # harness problem, not a clean tree, and must not be read as success.
+  if [ "$RG_STATUS" -ge 2 ]; then
+    fail_hard "rg failed with status $RG_STATUS scanning the five target files"
+  fi
+
+  if [ -n "$HITS" ]; then
+    CLAUSE2_EVIDENCE="$(printf '%s\n' "$HITS" | wc -l | tr -d ' ') process.env read(s) remain:"
+    CLAUSE2_HITS="$HITS"
+  else
+    CLAUSE2=PASS
+    CLAUSE2_EVIDENCE="no process.env read in any of the 5 files — each value now arrives from validated config"
+  fi
+fi
+
+# ---------------------------------------------------------------------------
+# CLAUSE 3 — positive control: the scanner still finds a known hit.
+# ---------------------------------------------------------------------------
+if [ ! -f "$REPO_ROOT/$CONTROL" ]; then
+  CLAUSE3_EVIDENCE="positive control $CONTROL does not exist — cannot prove the scan matches anything"
+else
+  CONTROL_COUNT="$(cd "$REPO_ROOT" && rg -c 'process\.env' "$CONTROL" 2>/dev/null)"
+  CONTROL_COUNT="${CONTROL_COUNT:-0}"
+  if [ "$CONTROL_COUNT" -gt 0 ]; then
+    CLAUSE3=PASS
+    CLAUSE3_EVIDENCE="$CONTROL still contains $CONTROL_COUNT process.env line(s) — the pattern, tool, and invocation are proven to match"
+  else
+    CLAUSE3_EVIDENCE="$CONTROL contains ZERO process.env lines — either the composition root stopped reading the environment, or this scan is broken and clause 2's clean result is meaningless"
+  fi
+fi
+
+printf 'CLAUSE 1 (all 5 target files exist):                    %s — %s\n' "$CLAUSE1" "$CLAUSE1_EVIDENCE"
+printf 'CLAUSE 2 (no process.env in any of the 5):              %s — %s\n' "$CLAUSE2" "$CLAUSE2_EVIDENCE"
+if [ "$CLAUSE2" != PASS ] && [ -n "${CLAUSE2_HITS:-}" ]; then
+  printf '%s\n' "$CLAUSE2_HITS" | sed 's/^/    /'
+fi
+printf 'CLAUSE 3 (scanner proven to match in %s):   %s — %s\n' "$CONTROL" "$CLAUSE3" "$CLAUSE3_EVIDENCE"
+
+if [ "$CLAUSE1" = PASS ] && [ "$CLAUSE2" = PASS ] && [ "$CLAUSE3" = PASS ]; then
+  exit 0
+fi
+exit 1
--- a/scripts/done-means/750-l2b1-tool-readers-take-config.sh
+++ b/scripts/done-means/750-l2b1-tool-readers-take-config.sh
@@ -26,6 +26,14 @@
 # at all, because every value they need now arrives through a parameter or
 # through `MemoryToolDependencies`.
 #
+# ABSENCE ALONE IS NOT THE BAR. A reader that stopped reading `process.env` and
+# whose caller never started passing the value is not wired -- it is silently
+# running on its `?? default` forever, and clauses 1-3 go green on exactly that
+# state (observed on this branch: three of the four readers took parameters and
+# `server/main.ts` passed none of them). So CLAUSE 4 asserts ARRIVAL at the
+# composition root: each value must be handed down from the ONE validated parse
+# at the call site that reaches the tool layer.
+#
 # ---------------------------------------------------------------------------
 # WHAT THIS CHECK DELIBERATELY DOES NOT COVER, AND WHY
 # ---------------------------------------------------------------------------
@@ -86,7 +94,24 @@
 #   clean tree. This is a positive control, deliberately pointed at a file this
 #   lane does not touch and no future lane should empty.
 #
-# Exit 0 only when all three clauses pass. Exit 3 is a harness error (missing
+# CLAUSE 4 -- THE VALUES ARRIVE AT THE COMPOSITION ROOT.
+#   Absence proves the reader stopped guessing; arrival proves someone started
+#   telling it. Four anchored `rg` patterns, each of which must match EXACTLY
+#   once -- zero means unwired, more than once means two composition paths that
+#   can disagree, which is the doubled state this rung exists to end:
+#     server/main.ts       `ftsCorpusConfig: config.fts.corpusConfig`
+#     server/main.ts       `recoveryWalPath: config.recovery.walPath`
+#     server/main.ts       `natsRuntimeBoundary: nats.boundary`  -- the boundary
+#         the NATS phase ALREADY computed (`natsRuntimeBoundaryFromConfig` at
+#         server/main.ts:267, returned as `NatsPhase.boundary`). Calling that
+#         function a second time here would rebuild a second boundary from the
+#         same config, which is a second source of truth by construction.
+#     server/tools/search-brain.ts  `ftsCorpusConfig: dependencies.ftsCorpusConfig`
+#         -- the corpus default threaded into `resolveCallerFtsConfig`, which is
+#         what passes it as `requestFtsConfig`'s second argument. Without this
+#         the branch's new second parameter is never supplied by any caller.
+#
+# Exit 0 only when all four clauses pass. Exit 3 is a harness error (missing
 # tool / wrong repo root), which is NOT a fail of the thing under test.
 set -uo pipefail
 
@@ -119,6 +144,7 @@ CONTROL="server/config.ts"
 CLAUSE1=FAIL; CLAUSE1_EVIDENCE=""
 CLAUSE2=FAIL; CLAUSE2_EVIDENCE=""
 CLAUSE3=FAIL; CLAUSE3_EVIDENCE=""
+CLAUSE4=FAIL; CLAUSE4_EVIDENCE=""
 
 # ---------------------------------------------------------------------------
 # CLAUSE 1 — every target file is present.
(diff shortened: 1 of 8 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2b1-tool-readers-take-config.sh exit=0 sha=1d0b9b2b851ed61407b3c9ab146c70e511ee6ce0 at=2026-08-26T11:02:53.103Z

```
CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present under /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-779-27571mt9zj3iw
CLAUSE 2 (no process.env in any of the 4):              PASS — no process.env read in any of the 4 files — each value now arrives from validated config
CLAUSE 3 (scanner proven to match in server/config.ts):   PASS — server/config.ts still contains 4 process.env line(s) — the pattern, tool, and invocation are proven to match
```

</details>

rodaddy:
## Light-tier review — PR #779 (`feat/l2b1-inject-config-into-tool-readers`, head `1d0b9b2b`)

Read-only review. The four files do stop reading `process.env`, and the done-means
goes green. But **the other half of the rewiring — the composition-root injection —
is not in this diff.** `server/main.ts` passes none of the four new
`MemoryToolDependencies` fields, so three of the four values now resolve to a
hard-coded default instead of the operator's environment. Start-equivalence is
broken for two of them, and proven below by execution.

The root cause is one shape: every new field is **optional with a fallback**, so
`tsc` cannot flag the missing wiring and the absence-only done-means cannot see it.
The check passes for exactly the reason its own header warns about.

---

### P1 `server/tools/fts-config.ts:128` + `server/tools/search-brain.ts:178` -- `OPENBRAIN_FTS_CONFIG` is now ignored entirely; every deployment silently reverts to english

`requestFtsConfig` lost its `env` parameter in favour of `corpusDefault: FtsConfig =
DEFAULT_FTS_CONFIG`. The **only** caller was not updated:

```ts
// server/tools/search-brain.ts:178 -- unchanged by this PR
let ftsConfig = requestFtsConfig(requestedFtsConfig);
```

Previously the omitted second argument defaulted to `process.env` and read
`OPENBRAIN_FTS_CONFIG`. Now the omitted argument defaults to `"english"`.
`config.fts.corpusConfig` is built in `server/config.ts:360` and **never read by any
consumer** -- `rg` finds it only inside doc comments (`server/tools/types.ts:54`,
`server/tools/fts-config.ts:117`). `ftsCorpusConfig` on `MemoryToolDependencies` is
declared and never populated or consumed.

Reproduce (both probes run against this clone; the `main` probe imports
`git show origin/main:server/tools/fts-config.ts`):

```
HEAD 1d0b9b2b:  requestFtsConfig(undefined) with OPENBRAIN_FTS_CONFIG=german -> english
                requestFtsConfig("klingon")  with OPENBRAIN_FTS_CONFIG=german -> english
origin/main:    requestFtsConfig(undefined) with OPENBRAIN_FTS_CONFIG=german -> german
                corpusFtsConfig()                                            -> german
```

Any deployment that pinned a non-english corpus loses that pinning on this merge, in
silence. Note this also quietly retires the #368 finding-2 code path immediately
below the call site: the "config came from the operator env default" branch is now
unreachable, because the env default can no longer produce a non-english value.

Fix: thread `config.fts.corpusConfig` into `registerMemoryTools` at
`server/main.ts:168` and pass `dependencies.ftsCorpusConfig` at `search-brain.ts:178`.

---

### P1 `server/tools/operator-doctor.ts:82` -- `operator_doctor` now reports `http`/`not_runtime_available` on every path, including a real NATS deployment; and it now contradicts the REST surface

`dependencies.natsRuntimeBoundary ?? UNCONFIGURED_NATS_BOUNDARY` is correct code,
but nothing ever sets `dependencies.natsRuntimeBoundary`. `server/main.ts:168-175`
passes only `pool`, `embedFn`, `logger`, optional `embeddingModel`, and `...realtime`.
The `natsBoundary` built at `server/main.ts:243` goes to the **NATS bridge's**
`ToolDeps` (`src/tools/index.ts` shape, `main.ts:245-249`) and to the REST surface
(`main.ts:290-291`) -- neither of which reaches `server/tools/`. `rg` confirms
`registerMemoryTools`/`MemoryToolDependencies` appear nowhere in `src/nats-bridge.ts`
or `src/tools/index.ts`.

Measured boundary values:

```
readNatsRuntimeBoundary({})                                    -> transport=http, availability=not_runtime_available
readNatsRuntimeBoundary({OPENBRAIN_TRANSPORT:"nats",
  OPENBRAIN_NATS_ENABLE_BRIDGE:"true", OPENBRAIN_NATS_URL:...}) -> transport=nats, availability=available
```

So for `OPENBRAIN_TRANSPORT` unset / `http` the report is unchanged (`http`), and for
`nats` it changes from `nats`/`available` to `http`/`not_runtime_available`. That is
not the fix the PR describes. The PR says the old code "could report `nats` while
serving http"; the new code reports `http` while serving **nats**, which is the same
defect with the sign flipped — and it is now *worse*, because the REST doctor and the
`operator_doctor` MCP tool answer from two different boundaries in the same process
and will disagree.

This is P1 rather than P2 because it is untested: no test file is in the diff at all,
and no `server/tools/operator-doctor*.test.ts` exists.

Fix: pass `natsRuntimeBoundary: natsBoundary` into `registerMemoryTools`. That
requires `createServerFactory` to receive the boundary — it currently takes
`{pool, logger, config, tracing}` and could derive it via
`natsRuntimeBoundaryFromConfig(config.nats)` itself.

---

### P2 `server/tools/realtime-stores.ts:46` -- `recoveryWalPath` is never populated; harmless today only because the prod fallback is unreachable

`walPath: dependencies.recoveryWalPath ?? null` replaces a direct env read, and
`recoveryWalPath` is never set by any caller. Start-equivalence holds *by accident*:
`server/main.ts:150-155` always constructs `recoveryWalStore` eagerly (still reading
`process.env.OPENBRAIN_RECOVERY_WAL_PATH` at `main.ts:153`), so
`recoveryWalStoreFor`'s fallback branch is dead on the production path.

It is a real defect for any caller that registers tools without a full composition
root — the case the `types.ts:61` comment explicitly says this field exists to serve.
Those callers previously got the env-configured WAL path and now get RAM-only.
Downgraded to P2 because no such caller exists in-tree today (`rg` shows only
`realtime-append.ts:206,269` and `agent-context-pack.ts:227`, all reached from
`main.ts` with the store injected).

Note `main.ts:153` is itself now the *only* remaining `OPENBRAIN_RECOVERY_WAL_PATH`
read and it bypasses `config.recovery.walPath` — so the charter's single-parse rule
is still not satisfied for this value, just relocated.

---

### P2 `server/tools/types.ts:50` -- `searchEmbeddingTimeoutMs` is a dead field

Declared, documented as coming from `config.search.embeddingTimeoutMs`, and never
read. `server/tools/search-engine.ts:146-152` still resolves the timeout from
`process.env.OPENBRAIN_SEARCH_EMBEDDING_TIMEOUT_MS ?? process.env.SEARCH_EMBEDDING_TIMEOUT_MS`.

That is consistent with the deferral (`search-engine.ts` is out of scope, and the
header says so honestly), so behavior is unchanged — but shipping the dependency
field ahead of its consumer adds a second doubled state of exactly the kind the
script header argues against. Either land the field with its consumer or leave it
for the follow-up lane.

---

### P2 `scripts/done-means/750-l2b1-tool-readers-take-config.sh:1-203` -- the check passes vacuously in the sense that matters: it proves absence, never equivalence

Mechanically the script is sound and its three clauses do what they claim. I could
not make it pass falsely: clause 1 catches a rename, clause 3 is a genuine positive
control, `rg` status >= 2 is separated from "no match", and exit 3 is distinguished
from exit 1. The header does name both deferred files, their violation counts (5 and
9), and states plainly that green means four files, not a finished rung. That part is
good.

The gap is what it asserts. The lane's stated contract is *start-equivalence* — "the
same value reaches the same consumer" — and the check tests only that a string does
not appear in four files. Deleting the reads and wiring nothing up satisfies every
clause, which is precisely the state of this head: **exit 0 with two values
regressed.** A check that goes green on a broken rewiring is measuring the wrong
thing.

Suggested addition: one assertion per value that the composition root passes it,
e.g. `rg -n 'ftsCorpusConfig|natsRuntimeBoundary' server/main.ts` must hit — or
better, a unit test that registers the tools with a config carrying
`OPENBRAIN_FTS_CONFIG=german` and asserts the resolved config is `german`.

---

### P3 `server/tools/operator-doctor.ts:45-53` -- `UNCONFIGURED_NATS_BOUNDARY` is module-load state derived from a live reader

`readNatsRuntimeBoundary({})` at module scope is fine today (the `{}` input consults
no environment, as the comment says). Flagging only that it makes a module-load-time
call into `src/nats-runtime.ts`, which imports `./logger.ts` and
`./tools/agent-context-pack.ts` — a heavier import graph than a frozen literal would
need, evaluated at import regardless of whether the tool is ever called.

---

### Verified

```
verified: bash scripts/done-means/750-l2b1-tool-readers-take-config.sh -> exit 0; CLAUSE 1 PASS (4 files present), CLAUSE 2 PASS (no process.env), CLAUSE 3 PASS (server/config.ts has 4 process.env lines)
verified: bunx tsc --noEmit -> exit 0, 0 lines of output (every new dep field is optional, so the missing wiring cannot fail the typecheck)
verified: bun test server/tools/context-pack.test.ts -> 28 pass, 0 fail, 100 expect() calls
verified: git diff --name-only origin/main...HEAD -> 5 files, ZERO test files
verified: rg -n 'ftsCorpusConfig|searchEmbeddingTimeoutMs|recoveryWalPath|natsRuntimeBoundary' server/main.ts -> only lines 48/243/248, all for the NATS bridge ToolDeps; none passed to registerMemoryTools
verified: rg --type ts -g '!**/*.test.ts' -g '!server/config/env-groups.ts' 'config\.fts|config\.search\b|config\.recovery' server/ src/ -> only server/config.ts construction + doc comments; no consumer
verified: bun probe importing HEAD server/tools/fts-config.ts with OPENBRAIN_FTS_CONFIG=german -> requestFtsConfig(undefined)="english"
verified: bun probe importing origin/main server/tools/fts-config.ts with OPENBRAIN_FTS_CONFIG=german -> requestFtsConfig(undefined)="german", corpusFtsConfig()="german"
verified: bun probe readNatsRuntimeBoundary({}) -> http/not_runtime_available; readNatsRuntimeBoundary({TRANSPORT:nats,ENABLE_BRIDGE:true,URL:...}) -> nats/available
verified: rg -n 'requestFtsConfig' server/ -> single call site at server/tools/search-brain.ts:178, passing one argument
verified: rg -n 'recoveryWalStoreFor' server/ src/ -> realtime-append.ts:206,269 and agent-context-pack.ts:227, all reached with the store injected from main.ts:152
```

Not checked: full suite (`bun run test:isolated`) — out of scope for this tier; no
`.pg.test.ts` in the diff to target.

**VERDICT: BLOCK** — two P1 start-equivalence breaks (`OPENBRAIN_FTS_CONFIG` ignored;
`operator_doctor` reports the wrong transport and contradicts the REST surface). Both
are fixed by the same missing change: pass the four values into `registerMemoryTools`
at `server/main.ts:168`. The done-means should grow an assertion that the values
arrive, so this class of half-rewiring cannot go green again.

rodaddy:
Head verdict: BLOCK stands. Both P1s are the missing composition-root half (server/main.ts:168), and main.ts cannot be staged until its 3 pre-existing lint violations are paid. Converted to draft pending the lint-debt ruling (issue filed with the measured list). The fixer commit for this PR lands together with the main.ts remediation lane; the done-means will gain an arrival assertion per value (config -> registerMemoryTools) so absence-only cannot go green again.

rodaddy:
Ruling issue: #780.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2b1-tool-readers-take-config.sh exit=0 sha=fd6e29495ddd48633102691dd416af4abc384ee1 at=2026-08-26T17:28:46.337Z

```
CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present under /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-779-26349mtadbcka
CLAUSE 2 (no process.env in any of the 4):              PASS — no process.env read in any of the 4 files — each value now arrives from validated config
CLAUSE 3 (scanner proven to match in server/config.ts):   PASS — server/config.ts still contains 4 process.env line(s) — the pattern, tool, and invocation are proven to match
CLAUSE 4 (values arrive at the composition root):        PASS — all 4 values are passed exactly once from the composition root
CLAUSE 5 (injected values observed at their consumers):  PASS — 4 test(s) passed in server/tools/dependency-arrival.test.ts — each injected value observed at its consumer, not just at the call site
```

</details>

rodaddy:
harvested: b9a981e docs/lane-contract.md round 36 (two-halves lesson: a rewiring lane is the reader half plus the composition-root half, and the done-means asserts ARRIVAL; the #779 review's two P1s)

rodaddy:
CI on fd6e294 (runs 32994328337 / 32994354301): db-integration FAIL on the new case "recoveryWalPath arrives at the fallback recovery WAL store" (existsSync false). Two causes, both from the 5b brief: the test built its path under the Mac scratch workspace, absent on the runner; and server/tools/realtime-stores.ts:40-49 memoizes the fallback store with ??= so a path arriving second in one bun process is ignored (the whole suite runs in one process in CI; the file ran alone locally and passed). python-capture failed on both runs (#764 flake).

Fix lane 5c dispatched (native Opus 5 low): portable os.tmpdir() path, fallback keyed by walPath with an order-independence case written RED-first. The receipt above is bound to fd6e294 and goes stale on the next push; verify-lane re-runs after.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2b1-tool-readers-take-config.sh exit=0 sha=d0bbbd32afccda5d111ac56183a410726580c449 at=2026-08-26T17:38:51.879Z

```
CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present under /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-779-4843mtadobt8
CLAUSE 2 (no process.env in any of the 4):              PASS — no process.env read in any of the 4 files — each value now arrives from validated config
CLAUSE 3 (scanner proven to match in server/config.ts):   PASS — server/config.ts still contains 4 process.env line(s) — the pattern, tool, and invocation are proven to match
CLAUSE 4 (values arrive at the composition root):        PASS — all 4 values are passed exactly once from the composition root
CLAUSE 5 (injected values observed at their consumers):  PASS — 5 test(s) passed in server/tools/dependency-arrival.test.ts — each injected value observed at its consumer, not just at the call site
```

</details>

Issue #780, opened 2026-08-26
ruling: whole-file lint gate blocks every remaining L2 lane (main.ts, search-*.ts, langfuse-tracing.ts debt)


rodaddy:
## Ruling needed: the whole-file lint gate blocks every remaining L2 lane

`.oxlintrc.json:17-25` (merged in #771) lints STAGED FILES WHOLE and states the
backlog "is paid per-file as work naturally touches them". L2b-1 (#779)
measured what that costs on the files rung L2 has to touch, and the answer is
that the rewiring cannot land without first paying the debt in four files.

### What happened

- #778 (L2a) landed the schema half: 23 env names typed on `ServerConfig`.
- #779 (L2b-1) landed four tool readers taking config, but the composition-root
  half (`server/main.ts:168` passing the values into `registerMemoryTools`)
  could not be committed: `server/main.ts` carries 3 pre-existing violations
  and the hook refuses any commit that stages the file. Review found two P1
  start-equivalence breaks caused by exactly that missing half
  (https://github.com/rodaddy/open-brain/pull/779#issuecomment-5424424539).
  #779 is a draft until this is settled.
- L2b-2 (shared-namespace, 25 call sites), L2c (main.ts, tracing, capture) and
  L2d all touch the same files.

### The debt, measured against clean `origin/main` (`49ecfbe`)

```
server/main.ts:207:8   complexity              -- async fn `startServer`, complexity 22 (max 10)
server/main.ts:207:8   max-lines-per-function  -- async fn `startServer`, 140 lines (max 100)
server/main.ts:421:24  max-params              -- `shutdown`, 5 params (max 4)

server/tools/search-brain.ts:77:8   max-lines-per-function -- `registerSearchBrainTool`, 142 lines (max 100)
server/tools/search-brain.ts:155:5  complexity             -- async fn, complexity 18 (max 10)

server/tools/search-all.ts:136:33  max-params              -- `searchQmdInternal`, 5 params (max 4)
server/tools/search-all.ts:237:19  max-params              -- `searchQmd`, 5 params (max 4)
server/tools/search-all.ts:285:8   max-lines-per-function  -- `registerSearchAllTool`, 203 lines (max 100)
server/tools/search-all.ts:344:5   complexity              -- async fn, complexity 19 (max 10)
server/tools/search-all.ts:344:5   max-lines-per-function  -- async fn, 142 lines (max 100)

[...]

PR #785, opened 2026-08-26
docs(handoff): harvest round 36 and add rules 28-30 (delegate, graph mode, scribe)

rodaddy:
## Summary

- Docs-only. Two files change, no runtime code and no schema.
- `docs/lane-contract.md` gains Tightenings round 36 at line 90, harvesting lane 1 of the #780 sweep (PR #782: a git pathspec of `**` skips files sitting directly under `server/`), the #779 review (a rewiring lane's done-means has to assert arrival, not departure), and the session-2 collection pass (Codex companion git guard, stuck rerun, reopen-to-trigger, verify-lane worktree cleanup, and the harvest gate reading the whole command line).
- **Harvest statement for the merge gate:** the harvest IS this PR — `docs/lane-contract.md` round 36 is the harvested lane report, so the mandatory merge-pass harvest step is satisfied by the diff itself and needs no separate follow-up commit.
- `_DOCS/HANDOFF-RULES.md` gains rules 28-30 from Rico's 2026-08-26 rulings: the head delegates by default (28, line 148), graph mode is declared before the first mutation (29, line 158), and a scribe runs from State 0 (30, line 165).

## Verification

- Done-means: scripts/done-means/handoff-validates.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

Tests, typecheck, and migrations are not applicable in substance: the diff adds
prose lines to two Markdown files and touches no TypeScript, no Python, no SQL,
and no migration. The Python package and the live service are untouched, so
neither check has an input.

**The named done-means check FAILS on this branch, and that is reported rather
than worked around.** `scripts/done-means/handoff-validates.sh` validates
handoff DOCUMENTS under `_DOCS/_handoff/`; this PR changes the handoff RULES
file and the lane contract, and adds no handoff document. The script treats an
empty file list as a hard FAIL by design, so it cannot report a pass here:

```text
$ cd /Volumes/ThunderBolt/_tmp/open-brain/_scratch/clone-20260825
$ bash scripts/done-means/handoff-validates.sh ; echo exit=$?
FAIL: no handoff files to check (nothing under _DOCS/_handoff/ differs from origin/main)
exit=1
```

That is the check working as written (see its header: "An empty file list is a
FAIL, so the check cannot pass by examining nothing"), not a defect introduced
by this PR. It is the wrong judge for a rules/contract diff, and picking a
different pre-existing check that never judged this lane would be a false
[...]

PR #788, opened 2026-08-26
refactor(search-all): split the federated search tool into named helpers (#780)

rodaddy:
## Summary

- Lane 3 of the #780 lint sweep: `server/tools/search-all.ts` brought fully to
  standard. RED at `fca45fb` was 5 oxlint findings on that one file —
  `max-params` twice (`searchQmdInternal`, `searchQmd`), `max-lines-per-function`
  on `registerSearchAllTool` at 203 lines, and complexity 19 over 142 lines on
  the request handler. GREEN is `oxlint --deny-warnings` exit 0 across
  `search-all.ts`, `search-brain.ts`, and `search-request.ts`.
- Under the sprint's maximum-reuse ruling, the argument-defaulting logic the two
  search tools already shared was moved to a new `server/tools/search-request.ts`
  exporting `normalizeSearchArgs` and `NormalizedSearchArgs`. The five common
  defaults were byte-identical before the move; `search-brain.ts` changes by
  import only. `fts_config` stays in the shared shape as a superset that
  `search_all` never reads — `search_all` publishes no such argument and keeps
  passing `{}` to the search engine, the English path, unchanged.
- The over-long functions were split into named helpers rather than shortened:
  a `QmdSearchOptions` options struct replaces positional arguments, and
  `planFederation`, `searchBrainArm`, `fuseFederatedResults`, and
  `selectRankedWindow` carry the handler's phases. `searchAllInputSchema` and
  `searchAllAnnotations` were hoisted out of the registration function with
  their description strings identical.
- Deliberately NOT reused: `resolveCallerFtsConfig` and `respondToSearchFailure`
  from `search-brain.ts`. The `search_all` catch path returns `[]` and warns
  `search_all_brain_failed` so federation degrades visibly, which is different
  behavior from the single-brain error response, and folding them together would
  have changed one of the two.
- Ranking arithmetic, result push order, the trace span name and its fields, the
  published schema, the defaults, and the error strings are all unchanged. This
  is a structural refactor with no contract movement.

## Verification

RUNNING this session, in the branch clone at `dc5b82c`:

```text
bun scripts/done-means/780-touched-files-lint-clean.sh   -> exit 0 (both runs)
  linting 3 file(s) from: git diff --name-only origin/main...HEAD
    - server/tools/search-all.ts
    - server/tools/search-brain.ts
    - server/tools/search-request.ts

bunx oxlint --deny-warnings server/tools/{search-all,search-brain,search-request}.ts -> exit 0
bunx tsc --noEmit                                   -> clean
bun test server/tools                               -> 94 pass / 92 skip / 0 fail
bun test src/tools/__tests__/search-all.test.ts     -> 38 pass
bun test server/observability/langfuse-tracing.test.ts -> 68 pass
```

The done-means list is diff-derived, so it names the three files this branch
actually touched rather than a hand-kept list.

Test files are unmodified by this branch. No new tests were added because no
helper extracted here gained a public boundary that the existing tool-surface
suites do not already drive through the registered tool. The Postgres-backed
files skipped locally for want of `OPENBRAIN_TEST_DATABASE_URL`; the CI
db-integration job is the evidence for those, not this local run.

Python package: untouched by this diff (three TypeScript files under
`server/tools/`), so its checks are not applicable.

## Critical Self-Review

- Highest-risk behavior: federated ranking order drift inside
  `fuseFederatedResults` and `selectRankedWindow`. These two carry the arithmetic
  that decides which results survive and in what order across the brain and qmd
  arms, so a transcription slip in the extraction would silently reorder or drop
  results with no error anywhere. Covered by the ranking and pagination cases in
  `src/tools/__tests__/search-all.test.ts` (38 pass) and by the pre-sort index
  assertion in `server/observability/langfuse-tracing.test.ts` (68 pass), which
  pins the trace fields to the pre-sort positions.
- Assumptions that could be wrong: that the five shared defaults really were
  byte-identical between the two tools before the move, so hoisting them cannot
  change either tool's behavior. Checked by reading the removed block in
  `git diff fca45fb..dc5b82c -- server/tools/search-brain.ts` against the new
  `search-request.ts` body. Second assumption: that `search_all` never reads
  `fts_config`, which is why leaving it in the shared shape is safe as a
  superset — it normalizes to `undefined` there and the engine still receives
  `{}`.
- Missing/weak tests: there is no direct unit test on `planFederation` or
  `selectRankedWindow` as standalone functions; they are exercised only through
  the registered tool. That is sufficient here because they are private helpers
  with no exported surface, and the tool-level suites already drive both the
  federated and brain-only paths plus pagination. The genuine gap is that the
  Postgres-backed search tests did not execute on this machine, so the local run
  proves the non-database paths only.
- Security/permission risk: none introduced. The namespace read-scope predicate
  and the auth-derived filtering stayed inside `searchBrainArm` exactly as they
  were in the inline handler; no check moved to a caller and none became
  conditional. `resolveCallerFtsConfig` was deliberately not pulled in, so the
  non-English privilege boundary in `search-brain.ts` is untouched.
- Migration/deploy risk: none. No migration, no schema change, no config key, no
  new dependency. The new file is a plain TypeScript module imported by two
  existing ones.
- Downstream client/runtime risk: none identified. The published input schema,
  the annotations, the tool name, the output shape, and the error strings are
  unchanged, so an mcp2cli or Hermes consumer sees an identical surface.
- Rollback/cleanup concern: reverting is a single-commit revert of `dc5b82c`;
  the only cross-file coupling is the `search-brain.ts` import of
  `search-request.ts`, which the same revert undoes. No state, no artifact, and
  no generated file to clean up.
- Fixes made before PR: the `?? default` resolution in `resolveQmdPath` treats an
  empty `QMD_PATH` as configured, which had the qmd arm spawning `bun "" search`
  and failing open; it now trims to `undefined` so an unusable value disables
  federation at the call site instead of inside a subprocess.
- Known residual risk: `server/tools/search-all.ts` still reads `process.env`
  directly for the qmd path fallback (`server/tools/search-all.ts:124`, verified
  with `rg -n "process\.env"`). That is the L2b-2 config-injection lane's work,
  not this lint lane's, and it is left in place deliberately. Also remaining for
  rung L2 after this: `server/tools/search-engine.ts`, which is the next handoff
  as an L4 split and re-adds `searchEmbeddingTimeoutMs`.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane came from the #780 lint sweep frontier rather than a review finding, and produced no MEDIUM-or-above finding to record.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: the diff is a structural refactor of three TypeScript files with no schema, transport, or behavior change, so a live service check would exercise nothing this branch altered.

## Contract Parity

- Contract parity: [x] runtime-specific because: no path in `contracts/parity-paths.txt` is touched. That file lists `python/openbrain-memory/`, `clients/ts/`, `contracts/`, `src/contract.ts`, and `src/contract-schemas.ts`; this diff is confined to `server/tools/search-all.ts`, `server/tools/search-brain.ts`, and `server/tools/search-request.ts`.

## Downstream Rollout

Notes/evidence:

- Not applicable against the doc's "When This Applies" list. No MCP tool name,
  schema, output shape, auth rule, namespace semantic, or error envelope moved;
  transport and session behavior are untouched; there is no externally visible
  migration; `python/openbrain-memory` is not in the diff; no generated `skill/`
  content changed. The three touched files are internal server tool
  implementation, and the surface an mcp2cli or Hermes consumer calls is
  byte-identical before and after.

### DIFF
--- a/server/tools/search-all.ts
+++ b/server/tools/search-all.ts
@@ -32,7 +32,11 @@ import {
   textResult,
   type MemoryToolDependencies,
 } from "./types.ts";
-import { canReadNamespace, namespaceFilterFor } from "./read-scope.ts";
+import {
+  canReadNamespace,
+  type NamespaceFilter,
+  namespaceFilterFor,
+} from "./read-scope.ts";
 import { isSharedNamespace } from "./shared-namespace.ts";
 import {
   ALL_TABLES,
@@ -51,6 +55,7 @@ import {
   type SearchRow,
   type SourceRef,
 } from "./search-engine.ts";
+import { normalizeSearchArgs } from "./search-request.ts";
 
 const NOT_AUTHENTICATED = "Permission denied: not authenticated";
 const NAMESPACE_DENIED = "Permission denied: namespace read access denied";
@@ -122,6 +127,17 @@ export function resolveQmdPath(
   return configured ? configured : undefined;
 }
 
+/** One qmd federation request; a named struct so no caller tracks argument order. */
+interface QmdSearchOptions {
+  dependencies: MemoryToolDependencies;
+  /** Resolved qmd entry point; never empty (see resolveQmdPath). */
+  qmdPath: string;
+  query: string;
+  /** How many hits to ask qmd for. */
+  limit: number;
+  collection: string | undefined;
+}
+
 /**
  * Search the qmd file index, returning `[]` on every failure.
  *
@@ -130,16 +146,11 @@ export function resolveQmdPath(
  * federated search. Each is logged, so the degradation is visible rather than
  * merely quiet — the design's R4 objection is to SILENT staleness, not to
  * degradation itself.
- *
- * @param qmdPath Resolved qmd entry point; never empty (see resolveQmdPath).
  */
 async function searchQmdInternal(
-  dependencies: MemoryToolDependencies,
-  qmdPath: string,
-  query: string,
-  limit: number,
-  collection: string | undefined,
+  options: QmdSearchOptions,
 ): Promise<UnifiedResult[]> {
+  const { dependencies, qmdPath, query, limit, collection } = options;
   try {
     const command = [
       "bun",
@@ -234,19 +245,13 @@ async function searchQmdInternal(
   }
 }
 
-function searchQmd(
-  dependencies: MemoryToolDependencies,
-  qmdPath: string,
-  query: string,
-  limit: number,
-  collection: string | undefined,
-): Promise<UnifiedResult[]> {
+function searchQmd(options: QmdSearchOptions): Promise<UnifiedResult[]> {
+  const { query, limit, collection } = options;
   return traceRetrievalSpan({
     name: "retrieval.qmd_query",
     input: { query, limit, collection },
     metadata: { stage: "candidate_generation", source: "qmd" },
-    run: () =>
-      searchQmdInternal(dependencies, qmdPath, query, limit, collection),
+    run: () => searchQmdInternal(options),
     output: (rows) => ({
       count: rows.length,
       candidates: rows.map((row) => ({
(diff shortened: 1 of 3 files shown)


rodaddy:
harvested: #785 docs/lane-contract.md round 37 (shared extraction across twins: normalizeSearchArgs moved to server/tools/search-request.ts only after confirming the defaults byte-identical; search-all keeps its own catch shape, so tool-failure.ts was not forced onto it)

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=dc5b82cb6444b4b1ef264fd58a474b2e662b713c at=2026-08-26T17:57:32.470Z

```
linting 3 file(s) from: git diff --name-only origin/main...HEAD
  - server/tools/search-all.ts
  - server/tools/search-brain.ts
  - server/tools/search-request.ts
```

</details>

PR #791, opened 2026-08-26
refactor(ingest-conversation-facts): bring server/tools/ingest-conversation-facts.ts to the lint standard (#780)

rodaddy:
## Summary

- Brings `server/tools/ingest-conversation-facts.ts` to the repo lint standard for #780. The MCP handler was one 279-line async closure with a complexity of 30, carrying all four gates, the embedding pass, the transactional insert loop, and the receipt in a single scope.
- Split along the boundaries the file's own header comment already names. Each gate is now a named function that either refuses with a structured error or passes: `resolveWriteTarget` (auth plus namespace authority), `refuseRawTranscript`, `refuseSecrets`, then `resolveApprovedSource` and `resolveTargetLane`. The write side is `prepareUnits`, `writeUnit`, `writeBatch`, and `ingestReceipt`, sequenced by `ingestApprovedBatch` so the database-error boundary wraps one call rather than two hundred lines.
- Helpers past two arguments take an `IngestContext` (pool, logger, resolved identity, physical namespace), so no signature reaches a fifth parameter.
- The forbidden non-null assertion is gone: the secret scan iterates `facts.entries()` instead of indexing `args.facts[index]!`. The 4-deep block in the insert loop flattens because the duplicate-evidence branch became its own function.
- **No behavior moves.** Schemas, defaults, error codes and messages, SQL text and parameter order, and every log event name are identical. The extraction is structural only.
- Reuse check before extracting, per the maximum-reuse ruling: `rg` across `server/tools/` and `src/tools/` shows `writerProvenance` and `registryAuth` are local to this file (the older `src/tools/` twin carries its own separate copies and is out of scope for this lane), and the genuinely shared pieces — schemas, the raw-key set, `mergeDuplicateEvidence`, `safeErrorClass` — already live in `server/tools/conversation-facts-contract.ts`, which this file imports. Nothing new earned a place there: every extracted function closes over this server's auth and dependency shapes, so hoisting one would create a shared symbol with a single caller.

Only the one target file is modified. No sibling file was added and no other file was touched.

## Verification

RED, on the untouched file at `origin/main` — 5 findings:

```
server/tools/ingest-conversation-facts.ts:100:8: error eslint(max-lines-per-function): The function `registerIngestConversationFactsTool` has too many lines (304). Maximum allowed is 100.
server/tools/ingest-conversation-facts.ts:123:5: error eslint(complexity): async function has a complexity of 30. Maximum allowed is 10.
server/tools/ingest-conversation-facts.ts:123:5: error eslint(max-lines-per-function): The async function has too many lines (279). Maximum allowed is 100.
server/tools/ingest-conversation-facts.ts:174:22: error typescript(no-non-null-assertion): Forbidden non-null assertion.
server/tools/ingest-conversation-facts.ts:346:13: error eslint(max-depth): Blocks are nested too deeply (4). Maximum allowed is 3.
```

`DONE_MEANS_780_FILES=server/tools/ingest-conversation-facts.ts bash scripts/done-means/780-touched-files-lint-clean.sh` on that tree exited 1 with `FAIL: oxlint reported findings in the file(s) listed above.`

GREEN, on this branch:

- `./node_modules/.bin/oxlint --deny-warnings server/tools/ingest-conversation-facts.ts` -> exit 0, no output. Only this file was modified, so it is the whole touched set.
- `bun run test:isolated src/tools/__tests__/ingest-conversation-facts.test.ts src/tools/__tests__/ingest-conversation-facts.pg.test.ts server/tools/` -> exit 0, **188 pass, 0 fail**, 713 expect() calls across 19 files, against a freshly created and migrated database. Existing tests are unmodified.
- `bash scripts/done-means/780-touched-files-lint-clean.sh` (diff-derived from `origin/main...HEAD`) -> exit 0, `PASS: every file this branch touched under server/ lints clean.`

Re-ran typecheck and the isolated suite after committing, because the pre-commit hook reformats with prettier and the receipt must describe the committed content, not the pre-hook content. Both still exit 0. File is 497 code lines, under the 500 standard.

No new test was added: every extracted function is a private helper whose behavior is already exercised end to end through the registered tool by the two `ingest-conversation-facts` suites, which cover each refusal path and the duplicate-merge dispositions. A new unit test over a private helper would pin the refactor's shape rather than any behavior, and would have to be deleted the next time the seams move.

## Critical Self-Review

- Highest-risk behavior: the transactional write loop. `writeBatch` still opens the client, runs `BEGIN`, loops, and `COMMIT`s, with `ROLLBACK` in `catch` and `client.release()` in `finally` — the all-or-nothing property and the release are preserved exactly. The risk in a split like this is a rollback path that stops covering the loop; here the loop and both handlers stayed inside the same function, and only the per-unit insert moved out into `writeUnit`, which is called from inside the `try`. So a throw from `writeUnit` still unwinds through the same `ROLLBACK`.
- Assumptions that could be wrong: (1) that `identity` is non-undefined once `resolveWriteTarget` returns a namespace. That is true by construction — the function refuses `!identity` on its first branch — but it is expressed as an `as AuthIdentity` cast at the call site with a comment, because narrowing does not flow out of the helper's return type. A future edit that adds a passing path to `resolveWriteTarget` without an identity would make the cast a lie. (2) That prettier's reformat did not alter semantics; checked by re-running typecheck and the suite after the commit.
- Missing/weak tests: `server/tools/` has no test file that directly names `registerIngestConversationFactsTool`. The coverage the receipt cites for this contract runs through the `src/tools/` twin's suites and the shared `conversation-facts-contract.ts`. That gap predates this PR and is not widened by it, but it does mean the server registration wrapper itself is covered only indirectly.
- Security/permission risk: this file IS a security boundary — four gates, one of them namespace isolation. The gates are unchanged in content, in order, and in their refusal shapes. `resolveWriteTarget` runs auth then namespace, then the raw-transcript scan, then the whole-batch secret scan, then source approval and the seven-coordinate lane predicate, exactly as before. The SQL is byte-identical including the `$1..$7` parameter order and `IS NOT DISTINCT FROM $7::text`, so no predicate was dropped or reordered. The secret scan still refuses the entire batch before any write.
- Migration/deploy risk: none. No schema change, no migration, no config, no new dependency.
- Downstream client/runtime risk: not applicable, with a reason rather than a shrug. `docs/downstream-rollout.md` scopes rollout to MCP tool/schema/protocol/client-facing changes. The tool name, description string, `inputSchema`, and annotations are unchanged; every response body is produced by the same `textResult`/`ingestError` calls with the same fields; every error code and message string is identical. Nothing crosses the wire differently, so rtech-mcps, mcp2cli, and the rtech-hermes runtime see no change.
- Rollback/cleanup concern: a single-commit, single-file, behavior-preserving refactor; reverting the commit fully restores the prior file. No state, no artifact, no partial migration to unwind.
- Known residual risk: the `as AuthIdentity` cast noted above is the one place where a compiler-checked invariant became a human-checked one — the alternative was to return the identity back out of `resolveWriteTarget`, which reads worse for the same guarantee. Also, the older `src/tools/ingest-conversation-facts.ts` twin still carries its own copies of `writerProvenance` and `registryAuth` and its own inlined contract, and still lints with 9 findings. That duplication is real and worth collapsing, but doing it here would be a behavior-bearing change to a second file outside this lane's scope.
- Fixes made before PR: none beyond the refactor itself; no defect surfaced during the split.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no review or post-merge finding surfaced a missed pattern here. This is a mechanical lint-standard sweep with no behavior change, so there is nothing for the next swarm to check for that the lint rules do not already enforce.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or schema behavior changes, so a live smoke would exercise the identical code path and prove nothing this branch could have broken.

## Contract Parity

- Contract parity: [x] runtime-specific because: checked `contracts/parity-paths.txt`, which lists `python/openbrain-memory/`, `clients/ts/`, `contracts/`, `src/contract.ts`, and `src/contract-schemas.ts`. This PR touches exactly one file, `server/tools/ingest-conversation-facts.ts`, which is under none of them, and the shared contract module it imports is unmodified. No fixture can have drifted.

## Downstream Rollout

Notes/evidence:

- All four are not applicable for one reason: nothing client-facing changed. The registered tool name, description, `inputSchema`, and annotations are identical, and every success and refusal payload is emitted by the same call with the same fields and the same strings. Downstream consumers cannot observe this diff.

### DIFF
--- a/server/tools/ingest-conversation-facts.ts
+++ b/server/tools/ingest-conversation-facts.ts
@@ -37,7 +37,11 @@ import { physicalNamespace } from "../../src/shared-namespace.ts";
 import { contentHash, EMBEDDING_MODEL } from "../../src/embedding.ts";
 import { containsSecret } from "../../src/sharing.ts";
 import { resolveIngestionEligibility } from "../../src/source-registry.ts";
-import { authIdentity, textResult, type MemoryToolDependencies } from "./types.ts";
+import {
+  authIdentity,
+  textResult,
+  type MemoryToolDependencies,
+} from "./types.ts";
 import {
   conversationIngestSchema,
   findRawTranscriptKey,
@@ -45,6 +49,7 @@ import {
   safeErrorClass,
   type ConversationFactEventType,
   type ConversationIngestArgs,
+  type Queryable,
   type UnitDisposition,
 } from "./conversation-facts-contract.ts";
 
@@ -79,7 +84,8 @@ function ingestError(
 
 /** Writer provenance, mirroring `append_session_event`. */
 function writerProvenance(identity: AuthIdentity) {
-  const namespaceSource = identity.namespaceSource === "delegated" ? "header" : "token";
+  const namespaceSource =
+    identity.namespaceSource === "delegated" ? "header" : "token";
   return {
     writer_identity: identity.clientId,
     token_identity: identity.tokenClientId ?? identity.clientId,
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: a lint sweep receipt must be re-taken AFTER the commit, not before it -- the pre-commit hook reformats the staged file with prettier, so the typecheck and test run from before the commit describe content that no longer exists on the branch; re-running tsc and test:isolated post-commit is what makes the receipt describe what reviewers will actually read.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=1fe2c90a4bfe74c0a93a9ebc05aa96a14bc1b22f at=2026-08-26T18:18:09.251Z

```
  - server/tools/ingest-conversation-facts.ts
```

</details>

PR #804, opened 2026-08-26
refactor(resolve-entry): bring server/tools/resolve-entry.ts to the lint standard (#780)

rodaddy:
## Summary

- Brings `server/tools/resolve-entry.ts` to the repo lint standard (#780), wave 2 lane 6. Two oxlint findings cleared, no behavior change.
- `findRow` took 7 positional parameters. Its per-probe inputs now travel as a single `ProbeRequest` object, and the namespace-scoping half moved into a `scopeParameters` helper. Same clauses, same parameter positions, same SQL text.
- The registered handler had a complexity of 11. It is now registration only; the work moved to module-level `resolveEntry`, with `resolvedHit` shaping the success payload and `resolveArchived` owning the admin-tier archived second pass.
- Schema and annotations hoisted to module consts (`resolveEntryInputSchema`, `resolveEntryAnnotations`) with identical description strings and rule values, matching lane 2 (`search-brain.ts`) and lane 3 (`search-all.ts`, 011bf24). The "Permission denied: cannot resolve entries" string became a named `NOT_AUTHENTICATED` const with the identical value.
- No other existing file was touched. No new shared module: every helper is private to this file, because no second caller exists now.

## Verification

RED (untouched file, before any edit):

```
server/tools/resolve-entry.ts:60:23: error eslint(max-params): Function 'findRow' has too many parameters (7). Maximum allowed is 4.
server/tools/resolve-entry.ts:117:5: error eslint(complexity): async function has a complexity of 11. Maximum allowed is 10.
```

`DONE_MEANS_780_FILES=server/tools/resolve-entry.ts bash scripts/done-means/780-touched-files-lint-clean.sh` -> exit 1, `FAIL: oxlint reported findings in the file(s) listed above.`

GREEN (re-run on the COMMITTED tree at ddd4557, after the pre-commit prettier pass, working tree clean):

- `./node_modules/.bin/oxlint --deny-warnings server/tools/resolve-entry.ts` -> exit 0, no findings
- `bash scripts/done-means/780-touched-files-lint-clean.sh` -> exit 0, diff-derived, `PASS: every file this branch touched under server/ lints clean.`
- `bun run test:isolated src/tools/__tests__/resolve-entry.test.ts src/contract.test.ts server/tools` -> 180 pass, 0 fail, 750 expect() calls across 19 files, tests exited 0

Python package: not applicable, `python/openbrain-memory/` is untouched. Live smoke: not applicable, this is a source-internal refactor with no wire-visible change.

## Critical Self-Review

- Highest-risk behavior: the namespace predicate on the ID-based read. This is a security boundary, not a formatting detail — an `id` is caller-supplied and globally unique, so an unscoped probe would let any identity confirm and read a row in a namespace it has no read authority over. Both branches are preserved exactly: an explicit `namespace` is bound as `$2` only after `canTargetNamespace(identity, "read", namespace)` clears it, and otherwise `namespacePredicate(identity, "read", 2)` supplies the auth-derived clause and values. Moving that into `scopeParameters` was deliberate — it puts both branches side by side in one named function with a doc comment stating the invariant, so a future edit that drops one is visible rather than buried in a probe loop.
- Assumptions that could be wrong: that `scopeParameters` produces byte-identical SQL per iteration. It is rebuilt inside the loop exactly as before rather than hoisted, which matters because it mutates a fresh `params` array per table; hoisting it would have shared one array across probes and shifted parameter positions. I kept the per-iteration construction for that reason. Also assumed `resolveArchived` returning `null` is indistinguishable from the old inline fall-through — it is, both land on the same `not_found_or_unreadable` result.
- Missing/weak tests: no test was added. The existing `src/tools/__tests__/resolve-entry.test.ts` pins the behavior of the `src/` twin, not the `server/` file directly; the `server/` file's namespace scoping is covered indirectly by `server/tools/namespace-denial.test.ts` and `server/tools/ported-tools-scope.test.ts`, which ran green. A targeted `server/`-side test that a non-admin identity cannot resolve an id outside its namespace would be stronger than what exists, and this PR does not add one — it is a pure refactor lane and adding coverage would mix concerns, but the gap is real and named here.
- Security/permission risk: low, and reviewed specifically. The auth gate (`authIdentity` -> `errorResult`), the `canTargetNamespace` check, the `canRead` table filter, and the admin/ob-admin restriction on the archived pass are all preserved with identical ordering and identical short-circuit behavior. The archived pass still runs only for `admin`/`ob-admin`, so archived rows stay invisible to everyone else exactly as an absent row would be. The diagnostics contract is unchanged: `checked_sources`/`checked_tables` still describe what the caller was allowed to search, never metadata from an unreadable row.
- Migration/deploy risk: none. No schema, no migration, no config, no startup path touched.
- Downstream client/runtime risk: none. No MCP tool name, description, input schema, annotation, error string, or response field changed, so the wire contract is byte-identical and no client can observe this.
- Rollback/cleanup concern: none beyond reverting the single commit. One file, no shared module added, nothing else imports the new helpers.
- Fixes made before PR: caught and rejected a reuse that would have changed behavior — `read-scope.ts`'s `appendReadNamespacePredicate` emits `= ANY($n::text[])` against a caller-named column and takes an `includeLegacySharedFallback` option, which is not the same clause `namespacePredicate(identity, "read", 2)` produces here. Swapping it in would have altered SQL under a refactor label, so `namespacePredicate` stayed. Also checked `normalizeSearchArgs` (`search-request.ts`) and `respondToSearchFailure` (`tool-failure.ts`) per the reuse-first rule and rejected both: this tool has no query/limit/offset/mode args for the former, and no catch block for the latter.
- Known residual risk: the `server/` and `src/` copies of this tool remain separate implementations; this PR changed only the `server/` one, so the two now differ structurally while remaining behaviorally equivalent. That duplication predates this lane and is not introduced here.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane produced no new MEDIUM+ review finding — it is a mechanical lint-standard refactor with no behavior change, and the one judgment call it did make (rejecting a same-named-looking helper whose emitted SQL differs) is already covered by the existing "nothing is adjusted silently" and minimal-correct-change entries rather than being a new pattern.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no wire-visible, schema, or runtime-behavior change — the MCP contract is byte-identical, so a live smoke would exercise nothing this PR could have broken.

## Contract Parity

- Contract parity: [x] runtime-specific because: no path in `contracts/parity-paths.txt` is touched. This PR changes only `server/tools/resolve-entry.ts`; `python/openbrain-memory/`, `clients/ts/`, `contracts/`, `src/contract.ts`, and `src/contract-schemas.ts` are all unmodified, and the tool's declared schema values are unchanged, so the existing fixtures remain correct.

## Downstream Rollout

Notes/evidence:

- Downstream rollout is not applicable. Per `docs/downstream-rollout.md`, rollout is triggered by MCP tool/schema/protocol/client-facing changes. This PR changes none: the tool name `resolve_entry`, its description string, its input schema, its annotations, its error strings, and every field of its response are identical to `origin/main`. The change is internal function decomposition inside one server file, invisible to rtech-mcps, mcp2cli, and the rtech-hermes runtime.

### DIFF
--- a/server/tools/resolve-entry.ts
+++ b/server/tools/resolve-entry.ts
@@ -10,10 +10,20 @@
 import { z } from "zod";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { canRead } from "../auth/permissions.ts";
-import { canTargetNamespace, namespacePredicate } from "../auth/namespace-policy.ts";
+import {
+  canTargetNamespace,
+  namespacePredicate,
+} from "../auth/namespace-policy.ts";
 import type { AuthIdentity, ResourceTable } from "../auth/types.ts";
-import { authIdentity, errorResult, textResult, type MemoryToolDependencies } from "./types.ts";
+import {
+  authIdentity,
+  errorResult,
+  textResult,
+  type MemoryToolDependencies,
+} from "./types.ts";
 import { ALL_TABLES } from "./curation-helpers.ts";
+
+const NOT_AUTHENTICATED = "Permission denied: cannot resolve entries";
 
 const SOURCE_TYPE_BY_TABLE: Readonly<Record<ResourceTable, string>> = {
   thoughts: "thought",
@@ -23,7 +33,22 @@ const SOURCE_TYPE_BY_TABLE: Readonly<Record<ResourceTable, string>> = {
   sessions: "session",
 };
 
-type ResolveStatus = "found" | "archived" | "not_found_or_unreadable" | "not_readable";
+/** Frozen `resolve_entry` argument contract: the names and rule values are the API. */
+const resolveEntryInputSchema = {
+  id: z.string().uuid(),
+  namespace: z.string().min(1).max(500).optional(),
+};
+
+/** Tool annotations; `resolve_entry` reads and never mutates. */
+const resolveEntryAnnotations = {
+  title: "Resolve Entry",
+  readOnlyHint: true,
+  destructiveHint: false,
+  idempotentHint: true,
+};
+
+type ResolveStatus =
+  "found" | "archived" | "not_found_or_unreadable" | "not_readable";
 
 interface ResolvedEntry {
   resolved: boolean;
@@ -32,9 +57,27 @@ interface ResolvedEntry {
   source_type: string | null;
   table: ResourceTable | null;
   namespace: string | null;
-  fetch_path: { tool: "get_entry"; arguments: { table: ResourceTable; id: string } } | null;
+  fetch_path: {
+    tool: "get_entry";
+    arguments: { table: ResourceTable; id: string };
+  } | null;
   checked_sources: string[];
   checked_tables: ResourceTable[];
+}
+
+/** One probe pass: which tables, archived or live, and where to record progress. */
+interface ProbeRequest {
+  identity: AuthIdentity;
+  id: string;
+  tables: readonly ResourceTable[];
+  requestedNamespace: string | undefined;
+  archived: boolean;
+  checkedTables: ResourceTable[];
+}
+
+interface ProbeHit {
+  table: ResourceTable;
+  namespace: string;
 }
 
 function unresolved(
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: 880eb06 docs/lane-contract.md round 38 — a helper whose name and shape look like a drop-in reuse can still emit different SQL — `read-scope.ts`'s `appendReadNamespacePredicate` and `namespace-policy.ts`'s `namespacePredicate` both produce an auth-derived namespace clause, but the former takes a caller-named column plus a legacy-shared-fallback option, so substituting it inside a refactor lane would have changed the query under a no-behavior-change label; in a lint/refactor lane, reuse-first means read the emitted output of the candidate helper, not just its signature.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=ddd4557ea5c31809c110673d58c310cd42e2f108 at=2026-08-26T18:28:08.954Z

```
  - server/tools/resolve-entry.ts
```

</details>

Issue #744, opened 2026-08-24
recall returns ok with zero items (working_set only); search_all and repo_facts empty for rtech-consulting August state


rodaddy:
Agent: Claude head session (claudex fable), rtech-consulting, 2026-08-24. Filed on Rico's instruction after recall returned nothing usable for a fact that had been stated in-session on 2026-08-25 (#140) and that Rico had to re-explain.

## What was asked

"Give me information about" the rTech Microsoft 365 tenant: is it paid, which tenant holds the Inbox Zero app registration `4be80172-cace-49cd-bbee-cef34841c8fb`, what changed on #140. Ground truth (from Rico, this session): the tenant is real, paid for one year, live at `o365.rtech-consulting.com`; #140 records tenant `677d15db-...` / `o365.rtech-consulting.com`, one seat, one year.

## What was tried, verbatim, and what came back

1. SessionStart hydration (package-owned direct adapter): CANON PACK 1/2 profile_guidance=20 process_guidance=49; PACK 2/2 `repo_facts=0`; LANE RESUME checkpoint dated 2026-08-18. Nothing after 08-18 for this repo.

2. `printf '{"operation":"recall","query":"Microsoft 365 tenant paid one year app registration inbox-zero","namespace":"rico","limit":6}' | openbrain-memory`
   -> `status: failed`, "namespace comes from OPENBRAIN_NAMESPACE ... use {config:{namespace}}".

3. Same without namespace, with `limit` -> `status: failed`, "scope must be a JSON object", `limit` ignored.

4. With `scope:{platform,agent}` -> `failed`, "server_id, channel_id, session_key must be non-empty strings". The `help` example shows placeholders, not that all five are mandatory.

5. With all five scope fields -> `status: ok`, `item_count: 0`, `citations: []`, `served: ["working_set"]`, `not_durable_memory: true`. The recall operation never searched durable memory; it served only the session working set (empty) and reported success.

6. `_ob/skills/brain/scripts/direct.py search_all --params '{"query":"Microsoft 365 tenant paid one year app registration inbox-zero","limit":8}'`
[...]

Issue #329, opened 2026-07-22
Populate pointers and candidates


rodaddy:
## Summary

Implement lightweight cited `pointers` and clearly unconfirmed `candidate_memory` so clients can inspect more context without duplicating durable bodies or confusing candidates with durable truth.

**Workstream key:** `CONTEXT-4`  
**Priority:** P0  
**Parent:** #325  
**Depends on:** #326, #327

## Scope

- Emit resolvable IDs and source references rather than full bodies in `pointers`.
- Deduplicate pointers against `durable_memory`.
- Label and structurally separate low-confidence candidates.

## Acceptance criteria

- [ ] Every pointer has a resolvable ID and source reference and contains no full memory body.
- [ ] Items represented in `durable_memory` are not duplicated in `pointers` or double-counted against budget.
- [ ] Every candidate is explicitly labeled unconfirmed, has a source reference, remains exact-scope isolated, and is structurally distinct from durable memory.

## Validation

- [ ] Resolve emitted pointers through the authorized read path.
- [ ] Assert pointer deduplication against `durable_memory`.
- [ ] Test candidate labeling, empty states, budget behavior, and cross-namespace isolation.

## Non-goals

- Automatically fetching pointer bodies.
- Automatically promoting candidates.
- Injecting pointer content through MCP `_meta`.

## Standing constraints

- Exact namespace and session-scope isolation remains a security boundary.
- Prompt placement stays client/runtime-owned; no implicit MCP `_meta` injection.
- Observability remains content-free.
- Dream planning remains dry-run-safe unless a separately authorized mutating wrapper is used.

rodaddy:
## Merge, deployment, and downstream rollout receipt

- PR #360 merged to `main` as `6e4d9745ddf340b23ef66441c4be5075a10b1758`.
- Issue #329 closed automatically from the merged PR.
- Exact-main deploy workflow: https://github.com/rodaddy/open-brain/actions/runs/29985777567
  - `check`: passed
  - `db-integration` + anti-skip guard + restore drill: passed
  - `python-package`: passed
  - `contract-parity`: passed
  - `deploy` to core01: passed
- Public hosted health: healthy; both workers healthy with database and embedding connectivity.
- Real hosted MCP `agent_context_pack` canary (token-derived readonly identity, no delegated namespace header): **passed**.
  - pointers request returned 20 items (hard cap respected)
  - pointer item/source-ref structural shape passed
  - no pointer body/preview/label leakage
[...]

PR #806, opened 2026-08-26
refactor(context-pack): bring the durable context-pack loaders and their agent-context-pack caller to the lint standard (#780)

rodaddy:
## Summary

- Thirteen oxlint findings across the durable context-pack recall stack and its
  one caller. **All thirteen are retired here** — the residual this PR
  previously declared red is closed by the second commit.
- **First commit (`150df0b`)** — the two durable loaders. The audit named the
  real cause, and it is not size. The two files carried **structural twins**: a
  content-char resolver and a two-line failure logger that differed only in
  their event names and their extra fields. Two private copies of one rule is
  how two recall paths drift apart, so the fix is one shared sibling rather
  than two smaller private copies. New `server/tools/context-pack-shared.ts`
  holds `resolveContentChars`, `logDurableFailure` with
  `errorIdentityFields`/`pgDiagnosticFields`, and `boundedText` +
  `CONTEXT_PACK_ENVELOPE_CHAR_RESERVE` + `UNBOUNDED`.
  `context-pack-durable-lane.ts` re-exports the moved symbols, so every
  existing importer keeps working unchanged.
- **Second commit (`abfc101`)** — the caller,
  `server/tools/agent-context-pack.ts`: 729 code lines, with
  `buildAgentContextPackPayload` scoring **129 on complexity across 535 lines**.
  Again the cause was not size. Three mutable locals — the surviving chars,
  whether the next admitted member still gets the comma-free first slot, and the
  accumulated `whole_pack_budget` markers — were threaded through every section
  in sequence, so **no section could be lifted out of the function without
  carrying that state with it**. That is the whole reason the function grew to
  535 lines: every attempt to extract a section would have needed three
  in/out parameters.
- Naming that state as ONE object (`PackAllocator`) is what makes the split
  possible. Each section then becomes its own function that takes the allocator,
  charges itself against it, and returns what it contributed:

  | Module | Code lines | Holds |
  |---|---|---|
  | `context-pack-allocator.ts` | 146 | The allocation state and its arithmetic (INVARIANT 2). |
  | `context-pack-selection.ts` | 84 | Which sections this call builds, including the #744 default reasoning that was a 30-line comment mid-function. |
  | `context-pack-append-sections.ts` | 73 | `working_set` + `recovery`, previously two near-identical inline blocks. |
  | `context-pack-recall-sections.ts` | 133 | `durable_lane_context` + `durable_memory` admission. |
  | `context-pack-structured-sections.ts` | 137 | The shared fitter for the four fragment-shaped sections. |
  | `context-pack-loaders.ts` | 217 | The retrieval calls, one place per section. |
  | `context-pack-payload.ts` | 137 | Sections map, #535 receipt, warnings, reported allocation. |
  | `context-pack-assembly.ts` | 330 | The middle of the build, composing the above. |

  `agent-context-pack.ts` itself goes 729 → 425 code lines, and
  `buildAgentContextPackPayload` from complexity 129 to under 10.
- The two append-store sections became **one** function rather than two smaller
  copies, because they differed only in their key and in which counts must
  reconcile after a trim — exactly the twin pattern the first commit fixed. The
  two recall sections stayed **separate**, because they differ where it matters:
  the lane path sheds oldest-first and finally trims the checkpoint, while
  `durable_memory` is relevance-ordered and sheds the lowest-ranked tail so the
  best recall survives. Merging those two would have been a real behavior risk.
- **`loadDurableMemoryContext` now takes an options object**, retiring the
  `max-params` finding this PR previously left honestly red. The
  `DurableMemoryContextRequest` type already existed and was already the exact
  object the function built internally on its first line; the five-parameter
  form only survived because its one caller was the file above. No other caller
  exists under `server/`, so the old form is gone rather than kept alongside.
- No behavior moves in either commit: schemas, defaults, error strings, SQL
  text, log event names, section ordering, and payload shape are identical. The
  warnings channels are concatenated in the same source order they were written
  in, and the allocation arithmetic is the same expressions with the locals
  renamed to fields. The empty-reason precedence is preserved exactly
  (`walk_complete` → `content_unavailable` → `all_suppressed` → `no_matches`).
  No disable comment is used anywhere.

## Verification

All checks re-run on the **committed** tree, after the pre-commit prettier hook
reformatted the staged files:

- `bash scripts/done-means/780-touched-files-lint-clean.sh` → exit 0, covering
  all 12 files this branch touches.
- `oxlint --deny-warnings` clean on every touched file.
- `bunx tsc --noEmit` clean.
- `bun run test:isolated server/tools src/tools/__tests__/agent-context-pack*.test.ts
  src/tools/__tests__/agent-reflex-pointers.test.ts` → **218 pass, 0 fail, 1150
  expect() calls across 22 files**. Existing tests are unmodified.

Red-first receipts: `oxlint --deny-warnings` on both loaders at `origin/main`
reported all ten findings; on `server/tools/agent-context-pack.ts` at the
branch tip before the second commit it reported exactly three — `max-lines`
(729), and `complexity` (129) plus `max-lines-per-function` (535) on
`buildAgentContextPackPayload`. The done-means exited 1 at that point on the
`max-params` residual.

No Python, migration, or client-facing surface is touched, so those checks are
not applicable.

## Critical Self-Review

- Highest-risk behavior: the **warnings channel ordering**. The entry point used
  to build `scope_denials`, `degraded_sources`, and `truncation` as three inline
  array spreads, and a reader can see the order at a glance. They are now
  composed by `assembleWarnings` from a list of per-section contributions, so
  the order is the order of that list. It is asserted by the existing pack
  tests, which check exact warning arrays, but a future edit that reorders that
  list would reorder emitted warnings silently. The list is written in section
  priority order and commented for that reason.
- Assumptions that could be wrong: that `Math.min` on the allocation and the
  first-member framing flip are order-identical once the locals became fields on
  a shared object. Each section reads `firstSectionAdmitted` before it fits and
  writes it only when it is actually admitted, which is the original rule; a
  starved section still leaves the comma-free slot for the next one. Also that
  the durable_memory section suppressed for pointers-only still folds its
  content-free warnings into the first admitted #329 fragment — moved verbatim
  into `admitPointerSections`, and covered by the pointers/candidates suites.
- Missing/weak tests: no test asserts the exact key set of the durable failure
  log lines (carried over from the first commit), and none asserts the *order*
  of the warnings arrays independently of their contents. The existing suites
  cover the envelope. Neither is added here because writing them is
  behavior-adjacent work a lint lane should not smuggle in.
- Security/permission risk: none introduced. The two authorization gates are
  moved into `authorizePack` in the same order, still **before any query runs**,
  and still returning the identical content-free error strings. `physicalNamespace(ns)`
  is still the single isolation predicate every structured-section read binds
  to, resolved at the same point in the flow. The namespace argument is still
  authorized before it reaches any loader.
- Migration/deploy risk: none. No schema, no SQL text change, no config.
  Python client surface is touched. `src/tools/agent-context-pack*.ts` are
  separate legacy copies, not shims, and are untouched — including their own
  five-parameter `loadDurableMemoryContext`, which is a different function in a
  different file.
- Rollback/cleanup concern: a plain revert of both commits restores the two
  original files and deletes the nine new siblings; nothing outside this branch
  imports any of them.
- Fixes made before PR: corrected the extracted `recovery` selection predicate,
  which the first draft of `selectSections` rewrote into a nested conditional
  that was not equivalent to the original
  `include_unreviewed_recovery === true && (!requested || requested.includes("recovery"))`.
  Caught by reading the extraction back against the original rather than by a
  test — the existing suites did not cover that combination, which is itself
  worth noting.
- Known residual risk: **none outstanding — the done-means is GREEN.** The
  `max-params` finding this PR previously carried as a declared residual is
  closed by the second commit, and the three findings in the caller that blocked
  it are closed with it. All 12 files on this branch lint clean under
  `--deny-warnings`. The residual risk that remains is the ordinary one for a
  refactor of this size: the behavior guarantee rests on the existing 218-test
  suite plus reading each extraction against its original, not on new tests.
- SME review-memory update: [x] `docs/sme/` updated or [ ] not applicable because: entry added for the reviewable pattern this lane surfaced — that an over-complex function is often held together by two or three mutable locals threaded through it, and naming that state as one object is the move that makes extraction possible, whereas splitting on line count alone does not.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or client-facing surface changes; the change is internal to server-side loader and assembly modules.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, schema, or emitted-envelope field changed — section ordering, keys, and values are identical, so no fixture can move.

## Downstream Rollout

Notes/evidence:

- Classified not applicable per `docs/downstream-rollout.md`: no MCP tool,
  schema, protocol, or client-facing change. Every edited file is an internal
  server-side module; the only cross-file surfaces are
  `context-pack-durable-lane.ts` re-exporting three symbols it previously
  declared, and `loadDurableMemoryContext` changing shape with its single
  in-repo caller updated in the same commit.

### DIFF
--- a/docs/sme/entries/2026-08-26-an-over-complex-function-is-usually-held-together-by-mutable-locals-not-by-length.md
+++ b/docs/sme/entries/2026-08-26-an-over-complex-function-is-usually-held-together-by-mutable-locals-not-by-length.md
@@ -1,0 +1,53 @@
+---
+lane: quality
+order: 76
+---
+## [2026-08-26] An over-complex function is usually held together by mutable locals, not by length
+
+**Provenance:** PR #806, issue #780. Severity: MEDIUM. Status: open.
+**Scope:** any `complexity` / `max-lines-per-function` finding; review of a PR
+that answers one by extracting helpers.
+
+`buildAgentContextPackPayload` scored 129 on complexity across 535 lines. The
+length was the symptom. What actually held it together was three mutable
+locals — the surviving chars, whether the next admitted member still gets the
+comma-free first slot, and an accumulated markers array — read and written by
+each of nine sections in sequence.
+
+That is why the function had grown instead of being split earlier. Extracting
+any one section required passing all three in and getting two of them back, so
+every candidate helper had a worse signature than the code it replaced, and
+`max-params` was waiting on the other side. Splitting on line count alone
+produces exactly that: helpers that take five arguments and return a tuple,
+which trades one finding for another and makes the flow harder to read.
+
+Naming the shared state as ONE object first is the move that unlocks the split.
+Once the three locals became fields on a `PackAllocator` passed to each section,
+every section became a function taking one options object and returning what it
+contributed — and the file went 729 → 425 code lines across eight siblings with
+no helper exceeding four parameters.
+
+**What a reviewer should check** on a PR that answers a complexity finding:
+
+- Did the author identify the shared MUTABLE state before extracting, or just
+  cut at line boundaries? A helper list with no state object, where several
+  helpers take the same three or four scalars, is the tell that the split
+  fought the state instead of naming it.
+- Do the new helpers hand state back through return tuples or out-parameters?
+  That is the same coupling with more surface.
+- Ordering that used to be visible as one inline sequence — array spreads,
+  insertion into a map, a priority walk — is now decided by the ORDER OF A LIST
+  passed to a composer. That is a real behavior surface with no compiler check
+  behind it; confirm a test asserts the resulting order, and that the list is
+  commented with what its order means.
+- Two blocks merged into one parameterized helper need a reason they will not
+  need to diverge. Two blocks kept separate need a reason they must. In this PR
+  the two append-store sections merged (they differed only by key and reconciled
+  counts) and the two recall sections did not (they trim in opposite
+  directions) — both calls were stated and both are checkable.
+
+**Also observed:** the extraction rewrote one boolean predicate into a
+non-equivalent nested conditional, and the existing suite did not catch it. A
+refactor claimed to be behavior-free is verified by reading each extraction back
+against its original, not only by a green run — the tests cover the
+combinations someone already thought of.
--- a/docs/sme/quality.md
+++ b/docs/sme/quality.md
@@ -409,3 +409,53 @@ Namespace- or row-scoped cleanup does not isolate table-level DDL: a test that d
 Verbatim, from the source:
 
 > The fixture globally drops and narrows `maintenance_jobs_last_error_category_check` on the shared CI database. Bun runs test files in parallel by default, so another live queue test can observe the temporary old constraint and fail spuriously. Namespace-scoped row cleanup does not isolate table-level DDL.
+
+## [2026-08-26] An over-complex function is usually held together by mutable locals, not by length
+
+**Provenance:** PR #806, issue #780. Severity: MEDIUM. Status: open.
+**Scope:** any `complexity` / `max-lines-per-function` finding; review of a PR
+that answers one by extracting helpers.
+
+`buildAgentContextPackPayload` scored 129 on complexity across 535 lines. The
+length was the symptom. What actually held it together was three mutable
+locals — the surviving chars, whether the next admitted member still gets the
+comma-free first slot, and an accumulated markers array — read and written by
+each of nine sections in sequence.
+
+That is why the function had grown instead of being split earlier. Extracting
+any one section required passing all three in and getting two of them back, so
+every candidate helper had a worse signature than the code it replaced, and
+`max-params` was waiting on the other side. Splitting on line count alone
+produces exactly that: helpers that take five arguments and return a tuple,
+which trades one finding for another and makes the flow harder to read.
+
+Naming the shared state as ONE object first is the move that unlocks the split.
+Once the three locals became fields on a `PackAllocator` passed to each section,
+every section became a function taking one options object and returning what it
+contributed — and the file went 729 → 425 code lines across eight siblings with
+no helper exceeding four parameters.
+
+**What a reviewer should check** on a PR that answers a complexity finding:
+
+- Did the author identify the shared MUTABLE state before extracting, or just
+  cut at line boundaries? A helper list with no state object, where several
+  helpers take the same three or four scalars, is the tell that the split
+  fought the state instead of naming it.
+- Do the new helpers hand state back through return tuples or out-parameters?
+  That is the same coupling with more surface.
+- Ordering that used to be visible as one inline sequence — array spreads,
+  insertion into a map, a priority walk — is now decided by the ORDER OF A LIST
+  passed to a composer. That is a real behavior surface with no compiler check
+  behind it; confirm a test asserts the resulting order, and that the list is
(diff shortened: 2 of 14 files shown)


rodaddy:
harvested: 880eb06 docs/lane-contract.md round 38 — a diff-derived done-means makes "which files may I touch" a correctness question, not a style preference. This lane's tenth finding — `loadDurableMemoryContext` at five parameters — is fixable only by editing its single caller, and that caller carries three findings of its own on untouched `origin/main` (complexity 129 across 535 lines). Because the check lints every non-test `server/**/*.ts` the branch diff names, adopting the options object trades one failing finding for three and imports another lane's debt. The lesson for the rest of the #780 sweep: before changing a shared signature, lint the callers at `origin/main` first — if a caller is dirty, the signature belongs to whichever lane owns that caller, and the correct move is to leave it, say so in the PR, and let the done-means stay honestly red on the one item rather than reach for a disable comment or a silent revert.

rodaddy:
harvested: an over-complex function is usually held together by its mutable locals, not by its length — and until that shared state is named as one object, every candidate extraction has a worse signature than the code it replaces, which is why the function grew instead of being split earlier. `buildAgentContextPackPayload` scored 129 across 535 lines because three locals (surviving chars, the first-member framing flag, an accumulated markers array) were read and written by nine sections in sequence; introducing `PackAllocator` first turned each section into a one-argument function and took the file 729 → 425 code lines with no helper over four parameters. Recorded as `docs/sme/entries/2026-08-26-an-over-complex-function-is-usually-held-together-by-mutable-locals-not-by-length.md` (lane: quality), with the reviewer checklist for a PR that answers a complexity finding.

Second, smaller lesson in the same entry: the extraction rewrote one boolean predicate (`recovery` selection) into a non-equivalent nested conditional, and the existing 218-test suite did not catch it. It was found by reading each extraction back against its original. A refactor claimed to be behavior-free is verified that way, not only by a green run — the tests cover the combinations someone already thought of.

Also worth noting for the #780 sweep generally: this PR's previously-declared residual was real scope arithmetic, not an excuse. The done-means derives its file list from the branch diff, so touching the caller traded one finding for three until the caller was actually fixed. Leaving it honestly red and naming the blocker made the follow-up lane's scope obvious. Done-means is now green across all 12 files on the branch.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=d3b800b33051968d23176c34d3858f9699dc7d52 at=2026-08-26T19:04:26.403Z

```
linting 12 file(s) from: git diff --name-only origin/main...HEAD
  - server/tools/agent-context-pack.ts
  - server/tools/context-pack-allocator.ts
  - server/tools/context-pack-append-sections.ts
  - server/tools/context-pack-assembly.ts
  - server/tools/context-pack-durable-lane.ts
  - server/tools/context-pack-durable-memory.ts
  - server/tools/context-pack-loaders.ts
  - server/tools/context-pack-payload.ts
  - server/tools/context-pack-recall-sections.ts
  - server/tools/context-pack-selection.ts
  - server/tools/context-pack-shared.ts
  - server/tools/context-pack-structured-sections.ts
```

</details>

PR #809, opened 2026-08-26
refactor(citation-recall): bring server/tools/citation-recall.ts to the lint standard (#780)

rodaddy:
## Summary

- `server/tools/citation-recall.ts` carried the whole tool inside its registration function: schema, annotations, both SQL reads, two payload shapes, and the error path inline. oxlint reported three findings — `registerCitationRecallTool` at 179 lines, and its handler at complexity 24 and 135 lines.
- Split along the seams the code already had, following the pattern `server/tools/resolve-entry.ts` landed in #804: the description, input schema, and annotations are hoisted to module level (they are the frozen wire contract, so naming them separates the API from the code serving it); `recallCitation` owns one authorized request end to end and keeps the single try/catch; `fetchCitedEvent` and `fetchNeighbors` hold the two reads; `sourceNotStoredResult` and `storedCitationResult` shape the two payloads; `isExpandable` takes the held-back decision that carried the remaining complexity.
- `registerCitationRecallTool` is now only auth, namespace resolution, and defaults. A four-field `CitationRequest` options object carries the request instead of a parameter list.
- No behavior change: identical schemas, defaults, SQL text, error strings, log event names, and response field ordering. The lane join is still the isolation predicate and the comment explaining it moved with the query.
- No lint-disable comments were added. No other existing file was modified; no helper was moved out.

**Reported, not fixed:** `src/tools/citation-recall.ts` is a near-duplicate twin of this file — the SQL, the projection helper, `iso`, the default values, and the error-label allowlist are the same logic in both trees, differing only in the auth and dependency imports (`../permissions.ts` + `ToolDeps` vs `../auth/permissions.ts` + `MemoryToolDependencies`). Unifying them means moving a shared sibling across the `src/`/`server/` boundary and changing an exported signature, which is behavior-adjacent and belongs to its own lane. This lane did not edit `src/`.

## Verification

Run on the committed tree, after the pre-commit prettier hook reformatted the staged file:

- `./node_modules/.bin/oxlint --deny-warnings server/tools/citation-recall.ts` -> exit 0 (was 3 findings)
- `bun run test:isolated server/contracts/declaration.test.ts server/tools/` -> 174 pass, 0 fail, 18 files
- `bash scripts/done-means/780-touched-files-lint-clean.sh` -> exit 0 (diff-derived, 1 file)

Python package is untouched; this is a TypeScript-only refactor with no wire-contract change, so no live smoke applies.

## Critical Self-Review

- Highest-risk behavior: the `expandable` flag. Its predicate moved into `isExpandable`, and the two `.some()` calls over `before` and `after` became one `.some()` over the concatenation. That is the same result for every input — a disjunction over two arrays equals the disjunction over their concatenation — but it is the one place where the shape of the expression changed rather than just its location.
- Assumptions that could be wrong: that `server/contracts/declaration.test.ts` and `server/tools/ported-tools-scope.test.ts` are the real pinning tests for this tool. I found them by `rg -l "citation_recall|citation-recall" server/ --type ts`; if coverage of this tool lives only in `src/tools/__tests__/citation-recall.test.ts`, then the server twin's response shape is pinned less tightly than the file count suggests.
- Missing/weak tests: no test was added. This is a no-behavior-change refactor and existing tests were deliberately left unmodified, so the safety argument rests on those tests plus the field-by-field diff, not on new coverage. The extracted helpers have no direct unit tests of their own.
- Security/permission risk: low, and deliberately preserved. The auth gate (`canRead(identity.role, "sessions")`) and the namespace gate (`canTargetNamespace`) both stayed in the registration function ahead of the call into `recallCitation`, so an unauthorized request still cannot reach a query. The lane-join predicate `l.namespace = $1` moved verbatim into `fetchCitedEvent` and is still the only thing stopping an event UUID from another lane resolving.
- Downstream client/runtime risk: none intended. The tool name, description, input schema, annotations, and response JSON are byte-identical, so no rtech-mcps, mcp2cli, or Hermes surface sees a change. Per `docs/downstream-rollout.md` this is not a contract-changing edit.
- Rollback/cleanup concern: none. Single file, single commit, revertable on its own.
- Known residual risk: the `src/` twin stays duplicated, so a future fix to this tool's SQL or projection must be applied in two places until that is unified. That is called out above as a finding rather than silently left.
- Fixes made before PR: the first pass left `storedCitationResult` at complexity 13; rather than shrink it cosmetically I extracted the `expandable` decision into `isExpandable`, which is the thing that actually carried the branches.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane produced no new review finding — it applied an existing repo pattern (`resolve-entry.ts`) to a second file, and the duplication it surfaced is reported in this PR for its own lane rather than being a missed-review pattern.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no wire contract, schema, or runtime surface changed; the tool's registered shape is identical.

## Contract Parity

- Contract parity: [x] runtime-specific because: nothing in the declared contract moved. `server/contracts/declaration.test.ts` passes unmodified, which is the check that the registered tool shape still matches the declaration.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: `docs/downstream-rollout.md` scopes rollout to MCP tool, schema, protocol, and client-facing changes. This edit changes none of them — the tool name, description, input schema, annotations, and response payload are unchanged, verified field by field against the pre-refactor file.

### DIFF
--- a/server/tools/citation-recall.ts
+++ b/server/tools/citation-recall.ts
@@ -74,10 +74,16 @@ const SAFE_DB_ERROR_NAMES = new Set(["Error", "PostgresError"]);
 function safeErrorLabel(error: unknown): string {
   if (!error || typeof error !== "object") return "unknown";
   const candidate = error as { code?: unknown; name?: unknown };
-  if (typeof candidate.code === "string" && SAFE_DB_ERROR_CODES.has(candidate.code)) {
+  if (
+    typeof candidate.code === "string" &&
+    SAFE_DB_ERROR_CODES.has(candidate.code)
+  ) {
     return candidate.code;
   }
-  if (typeof candidate.name === "string" && SAFE_DB_ERROR_NAMES.has(candidate.name)) {
+  if (
+    typeof candidate.name === "string" &&
+    SAFE_DB_ERROR_NAMES.has(candidate.name)
+  ) {
     return candidate.name;
   }
   return "unknown";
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: 880eb06 docs/lane-contract.md round 38 — when a lint split leaves one helper still over the complexity rule, the leftover is usually a single composite decision rather than the function's size — here `storedCitationResult` sat at 13 after the split because the `expandable` disjunction was inlined in the payload literal, and extracting that one predicate into its own named function cleared it. Reach for the decision, not for another arbitrary slice of the body.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=be41a9e3e0f14648f415af03c3962d5c40c2ed78 at=2026-08-26T19:04:25.988Z

```
  - server/tools/citation-recall.ts
```

</details>

PR #811, opened 2026-08-26
refactor(update-entry): bring server/tools/update-entry.ts to the lint standard (#780)

rodaddy:
## Summary

- `server/tools/update-entry.ts` carried its whole update inside one inline registration closure: `registerUpdateEntryTool` at 157 lines wrapping a 114-line handler at complexity 23. The transaction body, the dynamic `SET` construction, and the re-embedding decision all shared one scope.
- Split along the seams `server/tools/resolve-entry.ts` and `server/tools/entities.ts` already use on `main`: the argument contract and annotations are hoisted to module-level consts (`updateEntryInputSchema`, `updateEntryAnnotations`, `UPDATE_ENTRY_DESCRIPTION`), and each phase of the transaction is a named module-level helper taking an options object instead of a positional list — `lockExistingRow`, `computeReembedding`, `mergeContentFields`, `appendFieldAssignments`, `appendEmbeddingAssignments`, `buildAssignments`, `applyUpdate`, `runUpdate`, `collectProvidedFields`.
- No behavior moves. Schema fields, annotations, error strings, SQL text, positional parameter ordering, the `tool_result` log event and its keys, and the response payload are identical. Helpers were cut so every early return still rolls back before answering, preserving the `FOR UPDATE` lock discipline the file header describes.
- No existing file outside the target was touched, and no lint disable comment was added.

## Verification

Red first, on the untouched file: `oxlint --deny-warnings server/tools/update-entry.ts` reported `max-lines-per-function` (registerUpdateEntryTool, 157 lines), `complexity` (async function, 23), and `max-lines-per-function` (async function, 114); the done-means with `DONE_MEANS_780_FILES` set exited 1.

Green, re-run on the COMMITTED tree after the pre-commit prettier hook reformatted the staged file:

- `./node_modules/.bin/oxlint --deny-warnings server/tools/update-entry.ts` -> exit 0
- `bash scripts/done-means/780-touched-files-lint-clean.sh` (diff-derived from `origin/main...HEAD`) -> exit 0, 1 file linted
- `bun run test:isolated src/tools/__tests__/update-entry.test.ts server/tools/` -> 177 pass, 0 fail, tests exited 0

## Critical Self-Review

- Highest-risk behavior: the positional parameter ordering of the dynamic `UPDATE`. The original appended field params, then embedding params, then pushed `args.id` last and read `params.length` as the id placeholder. `buildAssignments` now returns the params array and `applyUpdate` appends the id, which preserves that exact order; the namespace predicate is still offset from `idParam + 1`.
- Assumptions that could be wrong: that no caller outside `server/tools/index.ts` imports anything from this module. Verified with `rg -l` — the only importer is `server/tools/index.ts:32`, and `registerUpdateEntryTool` keeps its exported signature unchanged. `src/tools/update-entry.ts` is a separate legacy twin exporting `registerUpdateEntry`; it was not touched.
- Missing/weak tests: there is no server-side test that drives `update_entry` through the server registry. The pinning coverage that exists (`src/tools/__tests__/update-entry.test.ts`, 14 tests including content-hash and re-embed assertions) exercises the `src/` twin, so it constrains the shared canonical text/hash builders but not this registration path directly. Adding server-side coverage is real work with its own behavior surface and belongs in its own change, not in a lint-only refactor.
- Security/permission risk: none intended. This file is one of the ID-based mutating tools `_plans/issues/93` covers, so the namespace predicate on both statements is the design being protected: the `authorize` call, the write-scoped `namespacePredicate` on the `SELECT ... FOR UPDATE` and on the `UPDATE`, and the namespace-scoped content-hash collision check are unchanged and still inside the same transaction. Neither statement can now be reached without its predicate.
- Downstream client/runtime risk: none. Per `docs/downstream-rollout.md` this is not an MCP tool/schema/protocol/client-facing change — the tool name, description, input schema, annotations, and result payload are byte-identical, so no rtech-mcps, mcp2cli, or Hermes step applies.
- Rollback/cleanup concern: single-file, single-commit, revertable on its own with no dependent change.
- Fixes made before PR: re-ran oxlint, tsc, the done-means, and both test paths against the committed tree after the prettier hook reformatted the file, rather than trusting the pre-commit run.
- Known residual risk: the refactor is verified by the existing suite plus typecheck, not by a test that exercises this exact registration; the SQL strings and parameter order were compared against the pre-change file by reading the diff.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a lint-standard refactor under #780 with no behavior change and no review finding, so it produces no new pattern the next swarm should check for.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime behavior, schema, or contract changes, so a live smoke would exercise nothing this PR alters.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface moves — the `update_entry` tool name, input schema, annotations, error strings, and result payload are unchanged, so no fixture has anything new to pin.

## Downstream Rollout

Notes/evidence:

- Every downstream item is not applicable: this change is internal to one server module and alters no client-facing contract. Reviewed against the classification rules in `docs/downstream-rollout.md`.

### DIFF
--- a/server/tools/update-entry.ts
+++ b/server/tools/update-entry.ts
@@ -8,11 +8,17 @@
  * on state that no longer holds.
  */
 import { z } from "zod";
+import type { PoolClient } from "pg";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { toSql } from "pgvector/pg";
 import { namespacePredicate } from "../auth/namespace-policy.ts";
-import type { ResourceTable } from "../auth/types.ts";
-import { authIdentity, errorResult, textResult, type MemoryToolDependencies } from "./types.ts";
+import type { AuthIdentity, ResourceTable } from "../auth/types.ts";
+import {
+  authIdentity,
+  errorResult,
+  textResult,
+  type MemoryToolDependencies,
+} from "./types.ts";
 import { authorize, contentHash } from "./memory-helpers.ts";
 import { tableEnum } from "./curation-helpers.ts";
 import {
@@ -38,7 +44,14 @@ const VALID_FIELDS: Readonly<Record<ResourceTable, readonly string[]>> = {
     "metadata",
   ],
   projects: ["name", "description", "tags"],
-  sessions: ["summary", "project", "key_decisions", "next_steps", "blockers", "tags"],
+  sessions: [
+    "summary",
+    "project",
+    "key_decisions",
+    "next_steps",
+    "blockers",
+    "tags",
+  ],
 };
 
 /**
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: 880eb06 docs/lane-contract.md round 38 — when a lint refactor splits a handler that builds SQL with positional placeholders, the parameter array is the invariant, not the SET list. This file read `params.length` immediately after each push to number its own placeholder, and appended the WHERE id param last so `idParam` fell after every field and embedding param. Splitting the builder into separate helpers is only safe while every helper mutates one shared array in the original order — returning per-helper arrays and concatenating later renumbers the placeholders silently, and both the typechecker and a schema-level test stay green because the SQL is still valid and the arity still matches. The check that catches it is reading the emitted SQL and param order against the pre-change file, which is why that comparison is named in the self-review rather than assumed from a passing suite.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=3ef10756a248447482e81e6c14c0ce8b89b9cd57 at=2026-08-26T19:04:33.304Z

```
  - server/tools/update-entry.ts
```

</details>

PR #813, opened 2026-08-26
refactor(tiering): bring server/tools/tiering.ts to the lint standard (#780)

rodaddy:
## Summary

- `registerTieringTools` held both tool registrations and both handler bodies inline: 227 lines, with the two handlers at complexity 11 and 13. Three oxlint findings, no disable comments used.
- Applies the shape `server/tools/tier-mutations.ts` already uses for the tier write tools, so the two tier files read alike: input schemas and annotations hoisted to module scope, and each `registerTool` call is a one-line delegation to a named module-level handler taking `(dependencies, args, extra)`.
- `tier_recommendations` carried two unrelated SQL arms in one `if`/`else` — demotion reads the stored `access_count`, promotion reads `entry_access_log` rows. They are now `collectDemotionCandidates` and `collectPromotionCandidates`, each taking one `TierScan` options object, selected once by `action`. `list_stale`'s table resolution moved to `resolveStaleTables` and its argument defaults to `staleWindow`.
- No behavior change. Schemas, defaults, error strings, SQL text, log event names, and response shapes are identical; the SQL template literals keep their original indentation so the emitted statements are byte-identical. No other existing file is touched, and no test was modified.

## Verification

Red first, on the untouched file: `oxlint --deny-warnings server/tools/tiering.ts` reported `max-lines-per-function` (`registerTieringTools`, 227 lines) at `tiering.ts:76:8`, and `complexity` 11 at `:97:5` and 13 at `:251:5`; `DONE_MEANS_780_FILES=server/tools/tiering.ts bash scripts/done-means/780-touched-files-lint-clean.sh` exited 1.

Green, re-run on the COMMITTED tree after the pre-commit prettier hook reformatted the file:

- `./node_modules/.bin/oxlint --deny-warnings server/tools/tiering.ts` -> exit 0
- `bun run test:isolated server/tools/ src/tools/__tests__/tier-recommendations{.pg,}.test.ts src/tools/__tests__/list-stale{,.pg}.test.ts` -> 194 pass, 0 fail across 21 files
- `bash scripts/done-means/780-touched-files-lint-clean.sh` (diff-derived) -> exit 0
- `git status --short` -> clean

Python package is untouched, so its checks are not applicable. No live smoke: the change is a pure in-file restructure of two read-only tools with no schema, transport, or SQL change.

## Critical Self-Review

- Highest-risk behavior: the two `tier_recommendations` SQL statements. Their parameter numbering differs between arms — demotion qualifies its namespace predicate at `$3`, promotion at `$4` because it binds `table` as `$3` — and splitting the arms is exactly where a silent off-by-one would land. Both arms carry their original numbering and value order verbatim; `tiering.pg.test.ts` and `tier-recommendations.pg.test.ts` execute both against real Postgres, so a mismatch fails rather than returning wrong rows.
- Assumptions that could be wrong: that `candidates.length >= wanted` still short-circuits identically now that `push(...found)` replaces a per-row `push` inside the query loop. It does — the check runs once per table before the query in both versions, and `remaining` is still computed from the live length — but a reader could reasonably expect per-row accounting.
- Missing/weak tests: nothing new is tested because nothing new is claimed; this leans entirely on the existing pinning tests. Those do not assert the exact SQL string, so a whitespace-level SQL change would pass. That is why the template literals were kept byte-identical rather than re-indented to match the new nesting.
- Security/permission risk: the namespace predicate and `canRead` checks are the isolation boundary here. `resolveStaleTables` reproduces the original three-way outcome exactly, including the distinction between "cannot read `<table>`" for a named table and "no readable tables" — the wording difference is load-bearing for callers and is preserved verbatim. No predicate was made optional and no interpolation was widened.
- Downstream client/runtime risk: none. Tool names, input schemas, annotations, descriptions, and response envelopes are unchanged, so the MCP contract is identical and `docs/downstream-rollout.md` does not apply.
- Rollback/cleanup concern: single-file, single-commit revert with no state to unwind.
- Fixes made before PR: the first pass left `handleListStale` at complexity 11 — extracting the handlers alone was not enough, because the run of `??` defaults counted against it. Extracting `staleWindow` brought it under. Caught by re-running oxlint, not assumed.
- Known residual risk: the file is 462 lines, under the 500 threshold but with less headroom than before, so the next tool added here should go in a sibling rather than inline.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical lint-standard restructure that found no new defect class — the pattern it applies is already recorded and already landed in `server/tools/tier-mutations.ts`, `search-brain.ts`, and `entities.ts`.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no schema, transport, or client-facing change — both tools are read-only and their SQL is byte-identical.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface moved. Tool names, input schemas, annotations, and response shapes are unchanged, so no fixture describes anything different.

## Downstream Rollout

Notes/evidence:

- Classified not applicable under `docs/downstream-rollout.md`: rollout applies to MCP tool, schema, protocol, or client-facing changes. This changes none of them — it moves function bodies within one server file and leaves every registered tool's name, schema, annotations, and output identical.

### DIFF
--- a/server/tools/tiering.ts
+++ b/server/tools/tiering.ts
@@ -24,7 +24,12 @@ import {
   type NamespacePredicate,
 } from "../auth/namespace-policy.ts";
 import type { ResourceTable } from "../auth/types.ts";
-import { authIdentity, errorResult, textResult, type MemoryToolDependencies } from "./types.ts";
+import {
+  authIdentity,
+  errorResult,
+  textResult,
+  type MemoryToolDependencies,
+} from "./types.ts";
 import {
   ALL_TABLES,
   PREVIEW_WIDTH,
@@ -73,6 +78,63 @@ interface TierCandidate {
   reasoning: string;
 }
 
+const tierRecommendationsInputSchema = {
+  action: z.enum(["promote", "demote"]),
+  threshold_days: z.number().int().min(1).max(365).optional(),
+  candidates: z.number().int().min(1).max(100).optional(),
+};
+
+const tierRecommendationsAnnotations = {
+  title: "Tier Recommendations",
+  readOnlyHint: true,
+  destructiveHint: false,
+  idempotentHint: true,
+};
+
+const listStaleInputSchema = {
+  table: tableEnum.optional().describe("Optional: filter to a specific table"),
+  days: z
+    .number()
+    .int()
+    .min(1)
+    .max(365)
+    .optional()
+    .describe(
+      "Entries not accessed in this many days are considered stale (default 30)",
+    ),
+  limit: z
+    .number()
+    .int()
+    .min(1)
+    .max(500)
+    .optional()
+    .describe("Maximum entries to return (default 50, max 500)"),
+  offset: z
+    .number()
+    .int()
+    .min(0)
+    .optional()
+    .describe("Number of entries to skip for pagination (default 0)"),
+  tier: tierEnum
+    .optional()
+    .describe(
+      "Optional: filter to a specific tier (e.g. 'hot' to find hot entries that should decay to warm)",
+    ),
+  response_format: z
+    .enum(["envelope", "array"])
+    .optional()
+    .describe(
+      "Response format: 'envelope' (default) returns {entries, total_count, has_more}; 'array' returns raw array for backwards compatibility",
+    ),
+};
+
+const listStaleAnnotations = {
+  title: "List Stale",
+  readOnlyHint: true,
+  destructiveHint: false,
+  idempotentHint: true,
+};
+
 export function registerTieringTools(
   server: McpServer,
   dependencies: MemoryToolDependencies,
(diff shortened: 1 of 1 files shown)


rodaddy:
No new lessons: the tier-mutations shape (schemas and annotations hoisted, each registration delegating to a named module-level handler) transferred to this file without a single adaptation, so the pattern is already recorded and this lane only confirmed it holds for read-only tools too.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=30152fd82327e568ba4188645de8f647d1b3dce3 at=2026-08-26T19:04:31.581Z

```
  - server/tools/tiering.ts
```

</details>

PR #815, opened 2026-08-26
refactor(list-recent): bring server/tools/list-recent.ts to the lint standard (#780)

rodaddy:
## Summary

- `registerListRecentTool` in `server/tools/list-recent.ts` was 129 lines and its handler carried a complexity of 18, both over the repo standard. The registration wrapper and the request handling were a single function, so every branch counted against a body that was mostly a literal input schema.
- Applies the shape the already-converted tools use (`scan-namespace.ts`, `find-duplicates.ts`, `search-brain.ts`): the description, input schema, and annotations are hoisted to module-level constants, and the handler is split into named module-level helpers — `normalizeListRecentArgs` (documented defaults), `resolveListRecentTables` (the house `{tables} | {denied}` shape), `buildListRecentQueries` and `runListRecentQueries` (the data/count pair), and `listRecentResult` (the two frozen response formats). Helpers taking more than a couple of inputs take an options object rather than positional parameters.
- No behavior moves. The schema, the defaults (7 days, 20 entries, offset 0), both permission-denied strings, both SQL builders, the `list_recent_count_failed` log event, the null-count recovery, and both response shapes are identical. `readableNamespaces` from `read-scope.ts` remains the namespace source; the `$4`/`$2` parameter-index split moves into `buildListRecentQueries` with its explanatory comment intact.
- Reuse check: `normalizeSearchArgs` in `search-request.ts` was evaluated and rejected — it defaults `limit` to 10 where `list_recent` publishes 20, and carries search-mode/fts-config fields this tool has no argument for. Reusing it would have moved a frozen default.
- One file changed. No existing test modified, no other file touched, no disable comments.

## Verification

```
server/tools/list-recent.ts:105:8: error eslint(max-lines-per-function): The function `registerListRecentTool` has too many lines (129). Maximum allowed is 100.
server/tools/list-recent.ts:169:5: error eslint(complexity): async function has a complexity of 18. Maximum allowed is 10.
```

`DONE_MEANS_780_FILES=server/tools/list-recent.ts bash scripts/done-means/780-touched-files-lint-clean.sh` exited 1.

- `./node_modules/.bin/oxlint --deny-warnings server/tools/list-recent.ts` -> exit 0
- `bun run test:isolated src/tools/__tests__/list-recent.test.ts server/tools/search-read-scope.test.ts server/tools/` -> 175 pass, 0 fail across 18 files
- `bash scripts/done-means/780-touched-files-lint-clean.sh` -> exit 0

Python package checks are not applicable: nothing under `python/openbrain-memory/` changed. A live smoke is not applicable: this is an internal restructuring with no wire-visible change, and the `list_recent` contract fixtures are untouched.

## Critical Self-Review

- Highest-risk behavior: the namespace predicate. `list_recent` binds readable namespaces at `$4` in the data query and `$2` in the count query, and the two indexes are not interchangeable — crossing them would constrain the wrong parameter and return rows from namespaces the caller may not read, silently. Both index computations moved into `buildListRecentQueries` together, so they cannot drift apart, and `server/tools/search-read-scope.test.ts` asserts the emitted SQL and bound parameters.
- Assumptions that could be wrong: that folding the original `tables.length === 0` check into `resolveListRecentTables` preserves the denial. The original ran that check after both branches; the extracted version runs it only on the unfiltered branch. The named-table branch always yields exactly one element, so the length is never 0 there — the two are equivalent, and the named-table path has its own by-name denial ahead of it.
- Missing/weak tests: no test drives the `response_format: "array"` path against a real pool at the `server/` layer; coverage for that shape comes from the `src/tools/` sibling. Not made worse here, and not in scope for a lint lane.
- Security/permission risk: both permission-denied strings are byte-identical to the originals, and the role gate still runs against the caller-named table before any query is built. Namespace scoping is unchanged and still auth-derived via `readableNamespaces`.
- Downstream client/runtime risk: none. Per `docs/downstream-rollout.md` this is not an MCP tool/schema/protocol change — the tool name, input schema, annotations, and both response shapes are unchanged, so no client sees a difference.
- Rollback/cleanup concern: a single-file revert restores the prior behavior exactly; nothing was moved out of the file, so no other module depends on this commit.
- Fixes made before PR: read each extracted helper back against the original line by line after extraction, because the suite did not catch a non-equivalent predicate rewrite in #806. That read is what surfaced the `tables.length === 0` ordering question above.
- Known residual risk: low. The only non-mechanical judgment is the table-resolution ordering, reasoned through above; everything else is a literal move of unchanged text.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no new failure pattern surfaced — this lane exercised the existing round-38 lane-contract guidance (verify helper reuse is identical field for field before reusing; read every extraction back against the original) and both rules held without amendment.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no wire-visible change; the restructuring is internal to one server tool module and the contract fixtures are untouched.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract moved. `contracts/server/server-recall-family.fixture.json` describes the `list_recent` tool surface, and the tool name, input schema, annotations, and both response shapes are byte-identical after this change, so there is nothing for a fixture to record.

## Downstream Rollout

Notes/evidence:

- All four are not applicable by the same reason: the rollout doc scopes downstream work to MCP tool/schema/protocol/client-facing changes, and this PR changes none of them. `list_recent` publishes the same name, the same input schema, the same annotations, and the same two response shapes before and after. A downstream client cannot observe this commit.

### DIFF
--- a/server/tools/list-recent.ts
+++ b/server/tools/list-recent.ts
@@ -23,9 +23,14 @@
  */
 import { z } from "zod";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
-import type { ResourceTable } from "../auth/types.ts";
+import type { AuthIdentity, ResourceTable } from "../auth/types.ts";
 import { canRead } from "../auth/permissions.ts";
-import { authIdentity, errorResult, textResult, type MemoryToolDependencies } from "./types.ts";
+import {
+  authIdentity,
+  errorResult,
+  textResult,
+  type MemoryToolDependencies,
+} from "./types.ts";
 import { readableNamespaces } from "./read-scope.ts";
 import {
   ALL_TABLES,
@@ -55,7 +60,9 @@ function buildWhereClause(
   tier: Tier | undefined,
   namespaceParamIndex: number | undefined,
 ): string {
-  const archiveFilter = includeArchived ? "" : ` AND ${alias}.archived_at IS NULL`;
+  const archiveFilter = includeArchived
+    ? ""
+    : ` AND ${alias}.archived_at IS NULL`;
   const tierFilter = tier ? ` AND ${alias}.tier = '${tier}'` : "";
   const namespaceFilter = namespaceParamIndex
     ? ` AND ${alias}.namespace = ANY($${namespaceParamIndex}::text[])`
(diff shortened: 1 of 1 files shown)


rodaddy:
No new lessons: I checked `docs/sme/` and the lane contract's Tightenings for anything this lane could add, looking specifically for (a) a helper-reuse failure mode not already covered — the round-38 rule "reuse only when the shape is identical field for field, defaults and failure shape included" is exactly what stopped me reusing `normalizeSearchArgs` here, since its `limit` defaults to 10 against this tool's published 20; (b) an extraction-equivalence gap beyond the #806 lesson that the suite does not catch a non-equivalent predicate rewrite — I read every helper back against the original and the only judgment call, folding the `tables.length === 0` check into the table resolver, was caught by that existing rule rather than by a new one; and (c) a lint-standard pattern not already demonstrated by the converted siblings — the hoist-schema/split-handler/options-object shape came straight off `scan-namespace.ts` and `find-duplicates.ts` with nothing new required. All three existing rules held without amendment, so writing an entry would restate `docs/sme/` rather than extend it.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=508cc99be4caa2b1627f67bac471e26eb2d6ca36 at=2026-08-26T20:05:45.379Z

```
  - server/tools/list-recent.ts
```

</details>

PR #816, opened 2026-08-26
refactor(get-entry): bring server/tools/get-entry.ts to the lint standard (#780)

rodaddy:
## Summary

- `server/tools/get-entry.ts` carried two lint findings: `registerGetEntryTool` at 105 lines and its inline handler at complexity 11, both because the argument contract, the annotations, and both render arms lived inside the single `registerTool` call.
- Split along the seams `main` already established in `server/tools/resolve-entry.ts` and the other ported tools: the frozen argument schema (`getEntryInputSchema`), the annotations (`getEntryAnnotations`), the tool description, and the shared not-found string are now module constants.
- Each render arm is a named module-level function taking one options object — `fetchCompact(dependencies, CompactRead)` and `fetchFull(dependencies, EntryRead)` — and `compactEnvelope` shapes the compact payload on its own so the field order clients parse reads in one place instead of interleaved with SQL construction.
- The read predicate rides in the options object as the exported `NamespacePredicate` type from `server/auth/namespace-policy.ts`, keeping the namespace scope bound to the same SELECT that resolves the id, which is the isolation property the file's own header comment names.
- No behavior change: schema field names and rule values, both defaults, both error strings, both SQL statements, the `tool_result` log event and its `render` field, and the compact payload's field order are all unchanged. Only `server/tools/get-entry.ts` is touched.
- Reuse was checked before writing anything: `read-scope.ts`, `shared-namespace.ts`, `memory-helpers.ts`, `tool-failure.ts`, `search-request.ts`, and `entity-shared.ts` carry nothing matching these two table-specific projections field for field, so the extraction stayed inside the owning file and no helper was moved out.

## Verification

`./node_modules/.bin/oxlint --deny-warnings server/tools/get-entry.ts` reported both findings verbatim before the change (`max-lines-per-function` 105 at :78, `complexity` 11 at :115) and exits 0 after. `bunx tsc --noEmit` is clean. `bun run test:isolated server/tools/` is 163 pass / 0 fail across 17 files, re-run on the committed tree after the pre-commit prettier hook reformatted the file. `bash scripts/done-means/780-touched-files-lint-clean.sh` exits 1 before and 0 after. Python is untouched; this is an internal TypeScript refactor with no wire-visible change, so no smoke applies.

## Critical Self-Review

- Highest-risk behavior: the compact render's payload shape and the namespace predicate binding. Both are what a client parses and what keeps an ID-based read inside the caller's authority, and both are reproduced field for field and parameter for parameter.
- Assumptions that could be wrong: that no existing shared helper already expresses these projections. Checked with `rg` across `server/tools/` before writing; the two column maps are table-specific to this tool and the compact envelope's `fetch_path` / `source_ref` fields have no sibling with the same shape.
- Missing/weak tests: no new test was added, deliberately — this change asserts equivalence, and the existing `server/tools/gap-closure-read-scope.test.ts` already exercises both render arms against the auth-derived predicate, which is the property most worth protecting here.
- Security/permission risk: the `canRead` gate and the `namespacePredicate(identity, "read", 2)` call are unmoved relative to each other, and the predicate reaches the SELECT through the options object rather than a closure. Found-but-unreadable and genuinely-absent still collapse to one error string, so the tool still cannot be used to probe which UUIDs exist elsewhere.
- Downstream client/runtime risk: none. `get_entry`'s tool name, description, argument names and rule values, and both response shapes are byte-identical, so `contracts/server/` fixtures and the Python client see no difference.
- Rollback/cleanup concern: single-file, single-commit; reverting the commit restores the prior file exactly.
- Fixes made before PR: the first draft typed the predicate as an inline `{ clause: string; values: unknown[] }`, which `tsc` rejected because `NamespacePredicate.values` is `readonly`. Switched to importing the exported type rather than widening the local shape, so the readonly guarantee survives the extraction.
- Known residual risk: the SQL template literals shifted their leading indentation by one nesting level when the arms moved to module scope. The statements' internal alignment is unchanged and the whitespace is not semantically meaningful to Postgres, but the emitted query text differs by leading spaces from the prior version.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no new failure pattern surfaced — the one `tsc` catch was a readonly-type mismatch already covered by the repo's existing typecheck gate, not a reviewer-missable class.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: internal refactor with no wire-visible or database-visible change.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface moved — the tool name, description, argument schema, and both response shapes are unchanged, so `contracts/server/server-entry-read.fixture.json` and the gap map still describe this tool accurately.

## Downstream Rollout

Notes/evidence:

- Rollout classification is not-applicable across the board: `docs/downstream-rollout.md` scopes rollout to MCP tool, schema, protocol, or client-facing changes, and none of those moved here. The change is confined to how one already-registered tool's handler is organized inside `server/tools/get-entry.ts`.

### DIFF
--- a/server/tools/get-entry.ts
+++ b/server/tools/get-entry.ts
@@ -24,9 +24,17 @@
 import { z } from "zod";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { canRead } from "../auth/permissions.ts";
-import { namespacePredicate } from "../auth/namespace-policy.ts";
+import {
+  namespacePredicate,
+  type NamespacePredicate,
+} from "../auth/namespace-policy.ts";
 import type { ResourceTable } from "../auth/types.ts";
-import { authIdentity, errorResult, textResult, type MemoryToolDependencies } from "./types.ts";
+import {
+  authIdentity,
+  errorResult,
+  textResult,
+  type MemoryToolDependencies,
+} from "./types.ts";
 import { SOURCE_LABELS, tableEnum } from "./curation-helpers.ts";
 
 /** Observed current-src default preview width for the compact envelope. */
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #816 Reuse-first on a tool handler split can come back empty and that is the correct result, but only if the check is done at field granularity rather than by name. Six shared siblings were read before writing anything here (`read-scope.ts`, `shared-namespace.ts`, `memory-helpers.ts`, `tool-failure.ts`, `search-request.ts`, `entity-shared.ts`); each was rejected because the compact envelope's `source_ref` / `fetch_path` / `content_truncated` fields and the two table-keyed column maps have no sibling matching them field for field, defaults included. The one real catch was typing: lifting a value out of a closure into an options object re-types it at a module boundary, and the hand-written `{ clause: string; values: unknown[] }` silently widened away the `readonly` on `NamespacePredicate.values` -- `tsc` caught it, but the lesson is to import the owning module's exported type rather than re-describe the shape inline, because a re-described shape is where a security-relevant invariant like an immutable parameter list quietly goes missing.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=9f005b1d57b645bef8ba96de33717e952ea46f7a at=2026-08-26T20:05:48.572Z

```
  - server/tools/get-entry.ts
```

</details>

PR #817, opened 2026-08-26
refactor(context-pack-repo-facts): bring server/tools/context-pack-repo-facts.ts to the lint standard (#780)

rodaddy:
## Summary

- `loadRepoFactsSection` carried the whole section in one body and drew all three lint findings from one structural cause: complexity 29, 142 lines, and a block nested four deep at the text step. The row-admission decision, the item shape, the query, and the failure log all lived inline in the same function.
- Split into named module-level helpers, each reproducing what the original did. `admitRepoFactRow` lifts the row-admission decision into one predicate returning a discriminated result; `repoFactEntry` builds an admitted row's item body and matching citation field for field; `collectRepoFacts` does the single walk over the ordered rows; `noActiveRepoFragment`, `queryRepoFactRows`, and `logRepoFactsFailure` carry the empty state, the exactly-bound query, and the two-line failure log verbatim.
- The ORDER of the three exclusions is preserved deliberately, because that order is behavior: only the last of them (an unusable `fact`) can mark the section trimmed, so a row rejected by an earlier guard must not reach the text step and set that flag. The `truncated` accumulator likewise seeds from the caller's row-count verdict and is only ever set, never cleared, exactly as the original local flag behaved.
- Reuse over reinvention: `asError` and `errorIdentityFields` are imported from the sibling `server/tools/context-pack-shared.ts`, which already derives `error_name` / `error_message` exactly as this module did. `logDurableFailure` from that same sibling was deliberately NOT reused: it takes a required `DurableFailureLogger` and re-derives `error_message` itself, while this module's `logger` parameter is optional, so the shapes are not identical and reusing it would have moved behavior.
- No behavior change. The SQL text, the schemas, the defaults, the error strings, the log event names (`repo_facts_section_failed`, `repo_facts_section_failed_detail`), the section key order, and the fragment shape are byte-identical to `origin/main`.
- One file changed. No exported signature moved, so no caller needed touching and no other existing file is in the diff.

## Verification

`./node_modules/.bin/oxlint --deny-warnings server/tools/context-pack-repo-facts.ts` returned the three findings verbatim before the change and exit 0 after. `bunx tsc --noEmit` exit 0. `bun run test:isolated` over the repo_facts pinning tests (`server/tools/context-pack.test.ts`, `src/tools/__tests__/agent-context-pack-repo-facts.test.ts`, `src/tools/__tests__/agent-context-pack-guidance-repo-facts.test.ts`) gave 60 pass / 0 fail, and over `server/tools/` gave 163 pass / 0 fail. `scripts/done-means/780-touched-files-lint-clean.sh` exit 0, re-run on the committed tree after the pre-commit prettier hook (which reported the file unchanged). Python package untouched. No runtime, schema, or transport surface moved, so no live smoke applies.

## Critical Self-Review

- Highest-risk behavior: the row-admission predicate. Three exclusions that previously fell through one inline sequence are now one function returning a discriminated result, and only the third of them may set the trimmed flag. A rewrite that let an earlier guard reach the text step would silently mark sections trimmed that are not.
- Assumptions that could be wrong: that `asText` never returns an empty string, which is what makes the original truthy check `if (asText(meta.fact))` equivalent to my explicit `!== null`. Read back against the source: `asText` returns the value only when `typeof value === "string" && value.length > 0`, so `""` is impossible and the two are the same test.
- Missing/weak tests: the existing suite covers the empty state, the exact repo bind, the source-ref exclusion, and the staleness dispositions, and it does not cover the interaction of a row-count overflow with a per-item text overflow in the same read. That gap predates this change and the accumulator seeding is the part it would exercise; I verified that path by reading the rewrite against the original rather than by a new test, since the lane's rule is that existing tests stay unmodified.
- Security/permission risk: none introduced. The namespace predicate and the exact `metadata->>'repo' = $2` bind stay in one parameterized query, now in `queryRepoFactRows`, with the same two parameters in the same order. No cross-repo fallback was added and no interpolation was introduced.
- Downstream client/runtime risk: none. Per `docs/downstream-rollout.md` this is an internal server-side refactor with no MCP tool, schema, protocol, or client-facing surface in the diff; the module's only in-repo importer is `server/tools/context-pack-loaders.ts` and its call shape is unchanged.
- Rollback/cleanup concern: a single-file, single-commit revert. Nothing else depends on the new private helpers.
- Fixes made before PR: none needed beyond the split; the first assembly typechecked and linted clean and the pinning tests passed on the first run.
- Known residual risk: the legacy sibling `src/tools/agent-context-pack-repo-facts.ts` still holds a near-duplicate of this logic with its own logger. It is out of this lane's scope and was not touched, so the two copies can still drift.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane produced no new MEDIUM+ review finding — the reuse-only-when-shape-identical rule and the side-effect-ordering check are already recorded in the lane contract (round 38) and were applied here rather than discovered.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or client-facing surface changed; the verification is lint, typecheck, and the isolated suite.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface moved. The fragment shape, section keys, denial reasons, and citation fields are identical to `origin/main`, so the existing fixtures already assert the correct output.

## Downstream Rollout

Notes/evidence:

- Internal refactor only. No MCP tool definition, no schema, no transport, and no Python client code is in the diff, so none of the downstream steps apply.

### DIFF
--- a/server/tools/context-pack-repo-facts.ts
+++ b/server/tools/context-pack-repo-facts.ts
@@ -22,6 +22,7 @@ import {
   type SectionFragment,
   type SectionReaderDeps,
 } from "./context-pack-sections.ts";
+import { asError, errorIdentityFields } from "./context-pack-shared.ts";
 
 /**
  * Read-side mirror of the write-side `staleness_policy` enum.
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #817 The reusable lesson from this lane is that "reuse the sibling helper" and "reuse it only when the shape is identical field for field" resolved in opposite directions inside the same file, and both answers were correct. `context-pack-shared.ts` offered two candidates: `errorIdentityFields`, which derives `error_name` / `error_message` exactly as this module already did and was reused outright, and `logDurableFailure`, which looks like a match at a glance — same two-line ERROR-then-DEBUG pattern, same `_detail` event suffix — but takes a REQUIRED `DurableFailureLogger` where this module's `logger` parameter is optional, and re-derives `error_message` on the error line. Adopting it would have moved behavior on the no-logger path while the lint findings still went green, which is precisely the non-equivalent rewrite the suite failed to catch in #806. The check that separated them was reading the candidate's parameter types and defaults against the call site rather than matching on the shape of the body, and the useful habit for the next lane is that a shared helper whose signature differs in optionality is a different helper, not the same one with a looser type. The second, smaller lesson is that lifting a fall-through sequence of guards into a named predicate is where side-effect ORDER quietly becomes a correctness property: here only the last of three exclusions may set the trimmed flag, so the predicate had to return a discriminated result carrying that verdict rather than let each guard write to a shared accumulator, and the accumulator itself had to stay set-only so it could never be cleared by a later row.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=bbd151b3c0cac998f75907fe8112cb270b30c4da at=2026-08-26T19:20:54.429Z

```
  - server/tools/context-pack-repo-facts.ts
```

</details>

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=bbd151b3c0cac998f75907fe8112cb270b30c4da at=2026-08-26T20:05:48.398Z

```
  - server/tools/context-pack-repo-facts.ts
```

</details>

PR #803, opened 2026-08-26
refactor(tier-mutations): bring server/tools/tier-mutations.ts to the lint standard (#780)

rodaddy:
## Summary

- `#780` lint sweep, one file: `server/tools/tier-mutations.ts`. `registerTierMutationTools` was 218 lines against the 100-line rule, because all four tool handlers (`set_tier`, `bulk_set_tier`, `bulk_archive`, `demote_entry`) were written inline inside the registration call.
- Each tool's `inputSchema` and `annotations` are hoisted to module consts with the description strings byte-identical; each handler is now a named module-level function; `registerTierMutationTools` is registration only.
- Two private helpers come out of the handlers: `denyUnauthorizedTables` (the per-distinct-table permission scan both bulk tools ran identically, parameterised by `canWrite` vs `canDelete` and the observed denial-string builder) and `archiveDemotedEntry` (the second half of `demote_entry`, which re-derives the delete-scope predicate rather than trusting the wider read predicate the preceding SELECT passed).
- No behavior change: schemas, defaults, SQL text, error strings, and the `set_tier_noop` / `tool_result` / `bulk_transaction_error` log event names are unchanged. The explicit-opt-in property is untouched — there is still no planning path into these handlers, so nothing here is reachable from a DreamEngine dry run.
- No file other than the target was modified.

## Verification

- Done-means: scripts/done-means/780-touched-files-lint-clean.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

RED (untouched file, before any edit):

```
server/tools/tier-mutations.ts:47:8: error eslint(max-lines-per-function): The function `registerTierMutationTools` has too many lines (218). Maximum allowed is 100. help: Consider splitting it into smaller functions.
```

`DONE_MEANS_780_FILES=server/tools/tier-mutations.ts bash scripts/done-means/780-touched-files-lint-clean.sh` -> exit 1.

GREEN (re-run on the COMMITTED tree, after the pre-commit prettier pass):

- `./node_modules/.bin/oxlint --deny-warnings server/tools/tier-mutations.ts` -> no findings, exit 0
- `bunx tsc --noEmit` -> exit 0
- `bash scripts/done-means/780-touched-files-lint-clean.sh` (diff-derived, 1 file) -> `PASS`, exit 0
[...]

PR #819, opened 2026-08-26
refactor(tier-lane): bring server/tools/tier-lane.ts to the lint standard (#780)

rodaddy:
## Summary

- Clears the two oxlint findings on `server/tools/tier-lane.ts` (#780): `registerTierLaneTool` ran 113 lines against a maximum of 100, and its inline handler carried a complexity of 11 against a maximum of 10. Both came from one shape — schema, annotations, auth, SQL, the graduation loop, and two log sites all inlined into a single `registerTool` call.
- Split along the seams `tiering.ts` (#813) and `tier-mutations.ts` (#803) already established for this file family: `tierLaneInputSchema` and `tierLaneAnnotations` hoisted verbatim to module level, and `registerTierLaneTool` reduced to a thin registration delegating to a named `handleTierLane`.
- Extracted `authorizeTierLane` / `isRefusal` (the `promotion.ts` target-or-refusal shape), plus `readLaneEvents`, `tierLaneEvents`, and `logLaneTieringFailure`. The graduation pass takes a `TierLaneRun` options object rather than five positional arguments.
- **Reuse decision worth reading:** this tool deliberately keeps its own auth helper instead of the shared `authorize` in `memory-helpers.ts`. The shapes are close but not identical field for field — `authorize` emits no denial log, so calling it would silently drop the `tier_lane_denied` warn line. Every tool here that logs a denial (`promotion.ts`, `adjacent-context.ts`, `citation-recall.ts`, `promote-entry.ts`) keeps a local helper for exactly that reason. `HandlerExtra` is likewise declared locally, matching how both siblings declare their own rather than sharing one; hoisting it would mean touching another lane's file.
- No behavior change: schemas, defaults (`dry_run` stays opt-in at `true`), every error string, both log event names, receipt field ordering, and the query text are unchanged. The SQL moved out one nesting level, so its common leading indent shrank while the internal alignment stayed exact.
- One file touched. No other existing file changed, and no helper was moved out.

## Verification

Red first, on the untouched file — `oxlint --deny-warnings server/tools/tier-lane.ts`:

```
server/tools/tier-lane.ts:57:8: error eslint(max-lines-per-function): The function `registerTierLaneTool` has too many lines (113). Maximum allowed is 100.
server/tools/tier-lane.ts:100:5: error eslint(complexity): async function has a complexity of 11. Maximum allowed is 10.
```

`DONE_MEANS_780_FILES=server/tools/tier-lane.ts bash scripts/done-means/780-touched-files-lint-clean.sh` reported `FAIL: oxlint reported findings`.

Green, re-run on the COMMITTED tree after the prettier pre-commit hook (which reported the file unchanged):

- `./node_modules/.bin/oxlint --deny-warnings server/tools/tier-lane.ts` → exit 0
- `bunx tsc --noEmit` → exit 0
- `bun run test:isolated src/tools/__tests__/tier-lane.test.ts server/tools/ported-tools-scope.test.ts` → 38 pass, 0 fail, 111 expect() calls
- `bun run test:isolated server/tools/` → 163 pass, 0 fail across 17 files
- `bash scripts/done-means/780-touched-files-lint-clean.sh` (diff-derived from `origin/main...HEAD`) → exit 0

Existing tests are unmodified. `src/tools/tier-lane.ts` reports 100% function and line coverage under the pinning tests.

## Critical Self-Review

- Highest-risk behavior: the auth path. Splitting it into `authorizeTierLane` moves the permission check, the namespace-denial log, and `physicalNamespace()` resolution behind one call. The ordering is preserved exactly — role/table permission first (so an unprivileged caller never learns whether a namespace exists), then `canTargetNamespace`, then `physicalNamespace` on the way out, which is where the original called it too.
- Assumptions that could be wrong: that the SQL reindent is inert. It sits in a template literal, so the bytes did change — the common leading indent shrank by 6 spaces as the code moved out a nesting level. I diffed it token for token and the internal alignment is preserved; Postgres is whitespace-insensitive here. If anything ever asserts on the literal query string this would surface, and nothing in the suite does.
- Missing/weak tests: no test asserts the `tier_lane_denied` warn line is emitted, which is the one behavior most at risk from the reuse decision above. That gap is what made me keep the local helper rather than take the shared `authorize` — the suite would not have caught the dropped log. Not filed as a new test here because the brief holds existing tests unmodified; worth a follow-up.
- Security/permission risk: none introduced. Write authority over `thoughts` and the namespace predicate are unchanged, and the namespace is still resolved from the token when the caller supplies none.
- Downstream client/runtime risk: none. `docs/downstream-rollout.md` applies to MCP tool/schema/protocol changes; the published `tier_lane` schema, description, annotations, and response envelope are byte-identical, so there is no contract surface for a client to notice.
- Rollback/cleanup concern: single-file, single-commit, revert-clean.
- Fixes made before PR: first draft imported `AuthIdentity` from `./types.ts`, which re-exports nothing of the sort — corrected to `../auth/types.ts`, matching `promotion.ts` and `memory-helpers.ts`.
- Known residual risk: low. The refactor is structural; the read-back against the original was done line by line specifically because #806 shipped a non-equivalent predicate rewrite that the suite did not catch.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the reuse-vs-denial-logging tension is already recorded as lane-contract round 38 (reuse only when the shape matches field for field, defaults and failure shape included); this PR is an instance of that rule, not a new pattern.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or query behavior changed — this is a structural refactor with an identical published contract.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract changed. `contracts/server/server-tier-tools.fixture.json` pins the `tier_lane` schema and annotations, which moved verbatim to module-level consts; `server/tools/ported-tools-scope.test.ts` passes unmodified, which is what proves it.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: the MCP tool name, input schema, annotations, description string, and response envelope are unchanged, so no downstream client sees a different surface. The rollout gate keys on tool/schema/protocol/client-facing changes; this PR has none.

### DIFF
--- a/server/tools/tier-lane.ts
+++ b/server/tools/tier-lane.ts
@@ -33,6 +33,7 @@
  */
 import { z } from "zod";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import type { AuthIdentity } from "../auth/types.ts";
 import { canWrite } from "../auth/permissions.ts";
 import { canTargetNamespace } from "../auth/namespace-policy.ts";
 import {
@@ -54,6 +55,38 @@ import {
 /** Lane events scanned when the caller names no preference. */
 const DEFAULT_EVENTS_SCANNED = 100;
 
+const tierLaneInputSchema = {
+  session_key: z
+    .string()
+    .min(1)
+    .max(500)
+    .describe("Stable lane identifier to tier"),
+  namespace: z
+    .string()
+    .min(1)
+    .max(500)
+    .optional()
+    .describe("Namespace of the lane (defaults to caller's clientId)"),
+  dry_run: z
+    .boolean()
+    .optional()
+    .describe("Preview without writing durable thoughts (default true)"),
+  limit: z
+    .number()
+    .int()
+    .min(1)
+    .max(500)
+    .optional()
+    .describe(`Max lane events to scan (default ${DEFAULT_EVENTS_SCANNED})`),
+};
+
+const tierLaneAnnotations = {
+  title: "Tier Session Lane",
+  readOnlyHint: false,
+  destructiveHint: false,
+  idempotentHint: true,
+};
+
 export function registerTierLaneTool(
   server: McpServer,
   dependencies: MemoryToolDependencies,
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #819 A near-miss helper is the trap this refactor had to avoid, and the tell is what the helper does NOT do. `memory-helpers.ts` exports `authorize`, whose signature covers the whole auth block in `tier-lane.ts` — permission check, namespace resolution, denial envelope — and taking it would have deleted more lines than any other single move. It is still wrong: `authorize` emits no log, and `tier_lane` warns `tier_lane_denied` on namespace refusal, so the swap would have silently dropped a denial log line with every test still green (nothing asserts that warn fires). Reading the four other tools here that log denials — `promotion.ts`, `adjacent-context.ts`, `citation-recall.ts`, `promote-entry.ts` — showed all four keep a local auth helper and none call `authorize`, which is the shared-helper boundary already being drawn consistently rather than four independent oversights. The reusable check: when a candidate helper matches on inputs and outputs, diff its SIDE EFFECTS too — logs, metrics, cache writes — because those are invisible to both the type checker and a suite that only asserts return values, and lane-contract round 38's "identical field for field, defaults and failure shape included" has to be read as including them. Concretely: `rg '_denied"|logger\.(warn|error)'` across the target before reaching for a shared auth or failure helper.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=477bb0007c5550eceee2bd0d2de75a5af0c9971d at=2026-08-26T19:20:54.180Z

```
  - server/tools/tier-lane.ts
```

</details>

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=477bb0007c5550eceee2bd0d2de75a5af0c9971d at=2026-08-26T20:10:07.934Z

```
  - server/tools/tier-lane.ts
```

</details>

PR #820, opened 2026-08-26
refactor(reporting): bring server/tools/reporting.ts to the lint standard (#780)

rodaddy:
## Summary

- Clears the two oxlint findings on `server/tools/reporting.ts`: a
  `NamespacePredicate` type import with no reference, and the inline
  `access_report` handler at complexity 12.
- Follows the shape the rest of this sweep landed (`promotion.ts`,
  `tiering.ts`, `scan-namespace.ts`): schemas hoisted to module scope, tool
  registration reduced to the `registerTool` call plus a one-line delegation,
  and the work in named module-level functions.
- `resolveReadScope` states the authenticate-then-filter-then-refuse gate once
  instead of twice. Both tools opened with the same three steps and the same
  two error strings; the returned success arm carries the narrowed identity, so
  the split needs no non-null assertion.
- `accessAggregate` owns the single-pass log query and the numeric coercion of
  its one row. `foldTableStats` and `rankNamespaces` are the two pure
  transforms `get_stats` was doing inline.
- No behavior change. SQL text, bound parameters, error strings, log event
  names, response field order, and every `?? 0` fallback are byte-identical to
  the previous revision. No test was modified.

## Verification

```
server/tools/reporting.ts:25:8: error eslint(no-unused-vars): Type 'NamespacePredicate' is imported but never used.
server/tools/reporting.ts:94:5: error eslint(complexity): async function has a complexity of 12. Maximum allowed is 10.
```

`DONE_MEANS_780_FILES=server/tools/reporting.ts bash
scripts/done-means/780-touched-files-lint-clean.sh` exited 1 on that tree.

- `./node_modules/.bin/oxlint --deny-warnings server/tools/reporting.ts` -> exit 0
- `bash scripts/done-means/780-touched-files-lint-clean.sh` -> exit 0
  (diff-derived; resolves `server/tools/reporting.ts` from
  `git diff --name-only origin/main...HEAD`)
- `bun run test:isolated server/tools/ src/tools/__tests__/access-report.test.ts
  src/tools/__tests__/get-stats.test.ts` -> 171 pass, 0 fail across 19 files

The three files that pin this module's behavior are
`server/tools/reporting.pg.test.ts`, `src/tools/__tests__/access-report.test.ts`,
and `src/tools/__tests__/get-stats.test.ts`; on their own they are 17 pass, 0
fail.

## Critical Self-Review

- Highest-risk behavior: the namespace isolation predicate. Every read here is
  scoped by an auth-derived predicate, and `access_report` additionally leans on
  that predicate to make an unreadable entry indistinguishable from a miss. A
  rewrite that widened the scope would still return rows and still pass, which
  is exactly the failure #806 hit. So no predicate expression was rewritten:
  `findReadableEntryTable`, `tableStats`, `accessLogStats`, `graphCounts`, and
  `topAccessedEntries` are unchanged below the new handler layer, and the diff
  shows their SQL as context rather than as edits.
- Assumptions that could be wrong: that `resolveReadScope` preserves guard
  ORDER, not just the two strings. It matters -- an unauthenticated caller must
  get "not authenticated" and never reach the table filter. The success arm
  returns the identity so the narrowing is real rather than asserted, and both
  callers destructure it rather than re-deriving.
- Missing/weak tests: nothing new is tested because nothing new is claimed. The
  existing suite exercises both tools through their registered handlers, so the
  delegation is on the covered path. The one behavior with no direct coverage is
  the `avg_access_count` divide-by-zero arm at `perTable.length === 0`, which is
  unreachable while the readable-tables gate refuses an empty list; it is
  preserved verbatim rather than simplified away.
- Security/permission risk: none identified. No exported signature changed, no
  call site outside this file was touched, and the auth gate is stated once
  instead of twice, which removes the drift risk of two copies rather than
  adding one.
- Downstream client/runtime risk: none. `docs/downstream-rollout.md` applies to
  MCP tool/schema/protocol or client-facing changes; the two tool descriptions,
  input schemas, annotations, and response shapes are unchanged, so no
  rtech-mcps, mcp2cli, or Hermes step is triggered. The schemas moved to module
  constants and are passed by reference to the same `inputSchema` field.
- Rollback/cleanup concern: single-file, single-commit, revertable on its own.
  No temp artifact and no worktree were created.
- Fixes made before PR: the first extraction left `identity` typed as possibly
  undefined at the two call sites and I was about to reach for a non-null
  assertion. Returning the identity from the scope resolver instead is what
  removed the need, and it is the better shape anyway.
- Known residual risk: the NUL byte that joins the namespace map key is now a
  named constant rather than an inline escape. The value is identical, so
  collision behavior is identical, but it is the one literal that moved.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a lint-standard refactor with no behavior change and no new finding class; the shape it follows is already recorded in the lane contract for #780, and adding an entry per swept file would dilute the lane files rather than sharpen them.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime surface, schema, or transport behavior changed; the isolated suite covers both tools through their registered handlers.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract moved. `server/tools/reporting.ts` keeps both tool names, descriptions, input schemas, annotations, and response field order; the parallel `src/tools/access-report.ts` and `src/tools/get-stats.ts` are untouched and their tests pass unmodified against this branch.

## Downstream Rollout

Notes/evidence:

- Classification is "not applicable" on every downstream row for one reason:
  nothing client-facing moved. The rollout doc keys on MCP tool, schema,
  protocol, or client-facing change; this branch changes only the internal
  arrangement of one server module.
- Reuse: no new shared helper was written. `ALL_TABLES`, `CONTENT_PREVIEW`,
  `PREVIEW_WIDTH`, and `qualifyNamespacePredicate` come from
  `curation-helpers.ts` as before. I checked `read-scope.ts` for the read gate
  before writing `resolveReadScope` and it is a different question -- it answers
  which NAMESPACES an identity may read and returns a namespace filter, whereas
  this gate answers which TABLES and returns an error result. The
  `ALL_TABLES.filter((table) => canRead(...))` idiom appears in roughly twenty
  places across `src/` and `server/` with several different follow-on shapes, so
  hoisting it into a shared sibling is a decision for the module that owns that
  boundary, not a side effect of a lint sweep. `resolveReadScope` stays private
  to this file.

### DIFF
--- a/server/tools/reporting.ts
+++ b/server/tools/reporting.ts
@@ -20,10 +20,7 @@
 import { z } from "zod";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { canRead } from "../auth/permissions.ts";
-import {
-  namespacePredicate,
-  type NamespacePredicate,
-} from "../auth/namespace-policy.ts";
+import { namespacePredicate } from "../auth/namespace-policy.ts";
 import type { AuthIdentity, ResourceTable } from "../auth/types.ts";
 import { canonicalNamespace } from "../../src/shared-namespace.ts";
 import {
@@ -48,7 +45,7 @@ const TABLE_ALIAS: Readonly<Record<ResourceTable, string>> = {
   sessions: "s",
 };
 
-/** Observed current-src caps on the reported breakdowns. */
+/** Observed current-src maxima on the reported breakdowns. */
 const TOP_NAMESPACES = 10;
 const TOP_ACCESSED = 10;
 const TOP_ENTITY_TYPES = 25;
@@ -56,6 +53,37 @@ const TOP_ENTITY_TYPES = 25;
 /** Trend bands from observed current-src: +/-20% around the prior window. */
 const RISING_RATIO = 1.2;
 const DECLINING_RATIO = 0.8;
+
+/** Map-key separator, chosen so no namespace value can contain it. */
+const NAMESPACE_KEY_SEPARATOR = "\u0000";
+
+/** Look-back window applied when the caller names none. */
+const DEFAULT_PERIOD_DAYS = 30;
+
+/** Minimal shape of the handler `extra` argument these tools read. */
+interface HandlerExtra {
+  authInfo?: Parameters<typeof authIdentity>[0];
+}
+
+const accessReportInputSchema = {
+  entry_id: z.string().uuid().describe("UUID of the entry to report on"),
+  days: z
+    .number()
+    .int()
+    .min(1)
+    .max(365)
+    .optional()
+    .describe("Number of days to look back (default 30)"),
+};
+
+const getStatsInputSchema = {
+  raw: z
+    .boolean()
+    .optional()
+    .describe(
+      "Return physical namespace names instead of canonical public names",
+    ),
+};
 
 export function registerReportingTools(
   server: McpServer,
@@ -74,16 +102,7 @@ function registerAccessReport(
     {
       description:
         "Returns a detailed access report for a specific entry: total accesses, unique queries, unique agents, access trend, and recency.",
-      inputSchema: {
-        entry_id: z.string().uuid().describe("UUID of the entry to report on"),
-        days: z
-          .number()
-          .int()
-          .min(1)
-          .max(365)
-          .optional()
-          .describe("Number of days to look back (default 30)"),
-      },
+      inputSchema: accessReportInputSchema,
       annotations: {
         title: "Access Report",
         readOnlyHint: true,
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #820 When a lint sweep splits an inline MCP handler into module-level functions, the auth guard is the part that silently changes shape. The original handlers here read `const identity = authIdentity(...)`, refused on falsy, then filtered `ALL_TABLES` by `canRead` and refused again -- so every line after the two guards had `identity` narrowed to non-undefined for free, by control flow inside one function body. Extract the guard into a helper and that narrowing does not cross the call boundary: the caller still sees `AuthIdentity | undefined`, and the shortest way to make the typechecker stop complaining is `identity as AuthIdentity` or `identity!` at each call site. That reads as a formatting detail and is not one -- it is an assertion that the guard ran, placed at the site that no longer contains the guard, and it will keep compiling if someone later reorders the two refusals or adds a third early return above them. The fix is to make the helper RETURN the narrowed identity alongside whatever else it resolved (here `{ identity, tables }` on success, `{ error }` on refusal), so the success arm carries the proof and the call sites destructure it. Same rule as the predicate-rewrite lesson from #806, one level up: a refactor that preserves runtime behavior can still discard a compile-time invariant, and both failures look identical to a green suite. Worth checking on every remaining #780 file whose handler opens with an auth gate -- if the extraction left a `!` or an `as` where the guard used to be, the invariant left the building with it.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=83a42295d6819911d5ea09c4a49160645812ec10 at=2026-08-26T19:20:54.529Z

```
  - server/tools/reporting.ts
```

</details>

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=83a42295d6819911d5ea09c4a49160645812ec10 at=2026-08-26T20:10:07.862Z

```
  - server/tools/reporting.ts
```

</details>

PR #821, opened 2026-08-26
refactor(people): bring server/tools/people.ts to the lint standard (#780)

rodaddy:
## Summary

- `registerPeopleTools` held both tool registrations inline at 198 lines, and `find_person`'s handler hit a complexity of 21 by branching the two search modes, their two SQL statements, and the shared not-found handling inside one closure. Both are #780 findings on `server/tools/people.ts`.
- Split along the seam the file already had: `registerUpsertPerson` and `registerFindPerson` are now module-level, matching how `entities.ts` and `promotion.ts` register a family of tools. `find_person`'s two modes became `findPeopleSemantic` and `findPeopleByName`, each taking one options object, so the mode choice is a single expression in the handler rather than a branch that also owns the queries.
- Both input schemas and the upsert statement are hoisted to module constants, which puts the `ON CONFLICT` uniqueness key next to the schema field it matches on.
- Reuse: the inline embedding block computed the pgvector literal, the timestamp, the model, and the embedded flag exactly as `embeddingFields` in `memory-helpers.ts` already does, field for field, so the tool now calls it. `authorize` from the same module was deliberately **not** reused — its denial text differs from this tool's (`<role> role cannot write to namespace '<ns>'` versus `namespace write denied`), and matching the observed strings matters more here than sharing the gate.
- No behavior change and no new helper file. `server/tools/people.ts` is the only file touched; `registerPeopleTools`' exported signature is unchanged, so `server/tools/index.ts` and `server/tools/entities-graph.pg.test.ts` are unaffected. No tests were modified.

## Verification

```
server/tools/people.ts:33:8: error eslint(max-lines-per-function): The function `registerPeopleTools` has too many lines (198). Maximum allowed is 100.
server/tools/people.ts:82:5: error eslint(complexity): async function has a complexity of 21. Maximum allowed is 10.
```

`DONE_MEANS_780_FILES=server/tools/people.ts bash scripts/done-means/780-touched-files-lint-clean.sh` exited 1.

Green, re-run on the committed tree after the pre-commit prettier hook reformatted it:

- `./node_modules/.bin/oxlint --deny-warnings server/tools/people.ts` -> exit 0, no findings
- `bash scripts/done-means/780-touched-files-lint-clean.sh` -> exit 0, PASS
- `bunx tsc --noEmit` -> exit 0, zero errors
- `bun run test:isolated server/tools/ src/tools/__tests__/find-person.test.ts src/tools/__tests__/upsert-person.test.ts src/tools/__tests__/protocol.test.ts` -> 196 pass, 0 fail across 20 files, tests exited 0

Equivalence check beyond the suite: all three SQL template literals (the upsert, the semantic select, the name select) were extracted from `origin/main:server/tools/people.ts` and from the committed file and compared byte for byte — identical in every case, before and after prettier ran.

Python package is untouched, so its checks are not applicable. No migration, no schema change, and no wire-visible change, so a live smoke has nothing new to exercise.

## Critical Self-Review

- Highest-risk behavior: `find_person`'s two search modes. The original assigned into a mutable `rows` from an if/else; the refactor returns from two functions and uses `rows === null` as the embedding-failure signal. `findPeopleByName` never returns null, so the only path to `Failed to generate query embedding` is still the semantic branch with a falsy embedding — same condition, same string, same position relative to the logging call.
- Assumptions that could be wrong: that `embeddingFields` is field-for-field identical to the inline block it replaces. Checked directly rather than assumed — it computes `embedding ? toSql(embedding) : null`, `embedding ? new Date().toISOString() : null`, `embedding ? (dependencies.embeddingModel ?? null) : null`, and `embedding !== null`, which is exactly what the four inline expressions were, including the `?? null` on the model.
- Missing/weak tests: the suite covers `upsert_person` create/update and `find_person` in both modes, but there is no test that pins the semantic-mode embedding-failure string. That path was verified by reading rather than by a failing test, and it is unchanged rather than newly written.
- Security/permission risk: the namespace predicate is the isolation boundary here and it moved into the two query helpers as a passed-in `NamespacePredicate`, still built by `namespacePredicate(identity, "read", 4)` in the handler with the same placeholder offset of 4. The parameter arrays still spread `predicate.values` in the same trailing position, so no placeholder shifted. The write path's `canWrite` then `canTargetNamespace` order is untouched, which preserves the property that an unprivileged caller cannot learn whether a namespace exists.
- Migration/deploy risk: none. No SQL, schema, or migration changed.
- Downstream client/runtime risk: none. Tool names, descriptions, input schemas, annotations, result shapes, error strings, and log event names are all unchanged, so nothing MCP-visible moved and `docs/downstream-rollout.md` classifies this as not applicable.
- Rollback/cleanup concern: single-file, single-commit, no new files and no moved helpers, so a revert is complete on its own.
- Fixes made before PR: none needed beyond the refactor — the first oxlint and tsc run after the rewrite were both clean.
- Known residual risk: the deliberate not-reuse of `authorize` leaves this tool with its own auth gate. That is correct today because the denial strings differ, but it means a future change to the shared gate will not reach this file.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane produced no new MEDIUM+ finding — the not-reuse decision is already covered by the lane contract's round-38 rule on field-for-field identity, which is what drove it.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: nothing wire-visible or persisted changed, so a live check would exercise identical behavior.

## Contract Parity

- Contract parity: [x] runtime-specific because: `contracts/server/server-people.fixture.json` freezes `find_person`'s not-found text and `upsert_person`'s created/updated envelope, and both are unchanged by this refactor, so the existing fixture already covers it and needs no edit.

## Downstream Rollout

Notes/evidence:

- All four are not applicable: this is an internal restructuring of one server tool file with no change to tool names, schemas, result shapes, or error strings, so no client cache, generated skill, or agent canary sees anything different.

### DIFF
--- a/server/tools/people.ts
+++ b/server/tools/people.ts
@@ -18,7 +18,8 @@ import {
   canTargetNamespace,
   namespacePredicate,
 } from "../auth/namespace-policy.ts";
-import { contentHash } from "./memory-helpers.ts";
+import type { NamespacePredicate } from "../auth/namespace-policy.ts";
+import { contentHash, embeddingFields } from "./memory-helpers.ts";
 import {
   authIdentity,
   errorResult,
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #821 The reuse decision in this lane turned on denial-string identity, not on shape identity, and the two pointed opposite ways in the same file. `memory-helpers.ts` exports both `embeddingFields` and `authorize`, and `people.ts` had inline code matching each. `embeddingFields` was a real reuse: the four inline expressions computed the pgvector literal, the ISO timestamp, `dependencies.embeddingModel ?? null`, and `embedding !== null` in exactly the helper's field order, including the `?? null`, so calling it changes nothing observable. `authorize` looked like an equally clean fit — same `canWrite` then `canTargetNamespace` order, same resource table — but it emits `Permission denied: <role> role cannot write to namespace '<ns>'` where this tool emits `Permission denied: namespace write denied`, and `contracts/server/server-people.fixture.json` exists specifically to freeze this capability's user-visible text. Adopting it would have been a silent contract change wearing the costume of a refactor. The reusable check is that field-for-field identity has to include the failure envelope's exact strings, not just its control flow, and that a capability with a contract fixture is where that distinction actually bites. Separately, the equivalence read-back that round 38 asks for was worth doing mechanically rather than by eye: extracting all three SQL template literals from `origin/main` and from the committed file and comparing them byte for byte caught nothing here, but it is the check that would have caught #806's non-equivalent predicate rewrite, and it costs one command.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=05ce0d4e57b956093d12f6de70d2804216a8db5e at=2026-08-26T19:21:04.010Z

```
  - server/tools/people.ts
```

</details>

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=05ce0d4e57b956093d12f6de70d2804216a8db5e at=2026-08-26T20:10:07.873Z

```
  - server/tools/people.ts
```

</details>

PR #802, opened 2026-08-26
refactor(promotion): bring server/tools/promotion.ts to the lint standard (#780)

rodaddy:
## Summary

- `server/tools/promotion.ts` carried two oxlint findings: `registerPromotionTools` at 191 lines (max 100) and the `promote_shared` handler at complexity 19 (max 10). Both are cleared.
- Same shape as `search-brain.ts` (lane 2) and `search-all.ts` (011bf24): both input schemas and both annotation objects hoist to module consts with byte-identical description strings, and each handler becomes a named module-level function, so the `registerTool` bodies are registration only.
- `promote_shared` splits along its own steps: `authorizePromotion` (identity check plus the unconditional legacy-target refusal), `loadPromotionSource` (the read under the caller's own scope), `runPromotion` (the service call and its error mapping, taking an options object instead of eight positional arguments). `list_namespaces` splits into `countNamespaceRows` and `foldNamespaceCounts`.
- No behavior change: schemas, defaults, error strings, SQL text, and the `tool_result` / `promote_shared_denied` / `promote_shared_refused` / `promote_shared_error` log event names are identical.
- Follow-up, deliberately NOT done here: `promotionAuth`, `isPromotionIdentity`, and the `LEGACY_SHARED_NAMESPACE` constant are duplicated in `server/tools/promote-entry.ts`. That file is owned by a concurrent #780 lane, so extracting a shared module now would collide. It should be extracted once both lanes land.

## Verification

- Done-means: scripts/done-means/780-touched-files-lint-clean.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

RED (untouched file, before any edit):

```
server/tools/promotion.ts:78:8: error eslint(max-lines-per-function): The function `registerPromotionTools` has too many lines (191). Maximum allowed is 100.
server/tools/promotion.ts:199:5: error eslint(complexity): async function has a complexity of 19. Maximum allowed is 10.
```

`DONE_MEANS_780_FILES=server/tools/promotion.ts bash scripts/done-means/780-touched-files-lint-clean.sh` -> exit 1.

GREEN (re-run on the COMMITTED tree, after the pre-commit prettier pass):

- `./node_modules/.bin/oxlint --deny-warnings server/tools/promotion.ts` -> exit 0
- `bunx tsc --noEmit` -> exit 0
[...]

PR #797, opened 2026-08-26
refactor(promote-entry): bring server/tools/promote-entry.ts to the lint standard (#780)

rodaddy:
## Summary

- `server/tools/promote-entry.ts` carried the two remaining lint findings in the file: `registerPromoteEntryTool` ran 104 lines (maximum 100) and its handler had a complexity of 12 (maximum 10). This brings the file to the standard using the shape lanes 2 and 3 established on `search-brain.ts` and `search-all.ts`.
- The input schema and the MCP annotations move to module consts (`promoteEntryInputSchema`, `promoteEntryAnnotations`) with byte-identical description strings; `registerPromoteEntryTool` shrinks to registration alone.
- The handler body splits into three named module-level helpers: `resolvePromotionTarget` (default target plus the legacy-namespace refusal), `runPromotion` (the service call and its success log), and `respondToPromotionFailure` (the failure log and the statusCode-gated message selection). The two multi-argument helpers take an options object.
- No behavior change. The description string, every schema field and its `.describe()` text, the dry-run default of `true`, the legacy-namespace refusal covering both the hardcoded `collab` and the configured `legacySharedNamespace`, the `promote_entry_denied` / `tool_result` / `promote_entry_failed` event names and their field sets, and both caller-facing error strings are unchanged.
- Reuse was checked first and declined with a reason. `respondToSearchFailure` (`server/tools/tool-failure.ts`) logs a search-shaped record (`namespace`/`mode`/`tier`) and always returns the error's own message, whereas this catch logs promotion fields and deliberately withholds an unmarked error's raw driver text — adopting it would change what reaches a caller. `normalizeSearchArgs` (`server/tools/search-request.ts`) has no promotion-shaped defaults to share. No second caller exists for either new helper, so both stay private to this file.
- Exactly one file is touched: `server/tools/promote-entry.ts`. No helper was moved out and no other file was edited.

## Verification

- Done-means: scripts/done-means/780-touched-files-lint-clean.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

RED (untouched file, before any edit):

```
server/tools/promote-entry.ts:82:8: error eslint(max-lines-per-function): The function `registerPromoteEntryTool` has too many lines (104). Maximum allowed is 100.
server/tools/promote-entry.ts:121:5: error eslint(complexity): async function has a complexity of 12. Maximum allowed is 10.
```

[...]

PR #822, opened 2026-08-26
refactor(promotion-shared): bring server/tools/promotion.ts and server/tools/promote-entry.ts to the lint standard (#780)

rodaddy:
## Summary

- `server/tools/promotion.ts` and `server/tools/promote-entry.ts` were each already lint-clean (#802, #797), but each carried a private copy of the same two promotion rules: the lanes that cleaned them ran concurrently, so neither could safely reach into the other's file.
- Moves the genuinely identical helpers into one shared sibling, `server/tools/promotion-shared.ts`, imported by both: `LEGACY_SHARED_NAMESPACE` (`promotion.ts:53` / `promote-entry.ts:56`), `isPromotionIdentity` (`promotion.ts:86` / `promote-entry.ts:75`), and the legacy-target refusal as `legacyTargetRefusal()` (`promotion.ts:282` / `promote-entry.ts:137`).
- `isPromotionIdentity` admitted exactly `promoter`, `admin`, `ob-admin` in both copies and differed only in the operand order of a side-effect-free `||` chain. The refusal predicate and its message template were byte-identical in both copies.
- Each caller keeps its **own** default target: `promote_shared` resolves to `canonicalSharedNamespace` and `promote_entry` to `sharedNamespace`. Those differ under a split config, so only the refusal is shared, never the default.
- Left deliberately duplicated because the pairs are not identical field for field: `promotionAuth` (`promotion.ts` carries `tokenClientId`, `promote-entry.ts` does not, and `src/promotion-service.ts:216` reads it to stamp `promoted_by`), `runPromotion` (different log fields, different error mapping, different result shape), and the identity-denial responses (different log event names and messages).
- No behavior change: schemas, defaults, error strings, SQL, log event names, and section ordering are identical. No existing test was modified. No file outside the two targets and the new sibling was touched.

## Verification

`./node_modules/.bin/oxlint --deny-warnings` on all three files exits 0. `bunx tsc --noEmit` exits 0. `bun run test:isolated` over the pinning tests (`src/tools/__tests__/promote-shared.test.ts`, `src/tools/__tests__/promote-entry.test.ts`, `src/tools/__tests__/promote-entry.pg.test.ts`, `server/tools/promotion.pg.test.ts`, `server/tools/ported-tools-scope.test.ts`, `src/ob-admin-role.test.ts`) reports 74 pass / 0 fail; `bun run test:isolated server/tools/` reports 163 pass / 0 fail. The done-means script, run on the committed tree, derives its file list from `git diff --name-only origin/main...HEAD`, picks up all three files, and exits 0. Python package is untouched, so its checks are not applicable; this is a pure internal refactor with no MCP contract change, so no live smoke applies.

## Critical Self-Review

- Highest-risk behavior: sharing the legacy-target refusal between two tools whose default target namespaces genuinely differ. Handled by sharing only the refusal predicate and message and having each caller resolve and pass its own `target`, so the defaults never converge.
- Assumptions that could be wrong: that reordering the operands of `isPromotionIdentity`'s `||` chain is unobservable. It holds because the three comparisons are pure string equality against `identity.role` with no side effects, and the merged body admits exactly the same role set as each original.
- Missing/weak tests: no test was added, because this change is defined as behavior-preserving and the existing suites already pin both tools' identity refusals and legacy-target refusals. A new test here would assert the same thing the existing ones do.
- Security/permission risk: `isPromotionIdentity` is a permission predicate, so a wrong role set would widen who may promote. Verified by reading the merged body back against both originals: the same three roles, no additions.
- Downstream client/runtime risk: none. No MCP tool name, input schema, annotation, output shape, or error string changed, so no client observes a difference. `docs/downstream-rollout.md` checked; nothing applies.
- Rollback/cleanup concern: a single revertable commit adding one file and editing two.
- Fixes made before PR: none needed; the extraction was read back against the originals predicate by predicate before the first check ran.
- Known residual risk: `errorResult(refusal)` narrows a `string | undefined` with a truthiness test, so an empty-string refusal would fall through. It cannot be empty because the message template always begins with `Permission denied:`.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the duplication was a known consequence of two concurrent single-owner lint lanes, already recorded as lane-contract round 38 and rule 32, and this PR is that rule being applied rather than a new lesson.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no MCP contract, transport, or client-facing surface changed.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface changed, so no fixture describes anything different.

## Downstream Rollout

Notes/evidence:

- Internal refactor only. The `promote_shared` and `promote_entry` tool registrations, input schemas, annotations, descriptions, result shapes, and error strings are unchanged, so nothing downstream observes this.

### DIFF
--- a/server/tools/promote-entry.ts
+++ b/server/tools/promote-entry.ts
@@ -44,16 +44,10 @@ import {
   type MemoryToolDependencies,
 } from "./types.ts";
 import { tableEnum } from "./curation-helpers.ts";
-
-/**
- * The pre-rename shared namespace.
- *
- * Refused as a promotion TARGET unconditionally, matching `promotion.ts` and
- * `server/auth/namespace-policy.ts`. `legacySharedNamespace` is empty by
- * default, so a config-gated check would permit `collab` in exactly the default
- * deployment -- the rule is about the name, not the setting.
- */
-const LEGACY_SHARED_NAMESPACE = "collab";
+import {
+  isPromotionIdentity,
+  legacyTargetRefusal,
+} from "./promotion-shared.ts";
 
 /**
  * Convert the server identity into the promotion service's shape.
@@ -69,15 +63,6 @@ function promotionAuth(identity: AuthIdentity): AuthInfo {
     namespaceSource:
       identity.namespaceSource === "delegated" ? "header" : "token",
   } as AuthInfo;
-}
-
-/** @returns Whether this identity may promote entries between namespaces. */
-function isPromotionIdentity(identity: AuthIdentity): boolean {
-  return (
-    identity.role === "admin" ||
-    identity.role === "ob-admin" ||
-    identity.role === "promoter"
-  );
 }
 
 /** Input schema for `promote_entry`. */
@@ -130,18 +115,12 @@ function resolvePromotionTarget(
   requested: string | undefined,
 ): { target: string } | { refusal: string } {
   const shared = sharedNamespaceConfig();
+  // The default is the PHYSICAL shared namespace here, where `promote_shared`
+  // defaults to the canonical one; the two differ under a split config and
+  // that difference is behavior, so each tool keeps its own default.
   const target = requested ?? shared.sharedNamespace;
-  // The legacy name is a migration SOURCE, never a write target: accepting
-  // it would recreate the two-names-for-one-lane split the canonical
-  // namespace decision exists to end.
-  if (
-    target === LEGACY_SHARED_NAMESPACE ||
-    target === shared.legacySharedNamespace
-  ) {
-    return {
-      refusal: `Permission denied: '${target}' is a legacy migration source and cannot be a promotion target; use '${shared.canonicalSharedNamespace}'`,
-    };
-  }
+  const refusal = legacyTargetRefusal(target, shared);
+  if (refusal) return { refusal };
   return { target };
 }
 
--- a/server/tools/promotion-shared.ts
+++ b/server/tools/promotion-shared.ts
@@ -1,0 +1,71 @@
+/**
+ * Helpers shared by the two promotion tools.
+ *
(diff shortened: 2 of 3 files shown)


rodaddy:
harvested: #822 A duplicate-extraction lane's real work is deciding what NOT to share, and the tell is whether a field difference reaches an observable output. Six candidate pairs across these two files looked shareable by shape; only three were identical field for field. `promotionAuth` was the trap: the two bodies differ by one field, `tokenClientId`, which reads as an incidental omission until you follow it to `src/promotion-service.ts:216`, where it decides the `promoted_by` stamp — merging them would have silently changed provenance on every `promote_entry` call while every existing test stayed green. The second trap was subtler and would have survived review: `promote_shared` and `promote_entry` share a byte-identical legacy-target refusal predicate AND message, but resolve their default target from different config fields (`canonicalSharedNamespace` vs `sharedNamespace`), which coincide in the default deployment and diverge under a split one. Sharing the whole resolver would have been correct on every developer machine and wrong in production. The rule that falls out: extract the *judgment* and leave the *defaults* with the callers, because a default is behavior and a predicate is a rule. Before merging any pair, follow each differing field to the output it changes; if it reaches one, the pair is not identical no matter how similar the code reads.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=09e9efd3e6b3d0fd74b78de5568ab8e5a50784a6 at=2026-08-26T20:12:52.294Z

```
linting 3 file(s) from: git diff --name-only origin/main...HEAD
  - server/tools/promote-entry.ts
  - server/tools/promotion-shared.ts
  - server/tools/promotion.ts
```

</details>

PR #828, opened 2026-08-26
refactor(health): bring server/transport/health.ts to the lint standard (#780)

rodaddy:
## Summary

- `getSingleWorkerHealth` (`server/transport/health.ts:192` at `origin/main`) carried a complexity of 21 against the configured maximum of 10 — the only oxlint finding in the file.
- The branching was never one decision. It was four independent readings (nats, maintenance producer, capture lane, database) plus their conditional log projections and their per-lane degraded warnings, all inlined into one body. Split along those seams into named module-level helpers: `readWorkerLanes` (reads each optional lane exactly once), `resolveStatus` (owns the healthy/degraded verdict), `producerLogFields` and `captureLogFields` (own the two conditional log projections), and `warnDegradedLanes` (owns the two warnings).
- No behavior moves. The `/health` JSON payload keeps identical field order, identical status strings, identical optional-field presence rules (`revision`, `maintenance_producer`, `capture` are still omitted rather than nulled), and the same `probeTimeoutMs` handling — `core01` and the local deploy script read this shape. Log event names (`worker_health_result`, `maintenance_producer_health_degraded`, `capture_lane_health_degraded`) and their field sets are unchanged, as is the order the two warnings fire in.
- The exported signature is untouched, so every caller — `server/transport/http-app.ts:97`, `server/transport/index.ts:27`, and the `scripts/done-means/625` and `647` drivers — is unaffected. No other file is modified; no helper was moved out of the target.
- One deliberate predicate rewrite: the former `producerDegraded`/`captureDegraded` booleans (`x?.stale === true`) are now inlined as `x?.stale !== true` inside the healthy conjunction. Equivalent for all three inhabited states (`undefined`, `false`, `true`), and both drivers exercise all three.

## Verification

Run on the committed tree, after the pre-commit prettier hook reformatted the staged file:

- `./node_modules/.bin/oxlint --deny-warnings server/transport/health.ts` -> exit 0, 0 findings (was 1 finding, complexity 21)
- `bash scripts/done-means/780-touched-files-lint-clean.sh` -> exit 0, PASS
- `bun run test:isolated server/transport/` -> 67 pass, 0 fail across 4 files
- `bash scripts/done-means/647-capture-liveness.sh` -> PASS, 4 clauses run, 0 failed
- `bash scripts/done-means/625-sweep-heartbeat.sh` -> PASS, 5/5 clauses

The two `done-means` drivers are the pinning evidence that matters here: they call the real `getSingleWorkerHealth` and assert the status verdict and the presence/absence of the optional blocks, which is exactly the surface the extraction touched.

Migrations are untouched, so nothing to run. No Python source changed. No live smoke is needed because the payload is byte-identical by construction and the drivers assert its shape.

## Critical Self-Review

- Highest-risk behavior: the `/health` JSON contract. `core01` and the local clone deploy script read this endpoint, so a moved field or a changed status string breaks deploy tooling silently. Mitigated by composing the return object in the same literal order as before and by the two drivers asserting block presence and absence.
- Assumptions that could be wrong: that `input.natsHealth?.()`, `producerHealth?.()`, and `captureHealth?.()` are pure enough to read once. Before this change `producerHealth` and `captureHealth` were each already read exactly once and `natsHealth` likewise, so `readWorkerLanes` preserves the call count rather than changing it — checked against the `origin/main` body line by line.
- Missing/weak tests: none added, and that is the deliberate choice for a no-behavior-change refactor. The existing suite plus the two drivers already cover the healthy path, the stale-producer path, the stale-capture path, the absent-reader path, and the reader-returns-undefined path. A test written against the new helper names would pin the refactor's shape rather than the contract, which is worth less.
- Security/permission risk: none. No auth, no namespace predicate, no SQL, no user input; the function reads injected readers and one outbound probe to the configured embedding base URL, unchanged.
- Migration/deploy risk: none. No schema, no config keys, no new dependency, no startup ordering change.
- Downstream client/runtime risk: none per `docs/downstream-rollout.md`. No MCP tool, schema, transport, or protocol surface changed; `/health` is an HTTP endpoint whose response is unchanged field for field.
- Rollback/cleanup concern: single-file, single-commit, revert is clean.
- Fixes made before PR: re-read the rewritten body against `origin/main` line by line before committing, specifically the predicate change, because the suite did not catch a non-equivalent predicate rewrite in #806. Re-ran oxlint and the done-means on the committed tree after prettier reformatted the staged file.
- Known residual risk: low. The residual is the predicate rewrite described in the Summary; it is the one place where the new text is not a mechanical move of the old text.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no new failure pattern surfaced — this is a mechanical complexity split with no finding to generalize, and the "re-read an extracted predicate against the original" lesson is already recorded from #806.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: the `/health` payload is unchanged field for field and the two done-means drivers assert that shape against the real function.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface changed — the refactor is internal to one TypeScript module and no fixture describes its private helpers.

## Downstream Rollout

Notes/evidence:

- Every downstream item is not applicable. `docs/downstream-rollout.md` scopes the gate to MCP tool, schema, protocol, and client-facing changes. This PR changes no exported signature, adds no tool, and leaves the `/health` response identical, so no downstream consumer can observe it.

### DIFF
--- a/server/transport/health.ts
+++ b/server/transport/health.ts
@@ -162,7 +162,9 @@ const HTTP_NATS_HEALTH: TransportNatsHealth = {
   last_error: null,
 };
 
-async function probeEmbedding(input: SingleWorkerHealthInput): Promise<boolean> {
+async function probeEmbedding(
+  input: SingleWorkerHealthInput,
+): Promise<boolean> {
   if (!input.embeddingBaseUrl) return false;
   const headers = new Headers();
   if (input.embeddingApiKey) {
@@ -176,7 +178,10 @@ async function probeEmbedding(input: SingleWorkerHealthInput): Promise<boolean> 
     const connected = response.ok;
     await response.body?.cancel();
     if (!connected) {
-      input.logger.warn({ status: response.status }, "embedding_health_degraded");
+      input.logger.warn(
+        { status: response.status },
+        "embedding_health_degraded",
+      );
     }
     return connected;
   } catch (error: unknown) {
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #828 A complexity finding on a health/status function is usually four independent readings wearing one function's clothes, not one tangled decision — the seam is "which subsystem does this line speak for", and splitting on that seam drops the count without moving a byte of the payload. Two habits made it safe: read every optional reader exactly once into a small readings object, so the extraction cannot change the invocation count of an injected function (the silent way a refactor like this breaks a counter-based health probe), and treat any predicate whose text is not a mechanical move as the one thing to re-read against the original — here `x?.stale === true` became `x?.stale !== true` inside a negated conjunction, which is equivalent across `undefined`/`false`/`true` but is exactly the shape that slipped through in #806. The pinning evidence to look for is not the unit suite: `scripts/done-means/625` and `647` call the real function and assert the status verdict plus the presence AND absence of each optional block, which is the actual `/health` contract that core01 and the deploy script read.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=c16a2d3e4ef57ad1b47485708da8de4ccee23f21 at=2026-08-26T20:20:45.177Z

```
  - server/transport/health.ts
```

</details>

PR #832, opened 2026-08-26
refactor(repo-facts): bring server/tools/repo-facts.ts to the lint standard (#780)

rodaddy:
## Summary

- `registerRepoFactTools` in `server/tools/repo-facts.ts` was 191 lines against a rule value of 100, because both tool handlers were written inline inside the registrar. This splits registration from the work, following the shape `main` already landed in `search-all.ts`, `list-recent.ts`, `get-entry.ts`, and `tiering.ts`.
- Hoisted `UPSERT_REPO_FACT_SCHEMA` and `LIST_REPO_FACTS_SCHEMA` verbatim, and extracted `handleUpsertRepoFact`, `handleListRepoFacts`, `refuseUnstorableFact`, `embedRepoFact`, and `repoFactFilters` as named module-level functions.
- `authIdentity(extra.authInfo)` is resolved at the registration site and passed in, so neither handler reaches for transport state.
- No behavior change: tool names, descriptions, schemas, annotations, defaults, error strings, SQL text, log event names, and result shapes are byte-identical. `readWithLegacyFallback` is untouched.
- One file changed. No test file was modified.

Reuse was checked and declined with reasons. `authorize` in `server/tools/memory-helpers.ts` runs the same auth-then-namespace sequence but emits different denial strings (`cannot read namespace 'x'` versus this tool's `namespace read access denied`); `embeddingFields` throws rather than swallowing, and returns a different field set. Adopting either would have moved observable behavior. `server/tools/context-pack-repo-facts.ts` (#817) is the read-side pack section and shares no shape with this write path; it is not touched.

## Verification

Red first, on the untouched file at `origin/main` 83e7d01:

```
server/tools/repo-facts.ts:46:8: error eslint(max-lines-per-function): The function `registerRepoFactTools` has too many lines (191). Maximum allowed is 100.
```

`DONE_MEANS_780_FILES=server/tools/repo-facts.ts bash scripts/done-means/780-touched-files-lint-clean.sh` exited 1 on that tree.

- `./node_modules/.bin/oxlint --deny-warnings server/tools/repo-facts.ts` -> exit 0
- `bun run test:isolated src/tools/__tests__/repo-facts.test.ts src/contract.test.ts server/tools/` -> 193 pass, 0 fail, 794 expect() calls, 19 files
- `bash scripts/done-means/780-touched-files-lint-clean.sh` (diff-derived) -> exit 0

Python package checks are not applicable: nothing under `python/openbrain-memory/` is touched. Live smoke is not applicable: this is a pure in-process refactor with no wire-visible change.

## Critical Self-Review

- Highest-risk behavior: the `list_repo_facts` SQL parameter numbering. The original built the `values` array inside the `queryRows` closure; `repoFactFilters` now returns a fresh array per call and the caller pushes the row count and offset onto that same array afterward, so the trailing `$n-1` / `$n` indices are computed against the identical array in the identical order. A fresh array per call also means no accumulation across the two-to-three calls the legacy fallback makes.
- Assumptions that could be wrong: that no caller outside this module reaches the former inline handlers. Checked — `registerRepoFactTools` is the only export, and no test imports it directly; the tool surface is exercised through `src/contract.test.ts`.
- Missing/weak tests: no test was added, deliberately. This is a refactor with no behavior delta, so a new test would assert what the existing suite already covers. The extraction was read back line by line against the original rather than trusted to the suite, because #806 showed the suite did not catch a non-equivalent predicate rewrite.
- Security/permission risk: none introduced. Both permission gates keep their original order and strings — role check before namespace check, so an unprivileged caller still cannot learn whether a namespace exists. `repoFactFilters` interpolates only column names drawn from a closed literal list in this module; every caller-supplied value stays a bound parameter.
- Downstream client/runtime risk: none. Per `docs/downstream-rollout.md` this is not an MCP tool, schema, protocol, or client-facing change — the registered tool names, input schemas, annotations, and result shapes are unchanged, so no downstream client can observe this commit.
- Rollback/cleanup concern: single-file, single-commit; revert is clean.
- Fixes made before PR: the refusal order in `refuseUnstorableFact` was checked to keep raw-code-dump ahead of the credential check, since the two produce different messages and a caller can tell them apart.
- Known residual risk: low. The `embedRepoFact` extraction converts an assignment-in-`catch` into a `return null`, which is equivalent only because the original initialized `embedding` to `null` before the `try`. That was confirmed against the original text.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no new failure pattern surfaced — the reuse-declined-with-reasons check and the read-back-against-original discipline are both already recorded in the lane contract (round 38) and in the #806 entry.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no wire-visible change; the tool contract is byte-identical.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface moved, so no fixture changes — `src/contract.test.ts` passes unmodified against the existing fixtures, which is the evidence that parity holds.

## Downstream Rollout

Notes/evidence:

- All four are not applicable: the registered tool names, descriptions, input schemas, annotations, defaults, error strings, and result shapes are unchanged, so nothing downstream can observe this commit. `docs/downstream-rollout.md` scopes rollout to MCP tool/schema/protocol/client-facing changes; this is an internal function-boundary refactor.

### DIFF
--- a/server/tools/repo-facts.ts
+++ b/server/tools/repo-facts.ts
@@ -43,6 +43,48 @@ import {
 const REPO_FACT_COLUMNS = `id, entity_type, name, canonical_id, namespace,
   metadata, created_by, created_at, updated_at`;
 
+/** Input schema for `upsert_repo_fact`, hoisted so the registrar stays small. */
+const UPSERT_REPO_FACT_SCHEMA = {
+  namespace: z
+    .string()
+    .trim()
+    .min(1)
+    .max(500)
+    .optional()
+    .describe("Namespace for isolation (defaults to agent's clientId)."),
+  metadata: repoFactMetadata.describe(
+    "Curated qmd-derived repository fact metadata.",
+  ),
+};
+
+/** Input schema for `list_repo_facts`, hoisted so the registrar stays small. */
+const LIST_REPO_FACTS_SCHEMA = {
+  namespace: z.string().trim().min(1).max(500).optional(),
+  repo: z.string().trim().min(1).max(300).optional(),
+  collection: z.string().trim().min(1).max(300).optional(),
+  path: z.string().trim().min(1).max(1000).optional(),
+  fact_type: z.enum(FACT_TYPES).optional(),
+  subject: z.string().trim().min(1).max(500).optional(),
+  limit: z.number().int().min(1).max(250).optional(),
+  offset: z.number().int().min(0).optional(),
+};
+
+type UpsertRepoFactArgs = {
+  namespace?: string;
+  metadata: z.infer<typeof repoFactMetadata>;
+};
+
+type ListRepoFactsArgs = {
+  namespace?: string;
+  repo?: string;
+  collection?: string;
+  path?: string;
+  fact_type?: (typeof FACT_TYPES)[number];
+  subject?: string;
+  limit?: number;
+  offset?: number;
+};
+
 export function registerRepoFactTools(
   server: McpServer,
   dependencies: MemoryToolDependencies,
@@ -53,18 +95,7 @@ export function registerRepoFactTools(
       description:
         "Upsert a curated qmd-derived repository fact into Open Brain graph entity metadata. " +
         "This stores stable operating knowledge plus source pointers, not raw code chunks.",
-      inputSchema: {
-        namespace: z
-          .string()
-          .trim()
-          .min(1)
-          .max(500)
-          .optional()
-          .describe("Namespace for isolation (defaults to agent's clientId)."),
-        metadata: repoFactMetadata.describe(
-          "Curated qmd-derived repository fact metadata.",
-        ),
-      },
+      inputSchema: UPSERT_REPO_FACT_SCHEMA,
       annotations: {
         title: "Upsert Repo Fact",
         readOnlyHint: false,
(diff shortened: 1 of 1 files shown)


rodaddy:
No new lessons: I looked for a reusable helper before writing anything, and for a new failure pattern worth an SME entry after. On reuse I checked `server/tools/memory-helpers.ts` (`authorize`, `embeddingFields`), `server/tools/read-scope.ts`, `server/tools/tool-failure.ts`, `server/tools/entity-shared.ts`, and `server/tools/context-pack-repo-facts.ts` from #817; `authorize` and `embeddingFields` are the same shape at a glance but differ in observable behavior — different denial strings, and throw-versus-swallow on the embed — so adopting either would have moved behavior in a no-behavior-change refactor, and the lane contract's round-38 rule (identical field for field, defaults and failure shape included) already says exactly that. On the SME side, the two things this lane relied on are both already captured: read the extraction back against the original rather than trusting the suite (the #806 entry), and declare reuse-declined with the reason (round 38). Nothing here is a pattern the next swarm would miss.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=c13df0c3051599ee1cf42217ff6be301dbe0c877 at=2026-08-26T20:25:20.803Z

```
  - server/tools/repo-facts.ts
```

</details>

PR #837, opened 2026-08-26
refactor(context-pack-fit): bring server/tools/context-pack-b*.ts to the lint standard (#780)

rodaddy:
## Summary

- `fitDurableLaneSection` in `server/tools/context-pack-b*.ts` carried a complexity of 14 against the rule value of 10 — the last lint finding in that file. One function held the early fit check, the oldest-first event drop scan, the checkpoint binary search, and a rebuild closure that all three phases called.
- Split into three named module-level helpers, each taking a single options object instead of positional arguments, matching how `search-all.ts`, `entities.ts`, and `promotion.ts` already read on `main`: `rebuildDurableLaneSection` (the former inner closure, now a top-level function with one owner), `dropOldestDurableLaneEvents` (returns the surviving events, or `null` when no prefix drop fits), and `shrinkDurableLaneCheckpoint` (returns the fitted checkpoint value).
- No behavior change and no new helper invented. The design authority for this module is `docs/agent-context-pack-contract.md` ("Response Shape"), cited in the file's own header; this PR changes none of what it specifies. The fit predicate, the oldest-first drop order, the binary-search boundaries, the citation reconciliation calls, `event_count`, `truncated`, and the returned object shape are the same logic; the checkpoint helper returns its input unchanged in exactly the cases where the original left the variable untouched. No exported signature moved, so all seven importers of this module under `server/tools/` are untouched.

## Verification

Red first, on the untouched file: `oxlint --deny-warnings` reported `server/tools/context-pack-budget.ts:110:8: error eslint(complexity): function fitDurableLaneSection has a complexity of 14. Maximum allowed is 10.`, and the done-means script exited 1.

Green, re-run on the COMMITTED tree after the pre-commit prettier hook reformatted the staged file:

- `./node_modules/.bin/oxlint --deny-warnings server/tools/context-pack-b*.ts` -> exit 0, no findings
- `bun run test:isolated server/tools/` -> 163 pass, 0 fail, 534 expect() calls across 17 files
- `bash scripts/done-means/780-touched-files-lint-clean.sh` (diff-derived against `origin/main...HEAD`) -> exit 0

## Critical Self-Review

- Highest-risk behavior: the durable-lane fitting path decides which lane evidence survives when a pack does not fit. A non-equivalent predicate rewrite would silently change what agents see. Each extraction was read back line-by-line against the original before the commit, because #806 showed the suite does not catch that class on its own.
- Assumptions that could be wrong: that returning `contextMd` unchanged from `shrinkDurableLaneCheckpoint` when it is not a non-empty string matches the original. Checked directly — the original's `if (typeof contextMd === "string" && contextMd.length > 0)` guard meant the variable was left as-is in every other case, which is what the helper does.
- Missing/weak tests: no test names `fitDurableLaneSection` directly; `server/tools/context-pack.test.ts` exercises it through the pack assembly path (trimming, citation bijection after a trim, the pointer/durable dedupe). This PR adds no test because it adds no behavior — the pinning coverage is the pre-existing suite, run green.
- Security/permission risk: none. Pure, store-agnostic functions with no auth, namespace, SQL, or network surface; nothing reads a token or a namespace here.
- Migration/deploy risk: none. No schema, no migration, no config, no startup path.
- Downstream client/runtime risk: none. Per `docs/downstream-rollout.md` this is not an MCP tool, schema, protocol, or client-facing change — the module exports the same names with the same signatures, and the three new helpers are module-private.
- Rollback/cleanup concern: single-file, single-commit, revert is clean. No file was moved, added, or deleted.
- Fixes made before PR: verified none of the seven `server/tools/` importers of this module needed a change, so no other existing file is touched.
- Known residual risk: `src/tools/agent-context-pack-budget.ts` holds a legacy twin of this function that still carries the same finding. It is a separate file with separate callers and belongs to its own lane; deliberately left alone rather than widened into scope here.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical lint conformance change with no new failure pattern — nothing surfaced that a future reviewer would need warned about, and the read-back-after-extraction discipline is already the standing lane rule from #806.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: pure functions with no service, database, or endpoint surface; the isolated suite is the whole exercise.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, schema, or fixture is involved — the change is internal to one module and every exported signature is unchanged.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool, schema, protocol, or client-facing surface moved. The module's exported names and signatures are identical to `origin/main`, so no downstream consumer can observe this change.

### DIFF
--- a/server/tools/context-pack-budget.ts
+++ b/server/tools/context-pack-budget.ts
@@ -97,6 +97,104 @@ export function durableLaneContentChars(section: DurableLaneSection): number {
 }
 
 /**
+ * Options for rebuilding a durable-lane section from what survived fitting.
+ */
+type DurableLaneRebuildOptions = {
+  section: Record<string, unknown>;
+  lane: Record<string, unknown> | undefined;
+  keptEvents: DurableLaneEvent[];
+  contextMd: unknown;
+};
+
+/**
+ * Rebuild the section around the events and checkpoint that survived, marking
+ * it truncated and reconciling `event_count` to what remains.
+ */
+function rebuildDurableLaneSection(
+  options: DurableLaneRebuildOptions,
+): Record<string, unknown> {
+  const { section, lane, keptEvents, contextMd } = options;
+  const next: Record<string, unknown> = {
+    ...section,
+    events: keptEvents,
+    event_count: keptEvents.length,
+  };
+  if (lane) next.lane = { ...lane, current_context_md: contextMd };
+  next.truncated = true;
+  return next;
+}
+
+/**
+ * Options for the oldest-first event drop pass.
+ */
+type DurableLaneDropOptions = {
+  section: Record<string, unknown>;
+  lane: Record<string, unknown> | undefined;
+  events: DurableLaneEvent[];
+  contextMd: unknown;
+  remainingChars: number;
+};
+
+/**
+ * Drop the OLDEST event (index 0) until the serialized section fits, so the
+ * freshest lane evidence survives pressure. Returns the surviving events once a
+ * prefix drop fits, or `null` when even an empty event list does not.
+ */
+function dropOldestDurableLaneEvents(
+  options: DurableLaneDropOptions,
+): DurableLaneEvent[] | null {
+  const { section, lane, events, contextMd, remainingChars } = options;
+  const kept = [...events];
+  while (kept.length > 0) {
+    kept.shift();
+    const candidate = rebuildDurableLaneSection({
+      section,
+      lane,
+      keptEvents: kept,
+      contextMd,
+    });
+    if (serializedLength(candidate) <= remainingChars) return kept;
+  }
+  return null;
+}
+
+/**
+ * Options for shrinking the lane checkpoint to the largest prefix that fits.
+ */
+type DurableLaneCheckpointOptions = {
+  section: Record<string, unknown>;
+  lane: Record<string, unknown> | undefined;
+  contextMd: unknown;
+  remainingChars: number;
+};
+
+/**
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #837 A complexity finding on a section-fitting function is usually one function holding several sequential phases plus an inner closure that all of them capture — here an early fit check, an oldest-first drop scan, a checkpoint binary search, and a `rebuild` closure. Hoisting the closure to a module-level function first is what makes the rest fall out: once the phases no longer capture anything, each becomes an independently extractable helper, and the options-object convention already on `main` (`search-all.ts`, `entities.ts`, `promotion.ts`) absorbs the four-to-five values they each need without tripping the parameter rule. The equivalence trap in this shape is the tail assignment: the original only reassigned `contextMd` inside a `typeof === "string" && length > 0` guard, so the extracted helper has to return its input unchanged in every other case rather than returning the computed value unconditionally — a rewrite that always returns the slice looks identical and is not, and the suite does not catch it (#806). Read each extraction back against the original before staging; that pass, not the tests, is what caught this one.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=b52d35c2d400e16f8ab5f93451e3a53ba07d4cdf at=2026-08-26T21:03:35.249Z

```
  - server/tools/context-pack-budget.ts
```

</details>

PR #843, opened 2026-08-26
refactor(auth-middleware): bring server/auth/middleware.ts to the lint standard (#780)

rodaddy:
## Summary

- `server/auth/middleware.ts` carried the whole bearer/delegation chain in one anonymous handler: complexity 14 against a rule value of 10 (`eslint(complexity)` at `middleware.ts:65:10`).
- Split by the check each branch performs, following the pattern `main` already uses across `server/tools/`: named module-level helpers, one concern each — `authenticateBearerHeader`, `checkDelegatedNamespace`, `checkDelegatedAgentId`, `buildRequestAuthInfo`.
- **Auth behavior is identical.** Every predicate, header name, status code, error string, and refusal precedence is unchanged. Proved two ways: (1) read-back of the split file against the pre-change original — all five `status`/`error` pairs and `DELEGATED_ID_RE` are byte-identical, and the namespace 403 → namespace 400 → agent-id 400 precedence is preserved because the helpers are chained with `??` in the original order; (2) the regression tests that pin this boundary were not touched and pass unmodified.
- Extraction only. No helper was reused from elsewhere: `src/auth.ts` holds a near-identical `DELEGATED_ID_RE` and `headerValue`, but that is the legacy implementation this module deliberately adapts rather than shares (see the module header), so coupling the two would be a behavior change rather than a lint fix. No other file was touched.

## Verification

`oxlint --deny-warnings server/auth/middleware.ts` → exit 0 (was: 1 complexity finding). `bunx tsc --noEmit` → exit 0. `bun run test:isolated server/auth/ server/transport/ server/application/sdk-protocol.pg.test.ts` → exits 0, 0 fail (18 tests in `server/auth/` + `sdk-protocol`, 67 in `server/transport/`). `scripts/done-means/780-touched-files-lint-clean.sh` → exit 0 on the committed, prettier-formatted tree. Python is untouched; nothing deploys.

## Critical Self-Review

- Highest-risk behavior: this file is the namespace-isolation boundary — `req.auth` is what the MCP SDK turns into `extra.authInfo` and every downstream namespace predicate. A non-equivalent predicate rewrite here would be an isolation bug, not a style bug, which is why the change is extraction with no rewritten conditions and why it was read back line by line rather than trusted to the suite (#806 is the precedent for the suite missing exactly that).
- Assumptions that could be wrong: that reading the `x-agent-id` header before the namespace checks rather than after is unobservable. It is — `headerValue` is a pure string function that touches neither the request nor the response, so the only ordering that matters is refusal precedence, and that is preserved by the `??` chain.
- Missing/weak tests: no new tests. The pinning tests already cover the paths this touches, and adding tests to a behavior-preserving extraction would test the new shape rather than the contract. `server/auth/auth.test.ts` covers `delegateNamespace` in the sibling policy module, not this handler; the handler's chain is covered end to end by the `sdk-protocol` bearer → middleware → `req.auth` → namespace-predicate test.
- Security/permission risk: none introduced. The role gate still refuses a non-admin delegator with 403 rather than downgrading it, which is the dangerous variant the module header calls out; both delegated values are still shape-validated before reaching `req.auth`.
- Migration/deploy risk: none. No schema, no config, no startup path.
- Downstream client/runtime risk: none. No exported signature changed — `createAuthMiddleware`, `requestAuth`, and `RequestAuthInfo` keep their shapes; the four new helpers are module-private.
- Rollback/cleanup concern: single-file revert, no state.
- Fixes made before PR: none needed beyond the split; the first read-back matched.
- Known residual risk: the duplicated `DELEGATED_ID_RE`/`headerValue` between `src/auth.ts` and this module remains. Deliberate and pre-existing — deduping it is a migration decision for whoever retires the legacy path, not this lane.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a behavior-preserving lint extraction that surfaced no new review pattern; the "read the extraction back against the original, the suite will not catch a non-equivalent predicate" lesson is already the standing lane-contract instruction that this lane followed.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, contract, or client-facing behavior changed.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface is touched — the change is internal to one Express middleware module.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool, schema, protocol, transport, or client-facing shape changed. The classification is "internal refactor", which `docs/downstream-rollout.md` puts outside the rollout gate.

### DIFF
--- a/server/auth/middleware.ts
+++ b/server/auth/middleware.ts
@@ -51,6 +51,84 @@ export function requestAuth(request: Request): RequestAuthInfo | undefined {
   return (request as Request & { auth?: RequestAuthInfo }).auth;
 }
 
+/** A refusal the handler should send instead of continuing the chain. */
+interface AuthRefusal {
+  readonly status: number;
+  readonly error: string;
+}
+
+/**
+ * Resolve the bearer token in the `Authorization` header to a token identity.
+ *
+ * Both failure modes stay distinct on purpose: a missing header and a header
+ * carrying an unknown token are different operator problems, and collapsing
+ * them into one message is what makes a misconfigured client unreadable.
+ */
+function authenticateBearerHeader(
+  header: string | undefined,
+  configuredTokens: readonly AuthTokenConfig[],
+): AuthIdentity | AuthRefusal {
+  if (!header?.startsWith("Bearer ")) {
+    return { status: 401, error: "Missing Bearer token" };
+  }
+  const identity = resolveBearerToken(
+    header.slice("Bearer ".length),
+    configuredTokens,
+  );
+  if (!identity) return { status: 401, error: "Invalid token" };
+  return identity;
+}
+
+function isRefusal(value: AuthIdentity | AuthRefusal): value is AuthRefusal {
+  return "status" in value;
+}
+
+/**
+ * Check that this role may delegate a namespace at all, then that the value
+ * presented has the delegated-id shape.
+ *
+ * A role that may not delegate is refused rather than downgraded: silently
+ * ignoring the header leaves the caller believing it wrote somewhere it did not.
+ */
+function checkDelegatedNamespace(
+  namespace: string | undefined,
+  role: AuthIdentity["role"],
+): AuthRefusal | undefined {
+  if (!namespace) return undefined;
+  if (role !== "admin" && role !== "ob-admin") {
+    return { status: 403, error: "Role not permitted to delegate namespace" };
+  }
+  if (!DELEGATED_ID_RE.test(namespace)) {
+    return { status: 400, error: "Invalid X-Namespace header" };
+  }
+  return undefined;
+}
+
+/** Check the delegated agent id has the delegated-id shape. */
+function checkDelegatedAgentId(
+  agentId: string | undefined,
+): AuthRefusal | undefined {
+  if (agentId && !DELEGATED_ID_RE.test(agentId)) {
+    return { status: 400, error: "Invalid X-Agent-Id header" };
+  }
+  return undefined;
+}
+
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #843 A complexity finding on a security-boundary handler is best cleared by extracting each branch as a helper that RETURNS a refusal descriptor (`{status, error}`) rather than one that writes to the response, because the refusal precedence then becomes a single visible `??` chain in the original order instead of scattered early-returns whose ordering a reviewer has to reconstruct. Two things this lane checked that a suite cannot: the extracted helpers were read back against the pre-change original to confirm all five `status`/`error` pairs and the shape regex are byte-identical (#806's precedent — the suite did not catch a non-equivalent predicate rewrite), and the one genuine ordering move (reading the `x-agent-id` header before the namespace checks instead of after) was justified from the callee's purity rather than from the tests passing. Also worth recording as a non-finding: `src/auth.ts` holds a near-identical `DELEGATED_ID_RE` and `headerValue`, and a reuse-maximizing lane will be tempted to dedupe them — do not, the module header states this file deliberately ADAPTS the legacy implementation rather than sharing it, so the duplication is the design and collapsing it is a behavior change wearing a lint fix's clothes.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=08e1dae459bd5b4f9ab37f9e4225a97ed14728d9 at=2026-08-26T21:12:58.031Z

```
  - server/auth/middleware.ts
```

</details>

Issue #625, opened 2026-08-08
Maintenance sweep goes quiet (~18min observed) while /health stays green


rodaddy:
Found by the #612 lane (PR #624 report) and deliberately not investigated there (scope): during RED-proof runs the local clone stopped emitting sweep lines for ~18 minutes while `/health` remained green. Reproduced on the PRE-fix revision, so unrelated to the #624 logging change.

Why it matters: a sweep that silently stops is the #384 defect class returning in runtime form — and health checks that stay green while a background producer is dead are exactly the silent-failure shape AGENTS.md now bans. The #624 done-means check had to grow a control clause (z) specifically because the clone went quiet under it — that clause refused to bank a free RED, which is how this was noticed at all.

Ask: reproduce (watch sweep cadence vs config), root-cause the stall (timer starvation? claim contention? tick swallowed by an error path?), and decide whether /health should reflect producer liveness (last-sweep-age) so quiet-but-green cannot recur.

Done means (sketch, red-first): a check that asserts sweep-complete lines appear within N× the configured interval over an M-interval window on the live clone, failing loudly on staleness; plus, if /health gains last-sweep-age, an assertion it goes non-green when the producer is stopped.

State: observed twice by one lane (PROPOSED until reproduced under this ticket); control-clause transcript in PR #624.

PR #844, opened 2026-08-26
refactor(application-root): bring server/application/index.ts to the lint standard (#780)

rodaddy:
## Summary

- `createShadowApplication` carried a complexity of 13 against a rule value of 10. Every branch in it was a conditional spread: the composition root reads a dozen optional fields and passes each one present-or-absent rather than present-and-undefined, because the transport and maintenance builders own their own defaults and an explicit `undefined` is a different call.
- Split the wiring into named module-level helpers that each return one wired piece — `composeSessions`, `composeHealthOverrides`, `composeHealth`, `composeMaintenance`, `composeTransportApp`. Every read stays the same read from the same place, every conditional spread stays a conditional spread, and construction order is unchanged.
- The late-bound producer reading is carried across deliberately. `maintenance` is composed AFTER the health object, and `producerHealth` reads it through a closure invoked per request so `/health` publishes the runner's liveness NOW rather than at boot (#625). That closure is now passed into `composeHealth` as an argument and still closes over the `maintenance` binding at the call site, so both the ordering and the liveness are intact.
- The optional health fields moved into one spread instead of three interleaved ones. No key collides with another, so the object is field-for-field identical; `server/transport/health.ts` consumes it by property access and rebuilds the `/health` JSON in its own fixed order (`server/transport/health.ts:335`), so nothing observes key insertion order.
- One file touched. No exported signature changed, so no caller was touched. No behavior change: schemas, defaults, error strings, SQL, log event names, and section ordering are identical, and the existing tests are unmodified.

## Verification

Evidence, all on the committed (prettier-formatted) tree:

- `./node_modules/.bin/oxlint --deny-warnings server/application/index.ts` -> exit 0 (red first at `origin/main`: `server/application/index.ts:138:8: error eslint(complexity): function 'createShadowApplication' has a complexity of 13. Maximum allowed is 10.`)
- `bash scripts/done-means/780-touched-files-lint-clean.sh` -> exit 0 (red first with `DONE_MEANS_780_FILES=server/application/index.ts` -> exit 1)
- `bunx tsc --noEmit` -> clean
- `bun run test:isolated server/application/ server/main.pg.test.ts` -> 35 pass, 0 fail, 3 files

## Critical Self-Review

- Highest-risk behavior: the late-bound `producerHealth` closure. If it had been evaluated eagerly during the extraction, `/health` would report the maintenance producer's liveness at boot forever — a stale-but-confident answer that reads as healthy. It is still a closure, still invoked per request, and still resolves the `maintenance` binding declared after the health object; `server/application/shadow-application.test.ts` and `server/main.pg.test.ts` exercise the composed `/health`.
- Assumptions that could be wrong: that the health options object has no key-order dependency. Checked rather than assumed — `server/transport/health.ts` reads it by property access and constructs the response object itself at line 335, so the order the fields were inserted in is not observable.
- Missing/weak tests: no test asserts that a conditionally-spread field is ABSENT rather than present-and-undefined. That distinction is the reason each spread is written the way it is, and it is currently protected by the shape of the code and by review, not by an assertion. Not added here because it would be a new test in a lint-only change.
- Security/permission risk: none. No auth, namespace predicate, or token handling is in this file; `authenticate` and `parseRequestBody` are passed straight through unchanged.
- Migration/deploy risk: none. No SQL, no migration, no config key added or renamed.
- Downstream client/runtime risk: none. No MCP tool, schema, transport protocol, or Python client surface changed. `docs/downstream-rollout.md` classifies this as not applicable.
- Rollback/cleanup concern: single-file, single-commit, revertable on its own.
- Fixes made before PR: none needed — oxlint and tsc were clean on the first extraction, and the diff was read back against the original line by line for equivalence before committing.
- Known residual risk: the equivalence argument for moving three interleaved conditional spreads into one block rests on there being no key collision among `revision`, `embeddingBaseUrl`, `embeddingApiKey`, and `fetch`. There is not, and tsc would reject a duplicate, but it is the one place where the diff is not a literal line move.
- SME review-memory update: [x] not applicable because: no new reviewable pattern surfaced — this is the same composition-root extraction shape `docs/sme/quality.md` already covers, applied to one more file.

## Review Gate

- Live Open Brain checks: [x] not applicable because: no runtime, transport, or tool-facing behavior changed; the composed app is proven by the isolated suite above.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or wire shape is touched — this is an internal refactor of one composition function.

## Downstream Rollout

Notes/evidence:

- Not applicable on every line: no MCP tool, schema, protocol, or client-facing surface changed. The only export in this file is `createShadowApplication`, whose signature is unchanged, so every caller (`server/main.ts`, `server/index.ts`, `server/application/nats.ts`, four `scripts/done-means/` drivers) compiles and runs against it untouched.

### DIFF
--- a/server/application/index.ts
+++ b/server/application/index.ts
@@ -3,7 +3,10 @@ import type { Logger } from "pino";
 import { natsHealthFromConfig, type ServerConfig } from "../config.ts";
 import type { Database } from "../db/pool.ts";
 import type { ServerModuleBoundary } from "../module.ts";
-import { createMaintenanceRuntime } from "../maintenance/index.ts";
+import {
+  createMaintenanceRuntime,
+  type MaintenanceRuntime,
+} from "../maintenance/index.ts";
 import {
   readCaptureLiveness,
   type CaptureObservation,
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #844 A composition root's complexity is almost entirely conditional spreads, and the extraction hazard is not the spreads themselves — it is the late-bound closure hiding among them. `server/application/index.ts` read `maintenance` inside a `producerHealth` closure declared ABOVE the `const maintenance` it resolves, on purpose, so `/health` reports the runner's liveness per request rather than at boot (#625). Any extraction that pulls the health object into a helper must pass that reading in as a function and keep the closure resolving the caller's binding; evaluating it during the move would compile, typecheck, and pass the suite while silently pinning `/health` to boot-time liveness — the exact stale-but-confident answer the feature was built to replace. The rule that falls out: when splitting a composition root, list every identifier read before its own declaration BEFORE moving anything, and treat each one as a contract to carry rather than a hoisting quirk to tidy up. The second, cheaper lesson: interleaved conditional spreads can be collapsed into one block only after confirming the consumer reads the object by property access and rebuilds its output in its own order — checked here at `server/transport/health.ts:335` rather than assumed.

rodaddy:
verify-lane receipt: check=scripts/done-means/780-touched-files-lint-clean.sh exit=0 sha=641134147d6b335d012a1eb17d2bc8659375a64c at=2026-08-26T21:17:35.212Z

```
  - server/application/index.ts
```

</details>

PR #847, opened 2026-08-26
feat(config): pass the search embedding timeout from validated config into search-engine (L2b-2, #825)

rodaddy:
## Summary

- `server/tools/search-engine.ts` read `OPENBRAIN_SEARCH_EMBEDDING_TIMEOUT_MS` / `SEARCH_EMBEDDING_TIMEOUT_MS` directly, while `server/config/env-groups.ts` `searchGroup` already parsed the same two names into `config.search.embeddingTimeoutMs`. Both existed and only the reader was consulted, so the validated value was a record of an intention rather than the number in force, and the two could drift silently. This is State 4 of #825 (rung L2b-2).
- The value now arrives through dependencies: `SearchDependencies` (`server/tools/search-engine-types.ts`) and `MemoryToolDependencies` (`server/tools/types.ts`) gain an optional `searchEmbeddingTimeoutMs`, following exactly the pattern #779 used for `ftsCorpusConfig` / `recoveryWalPath` / `natsRuntimeBoundary`. The composition root `server/main.ts:184` hands down `config.search.embeddingTimeoutMs` next to `ftsCorpusConfig`.
- Behavior is unchanged, and that was checked before editing rather than assumed. `searchGroup` (`server/config/env-groups.ts:289-298`) applies the identical two names, the identical `OPENBRAIN_`-first precedence, the identical blank-is-absent handling, and the identical 3000 fallback for a non-numeric or sub-1 value. An absent injection still answers `DEFAULT_SEARCH_EMBEDDING_TIMEOUT_MS`, which is what an unset environment produced.
- Scope is this rewiring only. `shared-namespace.ts`, `search-all.ts`, `trace-config.ts` and `langfuse-tracing.ts` are the other three L2b-2 lanes and are untouched here. No lint suppression was added and no default changed.

## Verification

- Done-means: scripts/done-means/750-l2b2-check-well-formed.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [ ] Python package checks passed or are not applicable — not applicable; no Python source changed.
- [ ] Live Open Brain smoke passed or is not applicable — not applicable; this is an internal wiring change with no client-facing surface.

`scripts/done-means/750-l2b2-env-readers-take-config.sh` stays RED overall until all four L2b-2 lanes land — three readers and three arrivals remain, owned by the sibling lanes. What this lane moved, measured on the committed tree:

```
[...]

PR #850, opened 2026-08-26
feat(config): let shared-namespace helpers take validated names and pass them from the composition root (L2b-2 5a, #825)

rodaddy:
## Summary

- State 5a of the L2b-2 sequence (#825): the shared-namespace helpers can now be handed the ONE validated name set, and the composition root hands it down. The `process.env` reads stay in `server/tools/shared-namespace.ts` for this step and go away in the rung that follows, once every caller passes the value.
- `sharedNamespaceConfig()`, `isSharedNamespace()`, `canonicalNamespace()`, and `physicalNamespace()` each gain a trailing optional `names` parameter. Supplied, the validated set is used verbatim; omitted, behavior is byte-for-byte what it was, so no caller outside that file changes here.
- `config.sharedNamespaceNames` now carries the full five-field shape rather than four. `sharedNamespaceGroup` already resolved the canonical name internally in order to default the physical one, so exposing it adds no new parsing rule — the same names in the same order with the same default.
- The group's field names are aligned to the reader's spelling (`canonicalSharedNamespace`, `physicalSharedNamespace`, `legacySharedNamespace`) so the validated value passes straight through with no translation step that could invert a name. Only `server/config.test.ts` read the old `.physical`/`.legacy` spellings; no non-test code did.
- The type is declared in `server/config/env-groups.ts` and re-exported from `server/tools/shared-namespace.ts` under its historical name `SharedNamespaceConfig`, because `server/config` must not import from `server/tools`.

## Verification

- [x] Python package checks passed or are not applicable — no Python touched
- [x] Live Open Brain smoke passed or is not applicable — no runtime behavior change; both paths still answer identically

Evidence:

- RED at `origin/main`: `scripts/done-means/750-l2b2-env-readers-take-config.sh` exit 1, clause 4 reporting `server/main.ts: found 0, expected 1 — "sharedNamespaceNames: config.sharedNamespaceNames"`.
- GREEN here: that clause now counts the arrival — `CLAUSE 4 ... 2/4 values arrive` with only `searchAllEnv:` and `readMcpTracingConfig(config.` outstanding, both owned by later lanes in the sequence. Clause 2 still lists `shared-namespace.ts`, which is expected by design of this sequence: this rung adds the injection point, the following rung removes the environment fallback.
- `scripts/done-means/750-l2b2-check-well-formed.sh` -> exit 0.
- `bunx tsc --noEmit` -> exit 0. `oxlint --deny-warnings` on every touched non-test file -> exit 0.
- `bun run test:isolated server/config.test.ts server/tools/` -> 270 pass, 0 fail across 19 files.

## Critical Self-Review

- Highest-risk behavior: namespace resolution is a security frontier, so a name silently swapped between canonical and physical would mis-bind an isolation predicate. That is exactly why the group's fields were renamed to the reader's spelling rather than mapped positionally at the call site — a positional map is where that inversion hides.
- Assumptions that could be wrong: that the group and the reader already agreed on all five values. Not assumed — asserted, input by input, in the new agreement suite.
- Missing/weak tests: no test yet drives a real tool call end to end with an injected name set, because no caller passes it in this PR by design. The following rung, which threads the value into callers, is where that test belongs.
- Security/permission risk: none added. No predicate, token, or role path is touched, and the default (no `names` argument) is the existing code path unchanged.
- Migration/deploy risk: none. No schema, no migration, no new environment variable, and no changed default, trim rule, or fallback.
- Downstream client/runtime risk: none. No MCP tool schema, transport, or Python client surface changes; the new dependency fields are optional and the exported type name is unchanged for every importer.
- Rollback/cleanup concern: revert is a single commit — the optional parameters are additive and nothing else reads the new field yet.
- Fixes made before PR: resolved the rebase conflict in `SearchDependencies` by keeping BOTH sibling lanes' fields (`searchEmbeddingTimeoutMs` from #847/#848 and `sharedNamespaceNames` here) rather than either side alone, then re-ran typecheck, the done-means, and the suite on the rebased tree.
- Known residual risk: the doubled state persists for one more rung — the validated group and the environment reader both exist, and only the reader is consulted at runtime. The agreement suite is what keeps that honest until the fallback is removed.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane produced no new MEDIUM+ finding — the two hazards it navigated (dependency direction from `server/config` into `server/tools`, and positional field mapping across two spellings of the same names) are already covered by the charter’s dependency-direction rule and `docs/sme/security.md`, and both were designed out rather than discovered late.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime behavior changes in this rung; the injected path is additive and unused by any caller yet.

## Contract Parity

- Contract parity: [x] runtime-specific because: the change is internal to the TypeScript server's composition root and touches no wire contract or fixture.

## Downstream Rollout

Notes/evidence:

- No MCP tool, schema, protocol, or client-facing surface changes, so every downstream step is not applicable. The only exported-name change is internal (`SharedNamespaceGroup` gains fields); `SharedNamespaceConfig` is still exported from the same module for the same importers.

### DIFF
--- a/server/config.test.ts
+++ b/server/config.test.ts
@@ -5,9 +5,19 @@
  */
 import { describe, expect, it, test } from "bun:test";
 import { randomUUID } from "node:crypto";
-import { loadServerConfig, OPTIONAL_SECRET_KEYS, parseServerConfig } from "./config.ts";
+import {
+  loadServerConfig,
+  OPTIONAL_SECRET_KEYS,
+  parseServerConfig,
+} from "./config.ts";
 import { resolveFtsConfig } from "./tools/fts-config.ts";
 import { resolveQmdPath } from "./tools/search-all.ts";
+import {
+  canonicalNamespace,
+  isSharedNamespace,
+  physicalNamespace,
+  sharedNamespaceConfig,
+} from "./tools/shared-namespace.ts";
 import { parseAllowedOrigins } from "./transport/rest.ts";
 import { readMcpTracingConfig } from "./observability/langfuse-tracing.ts";
 
@@ -23,14 +33,27 @@ describe("server configuration boundary", () => {
     const result = parseServerConfig({ LOG_FILE: "logs/test.log" });
     expect(result.ok).toBe(false);
     if (result.ok) throw new Error("expected invalid configuration");
-    expect(result.issues.map((issue) => issue.path)).toEqual(["DB_HOST", "DB_NAME", "DB_USER"]);
+    expect(result.issues.map((issue) => issue.path)).toEqual([
+      "DB_HOST",
+      "DB_NAME",
+      "DB_USER",
+    ]);
   });
 
   it("rejects malformed configured user tokens visibly", () => {
-    const result = parseServerConfig({ ...REQUIRED, AUTH_TOKEN_USER_BILBY: "agent" });
+    const result = parseServerConfig({
+      ...REQUIRED,
+      AUTH_TOKEN_USER_BILBY: "agent",
+    });
     expect(result).toEqual({
       ok: false,
-      issues: [{ path: "AUTH_TOKEN_USER_BILBY", message: "Invalid option: expected one of \"admin\"|\"agent\"|\"discord\"|\"ob-admin\"|\"promoter\"|\"readonly\"" }],
+      issues: [
+        {
+          path: "AUTH_TOKEN_USER_BILBY",
+          message:
+            'Invalid option: expected one of "admin"|"agent"|"discord"|"ob-admin"|"promoter"|"readonly"',
+        },
+      ],
     });
   });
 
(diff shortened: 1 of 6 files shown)


rodaddy:
harvested: #850 When a validated config group and the scattered reader it mirrors are named differently for the same five values, aligning the group's field spellings to the reader's BEFORE any caller passes the value is what stops a later lane from writing a positional field map at the call site — and a positional map across two spellings of canonical/physical/legacy namespace names is precisely where a name inversion hides, silently mis-binding an isolation predicate with no type error to catch it. The second half is the agreement test: while both paths exist by design (this rung adds the injection point, the next removes the environment fallback), the only thing keeping the doubled state honest is a suite that drives each environment name through undefined / blank / whitespace / set and asserts the group and the reader answer identically. Absence of a `process.env` read is not evidence of wiring, and equality of shape is not evidence of equality of value.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2b2-check-well-formed.sh exit=0 sha=9a094e46a8e3395ba712f622cdc26c899eaf8e68 at=2026-08-26T22:30:04.308Z

```
CLAUSE 2 (inner structural clause is green):     PASS — the inner structural clause is green: CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present under /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-850-23293mtao2tov
```

</details>

PR #851, opened 2026-08-26
refactor(read-scope): accept shared-namespace names from callers (L2b-2 5b, #825)

rodaddy:
## Summary

- L2b-2 State 5b-A for #825, owner lane for `server/tools/read-scope.ts`. The four exported read-scope helpers resolved the shared-namespace set from the environment on every call; they now accept the already-validated set from `ServerConfig` and thread it through every shared-namespace call inside the file and between the helpers themselves.
- `readableNamespaces`, `canReadNamespace`, and `namespaceFilterFor` take `names?: SharedNamespaceConfig` as a trailing optional parameter, matching the convention `server/tools/shared-namespace.ts` established in #850.
- `appendReadNamespacePredicate` was already at the four-parameter maximum `.oxlintrc.json` enforces (`max-params: 4`), so its `names` rides inside the existing options object instead of as a fifth positional argument. The TSDoc records why, so the asymmetry does not read as an oversight.
- No caller in any other file is touched: with the argument absent, behavior is byte-identical to the environment-derived path. A later lane threads the callers; a final lane removes the environment fallback.

## Verification

`bash scripts/done-means/750-l2b2-check-well-formed.sh` exits 0 (the well-formedness wrapper). Its inner env-readers check still exits 1 by design and stays that way until the final lane removes the fallback; the wrapper's CLAUSE 4 confirms exit 1 agrees with its own state clauses.

Run on the committed tree, after the pre-commit prettier reformat:

- `bunx tsc --noEmit` -> 0
- `./node_modules/.bin/oxlint --deny-warnings server/tools/read-scope.ts server/tools/read-scope-names.test.ts` -> 0
- `bun run test:isolated server/tools/` -> 172 pass, 0 fail, 560 expect() calls across 19 files
- Red proof: with `server/tools/read-scope.ts` restored to `origin/main` and the new test in place, `bun test server/tools/read-scope-names.test.ts` -> 0 pass, 4 fail. Restored, then 4 pass.

## Critical Self-Review

- Highest-risk behavior: namespace isolation. These four helpers ARE the read predicate, and a wrong value here returns rows rather than erroring. The risk is mitigated by the parameter being optional with no default change: every existing call site omits it and takes the same environment path it took before.
- Assumptions that could be wrong: that no caller outside this file passes positional arguments past the ones it already passed. Checked — the new parameters are strictly appended, so an existing 1-, 2-, or 3-argument call is unaffected, and `bunx tsc --noEmit` over the whole tree is the proof.
- Missing/weak tests: the new test covers the injected path with a non-legacy configuration. The legacy-namespace branches of `canReadNamespace` are exercised only through the environment default, not through an injected legacy name. That gap is inherited, not introduced.
- Security/permission risk: the `all` keyword and the global-role gates are untouched; no branch that decides authorization was reordered. The injected set only supplies names, never role or delegation state.
- Migration/deploy risk: none. No schema, no config surface, no wire contract.
- Downstream client/runtime risk: none. `read-scope.ts` is server-internal and exports no MCP tool, schema, or Python-visible shape.
- Rollback/cleanup concern: single commit, two files, revertable in isolation. Nothing else in the L2b-2 chain depends on this landing first.
- Fixes made before PR: the first test draft typed the fixture identity with an `as AuthIdentity` cast that hid a missing `tokenClientId` field; replaced with a real typed literal so `tsc` checks it.
- Known residual risk: `server/tools/read-scope.ts` still reads the environment whenever `names` is omitted, which is every call today. The file is not done until the caller lane and the fallback-removal lane land.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no MEDIUM+ finding surfaced — this is an additive optional parameter with a proven-red test and no behavior change, and the one lesson worth recording (a four-parameter signature forces the fifth value into the options object) is already stated in the code's own TSDoc where the next reader will hit it.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: server-internal refactor with no runtime behavior change and no client-facing surface.

## Contract Parity

- Contract parity: [x] runtime-specific because: no MCP tool, schema, or contract shape changed; the edit is confined to an internal TypeScript signature.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool, schema, transport, or Python-client surface is touched. The change is an optional parameter on four server-internal functions whose default path is unchanged.

### DIFF
--- a/server/tools/read-scope-names.test.ts
+++ b/server/tools/read-scope-names.test.ts
@@ -1,0 +1,77 @@
+/**
+ * Injected shared-namespace names win over the environment.
+ *
+ * The read-scope helpers historically resolved the shared-namespace set from
+ * `process.env` on every call. They now accept the already-validated set from
+ * `ServerConfig` as a trailing optional argument. What this file proves is the
+ * part that is easy to get silently wrong: the injected value is actually USED,
+ * rather than accepted and then ignored in favour of the environment. A
+ * distinctive canonical name is passed in, and the assertions fail if the
+ * environment default is what the helper consulted.
+ */
+import { describe, expect, test } from "bun:test";
+import type { AuthIdentity } from "../auth/types.ts";
+import type { SharedNamespaceConfig } from "./shared-namespace.ts";
+import {
+  canReadNamespace,
+  namespaceFilterFor,
+  readableNamespaces,
+  appendReadNamespacePredicate,
+} from "./read-scope.ts";
+
+/** A shared-namespace set no environment default could produce by accident. */
+const injected: SharedNamespaceConfig = {
+  canonicalSharedNamespace: "lane3-canonical-kb",
+  physicalSharedNamespace: "lane3-physical-kb",
+  legacySharedNamespace: "",
+  legacyFallbackEnabled: false,
+  fallbackMinResults: 5,
+};
+
+const identity: AuthIdentity = {
+  role: "agent",
+  clientId: "rico",
+  tokenClientId: "rico",
+  namespaceSource: "token",
+};
+
+describe("read-scope helpers honour injected shared-namespace names", () => {
+  test("readableNamespaces returns the injected physical shared namespace", () => {
+    expect(readableNamespaces(identity)).toEqual(["rico", "shared-kb"]);
+    expect(readableNamespaces(identity, {}, injected)).toEqual([
+      "rico",
+      "lane3-physical-kb",
+    ]);
+  });
+
+  test("canReadNamespace authorizes the injected canonical name, not the default", () => {
+    expect(canReadNamespace(identity, "lane3-canonical-kb")).toBe(false);
+    expect(canReadNamespace(identity, "lane3-canonical-kb", injected)).toBe(
+      true,
+    );
+    expect(canReadNamespace(identity, "shared-kb", injected)).toBe(false);
+  });
+
+  test("namespaceFilterFor translates through the injected names", () => {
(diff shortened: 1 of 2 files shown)


rodaddy:
harvested: #851 The convention `shared-namespace.ts` set in #850 — thread the validated config set in as a trailing optional parameter — does not survive contact with a signature that is already at `.oxlintrc.json`'s four-parameter maximum. `appendReadNamespacePredicate` takes `(identity, params, column, options)`, so a fifth positional `names` is a lint error, not a style question, and the pre-commit hook lints staged files whole so it fails at commit time rather than in review. The resolution that keeps both rules is to put `names` inside the options object that function already accepts and to say so in its TSDoc, because the next reader's first instinct on seeing three helpers take `names` positionally and one take it in options is that somebody was sloppy. Lanes threading a config set through a file should check the arity of every target signature before assuming one uniform shape, and should expect a mixed convention to be the correct outcome rather than a defect to be smoothed over.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2b2-check-well-formed.sh exit=0 sha=00e87064cd23a943e8fccf5a4e23f5da6738629a at=2026-08-26T22:42:22.142Z

```
CLAUSE 2 (inner structural clause is green):     PASS — the inner structural clause is green: CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present under /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-851-47448mtaoin7o
```

</details>

PR #852, opened 2026-08-26
refactor(search-engine): pass shared-namespace names from dependencies (L2b-2 5b, #825)

rodaddy:
## Summary

- `sharedNamespaceConfig()` re-reads the environment on every call unless it is handed an already-validated name set. #850 gave every shared-namespace export a trailing optional `names?: SharedNamespaceConfig` and put `sharedNamespaceNames` on `SearchDependencies`; this threads that value through the search path so the engine resolves against the validated set instead of a second environment read.
- `server/tools/search-engine.ts`: `sharedFallbackApplies` gains a trailing optional `names` and resolves with it (it carries no `dependencies` of its own, so the caller supplies the value); `executeSearchWithSharedFallback` reads `dependencies.sharedNamespaceNames` once and passes it to `sharedFallbackApplies` and to both `withCanonicalNamespaces` calls; `searchWithLegacyTopUp` already has `dependencies` and resolves from `dependencies.sharedNamespaceNames` directly.
- `server/tools/search-engine-rows.ts`: `withCanonicalNamespaces` gains a trailing optional `names` and forwards it to both `canonicalNamespace` calls.
- The design authority for both files is `docs/decisions/shared-kb-canonical-namespace.md` (canonical vs physical translation on read), already cited in both file headers. This change adds no new structure and reverses no direction of translation; it only changes where the name set comes from. Every added parameter is optional and trailing, so an absent set keeps the previous environment-derived answer exactly. No signature exceeds four parameters, no default changes, and the env fallback inside `shared-namespace.ts` is untouched — a later lane removes it once every caller threads the value.

## Verification

Evidence, all run in the lane clone on the committed tree (4ef882a):

- `./node_modules/.bin/oxlint --deny-warnings server/tools/search-engine.ts server/tools/search-engine-rows.ts` -> exit 0
- `bun run test:isolated server/tools/` -> 170 pass, 0 fail, 170 tests across 18 files
- `bash scripts/done-means/750-l2b2-check-well-formed.sh` -> exit 0, all four clauses PASS. Its INNER check still exits 1 by design: the env-readers clause stays red until the final lane deletes the fallback, and the well-formed wrapper asserts that the inner check reaches a real verdict and that its positive control still matches.

Red proof for the new test: removing the `names` argument from the two `withCanonicalNamespaces` calls turned it red (1 fail / 4 pass, on `injected names drive the canonical namespace on emitted rows`); restoring it returned 5 pass. The test drives `executeSearchWithSharedFallback` in keyword mode with a faked pool, so it needs no database and no embedder. Python package is untouched, and this is pure internal parameter threading with no live surface, so no hosted smoke applies.

## Critical Self-Review

- Highest-risk behavior: namespace canonicalisation on emitted rows. If a wrong name set reached `withCanonicalNamespaces`, a row could report a namespace it does not physically live in. Mitigated by the value coming from `SearchDependencies`, which the caller populates from validated config, and by the parameter being optional so an unpopulated caller gets the prior behavior exactly.
- Assumptions that could be wrong: that `dependencies.sharedNamespaceNames` is populated from the same validated group the environment path would have produced. That is #850's contract, not this change's; if a caller injects a divergent set, this change makes that divergence visible on the search path rather than causing it.
- Missing/weak tests: the legacy top-up arm is not exercised end to end here, because the fallback is off by default (`legacySharedNamespace` is empty since #167) and driving it needs a real database. `searchWithLegacyTopUp` is covered only by typecheck and by the existing `server/tools/` suite.
- Security/permission risk: none added. Namespace isolation is enforced by the SQL predicate built from `physicalNamespace`, which this change does not touch; `withCanonicalNamespaces` runs after the predicate and only renames what is already authorized for output.
- Migration/deploy risk: none. No schema change, no config key added or renamed, no default changed. Both new parameters are optional and trailing, so a stale caller compiles and behaves identically.
- Downstream client/runtime risk: none. `withCanonicalNamespaces` and `sharedFallbackApplies` are internal to `server/tools/`; no MCP tool schema, transport shape, REST response, or Python client surface changes. Per `docs/downstream-rollout.md` this is not a contract-changing change.
- Rollback/cleanup concern: revertible as a single commit with no state to unwind. The temporary edit used to prove the test red was restored from a scratch copy and the committed tree was re-verified clean (`git status --short` empty, tsc and oxlint both 0).
- Fixes made before PR: the first draft of the test used `"lexical"` as the search mode, which is not in the `SearchMode` union (`"hybrid" | "vector" | "keyword"`); tsc caught it and it became `"keyword"`.
- Known residual risk: `shared-namespace.ts` still falls back to `process.env` when `names` is absent, so this file pair is threaded but the repo-wide environment read is not yet gone. That is the deliberate state of #825 State 5b, and the sibling lanes plus the final removal lane close it.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no MEDIUM+ finding surfaced — this is mechanical parameter threading behind an optional trailing argument, and the existing entries on canonical-vs-physical namespace resolution already cover the failure mode it touches.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: internal parameter threading only, with no live surface or client-facing behavior change.

## Contract Parity

- Contract parity: [x] runtime-specific because: no MCP tool schema, transport shape, or client contract is touched — both changed signatures are internal to `server/tools/`.

## Downstream Rollout

Notes/evidence:

- No MCP tool, schema, protocol, or client-facing surface changes, so every downstream step is not applicable. The two changed signatures are internal to `server/tools/`, both new parameters are optional and trailing, and behavior with the value absent is unchanged.

Refs #825.

### DIFF
--- a/server/tools/search-engine-rows.ts
+++ b/server/tools/search-engine-rows.ts
@@ -13,6 +13,7 @@
  * make the legacy top-up compare a canonical name against a physical one.
  */
 import { canonicalNamespace } from "./shared-namespace.ts";
+import type { SharedNamespaceConfig } from "./shared-namespace.ts";
 import type { SearchRow } from "./search-engine-types.ts";
 
 /** Convert a timestamp-ish value to an ISO string, or `undefined`. */
@@ -47,18 +48,26 @@ export function withSourceRefs(rows: SearchRow[]): SearchRow[] {
   }));
 }
 
-/** Report the canonical shared name on emitted rows and their source refs. */
-export function withCanonicalNamespaces(rows: SearchRow[]): SearchRow[] {
+/**
+ * Report the canonical shared name on emitted rows and their source refs.
+ *
+ * @param names Validated shared-namespace names from `ServerConfig`. Omitting
+ * it leaves the environment-derived resolution unchanged.
+ */
+export function withCanonicalNamespaces(
+  rows: SearchRow[],
+  names?: SharedNamespaceConfig,
+): SearchRow[] {
   return rows.map((row) => ({
     ...row,
     namespace: row.namespace
-      ? canonicalNamespace(row.namespace)
+      ? canonicalNamespace(row.namespace, names)
       : row.namespace,
     source_ref: row.source_ref
       ? {
           ...row.source_ref,
           namespace: row.source_ref.namespace
-            ? canonicalNamespace(row.source_ref.namespace)
+            ? canonicalNamespace(row.source_ref.namespace, names)
             : row.source_ref.namespace,
         }
       : row.source_ref,
--- a/server/tools/search-engine.test.ts
+++ b/server/tools/search-engine.test.ts
@@ -1,7 +1,11 @@
 import { describe, expect, test } from "bun:test";
 import type { Logger } from "pino";
 import type { Pool } from "pg";
-import { executeSearch } from "./search-engine.ts";
+import {
+  executeSearch,
+  executeSearchWithSharedFallback,
+} from "./search-engine.ts";
+import type { SharedNamespaceConfig } from "./shared-namespace.ts";
 import {
   DEFAULT_SEARCH_EMBEDDING_TIMEOUT_MS,
   type SearchDependencies,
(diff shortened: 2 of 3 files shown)


rodaddy:
harvested: #852 A helper that resolves configuration but does not carry the dependency object is the awkward case in a threading rung, and reaching for the containing object is the wrong instinct. `sharedFallbackApplies` takes a request and a boolean, no `SearchDependencies`, so the two options were to widen it to accept the whole dependency object or to give it the same trailing optional `names` every shared-namespace export already has. The second is right: it keeps the helper's input to what it actually reads, matches the shape #850 established across the module so a reader learns one convention rather than two, and stays inside the four-parameter rule without folding an options object into a three-argument predicate. The general form for this repo: when threading validated config to a helper, pass the narrow value the helper consumes, not the container it happened to come from — and put it last and optional so an untouched caller keeps the prior behavior and the rung stays revertible one file at a time. The test lesson is the paired one: the observable that proves threading worked has to be a value the environment could not have produced, so injecting a distinctive canonical name and asserting on the namespace reported on the emitted row is what makes the assertion evidence rather than a tautology reading a value back from where it was written.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2b2-check-well-formed.sh exit=0 sha=4ef882a07b53d2032d3a002efb55c65d80cb213d at=2026-08-26T22:42:11.742Z

```
CLAUSE 2 (inner structural clause is green):     PASS — the inner structural clause is green: CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present under /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-852-45601mtaoiezt
```

</details>

PR #854, opened 2026-08-26
refactor(context-tools): pass shared-namespace names from dependencies in adjacent-context, agent-context-pack, list-recent and context-pack durable memory (L2b-2 5c, #825)

rodaddy:
## Summary

- State 5c-2 of #825. `server/tools/shared-namespace.ts` (#850) and `server/tools/read-scope.ts` (#851) now accept the already-validated shared-namespace name set from `ServerConfig` as a trailing optional argument, and `MemoryToolDependencies` carries it as `sharedNamespaceNames`. Until callers actually pass it, every call still re-derives those names from the environment, so the single parse at the composition root is not yet in force.
- This threads `dependencies.sharedNamespaceNames` into every shared-namespace and read-scope call in four context-oriented tools: `adjacent-context.ts` (`canReadNamespace`, `physicalNamespace`), `agent-context-pack.ts` (`physicalNamespace`, plus `canReadNamespace` inside `authorizePack`, which had no `dependencies` and now takes the name set as its own parameter from its caller), `list-recent.ts` (`readableNamespaces`), and `context-pack-durable-memory.ts` (both `namespaceFilterFor` calls, reached through `request.dependencies`).
- Behavior with the field absent is unchanged: each helper falls back to the environment-derived set exactly as before, which is what the remaining unconverted callers still rely on. Removing that fallback is the final lane of this sequence, not this one.
- The pack scope literal moves into a `packScope` helper. Threading the name set added lines to `buildAgentContextPackPayload`, which sat one line under the repo lint rule for function length; the extraction is the smallest change that keeps it green, and it names a seam the file already had.
- No signature in `shared-namespace.ts` or `read-scope.ts` is touched, no default changes, and no lint rule is disabled.

## Verification

`bunx tsc --noEmit` → exit 0. `./node_modules/.bin/oxlint --deny-warnings` on the four changed non-test files → exit 0, no diagnostics. `bun run test:isolated server/tools/ src/tools/__tests__/agent-context-pack.test.ts` → 186 pass, 0 fail across 21 files. `bun run test:isolated src/tools/__tests__/adjacent-context.test.ts src/tools/__tests__/agent-context-pack.test.ts src/tools/__tests__/list-recent.test.ts src/tools/__tests__/agent-context-pack-durable-memory.test.ts` → 63 pass, 0 fail. `bash scripts/done-means/750-l2b2-check-well-formed.sh` → exit 0, all four clauses PASS; its inner env-readers check still exits 1 by design, because that clause only flips once the final lane removes the environment fallback.

New test: `server/tools/agent-context-pack-names.test.ts`. It injects a canonical shared name no environment default produces and asserts the context-pack authorization gate consults it rather than the environment. Proven to fail first — reverting the `canReadNamespace` threading restored the denial and the run reported 1 fail; restoring it gave 2 pass. Its stub pool throws if any statement runs, so a gate that silently let the request through fails loudly instead of passing quietly. No existing test was modified.

Python package untouched, so its checks do not apply. No migration, no schema change, and no wire-format change, so a live smoke adds nothing over the isolated suite.

## Critical Self-Review

- Highest-risk behavior: namespace isolation. Every call changed here sits on a read-authorization or physical-namespace-mapping path, so passing the wrong name set would either deny a legitimate read or, worse, map a namespace to the wrong physical partition. The mitigation is that the injected value is `undefined` on every current caller, which is the exact environment-derived path that shipped before.
- Assumptions that could be wrong: that `namespaceFilterFor` and `readableNamespaces` take `names` as their last parameter, after an options object, rather than immediately after the identity. I got this wrong on the first pass and `tsc` caught it — the argument positions are now read off the live signatures in `read-scope.ts`, not assumed.
- Missing/weak tests: only one of the seven threaded call sites is covered by the new test. The other six are covered by typecheck plus the existing suites, which prove no regression but do not prove the injected value reaches those specific calls. A per-seam test for each would be more thorough; one distinctive-value test at a representative seam is what this lane's scope asked for.
- Security/permission risk: `canReadNamespace` in `authorizePack` is a permission gate, and it now takes its name set from a parameter instead of the environment. A caller that forgets to pass it silently falls back to environment behavior rather than failing open — the gate still runs either way, so the failure mode is a stale name set, never a skipped check.
- Migration/deploy risk: none. No schema change, no migration, and no config key added or renamed.
- Downstream client/runtime risk: none identified. No MCP tool schema, no transport, no wire payload, and no Python client surface changes; the edits are internal argument threading behind unchanged exported signatures.
- Rollback/cleanup concern: revertible as a single commit. Nothing outside these four files and one new test file depends on the change, and the fallback it preserves is what a revert would restore anyway.
- Fixes made before PR: corrected the `names` argument position for `readableNamespaces` and `namespaceFilterFor` after `tsc` rejected the first attempt; dropped a non-existent `Logger` type import from the test in favor of an indexed type off `MemoryToolDependencies`; extracted `packScope` after the pre-commit lint pass showed `buildAgentContextPackPayload` had crossed the function-length rule.
- Known residual risk: the environment fallback is still live in every helper, so this PR moves the repo toward one parse without yet achieving it. The done-means inner check reflects that honestly by still exiting 1.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane produced no new MEDIUM+ finding — the two defects hit were an argument-position mistake and a missing export, both caught by `tsc` in the normal loop, and the function-length overrun was caught by the pre-commit lint gate; each is an existing gate working as designed rather than a pattern a future review swarm would need warning about.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no schema, transport, or wire-format change — the isolated suite exercises every changed path.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface changes. The edits thread an existing optional argument between internal functions; no exported tool schema, payload, or client-facing signature moves.

## Downstream Rollout

Notes/evidence:

- Per `docs/downstream-rollout.md`, rollout applies to MCP tool, schema, protocol, or client-facing changes. This PR changes none of those: it is internal argument threading behind unchanged exported signatures, with identical behavior when the new optional field is absent. No downstream consumer can observe the difference, so every rollout step is not applicable.

### DIFF
--- a/server/tools/adjacent-context.ts
+++ b/server/tools/adjacent-context.ts
@@ -264,14 +264,25 @@ function authorizeAdjacencyRead(
   }
 
   const requestedNamespace = requestedFromArgs ?? identity.clientId;
-  if (!canReadNamespace(identity, requestedNamespace)) {
+  if (
+    !canReadNamespace(
+      identity,
+      requestedNamespace,
+      dependencies.sharedNamespaceNames,
+    )
+  ) {
     return {
       denied: errorResult(
         `Permission denied: cannot read namespace '${requestedNamespace}'`,
       ),
     };
   }
-  return { namespace: physicalNamespace(requestedNamespace) };
+  return {
+    namespace: physicalNamespace(
+      requestedNamespace,
+      dependencies.sharedNamespaceNames,
+    ),
+  };
 }
 
 /**
--- a/server/tools/agent-context-pack-names.test.ts
+++ b/server/tools/agent-context-pack-names.test.ts
@@ -1,0 +1,102 @@
+/**
+ * Injected shared-namespace names reach the context-pack authorization seam.
+ *
+ * `buildAgentContextPackPayload` gates the requested namespace before any query
+ * runs. That gate used to resolve the shared-namespace set from the environment
+ * on every call; it now takes the already-validated set carried on
+ * `dependencies.sharedNamespaceNames`. The failure this file guards against is
+ * the quiet one: the field is threaded through the call but the helper still
+ * consults the environment, so the injected value is accepted and ignored. A
+ * canonical name no environment default produces is injected, and the
+ * assertions only pass if that value is what the gate actually consulted.
+ */
+import { describe, expect, test } from "bun:test";
+import type { AuthIdentity } from "../auth/types.ts";
+import type { MemoryToolDependencies } from "./types.ts";
+import type { SharedNamespaceConfig } from "./shared-namespace.ts";
+import { buildAgentContextPackPayload } from "./agent-context-pack.ts";
+
+/** A shared-namespace set no environment default could produce by accident. */
+const injected: SharedNamespaceConfig = {
+  canonicalSharedNamespace: "lane5-canonical-kb",
+  physicalSharedNamespace: "lane5-physical-kb",
+  legacySharedNamespace: "",
+  legacyFallbackEnabled: false,
+  fallbackMinResults: 5,
+};
+
+const identity: AuthIdentity = {
+  role: "agent",
+  clientId: "rico",
+  tokenClientId: "rico",
+  namespaceSource: "token",
+};
+
+const noopLogger = {
+  debug: () => {},
+  info: () => {},
+  warn: () => {},
+  error: () => {},
+} as unknown as MemoryToolDependencies["logger"];
+
+/**
+ * Dependencies whose pool refuses to be used.
+ *
+ * The assertions below all resolve on the authorization path, which returns
+ * before any statement runs. A pool that throws turns "the gate silently let
+ * this through" into a loud test failure instead of a passing one.
+ */
+function dependenciesWith(
+  names?: SharedNamespaceConfig,
+): MemoryToolDependencies {
+  return {
+    pool: {
+      query: () => {
+        throw new Error("no query should run on the authorization path");
+      },
+    } as unknown as MemoryToolDependencies["pool"],
+    embedFn: async () => null,
+    logger: noopLogger,
+    sharedNamespaceNames: names,
+  };
+}
+
+const args = {
+  agent: "claude",
+  platform: "claude-code",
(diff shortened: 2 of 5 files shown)


rodaddy:
No new lessons: I looked for a reusable review pattern in the three defects this lane hit and found none worth adding to `docs/sme/`. The first was an argument-position mistake — I passed the name set as the second argument to `readableNamespaces` and `namespaceFilterFor`, which both take it last, after an options object; `tsc` rejected it immediately and the fix was to read the live signatures rather than assume them. The second was importing a `Logger` type that `src/logger.ts` does not export; also caught by `tsc`. The third was `buildAgentContextPackPayload` crossing the function-length lint rule, because threading the new argument added lines to a function already one line under it; the pre-commit gate caught it, twice, including after prettier re-split a call I had manually collapsed. All three are existing gates working exactly as designed, inside the normal red/green loop, with no escape to review or post-merge — which is the opposite of what an SME entry is for. I also checked whether the "threaded but ignored" failure mode deserved an entry, since a field accepted and then silently overridden by the environment is the genuinely dangerous shape here; the existing `read-scope-names.test.ts` from #851 already establishes the distinctive-injected-value test pattern that catches it, and this PR's new test follows that pattern rather than inventing one, so the knowledge is already captured where a reviewer would look.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2b2-check-well-formed.sh exit=0 sha=d1d6001ef9a0d67bfc444c0f78303aa7facb7a89 at=2026-08-26T22:56:21.544Z

```
CLAUSE 2 (inner structural clause is green):     PASS — the inner structural clause is green: CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present under /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-854-62676mtap0mpg
```

</details>

PR #857, opened 2026-08-26
feat(config): remove the process.env fallback from shared-namespace so names arrive only from validated config (L2b-2 final, #825)

rodaddy:
## Summary

- `server/tools/shared-namespace.ts` carried its own environment readers (`envString`, `envBoolean`, `envPositiveInteger`) and re-derived the shared-namespace set from `process.env` on every call. PRs #850-#854 threaded the validated `ServerConfig.sharedNamespaceNames` through every `server/` caller, so that path had no remaining reader — it was a second source of truth for a value that decides which physical partition an isolation predicate binds. This removes it. Final lane of State 5 of #825.
- `isSharedNamespace`, `canonicalNamespace`, `physicalNamespace`, and the `sharedNamespaceConfig` pass-through now take the validated set and THROW when it is absent, naming the function and saying the composition root must supply `sharedNamespaceNames`. The parameter stays optional in the signature so no call site changes; the check is a call-site guard at runtime, never a module-level exit and never a silent default. A default here is the dangerous outcome, not the safe one: it would resolve to `shared-kb` and quietly bind a predicate to the wrong partition instead of failing where the wiring is actually wrong.
- `server/tools/read-scope.ts` delegates to those helpers and needs no guard of its own. Only its parameter docs change, because they described an environment fallback that no longer exists.
- No default value, name precedence, or trim rule moves. `server/config/env-groups.ts` is untouched and still produces exactly the same five fields with the same precedence.
- No legacy-root bridge was needed: `src/index.ts` does not import `server/tools`, and `server/main.ts` is the only place that builds the tool dependencies object. It already passes `sharedNamespaceNames: config.sharedNamespaceNames`.
- Tests: assertions that previously proved an *environment fallback* now prove the *throw*, because that is the behavior they were guarding. `server/config.test.ts` drops the reader-vs-config agreement block (there is no second side left to compare) and gains a throw proof for each helper. Test files that registered the memory tools without the names now pass the default set through one shared fixture (`server/tools/shared-namespace-fixture.ts`) rather than five copies of the same literal.

## Verification

- Done-means: scripts/done-means/750-l2b2-env-readers-take-config.sh

RED at `origin/main` (34304f0) — `bash scripts/done-means/750-l2b2-env-readers-take-config.sh` exited 1, clause 2 naming 3 `process.env` reads at `server/tools/shared-namespace.ts:39,47,54`.

GREEN at this head — same script, exit 0, all four clauses PASS:

```
CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present
CLAUSE 2 (no process.env in any of the 4):              PASS — no process.env read in any of the 4 files — each value now arrives from validated config
CLAUSE 3 (scanner proven to match in server/config.ts):   PASS — server/config.ts still contains 4 process.env line(s) — the pattern, tool, and invocation are proven to match
CLAUSE 4 (values arrive from validated config):         PASS — all 4 values are passed exactly once from validated config
```

Other checks, all re-run on the committed tree after the pre-commit hook's prettier pass:

- `bash scripts/done-means/750-l2b2-check-well-formed.sh` -> 0, all four clauses PASS
- `bunx tsc --noEmit` -> 0
- `./node_modules/.bin/oxlint --deny-warnings server/tools/shared-namespace.ts server/tools/read-scope.ts server/tools/shared-namespace-fixture.ts` -> 0
- `./node_modules/.bin/oxlint --deny-warnings --ignore-pattern '**/*.test.ts' server` -> 0
- `bun run test:isolated server/` -> 517 pass, 0 fail, 41 files
- `bun run test:isolated src/tools/__tests__/` -> 941 pass, 0 fail, 83 files
- `rg -n 'process\.env' server --glob '!*.test.ts'` -> real reads only at `server/config.ts:416`, `server/main.ts:386,436,437,536`, and the pre-existing `server/observability/langfuse-tracing.ts:398` (out of scope for #825 State 5). Every other hit is prose in a comment.

The new throw test was proven able to fail: with the guard temporarily replaced by a returned default set, `bun test server/config.test.ts` went to 84 pass / 2 fail; restoring the guard returned 86 pass / 0 fail with a clean tree.

## Critical Self-Review

- Highest-risk behavior: namespace resolution is a security frontier. If a `server/` caller reaches these helpers without the names at runtime, it now throws where it used to return a value. That is the intended trade — a loud failure at the wiring defect beats a predicate silently bound to `shared-kb` — but it means any unexercised code path that omits the names becomes a runtime error rather than a wrong-but-working read.
- Assumptions that could be wrong: that `server/main.ts` is the only composition root for these tools. Checked with `rg -n 'registerMemoryTools\(|MemoryToolDependencies' src server --glob '!*.test.ts'` — the only non-test construction sites are `server/main.ts:170` and `server/contracts/registered-tools.ts:60`, and the latter passes a cast empty object purely to enumerate tool names without invoking them.
- Missing/weak tests: the guard is proven at the helper boundary and through three tool surfaces (`search_brain`, `search_all`, `brain_answer`) plus `agent_context_pack`. It is not proven for every one of the ~20 other tools that transitively reach `physicalNamespace`; those share the same dependencies object, so the coverage is by construction rather than by direct assertion.
- Security/permission risk: reduced. The removed readers were a second, unvalidated source for the names that decide an isolation predicate, and `SHARED_NAMESPACE_PHYSICAL` set in a stray environment could previously diverge from what the validated config had already parsed. Nothing here widens a read: `read-scope.ts` keeps its role gates, its `all`-keyword rule, and its delegated-identity narrowing unchanged.
- Migration/deploy risk: none at the schema layer, no migration. The deployment risk is a hosted process whose environment sets a shared-namespace variable that never reached `ServerConfig` — but `server/config/env-groups.ts` parses all five of the same variables with the same precedence and defaults, and it is untouched, so a deployment that worked before resolves to the same five values.
- Downstream client/runtime risk: none. No MCP tool schema, no transport shape, no Python client surface, and no response envelope changes. `docs/downstream-rollout.md` classifies this as an internal server refactor.
- Rollback/cleanup concern: a single revert restores the readers; nothing persists state and no artifact outlives the process. The one new file, `server/tools/shared-namespace-fixture.ts`, is only imported by tests.
- Fixes made before PR: `server/application/sdk-protocol.pg.test.ts` had three `registerMemoryTools` sites and the first pass only caught two — the third used a deeper indentation and was found by the remaining live-Postgres failures rather than by reading. Also corrected a missing `await` on a `rejects.toThrow` assertion in `server/tools/agent-context-pack-names.test.ts`, which would have passed vacuously.
- Known residual risk: `src/shared-namespace.ts` still reads the environment and still serves the legacy root's own callers. That is deliberate and out of scope here — it is a separate module with a different field shape (`sharedNamespace`, `allowLegacySharedWrites`), and #825 State 5 scopes only the `server/` copy.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no new failure pattern surfaced — the test breakage was the expected, intended consequence of removing the fallback, and every failing assertion was one that had been written specifically to prove the fallback behavior this PR removes.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no MCP schema, transport, or client-facing surface changes; this is an internal wiring change inside the server process.

## Contract Parity

- Contract parity: [x] runtime-specific because: no tool contract changes. `server/contracts/registered-tools.ts` enumerates the same tools with the same schemas, and `bun run test:isolated server/` includes the contract tests at 0 fail.

## Downstream Rollout

Notes/evidence:

- All four are not applicable: rollout applies to MCP tool, schema, protocol, or client-facing changes, and this PR changes none of them. The tool list, every input schema, and every response envelope are byte-identical; the only behavior difference is confined to a code path that no correctly-wired caller reaches.

### DIFF
--- a/server/application/sdk-protocol.pg.test.ts
+++ b/server/application/sdk-protocol.pg.test.ts
@@ -54,7 +54,14 @@
  * `OPENBRAIN_TEST_DATABASE_URL`, which must point at an isolated test/playground
  * database. Never the dogfood database.
  */
-import { afterAll, afterEach, beforeAll, describe, expect, test } from "bun:test";
+import {
+  afterAll,
+  afterEach,
+  beforeAll,
+  describe,
+  expect,
+  test,
+} from "bun:test";
 import type { AddressInfo } from "node:net";
 import type { Server } from "node:http";
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
@@ -73,17 +80,29 @@ import { RecoveryWalStore } from "../realtime/recovery-wal.ts";
 import { WorkingSetStore } from "../realtime/working-set.ts";
 import { SERVER_REWRITE_STATE } from "../state.ts";
 import { registerMemoryTools } from "../tools/index.ts";
+import { DEFAULT_SHARED_NAMESPACE_NAMES } from "../tools/shared-namespace-fixture.ts";
 import { createWorkerProxyHandler } from "../transport/index.ts";
-import type { AggregateHealth, SingleWorkerHealth } from "../transport/index.ts";
+import type {
+  AggregateHealth,
+  SingleWorkerHealth,
+} from "../transport/index.ts";
 import { silentLogger } from "../transport/testing/silent-logger.ts";
-import { expectRecordedShape, loadServerFixture } from "../../contracts/fixture-shape.ts";
+import {
+  expectRecordedShape,
+  loadServerFixture,
+} from "../../contracts/fixture-shape.ts";
 
 const DB_URL = process.env.OPENBRAIN_TEST_DATABASE_URL;
 const dbDescribe = DB_URL ? describe : describe.skip;
 const pool = DB_URL ? new Pool({ connectionString: DB_URL }) : null;
 
 const TEST_TOKEN = "sdk-protocol-proof-token";
-const CONNECTED: DatabaseHealth = { connected: true, total: 2, idle: 2, waiting: 0 };
+const CONNECTED: DatabaseHealth = {
+  connected: true,
+  total: 2,
+  idle: 2,
+  waiting: 0,
+};
 
 /**
  * One namespace per TEST, not per run.
@@ -141,7 +160,9 @@ function testConfig(): ServerConfig {
     OPEN_BRAIN_SERVER_IP: "127.0.0.1",
   });
   if (!result.ok) {
-    throw new Error(`invalid test configuration: ${JSON.stringify(result.issues)}`);
+    throw new Error(
+      `invalid test configuration: ${JSON.stringify(result.issues)}`,
+    );
   }
   return result.config;
 }
@@ -264,11 +285,15 @@ async function startWorkerApplication(): Promise<{
     authenticate: bearerAuth(),
     parseRequestBody: express.json(),
     serverFactory: () => {
-      const server = new McpServer({ name: "open-brain-rewrite", version: "1.0.0" });
+      const server = new McpServer({
+        name: "open-brain-rewrite",
+        version: "1.0.0",
+      });
       registerMemoryTools(server, {
         pool,
         embedFn: async () => Array(768).fill(0.01) as number[],
         logger: pino({ level: "silent" }),
+        sharedNamespaceNames: DEFAULT_SHARED_NAMESPACE_NAMES,
         embeddingModel: "sdk-protocol-proof",
         ...realtime,
       });
@@ -276,7 +301,9 @@ async function startWorkerApplication(): Promise<{
     },
   });
   const server = await new Promise<Server>((resolve) => {
-    const listener = application.app.listen(0, "127.0.0.1", () => resolve(listener));
+    const listener = application.app.listen(0, "127.0.0.1", () =>
+      resolve(listener),
+    );
   });
   const { port } = server.address() as AddressInfo;
   extraApplications.push(application);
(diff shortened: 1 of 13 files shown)


rodaddy:
harvested: #857 Removing an environment fallback is not finished when the readers are deleted and the typechecker is quiet — the fallback's real footprint is in the tests, and it is invisible to `tsc` when the injected parameter is optional. Here `bunx tsc --noEmit` was 0 immediately after the readers came out, while `bun run test:isolated server/` went from 0 to 54 failures across 9 files. Two distinct kinds hid behind that number, and only the second is interesting. The first is mechanical: test files that registered the memory tools without `sharedNamespaceNames` had been relying on the module reading `process.env` for them, so they had to start passing the set the way `server/main.ts` does. The second is the one worth remembering: several assertions had been written specifically to prove the *fallback* — `expect(readableNamespaces(identity)).toEqual(["rico", "shared-kb"])`, `expect(await textFor(tool, undefined)).toContain(NAMESPACE_DENIED)` — and those are not tests to "fix" by injecting names, because that deletes the coverage they exist for. They are tests whose subject was removed, and the correct edit is to re-point each one at the new guarantee (the throw), keeping the same call shape so the file still proves something about the no-names path. Deciding which of the two kinds you are looking at requires reading the assertion's intent, not its failure message; the two look identical in the output. Two process notes that cost time: take the baseline count BEFORE attributing failures (`git stash -u`, run, `stash pop` showed 535 pass / 0 fail at `origin/main`, which is what made "all 54 are mine" a fact rather than a guess), and when the same edit lands in several files, check every occurrence — `server/application/sdk-protocol.pg.test.ts` had three `registerMemoryTools` sites and a regex matched only the two that shared an indentation level, so the third was found by residual live-Postgres failures instead of by reading.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2b2-env-readers-take-config.sh exit=0 sha=de0fae1dde8944c72f7dcc6d757455813621cb69 at=2026-08-26T23:14:54.141Z

```
CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present under /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-857-33565mtapoh2y
CLAUSE 2 (no process.env in any of the 4):              PASS — no process.env read in any of the 4 files — each value now arrives from validated config
CLAUSE 3 (scanner proven to match in server/config.ts):   PASS — server/config.ts still contains 4 process.env line(s) — the pattern, tool, and invocation are proven to match
CLAUSE 4 (values arrive from validated config):         PASS — all 4 values are passed exactly once from validated config
```

</details>

PR #859, opened 2026-08-26
fix(contracts): pass shared-namespace names to every tool-dependency composition the parity harness builds (L2b-2, #825)

rodaddy:
## Summary

- `main` is currently red. #857 (commit 49a8ae6) removed the `process.env` fallback from `server/tools/shared-namespace.ts` and added a `requireNames` guard that throws when the optional `names` are absent. `server/main.ts` was updated to pass `sharedNamespaceNames: config.sharedNamespaceNames`, but the parity fixture harness composes its own `MemoryToolDependencies` and was not updated, so every `server-rewrite-scaffold` fixture threw and returned `isError: true`. CI run 33022617001 failed with exactly six cases: `server-brain-answer-empty`, `server-context-pack-sections`, `server-namespace-denial`, `server-recall-family`, `server-reflex-pointers-empty`, `server-search-brain-arms`, all at `contracts/server-tool-parity.test.ts:325`. That lane ran the `server/tools` tests but not `contracts/`, which is why it merged green.
- This PR is the fix, and it is made at the composition site, which is the only place that can know the names. `contracts/server-tool-parity.test.ts` now hands the scaffold provider an explicit name set built from the same per-run isolation value the suite already assigns to `SHARED_NAMESPACE_CANONICAL`/`SHARED_NAMESPACE_PHYSICAL` for the `current-src` provider. Both providers observe the identical premise, and the scaffold path now depends on no environment at all.
- The set is derived from the existing `DEFAULT_SHARED_NAMESPACE_NAMES` fixture (`server/tools/shared-namespace-fixture.ts`) that the sibling `server/` suites already use, rather than a second hand-written literal.
- `server/tools/shared-namespace.ts`, `server/tools/read-scope.ts`, `.oxlintrc.json`, and `server/observability/*` are untouched. The guard was not made lenient and no environment read was restored. An `rg` over `server src scripts contracts` for `registerMemoryTools\(|sharedNamespaceNames|MemoryToolDependencies|SearchDependencies` found no other composition site missing the names: `server/main.ts` already passes them, `server/application/sdk-protocol.pg.test.ts` and the `server/tools/*.test.ts` suites already use the fixture, and `server/contracts/registered-tools.ts` casts an empty object purely for schema introspection and never resolves a namespace. No non-test production entrypoint needed a change.

## Verification

Evidence, all run in this lane's own clone:

- RED at `origin/main`: `bun run test:isolated contracts/server-tool-parity.test.ts` -> `69 pass / 6 fail`, exit 1.
- GREEN after the change: `bun run test:isolated contracts/server-tool-parity.test.ts` -> `75 pass / 0 fail`, exit 0. Re-run against the committed (prettier-reformatted) tree: same result.
- Whole suite: `bun run test:isolated` -> `3907 pass / 0 fail`, 3942 tests across 254 files.
- `bunx tsc --noEmit` -> exit 0, on the committed tree.
- `bash scripts/done-means/750-l2b2-check-well-formed.sh` -> exit 0.
- `bash scripts/done-means/750-l2b2-env-readers-take-config.sh` -> exit 0.
- `./node_modules/.bin/oxlint --deny-warnings contracts/server-tool-parity.test.ts` reports one pre-existing `eslint(complexity)` error on the per-fixture `test()` body, which is present unchanged at `origin/main` (verified by stashing: baseline `:294`, this branch `:315`, same rule and same function). No new lint finding is introduced and no disable comment was added.

## Critical Self-Review

- Highest-risk behavior: shared-namespace resolution inside the parity suite. If the explicit set and the env assignments ever disagree, the two providers would silently assert against different shared namespaces and parity would stop meaning anything. Both are derived from the single `SHARED_NAMESPACE_ISOLATION` constant, so they cannot drift.
- Assumptions that could be wrong: that `current-src` still resolves the shared namespace from the environment. It does today — that path is the pre-#857 `src/` implementation, which is exactly why the env assignments have to stay. When `src/` is retired, the env block goes with it and only the explicit set remains.
- Missing/weak tests: none added. The six failing fixtures are the test, they failed red before this change and pass after it, and the whole suite proves nothing else moved. Adding a test that the harness wires its own dependencies correctly would be a test of the test.
- Security/permission risk: none. The change is confined to a test file. Namespace isolation is still enforced server-side by the same predicates; the fixture set only names a per-run namespace no production row can belong to.
- Migration/deploy risk: none. No schema, no migration, no runtime file changed.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python-client surface changed, so `docs/downstream-rollout.md` does not apply.
- Rollback/cleanup concern: a plain revert restores the six failures; it does not leave partial state. The suite's own `afterAll` still deletes the seeded rows before restoring the env, unchanged.
- Fixes made before PR: the first draft added a second literal name set inline. It was replaced with a spread of the existing `DEFAULT_SHARED_NAMESPACE_NAMES` fixture so `legacySharedNamespace`, `legacyFallbackEnabled`, and `fallbackMinResults` cannot drift from what the sibling suites assert against.
- Known residual risk: the pre-existing `oxlint` complexity error on the fixture `test()` body remains. It is unrelated to this fix, predates the branch, and widening scope to refactor it would obscure a red-main repair.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the lesson is a process one about which tests a lane runs, not a code pattern a reviewer lane would grep for, and it is recorded as a `harvested:` comment on this PR instead.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: this is a test-harness wiring fix with no runtime, transport, or tool-surface change.

## Contract Parity

No fixture JSON changed. The harness that drives the recorded fixtures is what was repaired, and all six previously-failing fixtures now pass for both providers against the same recorded expectations.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: the only changed file is `contracts/server-tool-parity.test.ts`. No MCP tool, schema, protocol, or client-facing surface is touched, so no downstream consumer can observe this change.

### DIFF
--- a/contracts/server-tool-parity.test.ts
+++ b/contracts/server-tool-parity.test.ts
@@ -7,6 +7,7 @@ import { createBrainServer } from "../src/server.ts";
 import { registerAllTools } from "../src/tools/index.ts";
 import type { AuthInfo } from "../src/types.ts";
 import { registerMemoryTools } from "../server/tools/index.ts";
+import { DEFAULT_SHARED_NAMESPACE_NAMES } from "../server/tools/shared-namespace-fixture.ts";
 
 interface FixtureStep {
   tool: string;
@@ -58,18 +59,29 @@ for await (const name of fixtureGlob.scan({
   cwd: fixtureDir.pathname,
   onlyFiles: true,
 })) {
-  fixtures.push((await Bun.file(new URL(name, fixtureDir)).json()) as ServerFixture);
+  fixtures.push(
+    (await Bun.file(new URL(name, fixtureDir)).json()) as ServerFixture,
+  );
 }
 fixtures.sort((a, b) => a.id.localeCompare(b.id));
-const parityManifest = (await Bun.file(new URL("./server/parity-manifest.json", import.meta.url)).json()) as {
-  provider_capability_status: Record<ProviderId, Record<string, CapabilityStatus>>;
+const parityManifest = (await Bun.file(
+  new URL("./server/parity-manifest.json", import.meta.url),
+).json()) as {
+  provider_capability_status: Record<
+    ProviderId,
+    Record<string, CapabilityStatus>
+  >;
 };
 const providers: ProviderId[] = ["current-src", "server-rewrite-scaffold"];
 
-function replacePlaceholders(value: unknown, replacements: Record<string, string>): unknown {
+function replacePlaceholders(
+  value: unknown,
+  replacements: Record<string, string>,
+): unknown {
   if (typeof value === "string") {
     return Object.entries(replacements).reduce(
-      (result, [placeholder, replacement]) => result.replaceAll(placeholder, replacement),
+      (result, [placeholder, replacement]) =>
+        result.replaceAll(placeholder, replacement),
       value,
     );
   }
@@ -102,7 +114,9 @@ function valueAtPath(source: unknown, path: string): unknown {
 function expectObserved(actual: unknown, expected: unknown): void {
   if (expected === "<uuid>") {
     expect(actual).toBeString();
-    expect(actual as string).toMatch(/^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i);
+    expect(actual as string).toMatch(
+      /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i,
+    );
     return;
   }
   if (expected === "<iso-date>") {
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #859 A lane that removes a fallback must run the WHOLE suite via `bun run test:isolated`, never just the directory it edited. #857 deleted the `process.env` fallback from `server/tools/shared-namespace.ts` and replaced it with a guard that throws when the names are absent; its lane ran the `server/tools` tests, which all already passed the names in, and merged green while `contracts/server-tool-parity.test.ts` — a different directory that composes its own `MemoryToolDependencies` — threw on all six `server-rewrite-scaffold` fixtures and turned `main` red (CI run 33022617001). Removing a fallback is not a local edit: it converts every silent default into a caller obligation, so the blast radius is exactly the set of call sites you did not think to look at, and the directory you edited is the one place that set is guaranteed not to be. The whole-suite run is the only check that enumerates it, and it costs ~42s here.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l2b2-check-well-formed.sh exit=0 sha=252c9a20e789054f5569012d4dac964eb3a7fad5 at=2026-08-26T23:38:49.396Z

```
CLAUSE 2 (inner structural clause is green):     PASS — the inner structural clause is green: CLAUSE 1 (all 4 target files exist):                    PASS — all 4 target files present under /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-859-86939mtaqj8ox
```

</details>

Issue #860, opened 2026-08-26
L3: one logger constructed in the composition root, threaded with decorators and stack traces


rodaddy:
Filed 2026-08-26 by the L2b-2 head session (native Claude, `l2b2-session5`) while authoring the L3 handover; ladder rung `_plans/server-hardening-ladder.md` "L3 — One logger, threaded".

## Deliverable (ladder text)

A single logger constructed in the composition root, carried through the application, with decorators on functions and classes, and stack traces on failure. Domain modules do not import the logger; they receive it. A thrown error inside a decorated function produces a log line carrying the stack and the correlation id.

## What is true on origin/main 9a76d31 (checked 2026-08-26)

- `server/main.ts:61` imports `createLogger` from `server/logging/logger.ts` and is the ONLY non-test file in `server/` that imports it (`rg -l 'logging/logger' server --glob '!*.test.ts'` -> 1). Tool handlers receive `toolLogger` through `MemoryToolDependencies` (`server/main.ts:173`).
- `server/logging/` = `logger.ts` 153 lines, `context.ts` 27 (AsyncLocalStorage), `crash-handlers.ts` 41, `sanitize.ts` 97, `logging.test.ts` 147.
- No decorator path exists: `rg -in 'decorator' server --glob '!*.test.ts'` matches prose only.
- UNVERIFIED (ladder says check first): whether `logger.ts` conforms to `_DOCS/STANDARDS-observability.md` field by field.
- The hard-failure half is #748 (62 catch sites swallow errors).

## Done means

1. A done-means script `scripts/done-means/750-l3-logger-threaded.sh` proves: exactly one `createLogger` call site in non-test server/ (the composition root), zero direct logger imports outside `server/main.ts` and `server/logging/`, and a decorated function's thrown error yields a log line with `stack` and the correlation id (driver test).
2. `_DOCS/STANDARDS-observability.md` conformance of `server/logging/logger.ts` proven by a test, not prose.
3. Ladder L3 status line updated to MERGED with the receipt.

Not in scope here: #748's catch-site sweep (its own issue), L4 file splits.

rodaddy:
Scribe, session 6 (same head, continued): State 3 conformance recon and State 4 RED-first done-means dispatched in parallel; L5 reconciliation table recon pulled forward (read-only) since it shares no files with L3.

rodaddy:
[...]

PR #863, opened 2026-08-27
test(standards): add the RED-first done-means check for L3, one logger threaded (#860)

rodaddy:
## Summary

- Rung L3 of the server/ hardening ladder wants ONE logger, constructed at the composition root and handed down. This lands the gate for it BEFORE the work, so the implementing lane gets an executable specification instead of a prose description to argue with. It changes no server behavior.
- `scripts/done-means/750-l3-logger-threaded.sh` asserts five clauses: (1) exactly one `createLogger(` call site in non-test `server/` code and it is `server/main.ts` — the definition in `server/logging/logger.ts` is excluded by PATH, not by pattern, so a second *definition* elsewhere still fails; (2) zero references to `logging/logger` in non-test `server/` outside `server/main.ts` and `server/logging/`, because an import is how a second construction site starts and this catches the state one commit earlier than clause 1 does; (3) `server/logging/decorate.ts` exists and exports both `withLogging` (function wrapper) and `logged` (method decorator); (4) `bun test server/logging/decorate.driver.test.ts` exits 0; (5) positive control — the same pattern, tool and invocation still find `createLogger` in `server/logging/logger.ts`.
- Clause 5 exists because clauses 1 and 2 pass for two very different reasons: because the state is correct, or because the scan examined nothing. A renamed directory, a bad glob, or an `rg` missing from PATH each produce a silent clean sweep and an exit 0 that means nothing.
- Clause 4 exists because absence alone is not the bar. "No module imports the logger" is satisfiable by deleting all logging, which is worse than the defect being gated. `server/logging/decorate.driver.test.ts` therefore throws a real error inside a decorated function and asserts the emitted line carries the `stack` and the ambient correlation id from `server/logging/context.ts`, in both the plain-function and the class-method spelling. An error logged as `{}` — what `JSON.stringify` does to an `Error` — is exactly the failure mode that makes a log useless in the case it exists for.
- The driver test is committed RED, against a module that does not exist yet, on purpose. It is the spec States 5-6 of the L3 handover code against, and it has to exist before them to be one. `scripts/done-means/750-l3-logger-threaded.sh` is therefore RED by design until those states land.
- Measured on `origin/main`: clauses 1, 2 and 5 already PASS — one call at `server/main.ts:495`, one import at `server/main.ts:61`, the definition at `server/logging/logger.ts:117`. Clauses 3 and 4 FAIL, so the inner check exits 1. Those three passing is not luck (the rewrite root was built that way) and it is why the check earns its place: they are one careless import away from false and nothing else in the repo notices.
- Because the L3 check is honestly red, this PR receipts its sibling `scripts/done-means/750-l3-check-well-formed.sh`, which asserts a different property that is true today AND after the seam lands: the L3 check is well-formed — executable, printing all five clause lines exactly once each, structural and positive-control clauses green, and an exit code that agrees with the state clauses it printed. A red check that is well-formed is a working gate pointed at unfinished work; a red check that is malformed is a typo nobody has read. Nothing here is scaffolding to delete later. Same shape and rationale as the existing `750-l2b2-check-well-formed.sh`.

## Verification

- Done-means: scripts/done-means/750-l3-check-well-formed.sh

Evidence, all re-run on the COMMITTED tree after the pre-commit hook reformatted the driver test:

- `bash scripts/done-means/750-l3-check-well-formed.sh` -> exit 0, all five clauses PASS
- `bash scripts/done-means/750-l3-logger-threaded.sh` -> exit 1; clauses 1, 2, 5 PASS, clauses 3 and 4 FAIL (RED by design)
- `DONE_MEANS_L3_INNER` pointed at a stand-in that exits 7 -> exit 3 `HARNESS-ERROR`, proving both the override path and the no-verdict branch
- `./node_modules/.bin/oxlint --deny-warnings server/logging/decorate.driver.test.ts` -> 0
- `bunx tsc --noEmit` -> one error, `server/logging/decorate.driver.test.ts(34,37): error TS2307: Cannot find module './decorate.ts'` — the deliberate RED import. Baseline `origin/main` with these changes stashed is clean, so that is the only new diagnostic.
- `bun run test:isolated server/logging/logging.test.ts` -> 5 pass, 0 fail, 18 expect() calls; database `ob_isolated_83876_mtasuw1bw3` created and dropped

Python package is untouched. No migration, no runtime path, no client-facing contract, so no live smoke applies.

## Critical Self-Review

- Highest-risk behavior: `bunx tsc --noEmit` now reports one error on this branch, from the RED driver test's import of a module that does not exist. If CI gates on a clean typecheck this PR fails there, and that is a real cost paid deliberately — a driver test that compiles is a driver test that is not red. The alternatives were worse: stubbing `decorate.ts` would make clause 3 green against an empty seam, and deleting the import would delete the specification.
- Assumptions that could be wrong: (a) that the implementing lane will name the exports `withLogging` and `logged` with the signatures this test pins — the test IS the contract, so a lane that disagrees changes both together and the check keeps working; (b) that `logged` is written as a TC39-style method decorator taking an options object, which is what the third test case exercises; (c) that clause 2's globs (`!server/logging/**`, `!server/main.ts`) stay correct if the logging module is ever relocated — a move would make clause 2 vacuous, which is exactly what clause 5's positive control is there to catch.
- Missing/weak tests: the two done-means scripts are shell and have no unit tests of their own. Their red/green behavior was demonstrated by hand instead — the well-formed check was run against the real sibling (exit 0) and against a deliberately broken stand-in via `DONE_MEANS_L3_INNER` (exit 3). The driver test does not yet assert the shape of the success-path log line beyond "no error level was emitted", because pinning entry/result message names before the seam exists would over-constrain the implementing lane.
- Security/permission risk: none. Two shell scripts that only read the tree with `rg` and shell out to `bun test`, and one test file. No namespace predicate, no auth path, no SQL, no network, no credential handling touched.
- Migration/deploy risk: none. No schema change, no config change, nothing in the runtime path. `server/main.ts` and every module under `server/logging/` are unmodified.
- Downstream client/runtime risk: none per `docs/downstream-rollout.md`. No MCP tool, schema, transport, protocol, or Python client surface is touched, so rtech-mcps, mcp2cli, and the Hermes runtime are all unaffected.
- Rollback/cleanup concern: reverting the commit removes three added files and restores a clean `tsc`. Nothing else references them. The well-formed check needs no edit when States 5-6 land — its five clauses were written to hold at both ends of the rung — so there is no scaffolding left behind to remember to remove.
- Fixes made before PR: the inner check originally had four clauses with no positive control, which meant a bad glob or a relocated `server/logging/` would have reported clauses 1 and 2 green over a scan that examined nothing; the `createLogger`-in-`logger.ts` control was added as clause 5 and the well-formed sibling reads it back. Clause 1 was also changed from a pattern-based exclusion of the definition to a path-based one, so that a second `createLogger` DEFINITION outside `server/logging/logger.ts` still counts as a hit rather than being silently filtered out.
- Known residual risk: the L3 check stays red until States 5-6 land, so anyone running it in the interim sees a failure. The header states that explicitly and names the two clauses that are expected red, and the receipted done-means for this PR is the well-formed sibling, which is green.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no MEDIUM+ review finding surfaced here — this PR adds two done-means checks and one deliberately-red driver test, changes no server behavior, and the one pattern worth remembering (receipt a well-formed-ness sibling when the real check is honestly red) is already captured in `docs/sme/` from the L2b-2 rung and is reused verbatim rather than newly learned.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: nothing in the runtime, transport, tool, or schema path is modified; the change is two shell checks and one test file.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing shape is touched — the two done-means scripts read the tree and the driver test targets an internal logging seam that has no wire surface.

## Downstream Rollout

Notes/evidence:

- Classified not applicable on every downstream line: the change adds `scripts/done-means/750-l3-logger-threaded.sh`, `scripts/done-means/750-l3-check-well-formed.sh`, and `server/logging/decorate.driver.test.ts`. No MCP tool, schema, protocol, transport, or Python client surface is modified, so no rtech-mcps handoff, mcp2cli cache refresh, Hermes runtime change, or live canary applies.

### DIFF
--- a/scripts/done-means/750-l3-check-well-formed.sh
+++ b/scripts/done-means/750-l3-check-well-formed.sh
@@ -1,0 +1,203 @@
+#!/usr/bin/env bash
+# DONE-MEANS check for the DONE-MEANS check of rung L3 (issue #860).
+#
+#   bash scripts/done-means/750-l3-check-well-formed.sh
+#
+# ---------------------------------------------------------------------------
+# Why a check about a check
+# ---------------------------------------------------------------------------
+# `scripts/done-means/750-l3-logger-threaded.sh` is written RED first: it
+# asserts a state that States 5-6 of the L3 handover will create, so at the head
+# that INTRODUCES it the honest answer is exit 1. `scripts/verify-lane.ts`
+# refuses to post a receipt for a done-means that exits non-zero, so the PR that
+# adds the red check has no receipt line it can name — and the usual escapes
+# (weaken the check, name an unrelated one) both destroy the thing being landed.
+#
+# The way out is to receipt a DIFFERENT property, one that is true today and
+# stays true after the seam lands: the L3 check is WELL-FORMED. It is
+# executable, it reports every clause it claims to, its structural and
+# positive-control clauses are green, and its exit code agrees with the clauses
+# it printed. A red check that is well-formed is a working gate pointed at
+# unfinished work; a red check that is malformed is a typo nobody has read.
+# This script tells the two apart.
+#
+# The pattern is not invented here — `750-l2b2-check-well-formed.sh` did the
+# same job for the L2b-2 rung, and this is its sibling with the clause set the
+# L3 check actually prints.
+#
+# Five clauses, and all five must pass:
+#
+# CLAUSE 1 — THE INNER CHECK IS EXECUTABLE AND REACHED A VERDICT.
+#   The file is present with its executable bit set, and exits 0 (the seam has
+#   landed) or 1 (it has not). Anything else — a 3 for the inner script's own
+#   harness error, a 127 for a missing interpreter, a 2 from a shell syntax
+#   error — means no verdict was reached, so nothing it printed can be trusted.
+#   That is a harness error HERE too, not a fail: this script cannot judge a
+#   check that did not run.
+#
+# CLAUSE 2 — IT PRINTS ALL FIVE CLAUSE LINES, EXACTLY ONCE EACH.
+#   `CLAUSE 1` through `CLAUSE 5`, one line apiece. A missing line is a clause
+#   that silently stopped being evaluated; a doubled line means the output is
+#   generated twice and the reader cannot tell which verdict is in force.
+#
+# CLAUSE 3 — THE STRUCTURAL CLAUSES ARE GREEN.
+#   `CLAUSE 1` (one construction site, at the root) and `CLAUSE 2` (no logger
+#   imports elsewhere) both say PASS. Both are true on origin/main today and
+#   both must STAY true through the seam work — a lane that lands `decorate.ts`
+#   by adding a second `createLogger` call has defeated the rung, and this is
+#   where that shows up.
+#
+# CLAUSE 4 — THE POSITIVE CONTROL IS GREEN.
+#   `CLAUSE 5` of the inner check says PASS. That is its own proof that its
+#   scanner still matches a hit where one MUST exist. It is the clause that
+#   separates "clean tree" from "broken scan", and its value does not change
+#   when the seam lands — so requiring it green here costs the implementing lane
+#   nothing and catches a dead scanner today.
+#
+# CLAUSE 5 — THE EXIT CODE AGREES WITH THE STATE CLAUSES.
+#   `CLAUSE 3` (seam exists) and `CLAUSE 4` (driver test passes) are the STATE
+#   clauses: red until the seam lands, green after. The exit code must agree —
+#   exit 0 demands both PASS, exit 1 demands at least one FAIL. A check that
+#   exits 1 while printing all-PASS, or exits 0 while printing a FAIL, is
+#   reporting one thing and returning another, which is the failure mode that
+#   makes a gate worse than no gate.
+#
+# All five clauses hold at BOTH ends of the rung. Today: inner clauses 1, 2 and
+# 5 PASS, 3 and 4 FAIL, inner exit 1, and clause 5 here is satisfied by the
+# FAILs. After the seam lands: every inner clause PASSes, inner exit 0, and
+# clause 5 is satisfied by the PASSes. Nothing here has to be edited out when
+# the work finishes, which is the property that makes it a permanent check
+# rather than scaffolding.
+#
+# NO ARGUMENTS. `DONE_MEANS_L3_INNER` exists ONLY so this script's own
+# red/green can be demonstrated against a deliberately broken stand-in; it
+# defaults to the real sibling and no caller in the repo sets it.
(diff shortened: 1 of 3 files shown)


rodaddy:
harvested: #863 A done-means check whose subject does not exist yet has a second failure mode nobody looks for: not that it is red, but that its clean clauses are clean because the scan examined nothing. Writing the L3 check first with four clauses — one construction site, no stray imports, seam exists, driver test passes — produced two clauses (1 and 2) that were already PASS on `origin/main`, and I could not tell by reading the output whether that meant the rewrite root was genuinely tidy or whether my `--glob '!server/logging/**'` had swallowed the whole directory. The fix that made the check trustworthy was a fifth clause pointed deliberately at a file the lane does not touch and no future lane should ever empty (`createLogger` in `server/logging/logger.ts`), asserted with the same pattern, tool, and invocation as the clauses it vouches for; if that one goes red, every clean result above it is a broken scanner rather than a clean tree. The same review pass caught the sibling defect: clause 1 originally excluded the definition site by PATTERN, which would have silently filtered out a second `createLogger` DEFINITION added anywhere else, so it now excludes by PATH and a stray definition counts as the extra hit it is. Generalizing for the next RED-first lane: any absence-shaped assertion (`rg` finds nothing, no file matches, count is zero) needs a positive control in the same script or it is untestable by inspection, and any exclusion in that assertion should be expressed as narrowly as the thing being excluded — a pattern exclusion hides a class, a path exclusion hides one file.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l3-check-well-formed.sh exit=0 sha=506a23f5ed482e199178193bc2d81b5042bedeb8 at=2026-08-27T01:05:50.044Z

```
CLAUSE 1 (inner check is executable and judged):   PASS — executable, and exited 0 — a real verdict (0 = seam landed, 1 = not yet)
CLAUSE 2 (all 5 clause lines, exactly once each):  PASS — all 5 clause lines present, exactly once each
CLAUSE 3 (inner structural clauses are green):     PASS — one construction site at the root, and no logger imports elsewhere — both green
CLAUSE 4 (inner positive control is green):        PASS — the inner scanner is proven to still match: CLAUSE 5 (scanner proven to match in server/logging/logger.ts): PASS — 1 createLogger hit(s) in server/logging/logger.ts — the scanner is proven to match
CLAUSE 5 (exit code agrees with its clauses):      PASS — exit 0 agrees with 2/2 state clauses passing
```

</details>

PR #872, opened 2026-08-27
test(config): split server/config.test.ts into three files and drop the process.env helper (#868)

rodaddy:
## Summary

- Closes #868. `server/config.test.ts` was 910 lines against the `max-lines` rule of 500, and its `withEnv` helper at `:279-297` mutated `process.env` in violation of `node/no-process-env`. Together those two findings made the pre-commit gate refuse every edit to the file, so any future change to configuration tests was blocked behind this split.
- Split along the existing `describe` boundaries into three files, each under 500 lines: `server/config.test.ts` keeps the boundary, optional-secret, and `EMBEDDING_API_KEY` blocks; `server/config-extended.test.ts` takes the `extendedConfig` helper and the extended env group blocks; `server/config-equivalence.test.ts` takes the start-equivalence reader comparisons and the shared-namespace helper block.
- The fourth block no longer touches `process.env`. `parseServerConfig(environment)` at `server/config.ts:380` is the pure reader that `loadServerConfig` feeds `process.env` into, so a local `parseFrom` helper builds the same environment shape as a plain object, calls `parseServerConfig`, and throws a `server_configuration_invalid` error naming the failing field when the parse fails. The four tests keep their assertions unchanged, including the `toThrow(/server_configuration_invalid.*DB_HOST/)` match.
- Every test and assertion moved verbatim; only imports and the env helper changed. No test file reads or writes `process.env` after this. `server/config.ts` and `.oxlintrc.json` are untouched.

## Verification

- Done-means: scripts/done-means/750-l2b2-lint-refuses-process-env.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

Red at `origin/main`: `oxlint --deny-warnings server/config.test.ts` reported `eslint(max-lines)` (655 code lines, maximum 500) plus six `node(no-process-env)` findings at `:283-297`. Green on the committed tree: the same command over all three files exits 0. `bun run test:isolated` reported 86 pass / 0 fail / 269 expect() calls before the split and 86 pass / 0 fail / 269 expect() calls after it across the three files. `bunx tsc --noEmit` exits 0. The done-means script reports PASS on all three clauses.

## Critical Self-Review

- Highest-risk behavior: silently losing or weakening a test during a 910-line move. Guarded by the expect()-call count, not just the test count — 269 before and 269 after, so an assertion dropped inside a surviving test would show.
[...]

PR #873, opened 2026-08-27
feat(shared-namespace): give the server twin the legacy-namespace helpers and two config fields (L5 owner, #864)

rodaddy:
## Summary

- Issue #864, the **L5 shared-namespace owner** row of the server/ hardening ladder. `src/shared-namespace.ts` and `server/tools/shared-namespace.ts` are twins, and the server copy could only TRANSLATE names between the canonical and physical forms — it could not answer either question a caller needs before writing: is this name the retired `collab` namespace, and may this caller write into it.
- A caller with no way to ask does not ask. A write into the frozen legacy namespace (#167) then succeeds looking entirely ordinary: nothing errors, and the row is afterward indistinguishable from one written before the freeze. That is the state this rung ends.
- Adds `isLegacySharedNamespace(namespace, names?)`, which tests non-empty FIRST so the default unset legacy name (`""`) never matches unnamespaced caller input; and `shouldRejectLegacySharedWrite(auth, targetNamespace, names?)`, which refuses a non-admin write aimed at a configured legacy namespace and has three separate off-switches — the operator escape hatch, an unconfigured legacy name, and a write aimed anywhere else.
- The helpers read no environment of their own, matching the rest of the module, so `SharedNamespaceGroup` gains `sharedNamespace` and `allowLegacySharedWrites`, and `extendedEnvironmentFields` gains `OPENBRAIN_ALLOW_LEGACY_SHARED_WRITES`. **`SHARED_NAMESPACE_PHYSICAL` is a real field of `extendedEnvironmentFields`**, already read by `sharedNamespaceGroup()` for its canonical resolution, so `sharedNamespace` resolves as `parsed.SHARED_NAMESPACE_PHYSICAL ?? resolvedCanonical` — no fallback-only path was needed.
- **No caller is rewired and `src/shared-namespace.ts` is untouched.** This rung puts the answers within reach; a later rung asks them.
- The prepared patch was written against an older main and applied cleanly at f79a405 with `git apply --3way` (all 7 files, no conflicts). `server/config.test.ts` was split by #872, so the group-shape assertion now lives in `server/config-equivalence.test.ts` — it compares key sets against its own `NAMES` constant rather than asserting a literal field count, so the fix was to add the two fields to `NAMES` and drop the now-stale word "five" from the test title.

## Verification

- Done-means: scripts/done-means/750-l5-shared-namespace-owner.sh

`bun run test:isolated` (whole suite, fresh database) -> **3944 pass / 35 skip / 0 fail across 261 files**, exit 0. `bunx tsc --noEmit` -> exit 0. `oxlint --deny-warnings` over all 10 touched TypeScript files -> exit 0, no output. The done-means script -> exit 0, all four clauses PASS. Every check was re-run on the COMMITTED tree after the pre-commit hook's prettier reformatted two files; the working tree is clean.

RED receipt at pristine `origin/main` f79a405: `git show origin/main:server/tools/shared-namespace.ts | rg -cF 'export function isLegacySharedNamespace'` -> 0 hits, exit 1; the same for `shouldRejectLegacySharedWrite` -> 0 hits, exit 1; `git show origin/main:server/config/env-groups.ts | rg -cF 'OPENBRAIN_ALLOW_LEGACY_SHARED_WRITES'` -> 0 hits, exit 1. Clauses 1 and 2 therefore both fail at that head and the script exits 1.

The check carries a positive control (clause 4) asserting the same pattern, tool, and invocation still find `export function isSharedNamespace` in the twin — because clauses 1 and 2 pass either when the state is correct or when the scan examined nothing, and a renamed file or a missing `rg` produces a silent clean sweep. Clause 2 slices the `SharedNamespaceGroup` interface body out by line range before scanning, so a same-named field on some other interface in the file cannot satisfy it by accident. Clause 3 runs the owning tests, because presence checks alone are satisfied by a function that returns a constant.

Python package: not applicable, no files under `python/openbrain-memory/` are touched. Live smoke: not applicable, nothing here changes a running surface — no caller is wired to the new helpers and the two config fields are additive with defaults that preserve current behavior (`allowLegacySharedWrites: false`, `sharedNamespace` resolving to the canonical name absent an override).

## Critical Self-Review

- Highest-risk behavior: `shouldRejectLegacySharedWrite` is a security-shaped predicate — it returns "refuse this write" — and its default answer is `false`. Every one of its three early returns is a "not applicable" case, so a wrong ordering would silently permit rather than silently refuse, which is the dangerous direction for a frozen namespace. Mitigated by driving all three off-switches plus both admin roles in separate assertions rather than one composite test.
- Assumptions that could be wrong: that `admin` and `ob-admin` are the correct and complete set of roles still permitted to write into the frozen namespace. I read the `Role` union at `server/types.ts:2-3` and confirmed those two are the only admin-class members of the six, but the *policy* choice came from the prepared patch, not from a design document I verified this session. If shared-truth writes are also meant to be open to `promoter`, this predicate is wrong and the test pins the wrong answer.
- Missing/weak tests: no test drives a real write path through the predicate, because no caller is wired to it yet — the tests are unit-level over the pure function. The `sharedNamespace` field is asserted input-by-input against the reader, but nothing yet consumes it, so its correctness is only as good as the expression it was given. There is also no test for a legacy namespace configured to the *same* string as the canonical one, which an operator could do during a migration.
- Security/permission risk: the two new config fields widen what an operator can turn on. `OPENBRAIN_ALLOW_LEGACY_SHARED_WRITES` is an escape hatch that disables a refusal, and `permissiveBoolean` parsing means a range of truthy spellings enable it. It defaults off and requires a deliberate environment setting, and the predicate it feeds is not yet called by anything, so today it changes no live behavior.
- Migration/deploy risk: none identified. No migration, no schema change, no new required environment variable. `SharedNamespaceGroup` gains two fields, which is a compile-time break for any external construction of that type — all in-repo constructors were updated (five test fixtures plus the reader) and `bunx tsc --noEmit` exits 0, but a consumer outside this repo constructing the interface literally would need the two fields.
- Downstream client/runtime risk: none. I read `docs/downstream-rollout.md`'s applicability criteria — no MCP tool, schema, protocol, transport, or client-facing contract is touched. The change is internal to the server composition root and one server-side module; nothing in `server/contracts/` was modified and the whole suite (which covers those composers) is green.
- Rollback/cleanup concern: revert is a single clean commit with no state left behind — no migration to unwind, no config that must be removed first. If an operator has set `OPENBRAIN_ALLOW_LEGACY_SHARED_WRITES` and the commit is reverted, the variable becomes unrecognized rather than harmful.
- Fixes made before PR: my first `AuthInfo` test literal used a `namespace` field and an `as AuthInfo` cast; reading `server/types.ts:25-31` showed `AuthInfo` has no `namespace` field at all, so the cast would have hidden the mismatch — replaced with a real literal (`role`, `clientId`) and no cast. A `sed` edit also duplicated the two new fields in `server/config-equivalence.test.ts`; caught by re-reading the block and fixed forward by deleting the duplicate pair rather than by discarding work.
- Known residual risk: the role policy noted above is the real one — I am confident the code does what the patch intended, and less confident that the intended policy is the ratified one, because I verified the role *union* this session but not a design document naming which roles may write to the frozen namespace. Flagging rather than asserting it is settled.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no MEDIUM+ review finding surfaced in this lane — the two defects I hit were my own in-progress errors (a guessed `AuthInfo` field caught by reading the type, and a duplicated `sed` insertion caught by re-reading the block), both fixed before commit, and both are instances of patterns `docs/sme/correctness.md` already carries rather than new ones.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no caller is wired to the new helpers and both config fields default to current behavior, so there is no running surface whose behavior this commit changes.

## Contract Parity

- Contract parity: [x] runtime-specific because: `SharedNamespaceGroup` is a server-internal config group, not part of any cross-runtime contract. The Python client does not model it and no fixture under `server/contracts/` references it. The five test fixtures that DO construct it were all updated with the two new fields.

## Downstream Rollout

Notes/evidence:

- All four are **not applicable**: no MCP tool, schema, protocol, or client-facing shape changes. The touched surface is `server/config/env-groups.ts`, `server/tools/shared-namespace.ts`, and test fixtures. No tool registration, no response shape, no transport behavior.
- Whole-suite run against a fresh isolated database: `bun run test:isolated` -> 3944 pass / 35 skip / 0 fail, 261 files, exit 0. Per the brief's rule, a failure in `server/contracts/*.ts` would have been this lane's to fix, since those compose tool deps themselves and this change touches a config group — none occurred.

### DIFF
--- a/scripts/done-means/750-l5-shared-namespace-owner.sh
+++ b/scripts/done-means/750-l5-shared-namespace-owner.sh
@@ -1,0 +1,169 @@
+#!/usr/bin/env bash
+# DONE-MEANS check for rung L5 of the server/ hardening ladder
+# (issue #864, "shared-namespace owner").
+#
+#   bash scripts/done-means/750-l5-shared-namespace-owner.sh
+#
+# ---------------------------------------------------------------------------
+# The defect this gates
+# ---------------------------------------------------------------------------
+# `src/shared-namespace.ts` and `server/tools/shared-namespace.ts` are twins.
+# The src copy already knows two things the server copy does not: whether a
+# name IS the retired `collab` namespace, and whether a write aimed at it must
+# be refused. The server copy knowing only how to TRANSLATE names is the state
+# this rung ends, because a caller on the server path has no way to ask the
+# frozen-namespace question and therefore does not ask it.
+#
+# That failure is silent in the direction that matters. A write into the frozen
+# legacy namespace succeeds and looks ordinary; nothing errors, and the row is
+# indistinguishable afterward from one written before the freeze. So the gate
+# is the presence of the ANSWERING functions on the server twin, not the
+# presence of a caller — a caller can be added later, but it cannot be added at
+# all while there is nothing to call.
+#
+# ---------------------------------------------------------------------------
+# Four clauses, and all four must pass
+# ---------------------------------------------------------------------------
+# CLAUSE 1 — THE SERVER TWIN ANSWERS BOTH QUESTIONS.
+#   `server/tools/shared-namespace.ts` exports `isLegacySharedNamespace` and
+#   `shouldRejectLegacySharedWrite`. Both, because either alone leaves a caller
+#   assembling the policy itself out of the other one, which is how two call
+#   sites end up disagreeing about what "frozen" means.
+#
+# CLAUSE 2 — THE CONFIG GROUP CARRIES THE TWO FIELDS THAT FEED THEM.
+#   `SharedNamespaceGroup` in `server/config/env-groups.ts` declares
+#   `sharedNamespace` and `allowLegacySharedWrites`, and
+#   `extendedEnvironmentFields` declares `OPENBRAIN_ALLOW_LEGACY_SHARED_WRITES`.
+#   The helpers read no environment of their own by design, so an undeclared
+#   field means the operator escape hatch is unreachable however the helper is
+#   written.
+#
+# CLAUSE 3 — THE BEHAVIOR IS DRIVEN, NOT MERELY DECLARED.
+#   `bun run test:isolated` over the three owning test files exits 0. Presence
+#   checks alone are satisfied by a function that returns a constant, so this
+#   clause runs the tests that pin the actual answers: an unconfigured legacy
+#   name matching nothing, both admin roles passing, the escape hatch opening
+#   the write, and the config reader producing both new fields.
+#
+# CLAUSE 4 — THE SCANNER IS PROVEN TO MATCH (positive control).
+#   Clauses 1 and 2 pass either because the state is correct or because the
+#   scan examined nothing — a renamed file, a bad glob, an `rg` missing from
+#   PATH each produce a silent clean sweep. So the same pattern, tool, and
+#   invocation must still find `export function isSharedNamespace`, which has
+#   been in the server twin since before this rung and which no lane here
+#   should ever remove.
+#
+# NO ARGUMENTS.
+set -uo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
+fail_hard() {
+  printf 'HARNESS-ERROR: %s\n' "$1" >&2
+  exit 3
+}
+
+command -v rg >/dev/null 2>&1 || fail_hard "rg (ripgrep) not on PATH"
+[ -d "$REPO_ROOT/.git" ] || [ -f "$REPO_ROOT/.git" ] || fail_hard "not a git repo at $REPO_ROOT"
+
+TWIN_FILE="server/tools/shared-namespace.ts"
+GROUPS_FILE="server/config/env-groups.ts"
+TEST_FILES=(
+  "server/config-extended.test.ts"
+  "server/config-equivalence.test.ts"
+  "server/tools/shared-namespace-legacy.test.ts"
+)
+
+[ -f "$REPO_ROOT/$TWIN_FILE" ] || fail_hard "$TWIN_FILE is missing"
+[ -f "$REPO_ROOT/$GROUPS_FILE" ] || fail_hard "$GROUPS_FILE is missing"
+
+CLAUSE1=FAIL; CLAUSE1_EVIDENCE=""
+CLAUSE2=FAIL; CLAUSE2_EVIDENCE=""
+CLAUSE3=FAIL; CLAUSE3_EVIDENCE=""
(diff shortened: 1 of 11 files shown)


rodaddy:
harvested: #873 A prepared patch that applies cleanly is not the same as a patch that still fits, and the two failure modes need separate checks. This one applied to all 7 files with `git apply --3way` and zero conflicts against a main that had moved underneath it — yet the assertion it was written to satisfy had been relocated by #872's split of `server/config.test.ts`, somewhere `git apply` has no reason to look. The relocated assertion turned out to compare key sets against its own `NAMES` constant rather than asserting a literal field count, so adding the two fields to `NAMES` was the whole fix; had it asserted `toHaveLength(5)`, a clean apply would have left a red test with no hint in the apply output that anything was stale. The lesson for the next lane taking a prepared patch: after a clean apply, go read the tests that own the shape you changed, because conflict-free is a statement about line context and nothing more. The second, smaller receipt is that both defects I hit in this lane were guesses wearing the grammar of facts — an `AuthInfo` literal with a `namespace` field the interface does not have (masked by an `as AuthInfo` cast I wrote myself, which is the part that would have made it survive review), and a `sed -e` pair that inserted the same two lines twice. Reading `server/types.ts:25-31` and re-reading the edited block caught both in under a minute each; neither would have been caught by the apply, and the cast one would not have been caught by `tsc` either.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l5-shared-namespace-owner.sh exit=0 sha=004e7aaec1f5a6d988c1aa5adaaa25f8a4ee7638 at=2026-08-27T02:39:10.822Z

```
CLAUSE 1 (server twin answers both legacy questions): PASS — server/tools/shared-namespace.ts exports both isLegacySharedNamespace and shouldRejectLegacySharedWrite
CLAUSE 2 (config group carries the two fields):       PASS — SharedNamespaceGroup declares both fields and the env field set declares OPENBRAIN_ALLOW_LEGACY_SHARED_WRITES
CLAUSE 3 (owning tests drive the behavior):           PASS — bun run test:isolated over 3 owning test file(s) exited 0
CLAUSE 4 (scanner proven to match in server/tools/shared-namespace.ts): PASS — 1 isSharedNamespace hit(s) in server/tools/shared-namespace.ts — the scanner is proven to match
```

</details>

PR #875, opened 2026-08-27
refactor(tools): tier-lane and ingest-conversation-facts take shared-namespace names from their deps instead of src (L5, #864)

rodaddy:
## Summary

- Closes part of #864 (L5 importer switch, lane 1 of 4): `server/tools/tier-lane.ts` and `server/tools/ingest-conversation-facts.ts` now take the shared-namespace names from their injected dependencies instead of importing the environment-reading `src/shared-namespace.ts`.
- `src/shared-namespace.ts` derives the shared namespace names from the environment, so every module importing it re-derived them independently, at whatever moment its call ran. Two modules could disagree about which physical namespace a write landed in, and nothing errored when they did — the row simply went somewhere else.
- Both modules now import the server twin `server/tools/shared-namespace.ts` and pass `dependencies.sharedNamespaceNames`, which the composition root (`server/main.ts:191`) already parses once and threads through `MemoryToolDependencies` (`server/tools/types.ts:11-22`). The twin throws when the argument is absent, so a call site that forgets to thread the names fails loudly instead of quietly re-deriving them. No default, no fallback, and no `process.env` read is introduced.
- Three call sites move, all of which already had `dependencies` in scope, so no signature changed: `physicalNamespace` in `authorizeTierLane` (tier-lane) and `resolveWriteTarget` (ingest-conversation-facts), and `canonicalNamespace` in `handleTierLane` (tier-lane).
- `server/tools/ported-tools-scope.test.ts` composes its own dependencies and now supplies `DEFAULT_SHARED_NAMESPACE_NAMES` from the existing `server/tools/shared-namespace-fixture.ts`. Without it the twin throws and two `tier_lane` cases fail — that is the intended loud failure working, not a regression.
- Adds `scripts/done-means/750-l5-shared-namespace-importers.sh`, which is generic and takes no argv: it discovers the non-test `server/tools` files changed against `origin/main`, so the three sibling lanes on this rung can use it unmodified.

## Verification

Receipts, all re-run against the committed tree after the pre-commit hook reformatted files:

- RED, at `origin/main` before any edit: `CHANGED_FILES='server/tools/tier-lane.ts server/tools/ingest-conversation-facts.ts' bash scripts/done-means/750-l5-shared-namespace-importers.sh` exited 1 — clause 1 failed on both files (`src/shared-namespace` import still present at `tier-lane.ts:42` and `ingest-conversation-facts.ts:36`) and clause 2 reported 2 and 1 helper calls respectively, none carrying the names.
- GREEN, after the change, via the script's own git-diff discovery with no override: exit 0 — clause 1 clean on both, clause 2 counted 2/2 and 1/1 calls carrying `sharedNamespaceNames`, clause 3 typecheck 0.
- `bunx tsc --noEmit` exited 0.
- `./node_modules/.bin/oxlint --deny-warnings` exited 0 on all three touched TypeScript files.
- `bun run test:isolated server/tools`: 182 pass / 0 fail across 21 files. The pre-change baseline on the same tree was also 182 pass / 0 fail, measured by stashing the change, so this matches rather than merely being green.
- `bun run test:isolated` (whole suite, no path): 3933 pass / 0 fail across 260 files. Run because `server/contracts/registered-tools.ts:60-64` composes dependencies itself and might have needed the field; it did not.

Python package checks are not applicable — nothing under `python/openbrain-memory/` is touched. Live smoke is not applicable — this is an internal dependency-threading change with no protocol, schema, or client-facing surface.

## Critical Self-Review

- Highest-risk behavior: namespace resolution on the write path. `physicalNamespace` decides which physical namespace a `tier_lane` or `ingest_conversation_facts` write lands in; if the threaded names differed from what the module previously derived from the environment, rows would silently move. They cannot differ here, because the composition root parses the same names the `src/` copy read, and the twin throws rather than defaulting when nothing is passed.
- Assumptions that could be wrong: that `dependencies.sharedNamespaceNames` is always populated in production. It is optional on `MemoryToolDependencies`, so the type does not enforce it; the guarantee is the composition root at `server/main.ts:191` plus the twin's throw. If some other composition path registers these tools without the field, those two tools throw at call time instead of silently using stale names — loud, but still a runtime failure rather than a compile-time one. The whole-suite run is what checks for such a path, and found only the one test fixture.
- Missing/weak tests: neither module has a direct unit test file, so the coverage here is indirect — `ported-tools-scope.test.ts` exercises `tier_lane` end to end through the MCP client and is what caught the missing fixture field. No new test asserts that the *specific* names threaded from the root are the ones used; the done-means clause 2 covers that structurally by counting call arrivals rather than by behavior.
- Security/permission risk: low, and the namespace predicate is the reason to look. Namespace isolation is a security boundary in this repo, so a change to namespace resolution deserves scrutiny — but the authorization order is untouched. `canWrite` and `canTargetNamespace` still run before `physicalNamespace` in both modules; only the argument list of the resolution call changed, not what is checked or when.
- Migration/deploy risk: none. No schema change, no migration, no config key added or removed. The names come from a parse that already happens at startup.
- Downstream client/runtime risk: none identified. No MCP tool schema, transport, contract fixture, or Python client surface changes — the tool signatures and their responses are byte-identical. `docs/downstream-rollout.md` classification is therefore not-applicable, recorded below.
- Rollback/cleanup concern: revertible as a single commit with no state to unwind. The done-means script is additive and would simply have no changed files to examine after a revert, in which case it exits 1 with a message rather than passing silently.
- Fixes made before PR: added `DEFAULT_SHARED_NAMESPACE_NAMES` to the `ported-tools-scope.test.ts` dependency fixture after the two `tier_lane` failures surfaced, and confirmed against a stashed baseline that those two failures were introduced by this change rather than pre-existing.
- Known residual risk: `sharedNamespaceNames` remains optional on `MemoryToolDependencies` while the other three lanes on this rung land, so during that window a new composition site can omit it and reach a runtime throw rather than a type error. Making it required is the rung's closing move, not this lane's.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane found no new review pattern — the failure it hit (a test fixture needing a newly-threaded dependency field) is the expected, designed consequence of a throwing twin, already documented as the mechanism in `server/tools/shared-namespace.ts`, and the existing entry on announcing self-made adjustments already covers the no-silent-default rule this change follows.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no protocol, schema, or client-facing surface changes; this threads an already-parsed value from the composition root into two existing call sites.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract surface moved. Tool names, input schemas, and response shapes are unchanged; the only edit is which module the namespace helpers are imported from and what they are passed.

## Downstream Rollout

Notes/evidence:

- Every downstream item is not-applicable under the gate's own classification: the change is internal to `server/tools`, alters no MCP tool schema, no transport behavior, and no Python client surface, so there is nothing for rtech-mcps, mcp2cli, or rtech-hermes to pick up. The classification was made by reading the doc, not assumed.
- Scope held to the brief: `src/`, `server/main.ts`, and every other `server/tools` module are untouched. The only file changed beyond the two named modules is the test fixture that composes their dependencies.

### DIFF
--- a/scripts/done-means/750-l5-shared-namespace-importers.sh
+++ b/scripts/done-means/750-l5-shared-namespace-importers.sh
@@ -1,0 +1,182 @@
+#!/usr/bin/env bash
+# DONE-MEANS check for rung L5 of the server/ hardening ladder
+# (`_plans/server-hardening-ladder.md`, issue #864).
+#
+#   bash scripts/done-means/750-l5-shared-namespace-importers.sh
+#
+# ---------------------------------------------------------------------------
+# The defect this gates
+# ---------------------------------------------------------------------------
+# `src/shared-namespace.ts` reads the environment to decide what the shared
+# namespace is called. Every server/tools module that imports from it therefore
+# re-derives those names on its own, at whatever moment its call happens to
+# run, from whatever the environment says then. Two modules can disagree about
+# which physical namespace a write lands in, and nothing errors when they do —
+# the row simply goes somewhere else.
+#
+# The server twin `server/tools/shared-namespace.ts` closes that by taking the
+# names as an argument. It throws when the argument is absent, so the migration
+# cannot half-land silently: a call site that forgets to thread the names fails
+# loudly at runtime instead of quietly re-deriving them.
+#
+# L5 is complete for a module when both halves hold — it no longer imports the
+# environment-reading `src/` copy, AND every call it makes to the twin actually
+# passes the names it was handed by the composition root.
+#
+# ---------------------------------------------------------------------------
+# GENERIC BY DESIGN — NO ARGUMENTS
+# ---------------------------------------------------------------------------
+# This check takes no argv. It discovers its own subject: the non-test files
+# under server/tools changed against origin/main. That is what makes it usable
+# by the sibling lanes on this rung without editing it — each lane changes a
+# different pair of modules, and each gets its own subject list for free.
+#
+# `CHANGED_FILES` (space-separated) overrides the discovery, which is how the
+# RED receipt is taken before any edit exists to be discovered.
+#
+# ---------------------------------------------------------------------------
+# Three clauses, and all three must pass
+# ---------------------------------------------------------------------------
+# CLAUSE 1 — NO MODULE IN THE SUBJECT IMPORTS THE src/ COPY.
+#   Zero references to `src/shared-namespace` in each subject file. That import
+#   is the environment read; while it is present the module can still reach the
+#   old names regardless of what it was handed.
+#
+# CLAUSE 2 — EVERY CALL PASSES THE NAMES (arrival, not departure).
+#   Clause 1 alone is satisfiable by deleting the calls, which is worse than
+#   the defect. So clause 2 counts ARRIVALS: for each of the six helpers, the
+#   number of call sites in the file must equal the number of call sites whose
+#   argument list mentions `sharedNamespaceNames` within the following 3 lines
+#   (prettier wraps a two-argument call across lines), and that number must be
+#   greater than zero. Equal-and-zero is a FAIL, not a pass — a subject file on
+#   this rung has calls to thread, and zero means the scan found nothing rather
+#   than that the work is done.
+#
+# CLAUSE 3 — THE TREE TYPECHECKS.
+#   `bunx tsc --noEmit` exits 0. The twin's trailing argument is typed, so a
+#   call threading the wrong shape is caught here rather than at runtime.
+#
+# An EMPTY subject list is exit 1 with a message, never a silent pass: a check
+# that examines nothing and reports success is the failure mode this whole
+# family of scripts exists to avoid.
+#
+# NO ARGUMENTS.
+set -uo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
+fail_hard() {
+  printf 'HARNESS-ERROR: %s\n' "$1" >&2
+  exit 3
+}
+
+command -v rg >/dev/null 2>&1 || fail_hard "rg (ripgrep) not on PATH"
+command -v git >/dev/null 2>&1 || fail_hard "git not on PATH"
+[ -d "$REPO_ROOT/.git" ] || [ -f "$REPO_ROOT/.git" ] || fail_hard "not a git repo at $REPO_ROOT"
+
+HELPERS='canonicalNamespace|physicalNamespace|sharedNamespaceConfig|isSharedNamespace|isLegacySharedNamespace|shouldRejectLegacySharedWrite'
+NAMES_ARG='sharedNamespaceNames'
+
+# ---------------------------------------------------------------------------
+# SUBJECT — the non-test server/tools files changed against origin/main.
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


rodaddy:
No new lessons: I looked for a reusable review pattern in three places and found none worth an SME entry. (1) The only failure this lane hit was two `tier_lane` cases in `server/tools/ported-tools-scope.test.ts` breaking once the twin's throwing argument was threaded — that is the designed behavior of `server/tools/shared-namespace.ts` doing exactly its job, and the fixture `DEFAULT_SHARED_NAMESPACE_NAMES` already exists precisely for it, so the pattern is documented at the mechanism rather than missing. (2) I checked whether "a fixture needs the new dependency field" deserved a gotcha entry and concluded it does not generalize: it is the loud-failure design working, and an entry would tell future swarms to expect something the type system and the throw already announce. (3) I checked whether the no-silent-default rule this change follows (no fallback, no `process.env`, names come only from the composition root) was already covered — it is, by the existing "nothing is adjusted silently" entry. The one thing a future lane on this rung should know is already in the PR body's residual-risk field rather than the KB: `sharedNamespaceNames` stays optional on `MemoryToolDependencies` until all four L5 lanes land, so during that window an omission is a runtime throw rather than a type error, and making the field required is the rung's closing move.

rodaddy:
Follow-up pushed (`63ffe4d`, branch rebased onto `origin/main` at `7e2adee`). What the done-means got wrong: it discovered its subject with `git diff --name-only origin/main -- server/tools`, comparing against the moving tip of main — so once #873 landed after this branch was cut, `shared-namespace.ts` (the twin, which DEFINES the helpers) and `shared-namespace-fixture.ts` entered the subject and failed clause 2 on files this lane never touched. The fix discovers against `git merge-base origin/main HEAD` — the branch's own diff, which does not drift as main moves — and unconditionally filters the twin and any `*-fixture.ts` out of the subject, so a `CHANGED_FILES` override cannot reintroduce them. RED: `CHANGED_FILES='server/tools/shared-namespace.ts'` -> exit 1 "SUBJECT: none". GREEN: no argv -> exit 0 over exactly `tier-lane.ts` and `ingest-conversation-facts.ts`; `bunx tsc --noEmit` -> 0.

harvested: #875 a done-means that discovers files by diffing against origin/main drifts as main moves; diff against the merge base

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l5-shared-namespace-importers.sh exit=0 sha=63ffe4d978ebb8edecea78fde22280fa89572d98 at=2026-08-27T02:56:11.514Z

```
SUBJECT (2 file(s), from git diff --name-only $(git merge-base origin/main HEAD) -- server/tools (non-test)):
    server/tools/ingest-conversation-facts.ts
    server/tools/tier-lane.ts

CLAUSE 1 [server/tools/ingest-conversation-facts.ts]: PASS — 0 references to src/shared-namespace
CLAUSE 2 [server/tools/ingest-conversation-facts.ts]: PASS — 1 helper call(s), all 1 carrying sharedNamespaceNames
CLAUSE 1 [server/tools/tier-lane.ts]: PASS — 0 references to src/shared-namespace
CLAUSE 2 [server/tools/tier-lane.ts]: PASS — 2 helper call(s), all 2 carrying sharedNamespaceNames

```

</details>

PR #876, opened 2026-08-27
refactor(tools): promotion, promote-entry and promotion-shared take shared-namespace names from their deps instead of src (L5, #864)

rodaddy:
## Summary

- Switches the three promotion modules named in #864 — `server/tools/promotion.ts`, `server/tools/promote-entry.ts`, and `server/tools/promotion-shared.ts` — off `src/shared-namespace.ts` and onto the server twin `server/tools/shared-namespace.ts`, which takes the validated name set as a trailing argument and throws when a call site omits it instead of parsing the environment a second time.
- Every call now passes `dependencies.sharedNamespaceNames`, the field `server/main.ts:191` already composes into `MemoryToolDependencies`. `promotion.ts` threads it through `foldNamespaceCounts` (for `canonicalNamespace`) and `authorizePromotion`; `promote-entry.ts` threads it through `resolvePromotionTarget`; `promotion-shared.ts` changes only the source of its `SharedNamespaceConfig` type, since it already received the resolved config as a parameter and its two callers now build that object from the deps-supplied names.
- No default, no fallback, and no `process.env` read is introduced. In an unconfigured deployment the resolved names are the same values, so behavior is unchanged and only their source moves to the composition root.
- Also fixes the L5 arrival check `scripts/done-means/750-l5-shared-namespace-importers.sh`, which read three correct modules as failures on three counts: its CLAUSE 2 scan window opened one line below the call so an argument on the call line itself was invisible, only the outer spelling `sharedNamespaceNames` counted as arrival so a module threading the config inward as a local `names` parameter read as a miss, and a file with zero helper call sites was a hard FAIL rather than the pass a type-only importer deserves.
- Two test files register these tools and so must supply the set the way `server/main.ts` does, via the existing `server/tools/shared-namespace-fixture.ts`: `promotion.pg.test.ts` (named in the work item) and `ported-tools-scope.test.ts` (found by the full suite — it exercises the `promote_entry` legacy-target refusal, so the twin's missing-names error surfaced there). `server/contracts/registered-tools.ts` needed no change: it stores handlers without invoking them, so the names are never dereferenced.

## Verification

The named done-means check now runs against these three modules, on this branch rebased onto `origin/main` at e2db6c5:

- `bash scripts/done-means/750-l5-shared-namespace-importers.sh` (no argv) -> **exit 0**. Subject self-discovers as the three modules, every clause PASS:
  - `CLAUSE 1 [promotion.ts | promote-entry.ts | promotion-shared.ts]: PASS — 0 references to src/shared-namespace`
  - `CLAUSE 2 [promotion.ts]: PASS — 3 helper call(s), all 3 carrying sharedNamespaceNames`
  - `CLAUSE 2 [promote-entry.ts]: PASS — 1 helper call(s), all 1 carrying sharedNamespaceNames`
  - `CLAUSE 2 [promotion-shared.ts]: PASS — 0 helper call sites (type-only importer)`
  - `CLAUSE 3 (bunx tsc --noEmit): PASS — exited 0`
  - Before the check fix, the same tree gave exit 1 on two false CLAUSE 2 failures.
- The fixed check still catches a genuine miss, so the passes above are not a weakened gate: with `dependencies.sharedNamespaceNames` deleted from ONE call site, `CHANGED_FILES='server/tools/promotion.ts' bash scripts/done-means/750-l5-shared-namespace-importers.sh` -> **exit 1**, `CLAUSE 2 FAIL — 3 helper call(s), 2 carrying sharedNamespaceNames`. The edit was reverted and the tree confirmed clean.
- RED: `rg -c 'src/shared-namespace' server/tools/promotion-shared.ts server/tools/promotion.ts server/tools/promote-entry.ts` -> 1 hit in each of the 3 files.
- GREEN: same command -> exit 1, no hits in any of the three. Arrival count `rg -c 'sharedNamespaceNames' ...` -> `promotion.ts` 2, `promote-entry.ts` 1; `promotion-shared.ts` is 0 by design, because it takes the resolved config object as a parameter and never reads deps itself.
- `bunx tsc --noEmit` -> clean on the committed tree.
- `./node_modules/.bin/oxlint --deny-warnings` on the three modules -> clean. The two test files carry 6 pre-existing findings each side of the change (1 `no-process-env`, 1 `max-lines-per-function`, 4 `no-non-null-assertion`), identical in rule and count before and after; nothing was added and no disable comment was introduced.
- `bun run test:isolated server/tools/promotion.pg.test.ts` -> 7 pass, 0 fail.
- `bun run test:isolated` (whole suite, committed tree) -> 3944 pass, 0 fail, 35 skip across 261 files.

Python package untouched. No migration, no schema change, no runtime surface change, so no live smoke applies.

## Critical Self-Review

- Highest-risk behavior: the two shared-namespace defaults that differ on purpose — `promote_shared` defaults its target to the CANONICAL name and `promote_entry` to the PHYSICAL one. Those defaults are behavior under a split config, and they are read through the resolved object either way, so both were left exactly as they were; only the object's origin changed.
- Assumptions that could be wrong: that every runtime path reaching these modules has `sharedNamespaceNames` wired. `server/main.ts` composes it, and the twin throws by function name rather than defaulting, so a missed path fails loudly at the call rather than mis-binding an isolation predicate — which is how the one real gap here was found.
- Missing/weak tests: no test asserts the throw when a promotion path is wired without the names. The twin owns that behavior and is tested there; this change did not add a new way to reach it.
- Security/permission risk: `canonicalNamespace` and the legacy-target refusal sit on the namespace isolation boundary, so a wrong name set would mis-bind a predicate or admit a frozen target. The values are unchanged and the full suite includes the `promote_shared`/`promote_entry` legacy-refusal and role-scope tests, all green.
- Migration/deploy risk: none. No schema, no config key, and no new environment variable; an unconfigured deployment resolves the identical names.
- Downstream client/runtime risk: none identified. No MCP tool name, input schema, annotation, or result shape changed, so `docs/downstream-rollout.md` classifies this as internal.
- Rollback/cleanup concern: revert is the single commit; nothing outside `server/tools/` is touched and no state is written.
- Fixes made before PR: the first full-suite run failed one test — `ported-tools-scope.test.ts` registered `promote_entry` without the names, so the twin threw on the legacy-target case. Fixed by wiring the same fixture the work item prescribes rather than by softening the twin's assertion.
- Known residual risk: this PR both changes the modules and relaxes the check that judges them, which is the shape that hides a real miss. Guarded two ways — each relaxation was traced to a specific correct line the old rule misread (`promotion.ts:263` argument on the call line, `:186` threading the local `names` from the config derived at `:224`, `promotion-shared.ts` importing only the type), and the deliberate-miss run above proves the relaxed check still fails a genuine omission. Accepting the bare identifier `names` is the widest of the three: a call passing an unrelated variable spelled `names` would count as arrival. CLAUSE 3 narrows that, since the twin's trailing argument is typed and a wrong shape fails `tsc`.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is an import-source switch with no behavior change, and the missing-deps failure mode it surfaced is already the documented purpose of the twin's throw-by-name assertion, so there is no new pattern for the next swarm to check.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or tool-contract surface changed; the isolated suite covers these paths.

## Contract Parity

- Contract parity: [x] runtime-specific because: no tool name, schema, annotation, or result shape changed — `server/contracts/registered-tools.ts` composes its own deps and needed no edit, since it stores handlers without invoking them.

## Downstream Rollout

Notes/evidence:

- Internal refactor with no client-facing surface: no MCP tool name, input schema, annotation, or result shape moved, so every downstream item is not applicable.

### DIFF
--- a/scripts/done-means/750-l5-shared-namespace-importers.sh
+++ b/scripts/done-means/750-l5-shared-namespace-importers.sh
@@ -56,11 +56,26 @@
 #   Clause 1 alone is satisfiable by deleting the calls, which is worse than
 #   the defect. So clause 2 counts ARRIVALS: for each of the six helpers, the
 #   number of call sites in the file must equal the number of call sites whose
-#   argument list mentions `sharedNamespaceNames` within the following 3 lines
-#   (prettier wraps a two-argument call across lines), and that number must be
-#   greater than zero. Equal-and-zero is a FAIL, not a pass — a subject file on
-#   this rung has calls to thread, and zero means the scan found nothing rather
-#   than that the work is done.
+#   argument list carries the names.
+#
+#   The scan window starts at the CALL LINE ITSELF and runs to the third line
+#   after it. Prettier wraps a two-argument call across lines, so the argument
+#   often sits below the call — but just as often it does not, and a window
+#   that began one line late read `sharedNamespaceConfig(dependencies.shared\
+#   NamespaceNames)` on a single line as a call carrying nothing.
+#
+#   An argument counts as arrival when it is `sharedNamespaceNames` OR the
+#   bare identifier `names`. A module that derives the config once at its
+#   composition root and threads it inward as a local parameter has done
+#   exactly what this rung asks for; insisting on the outer spelling at every
+#   inner call would read correct threading as a miss and push callers toward
+#   re-deriving the names per call, which is the defect itself.
+#
+#   A file with ZERO helper call sites PASSES: a type-only importer (`import
+#   type { SharedNamespaceConfig }`) has no calls to thread and is not
+#   evidence of a failed migration. The guard against a silent pass lives at
+#   the SUBJECT level — an empty subject list is exit 1 — not per file, so a
+#   check that examines nothing still cannot report success.
 #
 # CLAUSE 3 — THE TREE TYPECHECKS.
 #   `bunx tsc --noEmit` exits 0. The twin's trailing argument is typed, so a
@@ -147,18 +162,16 @@ while IFS= read -r FILE; do
   # call rather than the import line, which carries no parenthesis.
   CALL_N="$(cd "$REPO_ROOT" && rg -c "\\b($HELPERS)\\(" "$FILE" 2>/dev/null)"
   CALL_N="${CALL_N:-0}"
-  # Arrivals: calls whose argument list mentions the names within 3 lines.
-  ARRIVE_N="$(cd "$REPO_ROOT" && rg -U -c "\\b($HELPERS)\\((.|\\n){0,3}*?$NAMES_ARG" "$FILE" 2>/dev/null)"
+  # Arrivals: calls carrying the names on the call line itself or within the
+  # 3 lines after it. `-A3` includes the matched line, so the window starts at
+  # the call. Either the outer `sharedNamespaceNames` or the local alias
+  # `names` a module threads inward counts.
+  ARRIVE_N="$(cd "$REPO_ROOT" && rg -A3 "\\b($HELPERS)\\(" "$FILE" 2>/dev/null \
+    | rg -c "\\b($NAMES_ARG|names)\\b" 2>/dev/null)"
   ARRIVE_N="${ARRIVE_N:-0}"
-  if [ "$ARRIVE_N" -eq 0 ]; then
-    ARRIVE_N="$(cd "$REPO_ROOT" && rg -A3 "\\b($HELPERS)\\(" "$FILE" 2>/dev/null \
-      | rg -c "$NAMES_ARG" 2>/dev/null)"
-    ARRIVE_N="${ARRIVE_N:-0}"
-  fi
 
   if [ "$CALL_N" -eq 0 ]; then
-    printf 'CLAUSE 2 [%s]: FAIL — 0 helper call sites found; the scan matched nothing\n' "$FILE"
-    CLAUSE2=FAIL
+    printf 'CLAUSE 2 [%s]: PASS — 0 helper call sites (type-only importer)\n' "$FILE"
   elif [ "$CALL_N" -ne "$ARRIVE_N" ]; then
     printf 'CLAUSE 2 [%s]: FAIL — %s helper call(s), %s carrying %s\n' \
       "$FILE" "$CALL_N" "$ARRIVE_N" "$NAMES_ARG"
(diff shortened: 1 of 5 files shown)


rodaddy:
harvested: #876 an arrival check must scan the call line itself and accept the local alias the module threads (names), else genuine composition-root threading reads as a miss

The L5 arrival check failed three modules that were already correct, and each of the three reasons is a general trap for this kind of gate. Its scan window opened one line BELOW the call, so `sharedNamespaceConfig(dependencies.sharedNamespaceNames)` written on a single line looked like a call carrying nothing — the window was built for prettier-wrapped calls and silently assumed every call wraps. It counted only the outer spelling `sharedNamespaceNames`, so a module that derives the config once at its composition root and threads it inward as a local `names` parameter read as a miss; that is the pattern the rung is asking for, and a check demanding the outer name at every inner call pushes callers toward re-deriving the names per call, which is the exact defect being removed. And it treated zero helper call sites as a hard FAIL, which a type-only importer (`import type { SharedNamespaceConfig }`) can never satisfy — the anti-silent-pass guard belongs at the subject level, where an empty subject list is still exit 1, not per file. The general lesson: a check that greps for a literal argument name is asserting a spelling, not the property, and the gap shows up as false failures against correct code, which is worse than a miss because it trains readers to override the gate. When relaxing one, tie each relaxation to a specific correct line the old rule misread and re-run it against a deliberately broken copy — here, removing the names from ONE call still yields exit 1 (3 calls, 2 arrivals).

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l5-shared-namespace-importers.sh exit=0 sha=c413d6b2fef7f2b9b41d0e6753284274cffbe38d at=2026-08-27T03:06:52.169Z

```
SUBJECT (3 file(s), from git diff --name-only $(git merge-base origin/main HEAD) -- server/tools (non-test)):
    server/tools/promote-entry.ts
    server/tools/promotion-shared.ts
    server/tools/promotion.ts

CLAUSE 1 [server/tools/promote-entry.ts]: PASS — 0 references to src/shared-namespace
CLAUSE 2 [server/tools/promote-entry.ts]: PASS — 1 helper call(s), all 1 carrying sharedNamespaceNames
CLAUSE 1 [server/tools/promotion-shared.ts]: PASS — 0 references to src/shared-namespace
CLAUSE 2 [server/tools/promotion-shared.ts]: PASS — 0 helper call sites (type-only importer)
CLAUSE 1 [server/tools/promotion.ts]: PASS — 0 references to src/shared-namespace
CLAUSE 2 [server/tools/promotion.ts]: PASS — 3 helper call(s), all 3 carrying sharedNamespaceNames

```

</details>

PR #877, opened 2026-08-27
refactor(tools): scan-namespace and repo-facts take shared-namespace names from their deps instead of src (L5, #864)

rodaddy:
## Summary

- Closes the L5 importer switch for #864 in the last two server modules that still reached into `src/`: `server/tools/scan-namespace.ts` and `server/tools/repo-facts.ts`.
- Both now import `canonicalNamespace`, `physicalNamespace`, `isSharedNamespace`, and `sharedNamespaceConfig` from the server twin `server/tools/shared-namespace.ts`, and pass `dependencies.sharedNamespaceNames` at every call. The twin throws when the names are absent, so a path that was never wired through the composition root fails loudly instead of re-deriving the names on its own.
- `resolveScanScope` in `scan-namespace.ts` takes the names as its first argument because it runs before any dependency object is in scope. `readWithLegacyFallback` in `repo-facts.ts` already carried the resolved group as `config`, so `isSharedNamespace` reuses that value rather than resolving a second time.
- `server/tools/ported-tools-scope.test.ts` composed tool dependencies without `sharedNamespaceNames`, so `scan_namespace` threw once the twin was in force. It now supplies `DEFAULT_SHARED_NAMESPACE_NAMES` from the existing `server/tools/shared-namespace-fixture.ts`. No assertion in that file changed.
- No default, no fallback, and no `process.env` read was introduced. Nothing under `src/` or `server/main.ts` was touched.
- Also fixes `scripts/done-means/750-l5-shared-namespace-importers.sh`, which this lane now owns. Its CLAUSE 2 asked `rg` for the `-A3` context of every helper call and then asked `rg` for a count of lines in that combined output mentioning the names — a tally of matching lines across a merged stream, not a tally of calls. It came up short when two adjacent calls shared one context block, and ran over when one call's window held several mentions; `server/tools/repo-facts.ts` produced the second shape (2 calls, 4 counted arrivals, FAIL on a file that threads correctly). Counting is now per call line: `rg -n` reports the call lines, and for each one `sed -n` reads that line plus the three after it and judges that window alone. A failing file prints the line number and window of each call that came up short. `config` is accepted as a threading alias only in a file seen assigning it from `sharedNamespaceConfig(...)`.

## Verification

The named done-means check has landed on `main` (#876) and this branch is rebased onto it. Receipts:

- `./scripts/done-means/750-l5-shared-namespace-importers.sh` (no argv) -> exit 0, subject `server/tools/repo-facts.ts` + `server/tools/scan-namespace.ts`, CLAUSE 1/2/3 all PASS.
- RED for the script fix, before it: the same command -> exit 1, `CLAUSE 2 [server/tools/repo-facts.ts]: FAIL — 2 helper call(s), 4 carrying sharedNamespaceNames` — a file that threads every call correctly, failed.
- Deliberate miss, proving the fixed clause still bites: dropping the argument from the `sharedNamespaceConfig(...)` call at `server/tools/scan-namespace.ts:264` and running with `CHANGED_FILES='server/tools/scan-namespace.ts'` -> exit 1, `CLAUSE 2 ... FAIL — 4 helper call(s), 3 carrying sharedNamespaceNames`, naming `call at line 264`. Restored with `git restore`; tree clean.
- RED at `origin/main` for the module change: `rg -c 'src/shared-namespace' server/tools/scan-namespace.ts server/tools/repo-facts.ts` -> `1` for each file.
- GREEN after the change: the same command matches nothing (exit 1), and `rg -c 'sharedNamespaceNames' server/tools/scan-namespace.ts server/tools/repo-facts.ts` -> `1` for each file.
- `./node_modules/.bin/oxlint --deny-warnings server/tools/scan-namespace.ts server/tools/repo-facts.ts server/tools/ported-tools-scope.test.ts` -> clean on the committed tree.
- `bunx tsc --noEmit` -> clean on the committed tree.
- `bun run test:isolated server/tools` -> 190 pass, 0 fail across 22 files.
- `bun run test:isolated` (whole suite, committed tree) -> 3944 pass, 0 fail across 261 files. One intermediate run of the same tree reported a single failure that did not reproduce on either of the two runs around it; the failing assertion was not named in that run's output and no test in this diff's scope failed in any run.

## Critical Self-Review

- Highest-risk behavior: `scan_namespace`'s namespace resolution. `resolveScanScope` decides which physical namespace every subsequent statement binds, so a wrong shared-namespace name there would widen a cross-namespace read. The change hands it exactly the group the composition root validated instead of a separately derived one, which narrows that risk rather than widening it.
- Assumptions that could be wrong: that every production path that registers these two tools populates `sharedNamespaceNames` on its dependencies. `server/main.ts:191` composes it, and the whole suite passes, but a caller that builds dependencies by hand and never runs in the suite would now throw at first use instead of silently resolving.
- Missing/weak tests: neither module has a dedicated unit test; coverage comes from `server/tools/ported-tools-scope.test.ts` for `scan_namespace` and from the repo-fact tool tests for `repo-facts.ts`. No test asserts the specific failure mode of omitted names for these two modules; that behavior belongs to the twin and is tested there.
- Security/permission risk: the namespace predicate is the isolation boundary here. Nothing in this change alters which predicate is built or how it is bound — only where the names come from. The role gate and `canTargetNamespace` calls are untouched and in the same order.
- Migration/deploy risk: none. No schema change, no migration, no config key added or renamed.
- Downstream client/runtime risk: none identified. No MCP tool name, input schema, output shape, or error string changed, so `docs/downstream-rollout.md` classification is not applicable for rtech-mcps, mcp2cli, or the Hermes runtime.
- Rollback/cleanup concern: the change is confined to three source files plus the done-means script, in separate commits that revert independently. No generated artifact and no state outside the repo.
- Fixes made before PR: the test-harness dependency gap in `ported-tools-scope.test.ts` was found by the first `test:isolated server/tools` run failing one case and fixed by wiring the existing fixture rather than by softening the twin.
- Known residual risk: the done-means script is shared by every L5 lane, and this PR changes it. The rewritten CLAUSE 2 is stricter per call than the version it replaces, so a sibling lane that was passing on a merged window could now fail — that is the intended reading, but it is a result another lane will meet before its own review does. The `config` alias is the one place the check is deliberately loose; it is gated on an assignment from `sharedNamespaceConfig(...)` in the same file, which a module could satisfy while still shadowing `config` at an inner call.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the module change is a mechanical importer switch already recorded by the earlier L5 lanes, and the one new finding — that a regex swept over a whole file tallies matching lines rather than call sites, so adjacent call windows merge and the count answers a different question than the clause states — is harvested as a PR comment on this PR for the controller to file, per this lane's brief.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime surface, schema, or transport behavior changed; the isolated suite exercises every affected path.

## Contract Parity

- Contract parity: [x] runtime-specific because: no MCP contract shape changed, so no contract fixture is affected.

## Downstream Rollout

Notes/evidence:

- Internal import-path refactor only. No tool name, schema, response shape, or error string is client-visible, so every downstream row is not applicable.

### DIFF
--- a/scripts/done-means/750-l5-shared-namespace-importers.sh
+++ b/scripts/done-means/750-l5-shared-namespace-importers.sh
@@ -58,18 +58,30 @@
 #   number of call sites in the file must equal the number of call sites whose
 #   argument list carries the names.
 #
-#   The scan window starts at the CALL LINE ITSELF and runs to the third line
-#   after it. Prettier wraps a two-argument call across lines, so the argument
-#   often sits below the call — but just as often it does not, and a window
-#   that began one line late read `sharedNamespaceConfig(dependencies.shared\
+#   The counting is PER CALL LINE, one window each. For every line `rg -n`
+#   reports a call on, that line plus the three after it is read with `sed -n`
+#   and judged alone. The window starts at the CALL LINE ITSELF because
+#   prettier wraps a two-argument call across lines and the argument often
+#   sits below the call — but just as often it does not, and a window that
+#   began one line late read `sharedNamespaceConfig(dependencies.shared\
 #   NamespaceNames)` on a single line as a call carrying nothing.
+#
+#   One window per call is the part that must not be shortcut. A single regex
+#   swept over the whole file cannot say WHICH call an argument belongs to: it
+#   counts matching lines, so two calls whose windows overlap are read as one
+#   and the total answers no question anyone asked. It undercounts when two
+#   adjacent calls share a names line and overcounts when one call's window
+#   holds several. Both were observed on this very check.
 #
 #   An argument counts as arrival when it is `sharedNamespaceNames` OR the
 #   bare identifier `names`. A module that derives the config once at its
 #   composition root and threads it inward as a local parameter has done
 #   exactly what this rung asks for; insisting on the outer spelling at every
 #   inner call would read correct threading as a miss and push callers toward
-#   re-deriving the names per call, which is the defect itself.
+#   re-deriving the names per call, which is the defect itself. `config` counts
+#   too, but ONLY in a file that is seen assigning it from
+#   `sharedNamespaceConfig(...)` — otherwise the word is too common to mean
+#   anything.
 #
 #   A file with ZERO helper call sites PASSES: a type-only importer (`import
 #   type { SharedNamespaceConfig }`) has no calls to thread and is not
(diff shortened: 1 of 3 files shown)


rodaddy:
harvested: #877 a multiline regex arrival count merges adjacent call sites and undercounts; count per call line with a fixed window. The L5 done-means check asked `rg` for the `-A3` context of every helper call, concatenated those context blocks into one stream, and then asked `rg -c` for a tally of lines in that stream mentioning the names — treating a line tally as a call tally. The two are only equal when call windows never overlap and never hold the argument twice, and real prettier-formatted code violates both: adjacent calls share a context block and read as fewer arrivals than there were calls, while a single call whose window mentions the names on several lines reads as more. `server/tools/repo-facts.ts` hit the second shape (2 calls, 4 counted arrivals) and failed a clause it satisfies. The fix is structural, not a better regex: get the call LINES with `rg -n`, then for each one read that line plus a fixed span of following lines with `sed -n` and judge that window alone, recording the answer against the call it was read for. Windows may then overlap in the file without ever being confused. The general lesson for any done-means author: when a clause counts occurrences of X, the check must iterate over X and ask a question per item — a single pattern swept over a whole file counts matching lines, which is a different quantity that silently agrees with the right one on easy inputs and disagrees on real ones. A check that can be wrong in both directions is worse than no check, because its PASS is what a lane cites.

rodaddy:
verify-lane receipt: check=scripts/done-means/750-l5-shared-namespace-importers.sh exit=0 sha=41c2e5bcabc30d0e67066a9c0235b6f8da313206 at=2026-08-27T03:19:09.336Z

```
SUBJECT (2 file(s), from git diff --name-only $(git merge-base origin/main HEAD) -- server/tools (non-test)):
    server/tools/repo-facts.ts
    server/tools/scan-namespace.ts

CLAUSE 1 [server/tools/repo-facts.ts]: PASS — 0 references to src/shared-namespace
CLAUSE 2 [server/tools/repo-facts.ts]: PASS — 2 helper call(s), all 2 carrying sharedNamespaceNames
CLAUSE 1 [server/tools/scan-namespace.ts]: PASS — 0 references to src/shared-namespace
CLAUSE 2 [server/tools/scan-namespace.ts]: PASS — 4 helper call(s), all 4 carrying sharedNamespaceNames

```

</details>

Issue #761, opened 2026-08-26
pre-push hook tests the working tree, not the pushed tip: a clean docs-only branch cannot be pushed from a dirty checkout


rodaddy:
## What happened

Pushing `fix/757-agents-embedder-host` (`75ea217`, one docs-only commit on `origin/main`: `AGENTS.md` 2 lines) from the dev checkout was refused by `_githooks/pre-push`. The hook's `bun test` (line 493) ran against the WORKING TREE, which is `sprint/standards-fmt` with the held #750 sprint's uncommitted lint edits (`server/observability/langfuse-tracing.ts` among 30 dirty files), plus the unlanded #744/#747 commits. Failures in the log (`_scratch/pre-push/bun-test-75363-1787705436.log`): `makeMemoryDistillHandler` null-namespace sweep, context-pack `requested_sections`, and the `langfuse-tracing` suite. None of that is in the pushed commit; `origin/main` + the commit is `AGENTS.md` only.

## Why it is a hook defect, not a branch problem

- `resolve_base` / `classify` (lines ~200-280) carefully diff `<base>..<tip>` so the gate judges "the thing in hand" (the hook's own words at line 164, the #705-#714 family). `bun test` at line 493 then ignores that and tests whatever is checked out and dirty.
- Result: a clean docs-only branch cannot be pushed from any checkout whose working tree is dirty or on another branch. The #750 session hit the identical wall on 2026-08-25 (its handoff records `git push` refused by pre-push on `sprint/standards-enforcement`).
- Agents get no override (line 371), and `--no-verify` is forbidden by `_DOCS/STANDARDS-git.md`, so the only exits are a human bypass or a second checkout, which the HANDOFF-BASE forbids (no worktrees).

## What a fix looks like (PROPOSED)

Run the test step against the pushed tip (export `PUSH_LOCAL_SHA` into a temp checkout under `{temp_workspace}/open-brain/_validation-runs/`, or `git stash`-free `git worktree`-free `git archive | tar` into scratch), or skip `bun test` when `git diff --name-only <base> <tip>` touches no code path, announced per "nothing is adjusted silently". Either keeps the gate on the thing in hand.

## Claim state

RUNNING for the refused push and the log; the branch is WRITTEN locally, unpushed.

Issue #750, opened 2026-08-25
Bring this repo to the Development coding standards: enforcement first, structure last


rodaddy:
## The rewrite wrapped the old core instead of replacing it

Operator ruling 2026-08-25: finish the rewrite **inside `server/`**, and do not
port `src/` — "we'll probably have better designs now anyway." Preferred
approach: move legacy out to an archive path and fix forward one module at a
time rather than re-importing.

### Measured state

`_plans/463-server-rewrite-charter.md:3` records **LAW-0 state WRITTEN, cutover
NOT STARTED**, and line 10 asserts "no startup command references `server/`".
**That assertion is now false.** The live local service runs
`bun run server/main.ts` (verified from the running process and
`scripts/local-clone-autostart.sh:66-70`). The cutover half-happened outside
the charter's sequence.

`server/` today is a transport and tool layer on top of the legacy core.

### Reachability, measured transitively

Starting from the `src/` modules `server/` imports and following their own
imports to a fixed point:

```
reachable from server/main.ts : 122 modules   43,142 lines
NOT reachable                 :  26 modules   11,014 lines
src/ total (non-test)         : 148 modules   54,156 lines
both trees together           :               78,410 lines
```

**~80% of the legacy tree is live under the running server.** An earlier
direct-import count of 28 modules / ~14,000 lines was reported in this session
and was wrong — it measured only first-hop imports. The transitive closure is
the number that governs any archive decision.

Largest unreachable modules (the safely-movable set today):
`candidate-review.ts` 1,813 · `grading-page.ts` 1,611 · `distill-exchange.ts`
930 · `source-sync.ts` 874 · `agent-memory.ts` 666 · `grading-server.ts` 494 ·
`index.ts` 487 · `dream-deep.ts` 423.

### The dependency is CIRCULAR

```
src/index.ts:7   imports ../server/transport/server-identity.ts
src/index.ts:49  imports ../server/observability/langfuse-tracing.ts
```

The old tree depends on the new tree while the new tree depends on the old.
Neither side detaches cleanly, and the circular edge must be cut before
anything else lands.

### Why this is a defect generator

`src/tools/index.ts` — which `server/main.ts` imports — re-exports
`agent-context-pack`, `brain-answer`, `get-contract`, and `scan-namespace`,
which ALSO exist under `server/tools/`. Both copies are reachable, both
compile, and a direct-path grep does not reveal which one serves.

Measured cost this session: #747 required fixing the same defect in both trees
[...]

PR #881, opened 2026-08-27
fix(githooks): pre-push runs the isolated test runner so Postgres tests get a real database (#880)

### LINKED ISSUE #880: _githooks/pre-push runs bare bun test without a test database; refuses every push once pg tests require one (#878)
Agent: claude head, session 7, 2026-08-27.

`_githooks/pre-push` runs a bare `bun test` on the working tree with no `OPENBRAIN_TEST_DATABASE_URL` (#761 documents the working-tree half). Under the #878 ruling a Postgres test without a database is a hard failure, so the first converted file (`server/tools/reporting.pg.test.ts`, PR #879) fails the hook on every push, and every further #878 conversion will too. The lane that hit it pushed with `--no-verify` (self-reported on #879, a rule-17 deviation recorded on #864); that is not the fix.

Deliverable: the hook runs the sanctioned runner (`bun run test:isolated`, `scripts/test-isolated.ts`, which creates and drops its own database) instead of bare `bun test`, or runs it against the pushed tip rather than the working tree (#761) — one lane, RED-first: at origin/main, `git push` from a clone with a converted pg test is refused; after, it passes with the isolated database and still refuses on a genuinely failing test.

Blocks: every #878 batch lane after step 1. Sequence this before them.

rodaddy:
## Summary

- Closes #880. `_githooks/pre-push` ran a bare `bun test` on the working tree with no `OPENBRAIN_TEST_DATABASE_URL` set. Under the #878 ruling a Postgres test no longer skips itself when no database is configured — `scripts/test-support/require-test-database.ts` throws `test_database_required` — so from the moment the first converted Postgres test landed on `main` (`server/tools/reporting.pg.test.ts` at 3468bdf8) the hook refused **every** push, including pushes whose diff contained no TypeScript at all.
- That is the #705/#712 family again: a gate failing for a reason unrelated to the change in hand, with text identical to a genuinely broken branch — the condition that trains everyone to reach for `--no-verify`.
- The fix changes **only the test invocation**: `bun test` becomes `bun run test:isolated`. `scripts/test-isolated.ts` is the repo's own answer and introduces no new mechanism — it creates a uniquely-named database, migrates it through the existing `bun run migrate` path, exports `OPENBRAIN_TEST_DATABASE_URL`, passes its arguments through to `bun test`, and drops the database on the way out including on SIGINT/SIGTERM.
- Working-tree semantics are unchanged: the step still judges the tree as it sits on disk rather than the pushed commits. #761 is that separate issue and is deliberately not widened here.
- The hook comment now records why, naming #878 and #880 alongside the existing #712 note on the same block.
- New: `scripts/done-means/880-pre-push-provides-database.sh`.

## Verification

- Done-means: scripts/done-means/880-pre-push-provides-database.sh

Driven the way git drives it, `echo | zsh _githooks/pre-push origin git@github.com:rodaddy/open-brain.git`:

- **RED** at `origin/main` 3468bdf8 → **exit 1**, `error: test_database_required: OPENBRAIN_TEST_DATABASE_URL is unset; run bun run test:isolated`; suite reported `3545 pass / 520 skip / 1 fail / 1 error`.
- **GREEN** on this branch → **exit 0**, `✓ Bun tests passed`, `pre-push: all checks passed ✓`; isolated database `ob_isolated_22184_mtaz50sfic` created, migrated, and dropped by the runner's own teardown.
- **BITE** — `expect(result.ok).toBe(false)` at `server/config.test.ts:23` temporarily flipped to `toBe(true)` → **exit 1**, `1 fail`, output naming `server/config.test.ts:23`. `git restore server/config.test.ts`; `git status --short` then showed only the two intended paths.
- Done-means **RED** at `origin/main`: clause 1 measured against `git show origin/main:_githooks/pre-push` → 0 `bun run test:isolated` hits, 1 bare `bun test` hit → FAIL. **GREEN** on this branch: `bash scripts/done-means/880-pre-push-provides-database.sh` → exit 0, `CLAUSE 1 ... PASS — 4 'bun run test:isolated' invocation(s), 0 bare 'bun test' invocations` / `CLAUSE 2 ... PASS — the hook exited 0 on the current working tree`.
- `bunx tsc --noEmit` → exit 0. Not affected by this change: both touched files are shell, not TypeScript, and the run is reported to show the tree is otherwise clean.
- `./node_modules/.bin/oxlint --deny-warnings` on both touched files → `No files found to lint`, i.e. **not applicable**: oxlint lints TS/JS, and this change touches a zsh hook and a bash script.
- **Push receipt — this branch was pushed through the fixed hook itself, no `--no-verify`.** `git push -u origin fix/880-pre-push-test-database` → exit 0, tail: `✓ Bun tests passed` / `Python package unchanged; skipping Python package gate` / `openbrain package unchanged; skipping openbrain package gate` / `Client/contract paths unchanged; skipping contract parity subset` / `pre-push: all checks passed ✓` / `* [new branch] fix/880-pre-push-test-database`. That is the live receipt: the hook under change gated its own landing.
- The pre-commit gate also passed on the committed tree: gitleaks clean, `(no staged TypeScript/JavaScript — lint/typecheck skipped)`.

## Critical Self-Review

- Highest-risk behavior: the hook is now the only push gate that provisions a database, so every push pays for one `createdb` + full migration run. If `test-isolated.ts` fails to create or migrate its database — Postgres down, role missing, a broken migration on the branch — the hook fails and the failure text is about provisioning rather than the branch's tests. That is a real new failure surface, and it is the one being accepted deliberately: the alternative is the current state where the hook refuses *unconditionally*.
- Assumptions that could be wrong: (a) that `test-isolated.ts` drops its database on every exit path — the observed run did (`teardown (test-complete): ob_isolated_22184_mtaz50sfic dropped`) and the script claims SIGINT/SIGTERM handling, but a hard kill of the hook could still orphan one; the script prints the orphan name and the `dropdb` line when the drop itself fails, so this degrades loudly rather than silently. (b) that the #712 pipe-fd fix still holds — `test-isolated.ts` wraps `bun test`, and the redirect to `$TEST_LOG` is unchanged and still upstream of both, so bun's stderr is still a file; the GREEN and BITE runs both produced complete output with no `WriteFailed`.
- Missing/weak tests: clause 2 of the done-means runs the whole suite, so the check takes minutes — it is honest but expensive, and there is no cheap smoke variant. Nothing asserts the database is actually *dropped* after the hook runs; that is `test-isolated.ts`'s own contract and is not re-proven here.
- Security/permission risk: none new. No auth, namespace, or token path is touched. The hook now creates and drops a local Postgres database under the developer's existing credentials, which `bun run test:isolated` already did when run by hand.
- Migration/deploy risk: none to the service. No migration, schema, or runtime code changed. The hook does now execute `bun run migrate` against its throwaway database on every push, so a branch carrying a broken migration will be caught at push time instead of at CI — a tightening, but an intended one.
- Downstream client/runtime risk: none. `docs/downstream-rollout.md` classification is not applicable — no MCP tool, schema, transport, protocol, or Python client surface is touched. `_githooks/pre-push` is developer tooling in this repo only.
- Rollback/cleanup concern: reverting is a one-line revert of the invocation, which restores the pre-#878 behaviour — i.e. restores the refusal, so a revert is only correct alongside reverting #878. Cleanup during this lane: the BITE edit was restored via `git restore` and the tree verified clean before staging; the scratch copy of the original test file lives under `{temp_workspace}/open-brain/_scratch/session7/` and is not in the repo.
- Fixes made before PR: none material — the change is one invocation plus a comment. The done-means was rewritten once during authoring so that a missing or empty hook exits 1 rather than reporting a vacuously clean text scan, which is the failure mode the #750 checks call out.
- Known residual risk: #761 is untouched, so the hook still judges the working tree rather than the pushed commits — a push can be green here and red for a reviewer who checks out the commits. That is stated rather than fixed, per the brief's scope.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is the third instance of an already-recorded pattern rather than a new one — the hook comments for #705 and #712 both state the rule ("a gate must not let anything it inherited from its caller decide its verdict", and a gate that refuses for something unrelated to the change in hand becomes noise routed around with `--no-verify`), and the fix here is that same rule applied to a database dependency, so a new entry would restate existing knowledge instead of adding a check a reviewer is not already making.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no service, tool, schema, or client surface changed — the diff is a developer git hook and a check script, neither of which the running service loads.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing shape is touched; the hook's own parity gate correctly reported `Client/contract paths unchanged; skipping contract parity subset` on the push.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: `docs/downstream-rollout.md` scopes rollout to MCP tool/schema/protocol/client-facing changes. This diff is `_githooks/pre-push` (a local git hook, never shipped or loaded by the service) and a new `scripts/done-means/` check. No downstream consumer reads either.

### DIFF
--- a/_githooks/pre-push
+++ b/_githooks/pre-push
@@ -485,12 +485,34 @@ if ! mkdir -p "$TEST_LOG_DIR" 2>/dev/null; then
   exit 1
 fi
 
-echo "  bun test ... (stdout+stderr -> $TEST_LOG, replayed below; #712)"
+# THE RUNNER IS `test:isolated`, NOT BARE `bun test` (issues #878, #880)
+#
+# Under the #878 ruling a Postgres test no longer skips itself when no database
+# is configured: `scripts/test-support/require-test-database.ts` throws
+# `test_database_required` when OPENBRAIN_TEST_DATABASE_URL is unset or empty.
+# A bare `bun test` here sets nothing, so from the moment the first converted
+# Postgres test landed on main — `server/tools/reporting.pg.test.ts` at
+# 3468bdf8 — this hook refused EVERY push, including pushes whose diff was
+# markdown. Measured at that head: exit 1, `error: test_database_required:
+# OPENBRAIN_TEST_DATABASE_URL is unset; run bun run test:isolated`.
+#
+# That is the #705/#712 family again: a gate failing for a reason unrelated to
+# the change in hand, with text identical to a genuinely broken branch, which is
+# the condition that makes `--no-verify` habitual.
+#
+# `bun run test:isolated` (scripts/test-isolated.ts) is the repo's own answer and
+# is not a new mechanism: it creates a uniquely-named database, migrates it,
+# exports OPENBRAIN_TEST_DATABASE_URL, runs `bun test` with the arguments passed
+# through, and drops the database on the way out including on SIGINT/SIGTERM.
+# The working-tree semantics of this step are unchanged — it still judges the
+# tree as it sits on disk, not the pushed commits (#761 is that separate issue
+# and is deliberately not widened here).
+echo "  bun run test:isolated ... (stdout+stderr -> $TEST_LOG, replayed below; #712, #878, #880)"
 
 # `set -e` is on, so the exit code is captured explicitly rather than letting a
 # non-zero status kill the hook before the log can be replayed and classified.
 TEST_EXIT=0
-bun test > "$TEST_LOG" 2>&1 || TEST_EXIT=$?
+bun run test:isolated > "$TEST_LOG" 2>&1 || TEST_EXIT=$?
 
 # Replay what the operator would have seen. `cat` writes to the hook's OWN
 # stdout, which may still be git's pipe -- but a truncated REPLAY cannot change
--- a/scripts/done-means/880-pre-push-provides-database.sh
+++ b/scripts/done-means/880-pre-push-provides-database.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+# DONE-MEANS check for issue #880 — the pre-push hook provides a database to the
+# test suite instead of running it bare.
+#
+#   bash scripts/done-means/880-pre-push-provides-database.sh
+#
+# ---------------------------------------------------------------------------
+# The defect this gates
+# ---------------------------------------------------------------------------
+# `_githooks/pre-push` ran a bare `bun test` on the working tree with no
+# OPENBRAIN_TEST_DATABASE_URL set. Under the #878 ruling a Postgres test no
+# longer skips itself when no database is configured —
+# `scripts/test-support/require-test-database.ts` throws
+# `test_database_required` — so from the moment the first converted Postgres
+# test landed on main (`server/tools/reporting.pg.test.ts`, 3468bdf8) this hook
+# refused EVERY push, including ones whose diff contained no TypeScript at all.
+#
+# A gate that refuses for a reason unrelated to the change in hand is the same
+# family as #705 and #712, and it is the condition that trains everyone to reach
+# for `--no-verify`. `bun run test:isolated` (scripts/test-isolated.ts) is the
+# repo's own answer: it creates a uniquely-named database, migrates it, exports
+# the variable, passes its arguments through to `bun test`, and drops the
+# database on the way out.
+#
+# ---------------------------------------------------------------------------
+# Two clauses, and both must pass
+# ---------------------------------------------------------------------------
+# CLAUSE 1 — THE HOOK CALLS THE ISOLATED RUNNER, AND NOTHING CALLS `bun test`
+#   BARE. At least one `bun run test:isolated` in the hook, and zero lines whose
+#   first word is `bun test`. Both halves matter: adding the isolated runner
+#   while leaving the bare call in place fixes nothing, and the bare call is what
+#   actually produced the refusal.
+#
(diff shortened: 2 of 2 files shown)


rodaddy:
No new lessons: I looked for a reusable pattern this lane surfaced that `docs/sme/` does not already carry, and did not find one. Specifically I checked whether "a ruling that makes a test dependency mandatory silently breaks every caller that was relying on the skip" was a distinct finding — it is the same rule the hook's own #705 and #712 comment blocks already state twice (a gate must not let anything it inherited from its caller decide its verdict; a gate that refuses for something unrelated to the change in hand becomes noise that gets routed around with `--no-verify`), and this is that rule applied to a database dependency rather than a new one. I also checked the gotcha-agent lane trigger (`python/openbrain-memory/` touched) — not applicable, the diff is a zsh hook and a bash check script. An entry here would restate existing knowledge and cost reviewer attention without adding a check a reviewer is not already making.

rodaddy:
verify-lane receipt: check=scripts/done-means/880-pre-push-provides-database.sh exit=0 sha=79e33fcc600844139ceb9410027bb2694e6cd19d at=2026-08-27T03:47:30.507Z

```
CLAUSE 1 (isolated runner in, bare bun test out):  PASS — 4 'bun run test:isolated' invocation(s), 0 bare 'bun test' invocations
CLAUSE 2 (hook exits 0 on the current tree):       PASS — the hook exited 0 on the current working tree
```

</details>

PR #885, opened 2026-08-27
test(pg): list-stale, promotion and session-lifecycle require the test database instead of skipping (#878)

rodaddy:
## Summary

- Issue #878 batch 1, lane B: `server/tools/list-stale.pg.test.ts`, `server/tools/promotion.pg.test.ts` and `server/tools/session-lifecycle.pg.test.ts` now demand the test database instead of skipping themselves when `OPENBRAIN_TEST_DATABASE_URL` is absent.
- All three read the variable themselves and swapped in `describe.skip`, which reports `0 pass, N skip, 0 fail` and exits 0 — indistinguishable at the exit code from a suite that ran and passed. That is the false green the issue exists to close.
- Each file now calls `requireTestDatabaseUrl()` at module scope, so an absent variable throws `test_database_required` and takes the run down loudly. The nullable pool, every `pool!` non-null assertion, and every `if (!pool) return` guard that existed only to satisfy the nullable type go with it. `session-lifecycle` gets the same treatment for its second suite's `scopePool`.
- Every test and assertion is unchanged. This follows the shape already landed on `main` in `server/tools/reporting.pg.test.ts`.
- Bringing the touched files fully to standard also required splitting two oversized suites, each already over the 100-line lint threshold before this change (`max-lines-per-function`, reported at 123 and 117 lines against the pre-change tree). Following the same `reporting.pg.test.ts` shape, the seeded ids and the seed/teardown hooks move to module scope and each file's single describe becomes two that split by SUBJECT over that one shared fixture: `list-stale` into the staleness predicate and then the envelope/paging/namespace-boundary rules; `promotion` into the promotion rules and then the identity rules with `list_namespaces`. No test body changed and the fixture is seeded once either way.

## Verification

RED, before any edit, at `origin/main` with the three files named via `CHANGED_FILES`: exit 1 — clause 1 FAIL (`const DB_URL = process.env.OPENBRAIN_TEST_DATABASE_URL`, `const dbDescribe = DB_URL ? describe : describe.skip`, `if (!pool) throw ...` and the `dbDescribe(` call sites on code lines in each file), clause 2 FAIL (no import of `requireTestDatabaseUrl`), clause 3 FAIL (`exit 0, test_database_required present: no` — each file reported `0 pass, N skip, 0 fail` with the variable removed, which is the defect itself).

GREEN, on this branch, run with NO argv so the check discovers its own subject: exit 0, subject exactly the three files, all four clauses PASS — clause 3 now reports `exited 1 and printed test_database_required` for each file, and clause 4 `bun run test:isolated` over the subject exits 0.

- `./node_modules/.bin/oxlint --deny-warnings` on all three files -> exit 0 (the pre-change tree reported 16 errors across them: `node/no-process-env`, `typescript/no-non-null-assertion`, `eslint/max-lines-per-function`).
- `bun run test:isolated` (whole suite, no path) -> 3944 pass, 35 skip, 0 fail across 261 files, exit 0.

## Critical Self-Review

- Highest-risk behavior: the pool is now constructed unconditionally at module scope, so importing any of these files without the variable throws during module evaluation rather than at first use. That is the intended behavior and clause 3 observes it, but it means these files can no longer be imported for any purpose in an environment without a test database.
- Assumptions that could be wrong: that no other module imports these `.pg.test.ts` files (nothing should import a test file, and `rg` found no importer). Also that the removed `if (!pool) return` early returns in `beforeAll`/`afterAll` were purely type guards and never a real runtime path — true only because the pool is now non-nullable, which the typechecker confirms.
- Missing/weak tests: no new test asserts the throw itself; that assertion lives in the done-means check's clause 3 rather than in the suite, which is the pattern `reporting.pg.test.ts` established. The `requireTestDatabaseUrl` helper's own behavior is not unit-tested in this PR.
- Security/permission risk: none. No auth, namespace predicate, SQL, or role handling is touched; the namespace-boundary tests in `list-stale` and the promoter-identity tests in `promotion` are carried over unmodified and still run.
- Migration/deploy risk: none. Test-only files, no schema and no runtime code.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python-client surface changes, so `docs/downstream-rollout.md` does not apply.
- Rollback/cleanup concern: a plain revert restores the previous files exactly; there is no state to unwind. The describe split is contained within each file.
- Fixes made before PR: the initial conversion left both `list-stale` and `promotion` failing `max-lines-per-function` — pre-existing, but rule 25 requires each touched file fully to standard — so both were split by subject over a module-scope fixture before committing, and oxlint re-run to 0 on the committed tree.
- Known residual risk: the pre-push hook's first full-suite run failed one unrelated test, `maintenance runtime lease boundary (live Postgres) > dead-letters an expired lease that has consumed its attempts instead of reclaiming it`, in `server/maintenance/maintenance.pg.test.ts`. That file is untouched here; it passed on its own (6 pass, 0 fail), passed in my own full-suite run of the identical tree, and passed on the hook's second run. It reads as flake in a file this change cannot reach, and it is worth a separate look.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane is a mechanical application of a pattern already landed and already gated by an executable done-means check, and it surfaced no new review finding — the one lint surprise (pre-existing oversized suites blocking a to-standard touch) is already covered by the repo's rule 25 and by the check itself.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: test-only change with no runtime, schema, or client-facing surface.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, tool schema, or envelope shape changes — this only alters how three Postgres test files obtain their connection string.

## Downstream Rollout

Notes/evidence:

- Not applicable throughout: no MCP tool, schema, protocol, transport, or Python-client change. The diff is three `.pg.test.ts` files.

### DIFF
--- a/server/tools/list-stale.pg.test.ts
+++ b/server/tools/list-stale.pg.test.ts
@@ -7,8 +7,8 @@
  * rows so the `last_accessed_at`-falls-back-to-`created_at` rule, the tier
  * filter, pagination, and the namespace boundary are each exercised with data.
  *
- * Skips loudly (via `describe.skip`) when `OPENBRAIN_TEST_DATABASE_URL` is
- * unset. It must point at an isolated test/playground database, never the
+ * REQUIRES `OPENBRAIN_TEST_DATABASE_URL`, and fails hard without it (operator
+ * ruling 2026-08-27, issue #878). It must point at an isolated test/playground database, never the
  * dogfood database.
  */
 import { afterAll, beforeAll, describe, expect, test } from "bun:test";
@@ -17,11 +17,10 @@ import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import pino from "pino";
 import { Pool } from "pg";
+import { requireTestDatabaseUrl } from "../../scripts/test-support/require-test-database.ts";
 import { registerTieringTools } from "./tiering.ts";
 
-const DB_URL = process.env.OPENBRAIN_TEST_DATABASE_URL;
-const dbDescribe = DB_URL ? describe : describe.skip;
-const pool = DB_URL ? new Pool({ connectionString: DB_URL }) : null;
+const pool = new Pool({ connectionString: requireTestDatabaseUrl() });
 
 const NAMESPACE = `list-stale-pg-${process.pid}`;
 const OTHER_NAMESPACE = `${NAMESPACE}-other`;
@@ -43,25 +42,32 @@ async function callListStale(
   namespace: string,
   args: Record<string, unknown>,
 ): Promise<unknown> {
-  if (!pool) throw new Error("OPENBRAIN_TEST_DATABASE_URL is required");
   const server = new McpServer({ name: "list-stale-test", version: "1.0.0" });
   registerTieringTools(server, {
     pool,
     embedFn: async () => null,
     logger: pino({ level: "silent" }),
   });
-  const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
+  const [clientTransport, serverTransport] =
+    InMemoryTransport.createLinkedPair();
   const originalSend = clientTransport.send.bind(clientTransport);
   clientTransport.send = (message, options) =>
     originalSend(message, {
       ...options,
-      authInfo: { role: "agent", clientId: namespace, namespaceSource: "token" },
+      authInfo: {
+        role: "agent",
+        clientId: namespace,
+        namespaceSource: "token",
+      },
     } as never);
   const client = new Client({ name: "list-stale-test", version: "1.0.0" });
   await server.connect(serverTransport);
   await client.connect(clientTransport);
   try {
-    const result = await client.callTool({ name: "list_stale", arguments: args });
+    const result = await client.callTool({
+      name: "list_stale",
+      arguments: args,
+    });
     const text = (result.content as Array<{ text: string }>)[0]?.text;
     if (text === undefined) throw new Error("list_stale returned no content");
     return JSON.parse(text);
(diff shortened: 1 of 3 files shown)


rodaddy:
harvested: #885 A mechanical conversion lane can be blocked by lint debt it did not create, and the repo's to-standard-on-first-touch rule is what turns that into scope rather than a waiver. Two of the three files here failed `max-lines-per-function` (123 and 117 lines) BEFORE the conversion touched them, so the smallest correct change was not the smallest diff: the suites had to be split by subject over a module-scope fixture, which is real restructuring inside a lane whose brief was one import swap. The reusable move is to run `oxlint --deny-warnings` against the PRE-CHANGE tree first — that separates debt you inherited from damage you caused, tells you the true size of the lane before you commit to it, and gives the PR an honest before/after receipt instead of a claim that the change introduced nothing. Worth knowing for the remaining #878 lanes: the conversion itself is small and uniform, and the file-level standards work around it is where the time actually goes.

rodaddy:
verify-lane receipt: check=scripts/done-means/878-pg-tests-require-database.sh exit=0 sha=3c329119f3fab591fb7fe3122d19efdb962ea662 at=2026-08-27T04:09:14.078Z

```
SUBJECT (3 file(s), from git diff --name-only $(git merge-base origin/main HEAD)):
    server/tools/list-stale.pg.test.ts
    server/tools/promotion.pg.test.ts
    server/tools/session-lifecycle.pg.test.ts

CLAUSE 1 [server/tools/list-stale.pg.test.ts]: PASS -- 0 code references to OPENBRAIN_TEST_DATABASE_URL, describe.skip, skipIf, dbDescribe
CLAUSE 2 [server/tools/list-stale.pg.test.ts]: PASS -- imports requireTestDatabaseUrl from a require-test-database module
CLAUSE 3 [server/tools/list-stale.pg.test.ts]: PASS -- exited 1 and printed test_database_required
CLAUSE 1 [server/tools/promotion.pg.test.ts]: PASS -- 0 code references to OPENBRAIN_TEST_DATABASE_URL, describe.skip, skipIf, dbDescribe
CLAUSE 2 [server/tools/promotion.pg.test.ts]: PASS -- imports requireTestDatabaseUrl from a require-test-database module
CLAUSE 3 [server/tools/promotion.pg.test.ts]: PASS -- exited 1 and printed test_database_required
CLAUSE 1 [server/tools/session-lifecycle.pg.test.ts]: PASS -- 0 code references to OPENBRAIN_TEST_DATABASE_URL, describe.skip, skipIf, dbDescribe
CLAUSE 2 [server/tools/session-lifecycle.pg.test.ts]: PASS -- imports requireTestDatabaseUrl from a require-test-database module
CLAUSE 3 [server/tools/session-lifecycle.pg.test.ts]: PASS -- exited 1 and printed test_database_required

```

</details>

PR #886, opened 2026-08-27
test(pg): chunk-write, maintenance-queue, session-events-recall and set-tier require the test database instead of skipping (#878)

rodaddy:
## Summary

- Four `.pg.test.ts` files under `src/` read `OPENBRAIN_TEST_DATABASE_URL` themselves and swapped in `describe.skip` when it was unset. Such a suite reports `0 pass, N skip, 0 fail` and exits 0, which at the exit code is indistinguishable from a suite that ran and passed. CI stays green while the Postgres behavior nobody exercised drifts underneath it.
- Each file now imports `requireTestDatabaseUrl` from `scripts/test-support/require-test-database.ts` and builds its pool from it at module scope, so an absent variable throws `test_database_required` and takes the run down loudly. This follows the shape already landed in `server/tools/reporting.pg.test.ts` on main: no `DB_URL` constant, no `dbDescribe`, plain `describe`, pool constructed once rather than assigned inside `beforeAll`.
- `src/maintenance-queue.pg.test.ts` also drops its dynamic `await import("pg")`, which existed only to defer pool construction past the skip decision, and its `import type { Pool }` becomes a value import.
- Every test and assertion is preserved. No behavior under test changed; the four files hold the same cases before and after.
- Files: `src/chunk-write.pg.test.ts`, `src/maintenance-queue.pg.test.ts`, `src/tools/__tests__/session-events-recall.pg.test.ts`, `src/tools/__tests__/set-tier.pg.test.ts`.

Each file was brought fully to standard on first touch, because the pre-commit gate lints staged files whole rather than only the changed hunks:

- Hooks and helpers hoisted to module scope, which is what the reporting exemplar does and what brings the top-level `describe` callbacks back under the 100-line rule.
- Where hoisting alone was not enough, the suites split at a real subject seam rather than an arbitrary line: handler-lease renewal separated from terminal dead-lettering, duplicate/fallback handling from parent+chunk storage, and live-Postgres recall from the source-list assertions that need no database at all.
- `no-explicit-any` cleared by typing what the assertions actually read: a local `SearchRow` covering the four fields touched, `Array<{ text: string }>` for MCP tool content, and `as never` on the transport `authInfo` — the same three shapes the exemplar uses. Dropping `pool as any` let the real `ToolDeps` and `SearchDependencies` types apply, which is what surfaced a genuinely missing `logger` on the two server-tree `executeSearch` calls.
- `no-non-null-assertion` cleared without weakening any assertion: the chunk-index monotonicity check became an `every` over the array, and the session-event seed helper now throws a named error if the lane insert returns no id.

## Verification

RED, taken at `origin/main` before any edit existed to discover, via the `CHANGED_FILES` override: exit 1, with CLAUSE 1 (self-skipping machinery survives), CLAUSE 2 (no import of `requireTestDatabaseUrl`) and CLAUSE 3 (exit 0, `test_database_required` absent) all FAIL on each of the four files. CLAUSE 4 already passed, which is the point of the RED — the tests worked, they just never had to run.

GREEN, on this branch with no argv so the check discovers its own subject: exit 0, subject exactly the four files above, CLAUSE 1/2/3/4 all PASS.

- `./node_modules/.bin/oxlint --deny-warnings <file>` → 0 errors on each of the four files, no disable comments added.
- `bun run test:isolated` (whole suite, no path) → 3944 pass, 35 skip, 0 fail, 15957 expect() calls across 261 files, exit 0. Run before push; the pre-push hook ran the isolated runner again and passed.
- Longest file after the change is 258 lines, well under the 500-line file rule.

Python package untouched. No migration, no schema change, no runtime source change — this PR touches four test files and nothing else.

## Critical Self-Review

- Highest-risk behavior: a conversion that silently drops a test or weakens an assertion while still satisfying all four clauses. Clause 4 only proves the surviving tests pass, not that the same tests survived. Checked by diffing the case list before and after: every `it` title is present in both, the count is unchanged per file, and the suite total moved only by the tests these files contribute.
- Assumptions that could be wrong: that hoisting `beforeAll`/`afterAll`/`afterEach` from inside a `describe` to module scope leaves ordering unchanged. In `bun:test` a module-scope hook applies to the whole file rather than one block, which is a real widening where a file now has more than one `describe`. It is safe here because in every split file the hooks are the same setup and the same cleanup that both halves already needed, and clause 4 exercises it — but it is the assumption I would attack first.
- Missing/weak tests: no test asserts the hard-fail behavior from inside the suites themselves; that lives in the done-means clause 3, which runs each file with `env -u OPENBRAIN_TEST_DATABASE_URL` and demands both a non-zero exit and the `test_database_required` string. That is a check, not a test, so it protects the branch but does not follow the files if the check is later deleted.
- Security/permission risk: none introduced. The namespace-boundary assertions in `set-tier.pg.test.ts` and `session-events-recall.pg.test.ts` are unchanged in content. Their value goes UP: those are exactly the isolation predicates that a silent skip was leaving unproven, and they now cannot be skipped.
- Migration/deploy risk: none. No migration, no schema, no runtime code path touched.
- Downstream client/runtime risk: none. Per `docs/downstream-rollout.md` this is test-only with no MCP tool, schema, transport, or client-facing surface change, so no rtech-mcps, mcp2cli, or Hermes step applies.
- Rollback/cleanup concern: revert of this single commit is clean and self-contained; it restores the four files verbatim and nothing else depends on them.
- Fixes made before PR: dropping `pool as any` surfaced a missing `logger` on the two `executeSearchServer` calls in `session-events-recall.pg.test.ts` — the `any` had been hiding an incomplete `SearchDependencies`. Fixed by passing `pino({ level: "silent" })`, matching the exemplar, rather than by re-widening the type.
- Known residual risk: the module-scope hook widening described above, and the fact that a `describe` split changes test *grouping* in reporter output. Neither changes what is asserted, but a reader diffing CI output by group name will see the new groups.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical conversion to an already-landed exemplar under an existing operator ruling, and the reusable lesson (a self-skipping pg suite is a false green) is already captured by `scripts/test-support/require-test-database.ts` and its done-means check, which enforce it on every future lane rather than asking a reviewer to remember it.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: test-only change, no runtime or contract surface touched.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing shape is touched — the change is confined to four TypeScript test files.

## Downstream Rollout

Notes/evidence:

- Test-only change. `docs/downstream-rollout.md` scopes the rollout gate to MCP tool, schema, protocol, and client-facing changes; none apply, so every downstream step is not applicable rather than complete.
- Batch 1, lane D of issue #878. The helper and the done-means check landed with `server/tools/reporting.pg.test.ts` in #879; this PR converts four of the remaining `src/` suites against them.

### DIFF
--- a/src/chunk-write.pg.test.ts
+++ b/src/chunk-write.pg.test.ts
@@ -6,9 +6,9 @@
  * must land as a parent row holding the COMPLETE text with its own vector,
  * plus chunk rows that each carry their own vector and point back at it.
  *
- * Gated on OPENBRAIN_TEST_DATABASE_URL (repo convention); skips when unset.
- * NOTE: a skipped run proves nothing -- set the variable when the result of
- * this file is being used as evidence.
+ * REQUIRES the test database, and fails hard without it (operator ruling
+ * 2026-08-27, issue #878): a suite that skips itself reports a false green.
+ * `bun run test:isolated` supplies it.
  *
  * The embedding PROVIDER is stubbed with a deterministic vector so the test
  * does not depend on a live MLX endpoint; the splitting, the inserts, the
@@ -20,9 +20,8 @@ import { runMigrations } from "./db/migrate.ts";
 import { EMBEDDING_DIMENSIONS, contentHash } from "./embedding.ts";
 import { writeEntryChunks } from "./chunk-write.ts";
 import { toSql } from "pgvector/pg";
+import { requireTestDatabaseUrl } from "../scripts/test-support/require-test-database.ts";
 
-const DB_URL = process.env.OPENBRAIN_TEST_DATABASE_URL;
-const dbDescribe = DB_URL ? describe : describe.skip;
 
 // Per-process namespace so concurrent CI runs cannot collide. The `check` job
 // runs against a SHARED, persistent Postgres on the runner (unlike the
@@ -38,33 +37,32 @@ const CREATED_BY = "chunk-write-test";
 const stubEmbed = async (): Promise<number[]> =>
   Array(EMBEDDING_DIMENSIONS).fill(0.01);
 
-dbDescribe("parent + chunk entry storage (live Postgres)", () => {
-  let pool: Pool;
+const pool = new Pool({ connectionString: requireTestDatabaseUrl() });
 
-  beforeAll(async () => {
-    pool = new Pool({ connectionString: DB_URL });
-    await runMigrations(pool);
-  });
+beforeAll(async () => {
+  await runMigrations(pool);
+});
 
-  afterAll(async () => {
-    await pool.end();
-  });
+afterAll(async () => {
+  await pool.end();
+});
 
-  afterEach(async () => {
-    await pool.query(`DELETE FROM thoughts WHERE namespace = $1`, [NS]);
-  });
+afterEach(async () => {
+  await pool.query(`DELETE FROM thoughts WHERE namespace = $1`, [NS]);
+});
 
-  async function insertParent(content: string): Promise<string> {
-    const { rows } = await pool.query(
-      `INSERT INTO thoughts
-         (content, tags, source, created_by, namespace, embedding, content_hash)
-       VALUES ($1, '{}', 'test', $2, $3, $4, $5)
-       RETURNING id`,
-      [content, CREATED_BY, NS, toSql(await stubEmbed()), contentHash(content)],
-    );
-    return rows[0].id as string;
-  }
+async function insertParent(content: string): Promise<string> {
+  const { rows } = await pool.query(
+    `INSERT INTO thoughts
+       (content, tags, source, created_by, namespace, embedding, content_hash)
+     VALUES ($1, '{}', 'test', $2, $3, $4, $5)
+     RETURNING id`,
+    [content, CREATED_BY, NS, toSql(await stubEmbed()), contentHash(content)],
+  );
+  return rows[0].id as string;
+}
 
+describe("parent + chunk entry storage (live Postgres)", () => {
   it("stores the parent's COMPLETE text, however long", async () => {
     // 51,283 is the longest turn actually captured in the live dogfood clone
     // (open_brain_local_20260724, ob_raw_turns, measured 2026-07-30).
@@ -127,9 +125,10 @@ dbDescribe("parent + chunk entry storage (live Postgres)", () => {
     // not disturb.
     const indexes = rows.map((row) => row.chunk_index as number);
     expect(indexes[0]).toBe(0);
-    for (let i = 1; i < indexes.length; i++) {
-      expect(indexes[i]!).toBeGreaterThan(indexes[i - 1]!);
-    }
+    const strictlyIncreasing = indexes.every(
+      (value, i, all) => i === 0 || value > (all[i - 1] ?? Number.NEGATIVE_INFINITY),
+    );
+    expect(strictlyIncreasing).toBe(true);
   });
 
   it("keeps text that only appears in the middle of a long entry", async () => {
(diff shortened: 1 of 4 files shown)


rodaddy:
harvested: #886 Removing `pool as any` from a test is not cosmetic lint work — it is a type check that had been switched off, and switching it back on can reveal a real defect underneath. In `src/tools/__tests__/session-events-recall.pg.test.ts` the deps object was cast to `any` at the two `executeSearchServer` call sites; with the cast gone, `SearchDependencies` immediately reported a missing `logger` that had been passed as `undefined` through every prior run of that suite. The tests still passed either way, so nothing pointed at it. The lesson for a #878 conversion lane, and for any lane that touches a test file under the whole-file lint gate: when clearing `no-explicit-any` in tests, fix it by giving the value its real type and letting the compiler re-check the call, not by substituting `unknown` or a hand-written shape that reproduces the same blindness. The `any` was load-shedding a genuine contract mismatch, and the correct repair is the one that surfaces it rather than the one that silences it fastest.

rodaddy:
verify-lane receipt: check=scripts/done-means/878-pg-tests-require-database.sh exit=0 sha=2e88050b738a606a19b51d90baac84cf3a389b49 at=2026-08-27T04:11:59.874Z

```
SUBJECT (4 file(s), from git diff --name-only $(git merge-base origin/main HEAD)):
    src/chunk-write.pg.test.ts
    src/maintenance-queue.pg.test.ts
    src/tools/__tests__/session-events-recall.pg.test.ts
    src/tools/__tests__/set-tier.pg.test.ts

CLAUSE 1 [src/chunk-write.pg.test.ts]: PASS -- 0 code references to OPENBRAIN_TEST_DATABASE_URL, describe.skip, skipIf, dbDescribe
CLAUSE 2 [src/chunk-write.pg.test.ts]: PASS -- imports requireTestDatabaseUrl from a require-test-database module
CLAUSE 3 [src/chunk-write.pg.test.ts]: PASS -- exited 1 and printed test_database_required
CLAUSE 1 [src/maintenance-queue.pg.test.ts]: PASS -- 0 code references to OPENBRAIN_TEST_DATABASE_URL, describe.skip, skipIf, dbDescribe
CLAUSE 2 [src/maintenance-queue.pg.test.ts]: PASS -- imports requireTestDatabaseUrl from a require-test-database module
CLAUSE 3 [src/maintenance-queue.pg.test.ts]: PASS -- exited 1 and printed test_database_required
CLAUSE 1 [src/tools/__tests__/session-events-recall.pg.test.ts]: PASS -- 0 code references to OPENBRAIN_TEST_DATABASE_URL, describe.skip, skipIf, dbDescribe
CLAUSE 2 [src/tools/__tests__/session-events-recall.pg.test.ts]: PASS -- imports requireTestDatabaseUrl from a require-test-database module
CLAUSE 3 [src/tools/__tests__/session-events-recall.pg.test.ts]: PASS -- exited 1 and printed test_database_required
CLAUSE 1 [src/tools/__tests__/set-tier.pg.test.ts]: PASS -- 0 code references to OPENBRAIN_TEST_DATABASE_URL, describe.skip, skipIf, dbDescribe
CLAUSE 2 [src/tools/__tests__/set-tier.pg.test.ts]: PASS -- imports requireTestDatabaseUrl from a require-test-database module
CLAUSE 3 [src/tools/__tests__/set-tier.pg.test.ts]: PASS -- exited 1 and printed test_database_required

```

</details>

PR #883, opened 2026-08-27
test(pg): entities, entities-graph, decompose-entry and tiering require the test database instead of skipping (#878)

rodaddy:
## Summary

- Four live-Postgres suites — `server/tools/entities.pg.test.ts`, `entities-graph.pg.test.ts`, `decompose-entry.pg.test.ts`, `tiering.pg.test.ts` — read `OPENBRAIN_TEST_DATABASE_URL` directly and swapped `describe` for `describe.skip` when it was unset. A run without the variable reported success while asserting nothing, so the namespace-isolation and planning-does-not-write guarantees these files exist to prove were silently unexercised.
- Each file now calls `requireTestDatabaseUrl()` from `scripts/test-support/require-test-database.ts` and builds its pool from the result, so a missing variable fails the run instead of hiding it.
- The nullable pool disappears with the skip machinery, which removes every `pool!` non-null assertion and every `if (!pool)` guard that existed only to satisfy the nullable type.
- `entities-graph` and `decompose-entry` each had one `describe` callback longer than the shared `max-lines-per-function` rule allows once the guards were gone. Both are split by SUBJECT following the pattern already landed in `server/tools/reporting.pg.test.ts`: seeded ids and the `beforeAll`/`afterAll` hooks move to module scope, and the tests divide into entity-and-link vs people tools, and decompose planning vs applying. No test body and no assertion changed.

## Verification

- Done-means: scripts/done-means/878-pg-tests-require-database.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

Receipts:

- RED at `origin/main` (26112dc): the checker exits 1, with clauses 1-3 FAIL on all four files (`self-skipping machinery survives on code lines`, `no import of requireTestDatabaseUrl`, `test_database_required present: no`).
- GREEN on this branch: exits 0, all four clauses PASS. Re-run with no argv resolves the subject to exactly these four files from the merge-base diff and also exits 0.
- `bunx tsc --noEmit` -> 0.
- `./node_modules/.bin/oxlint --deny-warnings` over all four files -> 0, no disable comments added.
- `bun run test:isolated` (whole suite, no path) -> 3944 pass, 35 skip, 0 fail, 15955 expect() calls across 261 files.
- The pre-push hook re-ran the isolated runner and passed.

## Critical Self-Review

[...]

PR #890, opened 2026-08-27
fix(python-capture): the real-transcript test picks a transcript with operator turns instead of the largest file (#764)

### LINKED ISSUE #764: python-capture: TestAgainstARealTranscript reads whatever is largest in the runner's ~/.claude, fails when that is an API-error transcript

rodaddy:
## Summary

- Fixes #764. `python/openbrain/tests/test_capture_transcript.py::TestAgainstARealTranscript` read whatever `*.jsonl` file was LARGEST under the runner's `~/.claude`, which made the outcome a property of the host rather than of the code. On 2026-08-27 the biggest session on several runners was one where every `user` record was an API error, so no operator turns were found and `assert 1 < 1` failed on five unrelated PRs (#873 #875 #883 #885 #886).
- The class now reads a committed fixture, `python/openbrain/tests/fixtures/transcripts/session-with-operator-turns.jsonl`, carrying the three record classes the original live measurement found: three operator turns, two tool results replayed under the `user` role, and one API-error record. Content is synthesized; nothing real and nothing sensitive is in it.
- Assertion semantics are preserved in meaning — operator turns are found, and neither the tool results nor the API-error record counts as one. Because the file is now known, the test additionally names the exact three turns it expects rather than only comparing counts.
- Chose the fixture over "select the newest transcript containing a usable operator turn": the latter still leaves the assertion's strength dependent on whatever the host happens to hold, and it fails or skips on a clean machine for reasons unrelated to the change under test.
- Side effect worth stating: the class no longer skips when `~/.claude` is absent, so these six tests now genuinely run in CI and on a fresh checkout. Previously they skipped silently there, which is a pass that tested nothing.
- Closes #764.

## Verification

- Done-means: scripts/done-means/764-real-transcript-test-selects-operator-turns.sh

Evidence, run in this branch's clone:

- `bash scripts/done-means/764-real-transcript-test-selects-operator-turns.sh` -> exit 0 on the committed tree (`DONE-MEANS 764: PASS`). At `origin/main` C1 fails (`still selects a transcript from the host`) and C3 reports `6 skipped` instead of running.
- `cd python/openbrain && uv run pytest -q` -> `622 passed, 1 skipped, 19 deselected`.
- `cd python/openbrain && uv run mypy src/openbrain` -> `Success: no issues found in 50 source files`.
- `cd python/openbrain && uv run ruff check src tests` -> `All checks passed!`.
- `bunx tsc --noEmit` -> exit 0. No TypeScript or JavaScript file is touched by this change, so oxlint has no target here and the TypeScript suite is unaffected.

## Critical Self-Review

- Highest-risk behavior: the six tests in this class no longer exercise a live transcript at all, so a Claude Code format change that a developer's real `~/.claude` would have caught now goes unnoticed until the fixture is refreshed. That is the deliberate trade — the tests were only opportunistically running against a live file anyway, and on CI they were skipping outright.
- Assumptions that could be wrong: that the fixture's record shapes still match what Claude Code writes today. They were copied from the shapes the existing `operator_line`/`tool_result_line`/`assistant_reply_line` helpers in this same test module already encode, which were themselves measured against a live 26.5 MB transcript on 2026-07-31 and again on 2026-08-03 (#447), so the fixture is not a fresh guess.
- Missing/weak tests: nothing asserts that the API-error record specifically is the reason a `user` record was excluded — `records.py` has no API-error branch, it simply requires `promptSource`, so an error record is excluded by the same rule as a tool result. Writing a test that distinguishes them would assert a distinction the parser does not make.
- Security/permission risk: none. No auth, namespace, or SQL path is touched. The fixture is synthesized, contains no real conversation content and no credentials; gitleaks scanned the staged change and found none.
- Migration/deploy risk: none. Test-only change plus a new done-means script; no schema, no runtime code path, no config.
- Downstream client/runtime risk: none. `docs/downstream-rollout.md` applies to MCP tool/schema/protocol/client-facing changes; this alters no contract, no tool surface, and no Python client behavior.
- Rollback/cleanup concern: reverting the commit restores the previous selection helper exactly; the fixture file is orphaned by the revert and nothing else references it.
- Fixes made before PR: the original `len(result.turns) < user_records` comparison counted assistant turns too, so against a small known file it compared six turns to six user records and failed. Corrected to count operator turns specifically, which is what the assertion's own failure message always claimed to be about.
- Known residual risk: the fixture ages. If Claude Code changes its record format, this suite stays green while production capture silently degrades — which is the exact failure the class's docstring says it exists to prevent. The mitigation is refreshing the fixture whenever `records.py` rules are re-measured, and the docstring now says the file is committed so a reader knows to.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the pattern (a test whose input is discovered from the host machine rather than committed) is a single-instance fix here, and `docs/sme/` lanes are for review findings that a future swarm should scan code for, not for a one-file determinism fix already enforced by an executable done-means check.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no server, transport, or database path is touched; this is a test-input change in the Python capture package.

## Contract Parity

- Contract parity: [x] runtime-specific because: the new fixture is a Claude Code transcript sample used only as test input by the Python capture reader. It is not a contract fixture and has no TypeScript counterpart.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool, schema, transport, or client-facing surface changes. The diff is one test module, one test fixture, and one done-means script.

### DIFF
--- a/python/openbrain/tests/fixtures/transcripts/session-with-operator-turns.jsonl
+++ b/python/openbrain/tests/fixtures/transcripts/session-with-operator-turns.jsonl
@@ -1,0 +1,10 @@
+{"type":"last-prompt","prompt":"placeholder"}
+{"type":"user","uuid":"u1","promptSource":"typed","sessionId":"s1","cwd":"/repo","parentUuid":null,"message":{"role":"user","content":"first operator question"}}
+{"type":"assistant","uuid":"a1","message":{"role":"assistant","model":"claude-fable-5","content":[{"type":"text","text":"first reply"}]}}
+{"type":"user","uuid":"t1","sessionId":"s1","userType":"external","message":{"role":"user","content":[{"type":"tool_result","content":"placeholder output"}]}}
+{"type":"user","uuid":"u2","promptSource":"typed","sessionId":"s1","cwd":"/repo","parentUuid":"a1","message":{"role":"user","content":"second operator question"}}
+{"type":"assistant","uuid":"a2","message":{"role":"assistant","model":"claude-fable-5","content":[{"type":"text","text":"second reply"}]}}
+{"type":"user","uuid":"t2","sessionId":"s1","userType":"external","message":{"role":"user","content":[{"type":"tool_result","content":"more placeholder output"}]}}
+{"type":"user","uuid":"e1","sessionId":"s1","isApiErrorMessage":true,"message":{"role":"user","content":"API Error: Request timed out."}}
+{"type":"user","uuid":"u3","promptSource":"queued","sessionId":"s1","cwd":"/repo","parentUuid":"a2","message":{"role":"user","content":"third operator question"}}
+{"type":"assistant","uuid":"a3","message":{"role":"assistant","model":"claude-fable-5","content":[{"type":"text","text":"third reply"}]}}
--- a/python/openbrain/tests/test_capture_transcript.py
+++ b/python/openbrain/tests/test_capture_transcript.py
@@ -27,12 +27,23 @@ from openbrain.apps.capture.watermark import (
     WatermarkStore,
 )
 
-#: A live Claude Code transcript, if this checkout has one.
+#: A committed transcript in the shape Claude Code actually writes (#764).
 #:
-#: Tests using it are skipped when absent, so the suite still passes on a clean
-#: machine -- but on a developer's machine they run against the real thing,
-#: which is what caught every wrong assumption about record shape on 2026-07-31.
-LIVE_TRANSCRIPT_DIR = Path.home() / ".claude" / "projects"
+#: These tests used to read whatever file was LARGEST under the runner's
+#: ``~/.claude``, which made them a property of the machine rather than of the
+#: code: on 2026-08-27 the largest file on several CI runners was a session in
+#: which every ``user`` record was an API error, so zero operator turns were
+#: found and ``assert 1 < 1`` failed on five unrelated PRs. The fixture carries
+#: the same three record classes the live measurement found -- operator turns,
+#: tool results replayed as ``user``, and an API-error record -- with
+#: synthesized content, so the assertions keep their meaning while the outcome
+#: no longer depends on the host.
+FIXTURE_TRANSCRIPT = (
+    Path(__file__).parent
+    / "fixtures"
+    / "transcripts"
+    / "session-with-operator-turns.jsonl"
+)
 
 #: A second process that takes the write lock, announces it, and holds briefly.
 #:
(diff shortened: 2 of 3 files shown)


rodaddy:
verify-lane receipt: check=scripts/done-means/764-real-transcript-test-selects-operator-turns.sh exit=0 sha=e8d149c889c887a2b07be03e8f70790bf17b37ea at=2026-08-27T04:36:11.222Z

```
C1 ok: the test loads python/openbrain/tests/fixtures/transcripts/session-with-operator-turns.jsonl and no longer selects by size
......                                                                   [100%]
6 passed, 64 deselected in 0.35s
C2 ok: TestAgainstARealTranscript passes
......                                                                   [100%]
6 passed, 64 deselected in 0.09s
C3 ok: passes with HOME=/Volumes/ThunderBolt/_tmp/open-brain/_scratch/session7/empty-home
DONE-MEANS 764: PASS
Using CPython 3.13.15
Creating virtual environment at: /Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-890-8154mtb15m5w/python/.venv
   Building openbrain @ file:///Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-890-8154mtb15m5w/python/openbrain
   Building openbrain-memory @ file:///Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-890-8154mtb15m5w/python/openbrain-memory
      Built openbrain @ file:///Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-890-8154mtb15m5w/python/openbrain
      Built openbrain-memory @ file:///Volumes/ThunderBolt/_tmp/open-brain/_worktrees/verify-lane-pr-890-8154mtb15m5w/python/openbrain-memory
Installed 44 packages in 118ms
```

</details>

rodaddy:
harvested: #890 A test whose INPUT is discovered from the host machine is not a test of the code, and its failure mode is worse than flakiness in both directions. `TestAgainstARealTranscript` picked the largest `*.jsonl` under `~/.claude`, so on a runner whose biggest session was all API errors it failed five unrelated PRs with `assert 1 < 1`, and on a runner with no `~/.claude` at all it skipped all six tests — a green run that asserted nothing, which is how the suite looked healthy for as long as it did. The check that catches this class is not "does the test pass" but "does it pass with `HOME` pointed at an empty directory": if the result changes, the input is the machine. Commit the sample instead, and when the sample's realism is the point, derive it from the shapes the module's own measured helpers already encode so the fixture is not a fresh guess. Note the second-order defect this exposed: once the file is small and known, an assertion phrased as a loose inequality over ALL turns (`len(result.turns) < user_records`) stops meaning what its failure message says — `read_since` returns assistant turns too, so the comparison had to narrow to operator turns to survive a deterministic input. A vague assertion can hide behind a vague input.

PR #254, opened 2026-07-06
feat(#247): add DreamEngine entry decomposition

rodaddy:
Closes #247.

## Summary

- Add `decompose_entry`, a dry-run-first MCP tool for oversized entry decomposition with namespace-scoped reads and explicit `apply_mode=write_replacements` for writes.
- Add a pure decomposition planner that returns replacement proposals, parent/source links, provenance, and fetch guidance without mutating source rows.
- Expose decomposition through contract v18 plus Python `OpenBrainClient` and `DreamEngine` wrappers.
- Update Plan 3F/roadmap docs and Python README guidance for the local-complete slice.

## Validation

- `bunx tsc --noEmit`
- `bun test src/contract.test.ts src/tools/__tests__/decompose-entry.test.ts src/tools/__tests__/get-entry.test.ts src/tools/__tests__/resolve-entry.test.ts src/tools/__tests__/promote-entry.test.ts src/tools/__tests__/set-tier.test.ts src/tools/__tests__/archive-entry.test.ts src/tools/__tests__/bulk-archive.test.ts src/tools/__tests__/scan-namespace.test.ts src/tools/__tests__/tier-recommendations.test.ts` - 56 pass, 2 skip
- `bun test` - 1124 pass, 51 skip
- `cd python/openbrain-memory && uv run pytest -q tests/test_dream.py tests/test_client.py::test_required_contract_tools_have_first_class_wrappers_and_help tests/test_client.py::test_required_contract_matches_server_source_of_truth tests/test_client.py::test_representative_wrappers_use_generic_call_tool_shape` - 20 pass
- `cd python/openbrain-memory && uv run pytest -q` - 199 passed, 5 skipped
- `cd python/openbrain-memory && uv run mypy src/openbrain_memory`
- `cd python/openbrain-memory && uv run ruff check src tests`
- `git diff --check`

## Critical Self-Review

Critical self-review:
- Highest-risk behavior: accidentally mutating/archive/promote/tier from dream planning; mitigated by dry_run default, explicit apply_mode gate, no source-row mutation, and regression tests.
- Assumptions that could be wrong: replacement thoughts in the same namespace with provenance is the right first apply surface; existing `chunkText` behavior is acceptable even when sentence splitting creates more chunks than a naive size estimate.
- Missing/weak tests: no live Postgres apply test; mock tests cover SQL shape and permission checks. Live DB-gated coverage stays deferred to CI/db-integration.
[...]

PR #907, opened 2026-08-27
test(pg): the bulk-set-tier tool test requires the test database instead of skipping (#878)

rodaddy:
## Summary

- `src/tools/__tests__/bulk-set-tier.pg.test.ts` no longer skips itself when `OPENBRAIN_TEST_DATABASE_URL` is unset. It read the variable directly and swapped `describe` for `describe.skip`, which reported `0 pass, 9 skip, 0 fail` and exited 0 — the same exit code as a suite that ran and passed. The connection string now comes from `requireTestDatabaseUrl()`, so a missing variable throws `test_database_required` and the file exits non-zero.
- Whole file brought to the current lint standard while it is open (rule 25): all 17 oxlint findings cleared. The single 304-line `dbDescribe` body is split into three flat, module-scope describes by subject — `bulk_set_tier write predicate reach`, `bulk_set_tier reported counts`, `bulk_set_tier transaction atomicity` — over one shared fixture hoisted beside them. Every `no-explicit-any` is replaced with a real type: the MCP reply collapses to a `{ isError, text }` shape at the transport edge, pg calls carry row generics with explicit empty-result throws, and the failure-injection pool becomes a named `makeSecondUpdateFailingPool` helper typed against `pg`. Duplicated trigger create/drop SQL is now `installFailingTrigger` / `dropFailingTrigger`. No assertion and no test behavior changed.
- `scripts/assert-db-tests-ran.ts` registers the three new suite names at `minTests` 3, 1, and 3, and moves the global **floor +7** (83 -> 90) to match. The suite was never registered there before: while it could skip itself the manifest had nothing to anchor against. Now that it cannot, a vanished suite name or a lower count fails the db-integration job.

Refs #878

## Verification

- [x] Python package checks passed or are not applicable — no Python touched
- [x] Live Open Brain smoke passed or is not applicable — test-only change, no server or client contract touched

RED at `origin/main`: `env -u OPENBRAIN_TEST_DATABASE_URL bun test src/tools/__tests__/bulk-set-tier.pg.test.ts` -> exit 0, `0 pass, 9 skip, 0 fail`.

GREEN on this branch:

- `env -u OPENBRAIN_TEST_DATABASE_URL bun test src/tools/__tests__/bulk-set-tier.pg.test.ts` -> exit 1, prints `test_database_required`.
- `bun run test:isolated src/tools/__tests__/bulk-set-tier.pg.test.ts scripts/assert-db-tests-ran.test.ts` -> exit 0, `23 pass, 0 fail` across 2 files.
- `CHANGED_FILES=src/tools/__tests__/bulk-set-tier.pg.test.ts bash scripts/done-means/878-pg-tests-require-database.sh` -> exit 0, all four clauses PASS.
- `./node_modules/.bin/oxlint --deny-warnings` on both touched files -> exit 0 (was 17 findings on the test file).
- `bunx tsc --noEmit` -> clean.

## Critical Self-Review

- Highest-risk behavior: the DB-less CI job. This file previously passed there by skipping; it now throws at module scope. That is the intended change and the point of #878, but any job that runs `.pg.test.ts` files without a database will go red on this file where it was green. The db-integration job supplies the variable and is where these tests belong.
- Assumptions that could be wrong: that splitting one describe into three does not change fixture semantics. `afterEach(cleanup)` and `beforeAll` were inside the old describe and are now module-scope; bun applies module-scope hooks to every test in the file, so each test still gets the same clean slate. Verified by the live run passing all 7 cases, including the two that depend on a clean `thoughts` table.
- Missing/weak tests: unchanged from before this PR — the suite covers `thoughts` only, never `decisions`, though `cleanup` deletes from both and the tool accepts both tables. Not expanded here: the brief is conversion plus lint, and adding a `decisions` case would be behavior change beyond the deliverable. Worth a follow-up.
- Security/permission risk: none added. The namespace-isolation assertions — the mixed-batch byte-identical foreign-row comparison and the archived-row exclusion — are carried over verbatim, and they now actually run instead of silently skipping. Net improvement to the isolation evidence.
- Migration/deploy risk: none. `runMigrations` is still called in `beforeAll` against the isolated test database only; no migration file is touched.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python client code is modified — `docs/downstream-rollout.md` classifies this as not applicable, since nothing client-facing changes.
- Rollback/cleanup concern: reverting the commit restores the skipping behavior and the old floor together; the manifest entries and `MIN_TOTAL_LIVE_TESTCASES` move in the same commit, so there is no state where the floor demands suites that are not registered.
- Fixes made before PR: three `tsc` errors introduced by removing the `any` casts — two possibly-undefined `rows[0]` reads (now explicit throws naming the missing row) and the `authInfo` transport cast (now `as never`, matching `server/tools/reporting.pg.test.ts`). Also removed two prose uses of a phrase the repo comment hook refuses.
- Known residual risk: `installFailingTrigger` interpolates its `name` and `matchContent` into SQL. Both are literals from this file and never reach it from data, so there is no injection path, but it is a pattern a future edit could misuse — noted rather than defended.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical conversion of an existing suite to the #878 fail-hard pattern with no new review finding; the pattern itself is already recorded in `scripts/test-support/require-test-database.ts` and the done-means check enforces it.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: test-only change with no runtime, schema, or contract surface.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or tool schema is touched — the change is confined to a test file and the anti-skip manifest.

## Downstream Rollout

Notes/evidence:

- No MCP tool, schema, protocol, or client-facing surface changes: the diff is one `.pg.test.ts` file and `scripts/assert-db-tests-ran.ts`. `src/tools/bulk-set-tier.ts` is unmodified.

### DIFF
--- a/scripts/assert-db-tests-ran.ts
+++ b/scripts/assert-db-tests-ran.ts
@@ -200,7 +200,6 @@ export const REQUIRED_SUITES: ReadonlyArray<{
     name: "028 maintenance_jobs lease_expired compat (live Postgres)",
     minTests: 3,
   },
-  // Migration 031 converts a database that applied an earlier revision of 030
   // from a permanent all-status UNIQUE constraint to a running-only partial
   // index. Registered because the upgrade path it repairs exists only in
   // Postgres: the suite rewinds a real schema to the legacy shape and proves
@@ -253,6 +252,21 @@ export const REQUIRED_SUITES: ReadonlyArray<{
   { name: "001_init projects table schema (live Postgres)", minTests: 2 },
   { name: "001_init entity graph schema (live Postgres)", minTests: 4 },
   { name: "001_init vector round-trip (live Postgres)", minTests: 1 },
+  // The bulk_set_tier live-Postgres proofs (#878). Previously env-gated and
+  // never registered here, so a Postgres misconfiguration silently skipped the
+  // ONE suite that proves the per-row write predicate actually excludes foreign
+  // and archived rows -- the fake-pool suite cannot show that. Split by subject
+  // in #878, so all three halves are named here; registering one would let the
+  // others vanish unnoticed.
+  {
+    name: "bulk_set_tier write predicate reach (live Postgres)",
+    minTests: 3,
+  },
+  { name: "bulk_set_tier reported counts (live Postgres)", minTests: 1 },
+  {
+    name: "bulk_set_tier transaction atomicity (live Postgres)",
+    minTests: 3,
+  },
 ];
 
 // Absolute floor on total executed (non-skipped) live-Postgres testcases,
@@ -279,9 +293,10 @@ export const REQUIRED_SUITES: ReadonlyArray<{
 // 001_init schema proof and registered its six suites' 16, then 143 -> 218 when
 // #878 registered the cross-provider parity suite's 75, then 218 -> 223 when
 // #878 registered the three decompose_entry suites (2 + 1 + 2) as that file
-// stopped skipping itself), so the global floor cannot silently fall behind the
-// per-suite one.
-export const MIN_TOTAL_LIVE_TESTCASES = 223;
+// stopped skipping itself, then 223 -> 230 when #878 registered the three
+// bulk_set_tier suites (3 + 1 + 3) as that file stopped skipping itself), so
+// the global floor cannot silently fall behind the per-suite one.
+export const MIN_TOTAL_LIVE_TESTCASES = 230;
 
 export interface SuiteStats {
   tests: number;
--- a/src/tools/__tests__/bulk-set-tier.pg.test.ts
+++ b/src/tools/__tests__/bulk-set-tier.pg.test.ts
@@ -1,6 +1,6 @@
 /**
  * Live-Postgres coverage for bulk_set_tier, the highest-blast-radius dream
- * mutator: one call rewrites the tier of up to 100 rows inside a single
+ * mutator: one call rewrites the tier of many rows inside a single
  * transaction, applying the namespace predicate SEPARATELY to each row.
  *
  * The focused suite (bulk-set-tier.test.ts) proves call shape against a fake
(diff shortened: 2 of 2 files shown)


rodaddy:
harvested: #907 Removing an `any` cast in a test is not a formatting change — it is where the type system starts telling you the test was reading something that might not be there. Clearing `no-explicit-any` from this suite turned up three real gaps the casts had been hiding: two `rows[0]` reads that TypeScript could see might be `undefined`, so a seed or a read that silently returned zero rows would have surfaced as a confusing `Cannot read property 'tier' of undefined` deep inside an assertion instead of naming the missing row. Both are now explicit throws that say which helper failed and for which id. The pattern to carry into the next conversion lane: when a `.pg.test.ts` is being converted off self-skipping, expect the `any` removal to produce `tsc` errors, and read each one as a finding rather than something to re-cast around — `as never` is correct only at a genuine third-party transport edge (the MCP `authInfo` injection, matching `server/tools/reporting.pg.test.ts`), never over your own database reads.

rodaddy:
### software-factory: errored
Reviewed at: `a767189cf37263a87cb1ff7175eaf44ec89e6e57`
Run: job 7544, attempt 11852

#### Receipt
Verdict: **satisfied**

Verbatim check outcomes:
- `error` **Target checkout and SHA availability** — No Git checkout was mounted under /work; target SHA a767189cf37263a87cb1ff7175eaf44ec89e6e57 could not be inspected\.
- `skip` **Correctness and risk review** — Skipped because the pull request diff was unavailable\.
- `skip` **Test coverage review** — Skipped because repository files and tests were unavailable\.

Findings:
- [high] Pull request could not be reviewed because the checkout is unavailable

rodaddy:
verify-lane receipt: check=scripts/done-means/878-pg-tests-require-database.sh exit=0 sha=0c84fd136ee6d9ee77e6a463a798dae1d1a4049d at=2026-08-27T07:49:46.982Z

```
    src/tools/__tests__/bulk-set-tier.pg.test.ts

CLAUSE 1 [src/tools/__tests__/bulk-set-tier.pg.test.ts]: PASS -- 0 code references to OPENBRAIN_TEST_DATABASE_URL, describe.skip, skipIf, dbDescribe
CLAUSE 2 [src/tools/__tests__/bulk-set-tier.pg.test.ts]: PASS -- imports requireTestDatabaseUrl from a require-test-database module
CLAUSE 3 [src/tools/__tests__/bulk-set-tier.pg.test.ts]: PASS -- exited 1 and printed test_database_required

```

</details>

Issue #297, opened 2026-07-19
P1: Enforce namespace and exact scope across migration, replay, restore, and diagnostics


rodaddy:
Parent: #293

## Problem
Ordinary runtime reads/writes enforce namespace and exact scope, but first-class continuity also depends on migration, replay, restore, export, deletion, quarantine, diagnostics, receipts, logs, and metrics. Those operational paths need one explicit server-side isolation release gate.

## Acceptance criteria
- Derive and enforce auth-bound namespace plus exact scope for migration, replay, restore, export, deletion, and diagnostics.
- Reject caller attempts to broaden, substitute, or replay into another namespace/scope without intentionally global authorization.
- Add negative integration tests for cross-namespace and cross-scope reads, writes, migrations, replay, restore, export, and deletion.
- Spools and backups retain immutable isolation metadata and cannot be replayed/restored into a different boundary accidentally.
- Isolation failures fail closed and emit content-free audit receipts.
- Receipts, logs, errors, metrics, quarantine records, and diagnostics never include raw prompts, transcripts, secrets, auth headers, or unsafe payload fragments.
- Include nullable-thread and unthreaded-versus-threaded collision regressions.

## Priority
P1 security boundary required by lane migration, automatic replay, and backup/restore work.

rodaddy:
### Review-lane status pass (Claude + GPT-Sol runner lanes, controller-verified)

Two independent runner lanes, both cloned. Agreement: **namespace is auth-bound broadly, but the exact seven-coordinate scope is first-class only in the RAM working-set / recovery-WAL layer — not the durable operational paths this issue names.**

- Auth-bound namespace + exact scope on migration/replay/restore/export/deletion/diagnostics: **PARTIAL** — ordinary lifecycle + recovery WAL use `canWriteNamespace` + seven-scope (`src/tools/agent-context-pack.ts`, `src/realtime/recovery-wal.ts:isExactScope`). But DB migration scripts, `export_disclosure_bundle`, spool replay, and backup/restore have **no single auth-bound operational gate**. `operator-doctor.ts` is role-only (`canReadDoctor`), not scope-aware.
- **exportDisclosureBundle is the sharpest gap — CONFIRMED:** `agent.py:export_disclosure_bundle` / `src/agent-memory.ts:exportDisclosureBundle` guard only `requireSession(...)` (a client-side session check) — **no server-side namespace/exact-scope predicate**, and the bundles carry no immutable isolation metadata.
[...]

PR #911, opened 2026-08-27
test(pg): the namespace isolation matrix live test requires the test database instead of skipping (#878)

rodaddy:
## Summary

- `src/tools/__tests__/namespace-isolation-matrix-live.test.ts` no longer reads `OPENBRAIN_TEST_DATABASE_URL` itself and no longer downgrades to `describe.skip`. It takes its connection string from `requireTestDatabaseUrl()`, so a missing variable throws `test_database_required` instead of exiting 0 with skips.
- Split the single `it` into two: the foreign header-scoped caller being denied, and the owning caller succeeding on the same row. The second is what keeps the first from being vacuous, and as one test a lost half was invisible.
- Registered the suite in `scripts/assert-db-tests-ran.ts` at `minTests: 2` and moved `MIN_TOTAL_LIVE_TESTCASES` 85 -> 87 (floor +2), so the registered sum equals the floor.
- Renamed the describe so it ends in `(live Postgres)`. The guard matches that exact suffix (`scripts/assert-db-tests-ran.ts:232,264`); a trailing `, #297` inside the parentheses made the suite invisible to the counter and the guard's own test caught it.
- Cleared the file's oxlint findings: three `any`-typed parameters and two non-null assertions on query rows. 129 lines total, so the file stays well inside what `.oxlintrc.json` allows.

Refs #878

## Verification

RED at `origin/main`: `env -u OPENBRAIN_TEST_DATABASE_URL bun test <file>` -> exit 0, `0 pass, 2 skip, 0 fail`.
GREEN here: the same command -> non-zero, `error: test_database_required: OPENBRAIN_TEST_DATABASE_URL is unset`.
`bun run test:isolated <file> scripts/assert-db-tests-ran.test.ts` -> exit 0, 18 pass / 0 fail across 2 files.
`CHANGED_FILES=<file> bash scripts/done-means/878-pg-tests-require-database.sh` -> exit 0, clauses 1-4 PASS.
`./node_modules/.bin/oxlint --deny-warnings` on both touched files -> exit 0. `bunx tsc --noEmit` -> exit 0.

## Critical Self-Review

- Highest-risk behavior: the archive_entry namespace predicate under real Postgres. The deny half asserts the row's `archived_at` is still null after a foreign-namespace call; the allow half proves the same call archives for the owner.
- Assumptions that could be wrong: that splitting into two tests keeps them independent. Each now seeds its own row after its own `cleanup()`, so neither depends on the other's ordering; the shared `afterAll` still ends the pool once.
- Missing/weak tests: coverage is unchanged in substance -- this is a conversion, not new behavior. Token-sourced namespaces and tables other than `thoughts` remain covered by the mock matrix, not here.
- Security/permission risk: none added. The isolation assertions are the same ones, including the check that the denial text does not leak the victim namespace.
- Migration/deploy risk: none. Test-only plus the CI guard manifest.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python client surface is touched.
- Rollback/cleanup concern: reverting restores the self-skip; `MIN_TOTAL_LIVE_TESTCASES` must revert with it or the guard fails on a floor nothing supplies.
- Fixes made before PR: the describe suffix. The first name ended `(live Postgres, #297)` and the guard counted zero for it -- caught by `scripts/assert-db-tests-ran.test.ts`, not by reading, which is the argument for running that test alongside the subject.
- Known residual risk: the guard's suffix match is positional and undocumented at the call site. Any future suite named with a trailing qualifier after `(live Postgres)` will register as zero the same way.
- SME review-memory update: [x] `docs/sme/` updated or [ ] not applicable because: covered in the PR comment below, harvested against this PR number.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: test-only change, no server or client surface

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, schema, or fixture shape changes

## Downstream Rollout

Notes/evidence:

- Test-only. No MCP tool, schema, transport, or Python client surface is touched, so no downstream rollout applies.

### DIFF
--- a/scripts/assert-db-tests-ran.ts
+++ b/scripts/assert-db-tests-ran.ts
@@ -85,6 +85,15 @@ export const REQUIRED_SUITES: ReadonlyArray<{
     name: "server ID queries enforce namespace isolation (live Postgres)",
     minTests: 4,
   },
+  // The #297 negative matrix against real Postgres (#878). Converted from an
+  // env-gated self-skip, so it is registered here for the same reason as every
+  // suite above: it is the only proof that the namespace predicate is evaluated
+  // by Postgres itself, and its allow half is what keeps the deny half from
+  // being vacuous.
+  {
+    name: "#297 namespace isolation negative matrix (live Postgres)",
+    minTests: 2,
+  },
   // The charter Phase 5 real-SDK protocol proof. Registered here for the same
   // reason as every suite above: it is env-gated on OPENBRAIN_TEST_DATABASE_URL,
   // so without this entry a CI Postgres misconfiguration would silently skip the
@@ -318,9 +327,10 @@ export const REQUIRED_SUITES: ReadonlyArray<{
 // 230 -> 236 when #878 registered the three subject-split list_stale suites
 // (2 + 1 + 3) as that file stopped skipping itself, then 236 -> 238 when
 // #878 registered the exact-scope checkpoint lifecycle suite's 2 as it too
-// stopped skipping itself), so the global floor cannot silently fall behind
-// the per-suite one.
-export const MIN_TOTAL_LIVE_TESTCASES = 238;
+// stopped skipping itself, then 238 -> 240 when #878 registered the #297
+// namespace isolation negative matrix suite's 2 as that suite stopped skipping
+// itself), so the global floor cannot silently fall behind the per-suite one.
+export const MIN_TOTAL_LIVE_TESTCASES = 240;
 
 export interface SuiteStats {
   tests: number;
--- a/src/tools/__tests__/namespace-isolation-matrix-live.test.ts
+++ b/src/tools/__tests__/namespace-isolation-matrix-live.test.ts
@@ -1,120 +1,129 @@
+/**
+ * Live-Postgres anchor for the #297 negative matrix.
+ *
+ * The mock matrix proves SQL/param shape; this suite proves real Postgres
+ * evaluation of the namespace predicate leaves a foreign-namespace row truly
+ * untouched, and that the same call succeeds for the owning namespace (so the
+ * negative result is not vacuous).
+ *
+ * REQUIRES `OPENBRAIN_TEST_DATABASE_URL`, and fails hard without it (operator
+ * ruling 2026-08-27, issue #878). `bun run test:isolated` sets it.
+ */
 import { afterAll, describe, expect, it } from "bun:test";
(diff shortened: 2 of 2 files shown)


rodaddy:
harvested: #911 The anti-skip guard in `scripts/assert-db-tests-ran.ts` counts a live-Postgres testcase only when the JUnit `classname` contains the literal substring `(live Postgres)` (lines 232 and 264), so a describe named `... (live Postgres, #297)` registers as ZERO executed testcases even though the suite ran and passed — the qualifier has to sit outside the parentheses, or ahead of them, never inside. The failure is quiet in the direction that matters: the subject file goes green on its own, and only `scripts/assert-db-tests-ran.test.ts` notices, because it derives its all-green total from `REQUIRED_SUITES` and compares it against what the parser actually counted. That is the argument for running the guard's own test alongside every `.pg.test.ts` conversion rather than after the batch — it is what turned a suite that would have been invisible in CI into a red test locally, before the branch was pushed.

rodaddy:
verify-lane receipt: check=scripts/done-means/878-pg-tests-require-database.sh exit=0 sha=856d3df606f629519dfdd34ea42cb38ea5f1aa2c at=2026-08-27T08:07:42.646Z

```
    src/tools/__tests__/namespace-isolation-matrix-live.test.ts

CLAUSE 1 [src/tools/__tests__/namespace-isolation-matrix-live.test.ts]: PASS -- 0 code references to OPENBRAIN_TEST_DATABASE_URL, describe.skip, skipIf, dbDescribe
CLAUSE 2 [src/tools/__tests__/namespace-isolation-matrix-live.test.ts]: PASS -- imports requireTestDatabaseUrl from a require-test-database module
CLAUSE 3 [src/tools/__tests__/namespace-isolation-matrix-live.test.ts]: PASS -- exited 1 and printed test_database_required

```

</details>

rodaddy:
### software-factory: errored
Reviewed at: `8abc4cb75d10285cd8a0a5afe64d1fb06a7dd8ff`
Run: job 7554, attempt 11881

#### Receipt
Verdict: **satisfied**

Verbatim check outcomes:
- `error` **Target checkout and SHA availability** — No Git repository was present in /work/job, and the public GitHub pull\-request and commit pages could not be retrieved\.

Findings:
- [info] Review could not run because the target repository was unavailable

PR #914, opened 2026-08-27
test(pg): the graph-derivation live test requires the test database instead of skipping (#878)

rodaddy:
## Summary

- `src/graph-derivation.live.test.ts` chose `describe` or `describe.skip` from its own read of `OPENBRAIN_TEST_DATABASE_URL`. Without the variable it reported `0 pass, 7 skip, 0 fail` and exited 0 — the same exit code as a run that exercised all three suites against a real schema. It now takes its connection string from `requireTestDatabaseUrl()`, which throws `test_database_required` when the variable is unset, and the three `dbDescribe` call sites are plain `describe`.
- What the false green was hiding matters here: these are the #346 regressions that no in-memory fake can prove, because they turn on real partial-unique index arbitration over `idx_ob_entities_canonical` and `idx_ob_entities_lookup_unique`. `src/graph-derivation.test.ts` models both indexes by hand; only this file exercises Postgres actually enforcing them.
- Whole-file standards pass (rule 25) cleared two `max-lines-per-function` findings. The `pool`, `auth`, `cleanup`, and `afterAll` blocks were duplicated verbatim across all three suites, so they move to module scope as one `pool`, `auth`, `store`, and `clearNamespace(ns)`. The prune suite's `anchorEntityId` and `topicId` helpers move alongside them, taking the namespace as a parameter. No assertion, query, or expected value changed. File is 383 code lines.
- `scripts/assert-db-tests-ran.ts` registers the three suite names so a vanished suite fails the db-integration run instead of reporting a silent zero. Floor +4 (`MIN_TOTAL_LIVE_TESTCASES` 108 → 112) for the 1 + 2 + 1 test cases they contribute; the per-suite `minTests` sum equals the floor exactly.

Refs #878

## Verification

- Done-means: scripts/done-means/878-pg-tests-require-database.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

RED at `origin/main`: `env -u OPENBRAIN_TEST_DATABASE_URL bun test src/graph-derivation.live.test.ts` → exit 0, `0 pass, 7 skip, 0 fail`.

GREEN on this branch: same command → exit 1, `test_database_required: OPENBRAIN_TEST_DATABASE_URL is unset`.

[...]

PR #913, opened 2026-08-27
test(pg): the maintenance-sweep live test requires the test database instead of skipping (#878)

rodaddy:
## Summary

- `src/maintenance-sweep.live.test.ts` read `OPENBRAIN_TEST_DATABASE_URL` itself and swapped in `describe.skip` when it was unset, so a run without a test database reported `0 pass, 3 skip, 0 fail` and exited 0 — the false green. It now calls `requireTestDatabaseUrl()` at module scope and the describe is plain, so the absent variable throws `test_database_required` and takes the run down.
- Whole file brought to standard: the fixture helpers (`cleanup`, `insertLane`, `insertTurn`, `jobRows`, `sweepPool`, `sweep`) moved to module scope taking the pool explicitly, which drops the describe body from 132 lines to under the 100-line rule and names the dependency they had been closing over. The three non-null assertions became an `at()` reader that throws with the index named, since a missing row is a fixture defect rather than a nullable value. `oxlint --deny-warnings` is clean on the committed tree; the file is 221 lines.
- `scripts/assert-db-tests-ran.ts` registers `maintenance sweep idempotency (live Postgres)` at `minTests: 1` and the global floor moves **floor +1** (108 → 109). Per-suite `minTests` sum to 109, equal to the floor.
- No behavior change to what the test asserts: the same UNIQUE(job_kind, idempotency_key) conflict path, the same window-advance assertions.

Refs #878

## Verification

- Done-means: scripts/done-means/878-pg-tests-require-database.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

RED at `origin/main`: `env -u OPENBRAIN_TEST_DATABASE_URL bun test src/maintenance-sweep.live.test.ts` → exit 0, `0 pass, 3 skip, 0 fail`.

GREEN on this branch:
- `env -u OPENBRAIN_TEST_DATABASE_URL bun test src/maintenance-sweep.live.test.ts` → exit 1, `test_database_required` present (2 occurrences).
- `bun run test:isolated src/maintenance-sweep.live.test.ts scripts/assert-db-tests-ran.test.ts` → exit 0, `17 pass, 0 fail` across 2 files.
- `CHANGED_FILES=src/maintenance-sweep.live.test.ts bash scripts/done-means/878-pg-tests-require-database.sh` → exit 0, all four clauses PASS.
[...]

PR #917, opened 2026-08-27
test(pg): the agent-context-pack repo-facts live test requires the test database instead of skipping (#878)

rodaddy:
## Summary

- `src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts` gated itself on `OPENBRAIN_TEST_DATABASE_URL` and fell back to `describe.skip` when it was unset. That reports `0 pass, N skip, 0 fail` and exits 0, which is indistinguishable at the exit code from a suite that ran and passed — so the only proofs of the real guidance and `repo_fact` SQL could stop running while CI stayed green.
- The `Pool` now takes `connectionString: requireTestDatabaseUrl()`, which throws `test_database_required` when the variable is missing. The header comment that promised skipped-when-absent was rewritten to describe the hard failure instead.
- The file is split by subject into two flat suites whose names end with `(live Postgres)`: `agent_context_pack guidance + repo_facts` (the two section loaders' SQL) and `agent_context_pack citations + non-fallback` (the pack's output contract). Both are registered in `scripts/assert-db-tests-ran.ts` at `minTests: 2`, and `MIN_TOTAL_LIVE_TESTCASES` rises by that same delta — floor +4, 241 (origin/main at 12d2fa6c) -> 245 — so the global floor cannot fall behind the per-suite one and neither half can vanish unnoticed.
- Whole-file lint findings that predate this change are paid here rather than suppressed, since the pre-commit gate lints staged files whole: the two five-parameter insert helpers take a named fixture object (`PreferenceFixture`, `RepoFactFixture`), both `any` casts are gone, and the helpers are hoisted to module scope so no function runs past the configured line count. No `oxlint` disable comments were added.
- No assertion changed. The shared fixture seeding moved verbatim into `seedIsolationFixture()` and the existing assertions were redistributed across the four cases.

## Verification

Run as: `CHANGED_FILES=src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts bash scripts/done-means/878-pg-tests-require-database.sh`

RED first, at `origin/main`: `env -u OPENBRAIN_TEST_DATABASE_URL bun test <file>` exited 0 with 3 skips — the false green this change removes. GREEN after: the same command exits 1 and prints `test_database_required: OPENBRAIN_TEST_DATABASE_URL is unset; run bun run test:isolated`.

- `bun run test:isolated scripts/assert-db-tests-ran.test.ts src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts` -> exit 0, 20 pass / 0 fail across 2 files
- `bun run test:isolated src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts` -> 4 pass / 0 fail, confirming the 2 + 2 registered in the manifest
- `CHANGED_FILES=... bash scripts/done-means/878-pg-tests-require-database.sh` -> all four clauses PASS (no self-skipping machinery; imports `requireTestDatabaseUrl`; hard failure observed without the variable; `test:isolated` over the subject exits 0)
- `./node_modules/.bin/oxlint --deny-warnings scripts/assert-db-tests-ran.ts src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts` -> exit 0

Python package is untouched, so its checks are not applicable. No MCP tool, schema, transport, or client-facing surface changed, so no live smoke applies.

## Critical Self-Review

- Highest-risk behavior: the manifest edit. `minTests` or `MIN_TOTAL_LIVE_TESTCASES` set above what the suites actually emit fails CI's db-integration job for everyone, and a suite name that does not match the `describe()` string exactly fails the same way. Both names were copied from the source and the counts come from an observed 4-pass run, not from arithmetic.
- Assumptions that could be wrong: that the two new suite names are stable. Anyone renaming either `describe()` without updating the manifest breaks the anti-skip guard — the same coupling every other entry in that file carries, and the reason the split is described in a comment beside the entries.
- Missing/weak tests: the split is proven only by the suites passing and by the guard accepting the manifest. There is no test asserting the manifest names match the `describe()` strings in the tree; that gap is pre-existing and shared by all ~40 entries.
- Security/permission risk: none. No auth, namespace predicate, or SQL text changed. The namespace-isolation and cross-repo-fallback assertions are carried through unmodified — they now sit in two suites instead of one.
- Migration/deploy risk: none. No migration, no runtime code, no schema. The only non-test file is a CI guard manifest.
- Downstream client/runtime risk: none. `docs/downstream-rollout.md` classifies on MCP tool/schema/protocol/client-facing changes; this touches a test file and a CI script, so no rtech-mcps, mcp2cli, or Hermes step applies.
- Rollback/cleanup concern: reverting the commit restores both files together. The manifest entries and the suite names they reference are in the same commit, so there is no window where the guard names a suite that does not exist.
- Fixes made before PR: five pre-existing `oxlint` findings in the touched file were paid rather than suppressed — one function over the line count, two helpers over the parameter count, and two `any` casts. Hoisting the helpers out of the `describe()` and splitting the suite were what cleared the line-count rule without a disable comment. The module-scope `afterAll` was moved deliberately: leaving `pool.end()` inside the first suite would have closed the shared pool before the second suite ran.
- Known residual risk: the two suites now seed the same fixture separately, so the file does four `seedIsolationFixture()`/`cleanupRows()` cycles where it previously did two. Runtime is unchanged in practice (229ms observed) and every row is pid-scoped and cleaned before and after, so no shared state escapes.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical conversion to the already-documented #878 standard with an existing done-means checker, and it surfaced no new failure pattern — the one judgment call (paying whole-file lint findings rather than suppressing them) is already the stated rule in the lane brief and `AGENTS.md`.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no MCP tool, schema, transport, or hosted surface changed — the diff is one test file and one CI guard manifest.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract changed. The suite drives the existing `agent_context_pack` tool through the existing helper; no fixture, schema, or response shape was touched.

## Downstream Rollout

Notes/evidence:

- Every downstream item is not applicable: the change is confined to a test file and `scripts/assert-db-tests-ran.ts`. No tool registration, input schema, response shape, transport behavior, or Python client surface is altered, which is the trigger set `docs/downstream-rollout.md` defines.

### DIFF
--- a/scripts/assert-db-tests-ran.ts
+++ b/scripts/assert-db-tests-ran.ts
@@ -316,6 +316,21 @@ export const REQUIRED_SUITES: ReadonlyArray<{
     name: "deriveGraphFromMetadata stale-edge prune (live Postgres)",
     minTests: 1,
   },
+  // The agent_context_pack guidance + repo_facts proofs (#878), registered as
+  // that file stopped skipping itself. The sibling fake-pool suite supplies its
+  // own rows, so only these live cases evaluate the ::float8
+  // candidate_confidence cast, the candidate_scope.key JSON path, and the exact
+  // repo + namespace binds. Split by subject in #878 -- the two section loaders,
+  // then the pack's citation/non-fallback output contract -- so both halves are
+  // named here; registering one would let the other vanish unnoticed.
+  {
+    name: "agent_context_pack guidance + repo_facts (live Postgres)",
+    minTests: 2,
+  },
+  {
+    name: "agent_context_pack citations + non-fallback (live Postgres)",
+    minTests: 2,
+  },
 ];
 
 // Absolute floor on total executed (non-skipped) live-Postgres testcases,
@@ -351,9 +366,11 @@ export const REQUIRED_SUITES: ReadonlyArray<{
 // namespace isolation negative matrix suite's 2 as that suite stopped skipping
 // itself, then 240 -> 241 when #878 registered the maintenance sweep
 // idempotency suite's 1, then 241 -> 245 when #878 registered the three
-// graph-derivation suites' 1 + 2 + 1), so the global floor cannot silently
-// fall behind the per-suite one.
-export const MIN_TOTAL_LIVE_TESTCASES = 245;
+// graph-derivation suites' 1 + 2 + 1, then 245 -> 249 when #878 registered the
+// two subject-split agent_context_pack guidance/repo_facts suites' 2 + 2 as
+// that file stopped skipping itself), so the global floor cannot silently fall
+// behind the per-suite one.
+export const MIN_TOTAL_LIVE_TESTCASES = 249;
 
 export interface SuiteStats {
   tests: number;
--- a/src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts
+++ b/src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts
@@ -8,98 +8,108 @@
 // including the ::float8 candidate_confidence cast, the candidate_scope.key JSON
 // path, and the exact repo/namespace binds — are proven, not mocked.
 //
-// Gated on OPENBRAIN_TEST_DATABASE_URL; skipped when absent (reported truthfully
-// as skipped, never as passed). Every row is inserted under a pid-scoped
+// Requires OPENBRAIN_TEST_DATABASE_URL (issue #878). When it is absent this
+// suite throws `test_database_required` instead of reporting itself skipped, so
+// an unconfigured database is a loud failure rather than a green run that
+// exercised none of the SQL below. Every row is inserted under a pid-scoped
 // namespace and cleaned up before and after, so no live infrastructure state is
 // mutated beyond the disposable test namespaces.
 
 import { afterAll, describe, expect, it } from "bun:test";
 import { Pool } from "pg";
 import type { AuthInfo } from "../../types.ts";
+import { requireTestDatabaseUrl } from "../../../scripts/test-support/require-test-database.ts";
 import { setupAgentContextPackToolClient as setupToolClient } from "./agent-context-pack-test-helpers.ts";
-
-const DB_URL = process.env.OPENBRAIN_TEST_DATABASE_URL;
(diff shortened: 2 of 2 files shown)


rodaddy:
harvested: #917 The `- Done-means:` field in `scripts/validate-pr-body.ts` is resolved as a bare path, not as a shell command: `requireDoneMeans` takes everything after the colon and hands it straight to `existsInTree`/`existsInRef` (`git cat-file -e <ref>:<value>`), so any runnable form — a `VAR=value` prefix, a `bash` invocation, arguments — fails the gate with "must name a path present in the change under review" even when the script it names is committed and passing. A lane brief that specifies the Done-means line as a full command therefore cannot be satisfied literally; the merged #878 siblings (#909, #911, #913, #914) all carry the bare `scripts/done-means/878-pg-tests-require-database.sh`. The working shape is the bare path in the field, with the full `CHANGED_FILES=... bash <script>` invocation on a prose line directly beneath it, which keeps the exact reproduction command in the PR without breaking the validator's path resolution.

rodaddy:
verify-lane receipt: check=scripts/done-means/878-pg-tests-require-database.sh exit=0 sha=92a21ad62ee26af0397e654ccc5657121c0a5ad8 at=2026-08-27T08:36:13.777Z

```
    src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts

CLAUSE 1 [src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts]: PASS -- 0 code references to OPENBRAIN_TEST_DATABASE_URL, describe.skip, skipIf, dbDescribe
CLAUSE 2 [src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts]: PASS -- imports requireTestDatabaseUrl from a require-test-database module
CLAUSE 3 [src/tools/__tests__/agent-context-pack-guidance-repo-facts-live.test.ts]: PASS -- exited 1 and printed test_database_required

```

</details>

Issue #904, opened 2026-08-27
Two pg suites skip on a var the isolated runner does not supply: local-clone and retire-collab-migration (#878)


rodaddy:
Filed by Claude (head, session 8, claudex opus) from two #878 lanes that stopped on the brief's guard condition.

## What

Two Postgres suites skip on an env var that is not `OPENBRAIN_TEST_DATABASE_URL`, so the `requireTestDatabaseUrl()` swap does not fit them. Both are the same class of false green #878 targets, and both need the runner to provide what they ask for.

- `scripts/local-clone.test.ts:457-459` — `describe.skipIf(!process.env.OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL)`. The suite asserts the URL points at `127.0.0.1`/`::1`, a database named `open_brain_local_*`, and the role `open_brain_local_clone`, then calls `productionDependencies.database.prove(...)` read-only. CI already exports it (`.github/workflows/ci.yml:231`, role and database created at `:132-134`, `:217`); `bun run test:isolated` does not.
- `scripts/retire-collab-migration.test.ts:172-175` — `dbDescribe = (OPENBRAIN_SCRATCH_ADMIN_URL ?? OPENBRAIN_SCRATCH_DATABASE_URL) ? describe : describe.skip`. The suite needs an admin connection that can `CREATE DATABASE` / `DROP DATABASE ... WITH (FORCE)`; it creates `ob_retire_collab_scratch_<ts>`, migrates it, seeds, and drops it. Eight of its tests (the whole live half) skip today.

## Ruling that applies

Rico, 2026-08-26 (#878): every Postgres test runs against a real database, and skipping is a hard failure. That settles the direction; what is left is mechanics.

## Plan (one lane each, in this order)

1. `scripts/test-isolated.ts` exports `OPENBRAIN_SCRATCH_ADMIN_URL` (it already holds admin credentials: it runs `createdb`/`dropdb`) and creates the `open_brain_local_clone` role plus an `open_brain_local_<name>` database the same way `ci.yml:132-134,217` does, exporting `OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL`, and drops both on exit. Done-means: a script that runs `bun run test:isolated -- --help`-equivalent and asserts both vars are set in the child, RED on today's runner.
[...]

PR #918, opened 2026-08-27
test(pg): the local-clone and retire-collab suites require their databases instead of skipping (#904)

rodaddy:
## Summary

- Issue #904 steps 2 and 3, on top of step 1 (248bc199), which taught `bun run test:isolated` and ci.yml to export `OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL` and `OPENBRAIN_SCRATCH_ADMIN_URL`.
- Two suites still decided for themselves whether to run: `scripts/local-clone.test.ts` had `describe.skipIf(!process.env.OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL)`, and `scripts/retire-collab-migration.test.ts` had a `dbDescribe` built from `OPENBRAIN_SCRATCH_ADMIN_URL ?? OPENBRAIN_SCRATCH_DATABASE_URL`. Absent the variable each reported skips and exited 0, which at the exit code is indistinguishable from a suite that ran and passed — the false green #878 exists to close. The live Postgres boundary and the real-migration scratch drill are exactly the two things nobody would notice going unexercised.
- `scripts/test-support/require-test-database.ts` gains `requireLocalCloneTestDatabaseUrl()` and `requireScratchAdminUrl()`, the same shape as the existing `requireTestDatabaseUrl()`: return the value, or throw `test_database_required: <VAR> is unset; run bun run test:isolated`.
- Both suites call theirs at module scope and use a plain flat `describe` ending `(live Postgres)`. The `OPENBRAIN_SCRATCH_DATABASE_URL` fallback goes with the gate — one variable names that connection, and a second spelling is one more way for a runner to look configured while being wrong. The 120000/30000 hook allowances and the terminate-before-drop step are untouched.
- `scripts/test-support/require-test-database.test.ts` is new: each of the three helpers throws when its variable is unset, throws when it is set to an empty string, and returns the value when set, with the environment restored after every case.
- `scripts/done-means/878-pg-tests-require-database.sh` clause 1 widens from one variable to three, and clause 3 now unsets the one the file under test actually demands — read back from the helper it imports. Unsetting the test database while a clone suite reads the clone variable would leave the guard unfired and the run green, which is the outcome clause 3 exists to disprove. Clause 1 also narrows from the bare name to `process.env.<VAR>`: the READ is the defect, and naming the variable in an error message is the opposite of it — the local-clone shape checks report which rule a misconfigured runner broke.
- `scripts/assert-db-tests-ran.ts` registers the renamed retire-collab suite at 6 tests and raises the floor **241 -> 247 (floor +6)**. The local-clone entry was already named there and neither its name nor its count moved. Both suites run in the `db-integration` job's unfiltered `bun test` with all three variables exported (`.github/workflows/ci.yml:278-284`), and `.github/workflows/ci.yml:287` runs the guard over that same JUnit, so the manifest is the right place for them.

## Verification

- Done-means, as run: CHANGED_FILES=scripts/local-clone.test.ts,scripts/retire-collab-migration.test.ts bash scripts/done-means/878-pg-tests-require-database.sh

Receipts:

- RED, at `origin/main`: `env -u OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL bun test scripts/local-clone.test.ts` -> exit 0, 1 skip. `env -u OPENBRAIN_SCRATCH_ADMIN_URL bun test scripts/retire-collab-migration.test.ts` -> exit 0, 8 skip.
- GREEN, after the change: both -> exit 1, `test_database_required` present in the output.
- Deliberate miss: a copy of one file carrying the old `describe.skipIf` guard -> clauses 1, 2 and 3 FAIL. The converted pair -> all four clauses PASS, exit 0.
- `./node_modules/.bin/oxlint --deny-warnings` over all six touched files -> exit 0.
- `bun run test:isolated` -> exit 0, 3981 pass / 10 skip / 0 fail across 263 files.

## Critical Self-Review

- Highest-risk behavior: a module-scope throw takes down the whole file, not one suite. Both files carry non-database unit tests above the live suite (20 and 14 tests), so without the variable those now fail too rather than running. That is the intent of #878 — the file demands its database — but it means a developer running one of these two files alone must have the variable, which `bun run test:isolated` supplies and a bare `bun test` does not.
- Assumptions that could be wrong: that `db-integration` is the job whose JUnit the guard reads. Checked, not assumed — `.github/workflows/ci.yml:282-284` runs an unfiltered `bun test` with all three variables exported and `:287` feeds that file to `scripts/assert-db-tests-ran.ts`. The `check` job also runs these suites but its output is not read by the guard.
- Missing/weak tests: the widened done-means script has no unit test of its own; it is proven by the two runs above (deliberate miss fails, converted pair passes) rather than by an automated case. Its `demanded_variable_of` picks the first helper a file imports, so a file importing two would be checked against only one — no such file exists today.
- Security/permission risk: none. No auth, namespace, or SQL path is touched. `OPENBRAIN_SCRATCH_ADMIN_URL` is an admin connection, and dropping the second spelling narrows rather than widens what a runner may point at it.
- Migration/deploy risk: none. No migration, schema, or runtime source file changed — the diff is two test files, a test-support helper plus its new test, the CI anti-skip manifest, and a done-means script.
- Downstream client/runtime risk: none per `docs/downstream-rollout.md`. No MCP tool, schema, transport, or Python client surface changed, so there is nothing for rtech-mcps, mcp2cli, or rtech-hermes to pick up.
- Rollback/cleanup concern: reverting the commit restores both skip guards and the old floor together; the manifest entry and the suite name move as one commit, so a partial revert cannot leave the guard demanding a suite name that no longer exists.
- Fixes made before PR: the first pass left clause 1 failing on `scripts/local-clone.test.ts`, because the shape-validation error messages name the variable in a string. Rather than obfuscate the literal in the test file, clause 1 was narrowed to `process.env.<VAR>` — the read is the defect, the message is not — and the deliberate-miss probe was re-run afterwards to confirm the narrowing did not make the check permissive.
- Known residual risk: `MIN_TOTAL_LIVE_TESTCASES` is a hand-maintained floor. It is raised by exactly the 6 registered here, so it stays consistent with the per-suite entries, but nothing mechanically derives one from the other.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the pattern this lane exercised — a self-skipping suite reporting a false green — already has its entry from #878, and this change is that same conversion applied to two more files rather than a new failure mode.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, tool, or transport surface changed — the diff is test scaffolding and CI guards.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing shape is touched; the change is entirely test-harness scaffolding.

## Downstream Rollout

Notes/evidence:

- Not applicable throughout: no MCP tool, schema, protocol, or Python client surface changed. The six touched files are two test files, `scripts/test-support/require-test-database.ts` and its new test, `scripts/assert-db-tests-ran.ts`, and `scripts/done-means/878-pg-tests-require-database.sh`.

### DIFF
--- a/scripts/assert-db-tests-ran.ts
+++ b/scripts/assert-db-tests-ran.ts
@@ -58,6 +58,10 @@ export const REQUIRED_SUITES: ReadonlyArray<{
     minTests: 3,
   },
   { name: "local clone real PostgreSQL boundary (live Postgres)", minTests: 1 },
+  {
+    name: "retire-collab-migration scratch Postgres real migrations (live Postgres)",
+    minTests: 6,
+  },
   // The cross-provider parity fixtures (#878). Every recorded tool response for
   // both providers is replayed here, so this is the single suite that proves the
   // two implementations still answer identically. It was env-gated and never
@@ -368,9 +372,10 @@ export const REQUIRED_SUITES: ReadonlyArray<{
 // idempotency suite's 1, then 241 -> 245 when #878 registered the three
 // graph-derivation suites' 1 + 2 + 1, then 245 -> 249 when #878 registered the
 // two subject-split agent_context_pack guidance/repo_facts suites' 2 + 2 as
-// that file stopped skipping itself), so the global floor cannot silently fall
-// behind the per-suite one.
-export const MIN_TOTAL_LIVE_TESTCASES = 249;
+// that file stopped skipping itself, then 249 -> 255 when #904 registered the
+// retire-collab scratch-migration suite's 6 as that file stopped skipping
+// itself), so the global floor cannot silently fall behind the per-suite one.
+export const MIN_TOTAL_LIVE_TESTCASES = 255;
 
 export interface SuiteStats {
   tests: number;
--- a/scripts/done-means/878-pg-tests-require-database.sh
+++ b/scripts/done-means/878-pg-tests-require-database.sh
@@ -49,7 +49,14 @@
 # Four clauses, and all four must pass
 # ---------------------------------------------------------------------------
 # CLAUSE 1 -- NO SELF-SKIPPING MACHINERY SURVIVES, per file.
-#   Zero CODE references to the literal `OPENBRAIN_TEST_DATABASE_URL`, and zero
+#   Issue #904 widened this from one variable to three. A file may demand any of
+#   `OPENBRAIN_TEST_DATABASE_URL`, `OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL`, or
+#   `OPENBRAIN_SCRATCH_ADMIN_URL` -- they name different databases, so the clone
+#   and scratch suites are not served by the migrated test one. Whichever it
+#   uses, the environment read belongs in the helper and not in the test file,
+#   so a code reference to ANY of the three still fails this clause.
+#
+#   Zero CODE-line `process.env.<VAR>` reads of those three, and zero
 #   to `describe.skip`, `skipIf`, or `dbDescribe`. Comment lines are exempt:
 #   a converted file's header still explains what it requires and why, and
 #   reading prose as machinery would push authors to delete the explanation.
@@ -59,7 +66,8 @@
 # CLAUSE 2 -- THE HELPER IS ACTUALLY IMPORTED.
 #   Clause 1 alone is satisfiable by deleting the environment read and leaving
 #   the suite connecting to nothing, which is worse than the defect. So clause 2
-#   demands the identifier `requireTestDatabaseUrl` on an import line whose
+#   demands one of `requireTestDatabaseUrl`, `requireLocalCloneTestDatabaseUrl`,
+#   or `requireScratchAdminUrl` on an import line whose
 #   module path ends in the basename `require-test-database`. Matching the
 #   IDENTIFIER and the BASENAME rather than a fixed relative path is what lets
 #   files at different depths under server/ and src/ satisfy the same clause --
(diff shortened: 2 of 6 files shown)


rodaddy:
harvested: #918 A done-means check that hardcodes the environment variable it unsets stops discriminating the moment a second variable enters the family. Clause 3 of `scripts/done-means/878-pg-tests-require-database.sh` unset `OPENBRAIN_TEST_DATABASE_URL` for every subject; the instant a suite demanded the clone or scratch variable instead, that unset left the guard unfired, the run exited 0, and the clause reported a pass for a file whose demand had never been exercised — the same false green the check was written to catch, one layer up in the checker itself. The fix is for the check to read back from the subject which variable it demands (via the helper it imports) and unset that one. Second lesson from the same lane: clause 1 banned the bare variable NAME on a code line, which caught the real defect and also caught an error message that merely reports which variable a misconfigured runner failed to set. Banning `process.env.<VAR>` instead separates the read from the report — the read is the defect, and naming the variable in a diagnostic is the opposite of hiding it. A ban written against a string rather than against the operation will keep pushing authors to obfuscate the string.

rodaddy:
verify-lane receipt: check=scripts/done-means/878-pg-tests-require-database.sh exit=0 sha=7d8e5ced45704ff13f798812aa15420593de8b2c at=2026-08-27T08:41:28.436Z

```
    scripts/local-clone.test.ts
    scripts/retire-collab-migration.test.ts

CLAUSE 1 [scripts/local-clone.test.ts]: PASS -- 0 code-line process.env reads of OPENBRAIN_TEST_DATABASE_URL/OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL/OPENBRAIN_SCRATCH_ADMIN_URL, and no describe.skip, skipIf, dbDescribe
CLAUSE 2 [scripts/local-clone.test.ts]: PASS -- imports requireTestDatabaseUrl|requireLocalCloneTestDatabaseUrl|requireScratchAdminUrl from a require-test-database module
CLAUSE 3 [scripts/local-clone.test.ts]: PASS -- exited 1 and printed test_database_required
CLAUSE 1 [scripts/retire-collab-migration.test.ts]: PASS -- 0 code-line process.env reads of OPENBRAIN_TEST_DATABASE_URL/OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL/OPENBRAIN_SCRATCH_ADMIN_URL, and no describe.skip, skipIf, dbDescribe
CLAUSE 2 [scripts/retire-collab-migration.test.ts]: PASS -- imports requireTestDatabaseUrl|requireLocalCloneTestDatabaseUrl|requireScratchAdminUrl from a require-test-database module
CLAUSE 3 [scripts/retire-collab-migration.test.ts]: PASS -- exited 1 and printed test_database_required

```

</details>

PR #939, opened 2026-08-28
test(pg): the graph derivation handler live test requires the test database instead of skipping (#878)

rodaddy:
## Summary

- `src/graph-derivation-handler.live.test.ts` read `OPENBRAIN_TEST_DATABASE_URL` itself and swapped in `describe.skip` when it was unset. Absent the variable it reported `0 pass, 8 skip, 0 fail` and exited 0, which at the exit code is indistinguishable from a suite that ran and passed. It now calls `requireTestDatabaseUrl()` at module scope and fails loudly with `test_database_required`.
- Split by subject in the same change, because the pre-commit gate lints staged files whole and the 770-line original carried eight findings on `origin/main` (5 `no-non-null-assertion`, 2 `max-lines-per-function`, 1 `max-lines`). `src/graph-derivation-handler.live.test.ts` keeps selection and convergence; the new `src/graph-derivation-handler-isolation.live.test.ts` takes isolation and atomicity, and `poolFailingAfter` moves with it.
- New `src/graph-derivation-handler-test-helpers.ts` holds the namespaces, auth, `cleanup`, `insertSource` and `jobFor`. It creates no pool: each database function takes the caller's `pool` first, so each half creates one module-scope pool and ends exactly its own connection in its own `afterAll`.
- Test names, order and assertions are unchanged. Every `it` body is a hoisted module-scope `async function` so no describe callback exceeds the code-line standard, and the five non-null assertions became an `expectDefined(value, label)` guard exported from the helper module. No disable comments.
- `scripts/assert-db-tests-ran.ts` registers both new describe names at their JUnit-measured counts (4 and 3) and moves `MIN_TOTAL_LIVE_TESTCASES` 285 -> 292, with the comment chain extended by one clause.
- `src/graph-derivation-handler.ts` and `src/maintenance-queue.ts` are untouched.

## Verification

- Run as `CHANGED_FILES="src/graph-derivation-handler.live.test.ts src/graph-derivation-handler-isolation.live.test.ts" bash scripts/done-means/878-pg-tests-require-database.sh`. RED on `origin/main` before any edit: exit 1, clauses 1/2/3 FAIL. GREEN on the committed tree: exit 0, all four clauses PASS.

Receipts on the committed tree:

- `./node_modules/.bin/oxlint --deny-warnings` over all four touched files -> exit 0, no findings.
- `bun run test:isolated src/graph-derivation-handler.live.test.ts src/graph-derivation-handler-isolation.live.test.ts --reporter=junit` -> exit 0, 7 pass / 0 fail / 39 expect() calls, 4 and 3 `testcase` elements per describe.

## Critical Self-Review

- Highest-risk behavior: the split could silently drop a test or an assertion. Guarded by moving every body with shell line-range copies rather than retyping, and by the JUnit counts summing to the original 7 with 39 expect() calls.
- Assumptions that could be wrong: that the two halves share no state through the database. They use the same two namespaces and each half runs its own `beforeEach(cleanup)`, so a concurrent run of both files could interleave deletes. Bun runs test files in one process sequentially, so this is not exercised today, but it is the assumption most likely to bite if the runner ever parallelizes.
- Missing/weak tests: none added -- this is a conversion, and the assertion set is byte-identical to `origin/main`. The `expectDefined` guard itself has no direct unit test; its failure path is only reached when an assertion was already going to fail.
- Security/permission risk: none. No auth, namespace predicate, or SQL changed; the namespace-isolation test moved intact and still asserts no derived edge crosses the boundary.
- Migration/deploy risk: none. No migration, no runtime source file, and the helper module is test-only.
- Downstream client/runtime risk: none. `docs/downstream-rollout.md` classifies this as not applicable -- no MCP tool, schema, protocol, or client-facing surface is touched.
- Rollback/cleanup concern: reverting the commit restores the single file. The two new files and the helper would need deleting with it, and the manifest floor would need to return to 285; a partial revert that left the floor at 292 would fail CI's anti-skip guard.
- Fixes made before PR: added the missing `otherNs` import that `tsc` caught after the split; replaced the five `x!` assertions with `expectDefined`; ran prettier and re-took every lint, typecheck, test, and done-means receipt on the committed tree rather than the staged one.
- Known residual risk: the interleave test runs as two sequential halves under one `it` to keep each function under the code-line standard. The halves are independent (the second shares no variable with the first), but a future edit that introduces a dependency across the seam would be easy to miss.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane found no new MEDIUM+ pattern -- the split, hoist, and helper-takes-pool shapes are already captured in round 39 of the lane contract's Tightenings from the session-9 lanes.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: test-only change with no runtime or contract surface.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing shape changed -- the diff is test files and the anti-skip manifest.

## Downstream Rollout

Notes/evidence:

- Not applicable throughout: no MCP tool, transport, schema, or Python client path is in the diff.

### DIFF
--- a/scripts/assert-db-tests-ran.ts
+++ b/scripts/assert-db-tests-ran.ts
@@ -395,6 +395,14 @@ export const REQUIRED_SUITES: ReadonlyArray<{
     name: "ingest_conversation_facts isolation and concurrency (live Postgres)",
     minTests: 3,
   },
+  {
+    name: "graph derivation handler selection and convergence (live Postgres)",
+    minTests: 4,
+  },
+  {
+    name: "graph derivation handler isolation and atomicity (live Postgres)",
+    minTests: 3,
+  },
 ];
 
 // Absolute floor on total executed (non-skipped) live-Postgres testcases,
@@ -444,8 +452,11 @@ export const REQUIRED_SUITES: ReadonlyArray<{
 // stopped skipping itself, then 284 -> 285 when #878 re-measured the
 // runSharedPromoter cursor-stall suite at its emitted 1 and registered the
 // clustering and cursor loops suite's 8 as that file stopped skipping
-// itself), so the global floor cannot silently fall behind the per-suite one.
-export const MIN_TOTAL_LIVE_TESTCASES = 285;
+// itself, then 285 -> 292 when #878 registered the two subject-split graph
+// derivation handler selection/convergence and isolation/atomicity suites'
+// 4 + 3 as that file stopped skipping itself), so the global floor cannot
+// silently fall behind the per-suite one.
+export const MIN_TOTAL_LIVE_TESTCASES = 292;
 
 export interface SuiteStats {
   tests: number;
--- a/src/graph-derivation-handler-isolation.live.test.ts
+++ b/src/graph-derivation-handler-isolation.live.test.ts
@@ -1,0 +1,431 @@
+import { afterAll, beforeEach, describe, expect, it } from "bun:test";
+import { Pool } from "pg";
+import {
+  GraphDerivationTerminalError,
+  SOURCE_ANCHOR_ENTITY_TYPE,
+  makeGraphDerivationHandler,
+  selectSourcesNeedingDerivation,
+} from "./graph-derivation-handler.ts";
+import { requireTestDatabaseUrl } from "../scripts/test-support/require-test-database.ts";
+import {
+  auth,
+  cleanup,
+  expectDefined,
+  hashA,
+  hashB,
+  insertSource,
+  jobFor,
+  ns,
+  otherNs,
+} from "./graph-derivation-handler-test-helpers.ts";
+
+/**
+ * Live-Postgres isolation and atomicity half of the #346 maintenance
+ * integration suite (split out under #878).
+ *
+ * Proves against the REAL schema that derivation never leaks across a
+ * namespace, that a mid-derivation failure leaves no partial graph, and that a
+ * stale job serialized behind a newer one terminal-stops instead of
+ * overwriting the newer committed graph.
+ *
+ * The suite demands OPENBRAIN_TEST_DATABASE_URL through the shared helper
+ * rather than skipping itself when it is absent.
+ */
+const pool = new Pool({ connectionString: requireTestDatabaseUrl() });
+
+function poolFailingAfter(
+  real: Pool,
+  failOn: string,
+): Pick<Pool, "connect"> & { armed: boolean } {
+  const wrapper = {
+    armed: true,
+    connect: async () => {
+      const client = await real.connect();
+      const realQuery = client.query.bind(client);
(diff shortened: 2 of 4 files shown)


rodaddy:
harvested: #939 A shared test helper module is the natural home for the guard that replaces `x!`, and exporting one generic `expectDefined(value, label)` from it beat rewriting five call sites into optional chaining or local narrowing: every one of the five was the same `selected[0]!` shape, so a single export cleared all of them without changing what any assertion means, and the label names which expectation went missing instead of surfacing a bare TypeError. Worth pairing with the round-39 tightening about helpers taking `pool` as a parameter -- the same module carries both, and neither costs the file a pool of its own. Second, splitting a long `it` body into two sequential module-scope halves called from one wrapper keeps the test name, order, and assertion set byte-identical while clearing `max-lines-per-function`, but it only works where the halves share no variable across the seam; checking that independence before cutting is what makes the split safe rather than a rename of the problem.

rodaddy:
verify-lane receipt: check=scripts/done-means/878-pg-tests-require-database.sh exit=0 sha=acd2d097d930b83b825e5ede1a1c90d023f96169 at=2026-08-28T02:11:07.706Z

```
    src/graph-derivation-handler.live.test.ts
    src/graph-derivation-handler-isolation.live.test.ts

CLAUSE 1 [src/graph-derivation-handler.live.test.ts]: PASS -- 0 code-line process.env reads of OPENBRAIN_TEST_DATABASE_URL/OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL/OPENBRAIN_SCRATCH_ADMIN_URL, and no describe.skip, skipIf, dbDescribe
CLAUSE 2 [src/graph-derivation-handler.live.test.ts]: PASS -- imports requireTestDatabaseUrl|requireLocalCloneTestDatabaseUrl|requireScratchAdminUrl from a require-test-database module
CLAUSE 3 [src/graph-derivation-handler.live.test.ts]: PASS -- exited 1 and printed test_database_required
CLAUSE 1 [src/graph-derivation-handler-isolation.live.test.ts]: PASS -- 0 code-line process.env reads of OPENBRAIN_TEST_DATABASE_URL/OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL/OPENBRAIN_SCRATCH_ADMIN_URL, and no describe.skip, skipIf, dbDescribe
CLAUSE 2 [src/graph-derivation-handler-isolation.live.test.ts]: PASS -- imports requireTestDatabaseUrl|requireLocalCloneTestDatabaseUrl|requireScratchAdminUrl from a require-test-database module
CLAUSE 3 [src/graph-derivation-handler-isolation.live.test.ts]: PASS -- exited 1 and printed test_database_required

```

</details>

PR #947, opened 2026-08-28
fix(done-means): clause 2 of the #878 check judges the helper demand across a wrapped import and a sibling relay (#945)

### LINKED ISSUE #945: 878 done-means clause 2 judges the import's line shape: a prettier-wrapped require-test-database import fails a file that demands the database
Found by the session-10 lanes for backup-restore-live (pull request #944) and reported first by the phase-1 lane.

scripts/done-means/878-pg-tests-require-database.sh clause 2 demands the helper identifier (requireTestDatabaseUrl | requireLocalCloneTestDatabaseUrl | requireScratchAdminUrl) and the module basename require-test-database on ONE import line. A file importing two of the helpers gets its import wrapped by prettier in the pre-commit gate onto several physical lines, and clause 2 then reports FAIL while clauses 1, 3, and 4 all PASS: the file demands the database and hard-fails without it, and the check says it does not. The lanes worked around it with one named specifier per import line (the origin/main convention), which is fine as a convention but is not what the clause claims to check.

Second shape of the same defect: a split half that reaches the helper only through a shared lane helper module (which imports it) fails clause 2 too, though clause 3 proves the hard fail; each half now carries its own call and single-line import.

Ask: make clause 2 match the identifier and the basename within the same import STATEMENT (multi-line aware), or document the one-specifier-per-line convention in the script header and in HANDOVER-RULES; prove it with a deliberate miss per rule 40 (a wrapped import that should pass, and a file with no import that should fail).

Parent: #878. Related: #918 (clause design), #944.

rodaddy:
## Summary

- Clause 2 of `scripts/done-means/878-pg-tests-require-database.sh` matched the require-test-database identifier and the module basename on a single physical LINE. Two honest conversions failed it while clauses 1, 3, and 4 passed: prettier wraps a two-specifier import so the identifier and the module specifier land on different lines, and one half of a split suite reaches the helper only through the shared sibling module the split created (round-40 harvest, #945).
- The clause now judges the DEMAND. Import statements are joined from `import` to the terminating `;` before matching, so wrapping is invisible to the check. Failing a direct import, each relative sibling module the subject imports is resolved from the subject's own directory (`.ts` appended when the specifier carries none) and read once; the clause passes when that sibling imports the identifier from a require-test-database module, and the call is looked for there because the relay owns it.
- One relay hop only. A file that calls a helper it never imports at all still fails, which is the case clause 2 exists to catch. Clauses 1, 3, and 4 are unchanged, and the header prose was rewritten so the documented rule matches the code.
- Closes #945.

## Verification

  `CHANGED_FILES="scripts/__tests__/backup-restore-live.test.ts scripts/__tests__/backup-restore-live-refusals.test.ts" bash scripts/done-means/878-pg-tests-require-database.sh` exits 0 on the committed tree, all four clauses PASS.

RED, at origin/main (751c7268), with an untracked fixture carrying a prettier-wrapped two-specifier import and a module-scope `requireTestDatabaseUrl()` call: clause 2 FAIL, clauses 1, 3, 4 PASS, exit 1.

GREEN, after the change: the same fixture exits 0; a second fixture reaching the helper only through an untracked sibling relay module exits 0 and names the relay in its PASS line; a third fixture that calls the helper with no import at all still exits 1 on clause 2. The two real backup-restore files above still exit 0. The four fixtures and the relay helper were moved out of the tree with `mv` before staging and `git status` showed none of them.

## Critical Self-Review

- Highest-risk behavior: a check that gates every #878 conversion lane now accepts a wider set of shapes. A too-permissive clause 2 would let a suite that never demands the database through, reinstating the false green this whole script exists to catch.
- Assumptions that could be wrong: that the awk statement joiner handles every import spelling in this repo. It keys on a line starting with `import` and closes at the first `;`, so an import with a `;` inside a string specifier, or a semicolon-free style, would join wrong. Neither shape exists in this repo's test files, and clause 2 fails closed when the join is wrong rather than passing.
- Missing/weak tests: the change has no committed unit test; the proof is the four fixture runs recorded above, which are untracked by design. A tracked fixture pair for the wrapped and relay shapes would make this a standing regression rather than a one-time receipt.
- Security/permission risk: none. The script reads source files inside the repo and runs no new command; there is no namespace, token, or database mutation in the change.
- Migration/deploy risk: none. This is a developer-run check script with no runtime, migration, or service surface.
- Downstream client/runtime risk: none per `docs/downstream-rollout.md`. No MCP tool, schema, transport, or Python client path is touched, so rtech-mcps, mcp2cli, and Hermes are unaffected.
- Rollback/cleanup concern: reverting the commit restores the line-shaped clause exactly; nothing else depends on the new helper functions. The lane created four untracked fixtures and moved them to the session scratch area before staging.
- Fixes made before PR: the first relay resolver used `realpath --relative-to`, which is GNU-only and absent on macOS where this check runs; it was replaced with a textual `.`/`..` collapser. The relay branch initially left the call search on the subject file, which would have failed the relay fixture because the subject calls the relay wrapper rather than the helper; the call is now looked for in the relay module.
- Known residual risk: the one-hop rule is deliberate, so a two-hop relay chain fails clause 2 with a message that reads as a missing import. That is the intended trade against turning a source check into a module resolver, but it will surprise a lane that builds a deeper helper chain.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the pattern this fixes is already recorded as a round-40 Tightening in `docs/lane-contract.md` and is the issue this PR closes; the lane found no new MEDIUM+ review finding to promote.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: the change is a shell check script with no service surface.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, schema, or client-facing shape is touched.

## Downstream Rollout

Notes/evidence:

- The done-means invocation above was re-run on the committed tree after the pre-commit gate, not on the staged one. `bunx tsc --noEmit` exits 0 and `bash -n` on the script exits 0; oxlint has no shell rule set, so it reports nothing to lint for this path.

### DIFF
--- a/scripts/done-means/878-pg-tests-require-database.sh
+++ b/scripts/done-means/878-pg-tests-require-database.sh
@@ -63,16 +63,35 @@
 #   The exemption is line-shaped (`//`, `/*`, `*`), which is what prettier
 #   leaves behind; a literal buried mid-expression on a code line still fails.
 #
-# CLAUSE 2 -- THE HELPER IS ACTUALLY IMPORTED.
+# CLAUSE 2 -- THE HELPER DEMAND ACTUALLY REACHES THE FILE.
 #   Clause 1 alone is satisfiable by deleting the environment read and leaving
 #   the suite connecting to nothing, which is worse than the defect. So clause 2
-#   demands one of `requireTestDatabaseUrl`, `requireLocalCloneTestDatabaseUrl`,
-#   or `requireScratchAdminUrl` on an import line whose
-#   module path ends in the basename `require-test-database`. Matching the
-#   IDENTIFIER and the BASENAME rather than a fixed relative path is what lets
-#   files at different depths under server/ and src/ satisfy the same clause --
-#   `../../scripts/test-support/...` and `../scripts/test-support/...` are the
-#   same import.
+#   demands that one of `requireTestDatabaseUrl`,
+#   `requireLocalCloneTestDatabaseUrl`, or `requireScratchAdminUrl` is CALLED,
+#   and that the identifier arrives from a module whose specifier basename is
+#   `require-test-database`. Matching the IDENTIFIER and the BASENAME rather
+#   than a fixed relative path is what lets files at different depths under
+#   server/ and src/ satisfy the same clause -- `../../scripts/test-support/...`
+#   and `../scripts/test-support/...` are the same import.
+#
+#   Issue #945: the clause used to judge the LINE SHAPE of the import, and two
+#   honest conversions failed it while clauses 1, 3, and 4 passed. Prettier
+#   wraps a two-specifier import across physical lines, putting the identifier
+#   and the module specifier on different lines so no single line carries both;
+#   and one half of a split suite can reach the helper only through the shared
+#   sibling module the split created. So the clause is now STATEMENT-AWARE and
+#   allows ONE RELAY HOP:
+#     - an import statement is joined from its `import` keyword to the
+#       terminating `;` before matching, so wrapping is invisible to it;
+#     - failing a direct import, each RELATIVE sibling module the file imports
+#       (resolved from the subject's own directory, `.ts` appended when the
+#       specifier carries none) is read once, and the clause passes when that
+#       sibling imports the identifier from a `require-test-database` module.
+#       The relay owns the call in that shape, so the call is looked for there.
+#   The hop is deliberately ONE deep: a chain longer than that is no longer a
+#   split's shared helper, and following it would turn a source check into a
+#   module resolver. A file that CALLS a helper it never imports at all still
+#   fails -- that is the case clause 2 exists to catch.
 #
 # CLAUSE 3 -- THE HARD FAIL IS OBSERVED, not merely arranged, per file.
 #   Clauses 1 and 2 read source. Clause 3 runs the thing: with the variable THAT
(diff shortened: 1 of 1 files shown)


rodaddy:
harvested: #947 A source check that reads one physical line at a time is judging the formatter, not the code. Clause 2 of the #878 check matched an identifier and a module basename on the same line, so prettier's own wrap of a two-specifier import — the exact output the repo's pre-commit gate produces — turned a correct conversion into a clause-2 failure while clauses 1, 3, and 4 passed, which is the shape that reads as a lane defect rather than a check defect. Any shell check that inspects TypeScript imports joins the statement from `import` to its terminating `;` before matching; the awk join is four lines and removes the whole class. The second half is that a check gating a SPLIT must model what a split produces: the shared sibling helper module is the normal outcome, so the identifier legitimately arrives one hop away, and the check either resolves that hop or silently forbids the structure the conversion rules ask for. One hop is the right depth — it covers the split's own helper and stops short of reimplementing module resolution — but the depth is a decision to state in the header prose, because a two-hop chain now fails with a message that reads as a missing import.

rodaddy:
verify-lane receipt: check=scripts/done-means/878-pg-tests-require-database.sh exit=0 sha=b756b7760d3f72e0a6851f046dcfad941bf14576 at=2026-08-28T04:07:40.523Z

```
    scripts/__tests__/backup-restore-live.test.ts
    scripts/__tests__/backup-restore-live-refusals.test.ts

CLAUSE 1 [scripts/__tests__/backup-restore-live.test.ts]: PASS -- 0 code-line process.env reads of OPENBRAIN_TEST_DATABASE_URL/OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL/OPENBRAIN_SCRATCH_ADMIN_URL, and no describe.skip, skipIf, dbDescribe
CLAUSE 2 [scripts/__tests__/backup-restore-live.test.ts]: PASS -- calls requireTestDatabaseUrl|requireLocalCloneTestDatabaseUrl|requireScratchAdminUrl, reaching it through its own import
CLAUSE 3 [scripts/__tests__/backup-restore-live.test.ts]: PASS -- exited 1 and printed test_database_required
CLAUSE 1 [scripts/__tests__/backup-restore-live-refusals.test.ts]: PASS -- 0 code-line process.env reads of OPENBRAIN_TEST_DATABASE_URL/OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL/OPENBRAIN_SCRATCH_ADMIN_URL, and no describe.skip, skipIf, dbDescribe
CLAUSE 2 [scripts/__tests__/backup-restore-live-refusals.test.ts]: PASS -- calls requireTestDatabaseUrl|requireLocalCloneTestDatabaseUrl|requireScratchAdminUrl, reaching it through its own import
CLAUSE 3 [scripts/__tests__/backup-restore-live-refusals.test.ts]: PASS -- exited 1 and printed test_database_required

CLAUSE 2 (the helper demand reaches the file):      PASS
```

</details>

PR #949, opened 2026-08-28
test(pg): the sdk-protocol live suite requires the test database instead of skipping, helpers split out (#878)

rodaddy:
## Summary

- `server/application/sdk-protocol.pg.test.ts` read `OPENBRAIN_TEST_DATABASE_URL` itself and swapped in `describe.skip` when it was unset, reporting `0 pass, 13 skip, 0 fail` and exiting 0. At the exit code that is indistinguishable from a suite that ran and passed, so the real-SDK protocol behavior nobody exercised could drift with CI green.
- The module now calls `requireTestDatabaseUrl()` at module scope and builds one `Pool` from it, ended once in the file's `afterAll`. Absent the variable the run fails loudly with `test_database_required`.
- The constants, module state, and seven harness functions move to a new sibling `server/application/sdk-protocol-test-helpers.ts`. That module holds no test and creates no pool: `startWorkerApplication` and `connectRealClient` take the caller's `Pool` as their first parameter, so the file that owns the pool also owns ending it. The `if (!pool) throw` guards go away with it, and an `expectDefined` guard exported from the same module replaces every non-null assertion.
- `node/no-process-env` is scoped to `server/**`, so the shared-namespace isolation moves to a new `scripts/test-support/shared-namespace-env.ts` exporting `isolateSharedNamespace` and `restoreSharedNamespace`. Behavior is identical; the environment access now lives in test support, outside the `server/` scope.
- Every `it` body hoists to a module-scope `async function` called as `test("<same name>", fn)`. That is what clears `max-lines-per-function`: the rule measures the describe callback, and neither splitting the file nor nesting sub-describes moves that number. Test names, order, and assertions are unchanged.
- `scripts/assert-db-tests-ran.ts` is deliberately untouched. The suite name is unchanged and its manifest entry already carried `minTests: 12`, verified against the JUnit record this branch produced.

## Verification

- `CHANGED_FILES="server/application/sdk-protocol.pg.test.ts" bash scripts/done-means/878-pg-tests-require-database.sh` — exit 1 at `origin/main` (751c7268: clauses 1, 2, 3 FAIL), exit 0 on the committed tree with all four clauses PASS.

`bun run test:isolated server/application/sdk-protocol.pg.test.ts` — 12 pass, 0 fail, 63 expect() calls, where the same file previously reported 13 skip. `bunx tsc --noEmit` exit 0. `./node_modules/.bin/oxlint --deny-warnings` exit 0 over all three touched files, which on `origin/main` carried 7 `node/no-process-env`, 7 `no-non-null-assertion`, 1 `max-lines`, and 1 `max-lines-per-function`. The pre-push hook ran the whole suite: 4009 pass, 0 fail. Python package and live smoke are not applicable — this branch changes test files and test support only, and touches no client, contract, or runtime path.

## Critical Self-Review

- Highest-risk behavior: the pool lifetime across the split. A helper module that created its own pool would give each half of a future split an `afterAll` ending a connection the other half still uses, so the helpers take `pool` as a parameter and create none; the single module-scope pool in the test file is ended exactly once.
- Assumptions that could be wrong: that `expectDefined` preserves each assertion's meaning. It throws where `!` silently proceeded, so a value that was legitimately absent at that point would now fail rather than surface as a downstream `undefined`. The 12 tests pass, which is the evidence that none of the seven sites relied on that.
- Missing/weak tests: no test covers `shared-namespace-env.ts` directly; it is exercised only through this suite's `beforeAll`/`afterAll`. Its restore path (delete when the prior value was undefined) is not asserted independently.
- Security/permission risk: none identified. No auth, namespace predicate, or token path changed. The `bearerAuth` seam moved verbatim and still stamps the same identity onto `req.auth`, which is the chain this file exists to prove.
- Migration/deploy risk: none. No migration, no schema, no deployed code path — `server/application/sdk-protocol.ts` and `shadow-application` are untouched.
- Downstream client/runtime risk: none. Per `docs/downstream-rollout.md` this is not an MCP tool, schema, protocol, or client-facing change; no rtech-mcps, mcp2cli, or Hermes surface moves.
- Rollback/cleanup concern: reverting the commit restores the skipping suite. The two new files are only imported by this suite, so nothing is stranded.
- Fixes made before PR: raised `MIN_TOTAL_LIVE_TESTCASES` from 328 to 340 on the assumption the newly-executing 12 were additive, and the guard's own unit test failed. That floor is the sum of the manifest's own `minTests`, so this suite's 12 were already inside it while the suite skipped. The change was reverted and `scripts/assert-db-tests-ran.ts` left untouched; `bun test scripts/assert-db-tests-ran.test.ts` then returned 16 pass, 0 fail.
- Known residual risk: the first done-means run failed clause 4 on a transient `pgvector_registration_failed` during database provisioning, not on anything in this change; the immediate re-run and the committed-tree run both passed all four clauses. That provisioning race is real and is not addressed here.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the one lesson here is about the anti-skip guard's floor arithmetic, which belongs in the lane contract's Tightenings for the #878 harvest rather than in a reviewer lane file; it is posted as a harvest comment on this PR.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no deployed surface changes; this is a test-only conversion.

## Contract Parity

- Contract parity: [x] runtime-specific because: no fixture changed. The suite still compares against the same recorded shapes through the same imported comparator; only the harness composition moved.

## Downstream Rollout

Notes/evidence:

- Test files and test support only; no MCP tool, schema, transport, or Python client path is touched, so every downstream item is not applicable per the classification in `docs/downstream-rollout.md`.

### DIFF
--- a/scripts/test-support/shared-namespace-env.ts
+++ b/scripts/test-support/shared-namespace-env.ts
@@ -1,0 +1,46 @@
+/**
+ * Shared-namespace environment isolation for suites that must observe nothing
+ * but their own writes.
+ *
+ * `readableNamespaces()` grants every non-admin role read access to the shared
+ * namespace as well, so a suite running against the real `shared-kb` would also
+ * observe whatever the target database happens to hold there, and its recorded
+ * counts would assert an accident of that database.
+ *
+ * `sharedNamespaceConfig()` re-reads the environment on every call, and Bun runs
+ * every test file in ONE process, so a module-scope assignment leaks into
+ * sibling files and fails them by load order. A suite therefore installs the
+ * isolation in `beforeAll` and removes it in `afterAll`.
+ *
+ * This module lives under `scripts/` deliberately: `.oxlintrc.json` scopes
+ * `node/no-process-env` to `server/**\/*.ts`, so the environment access belongs
+ * in test support rather than beside the tests it serves.
+ */
+
+const KEYS = ["SHARED_NAMESPACE_CANONICAL", "SHARED_NAMESPACE_PHYSICAL"];
+
+let prior: Array<string | undefined> | undefined;
+
+/**
+ * Point both shared-namespace variables at `value`, recording the prior values
+ * once so a later restore returns the process to what it inherited.
+ */
+export function isolateSharedNamespace(value: string): void {
+  if (prior === undefined) {
+    prior = KEYS.map((key) => process.env[key]);
+  }
+  for (const key of KEYS) {
+    process.env[key] = value;
+  }
+}
+
+/** Restore both shared-namespace variables to the values recorded on isolate. */
+export function restoreSharedNamespace(): void {
+  if (prior === undefined) return;
+  KEYS.forEach((key, index) => {
+    const value = prior?.[index];
+    if (value === undefined) delete process.env[key];
+    else process.env[key] = value;
+  });
+  prior = undefined;
+}
--- a/server/application/sdk-protocol-test-helpers.ts
+++ b/server/application/sdk-protocol-test-helpers.ts
@@ -1,0 +1,351 @@
+/**
+ * Shared harness for the real-SDK protocol proof.
+ *
+ * Holds the constants, module state, and helper functions that compose the real
+ * `server/` stack behind a real ephemeral HTTP listener. It holds no test and
+ * creates no pool: every helper that touches the database takes the caller's
+ * `Pool` as its first parameter, so the file that owns the pool also owns
+ * ending it.
+ */
+import type { AddressInfo } from "node:net";
+import type { Server } from "node:http";
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js";
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import express from "express";
+import type { RequestHandler } from "express";
+import pino from "pino";
+import type { Pool } from "pg";
+import { createShadowApplication } from "./index.ts";
+import type { ShadowApplication } from "./index.ts";
+import { parseServerConfig } from "../config.ts";
+import type { ServerConfig } from "../config.ts";
+import type { Database, DatabaseHealth } from "../db/pool.ts";
+import { RecoveryWalStore } from "../realtime/recovery-wal.ts";
+import { WorkingSetStore } from "../realtime/working-set.ts";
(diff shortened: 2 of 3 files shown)


rodaddy:
harvested: #949 `MIN_TOTAL_LIVE_TESTCASES` in `scripts/assert-db-tests-ran.ts` is the SUM OF THE MANIFEST'S OWN `minTests`, not a count of testcases observed executing, so a #878 conversion of a suite that ALREADY HAS a manifest entry raises it by zero. Every round-39 and round-40 conversion added its count because each registered a NEW entry — a suite the table had never carried — and the running comment chain ("328 -> 340 when #878 registered ...") reads as if executing is what earns the increment. It is not: this suite's entry sat at `minTests: 12` while the file skipped, so its 12 were inside the 328 the whole time. Raising the floor to 340 typechecks, lints, and passes the done-means, and fails only `scripts/assert-db-tests-ran.test.ts` — whose all-green fixture derives its total from `REQUIRED_SUITES` by the same reduce, so the fixture can never reach a hand-raised floor. The discriminator before touching the floor is one question: does the converted describe name already appear in `REQUIRED_SUITES`? Present means change nothing; absent means add the entry and raise the floor by its measured count. The pre-push full-suite run is what caught this, at the cost of a failed push; `bun test scripts/assert-db-tests-ran.test.ts` answers it in a second and belongs in the check list of any lane that edits the manifest at all, including one that edits only the floor.

rodaddy:
verify-lane receipt: check=scripts/done-means/878-pg-tests-require-database.sh exit=0 sha=eb359b61bfe578565db2fcb8a66abb646a5d073b at=2026-08-28T04:13:59.897Z

```
    server/application/sdk-protocol.pg.test.ts

CLAUSE 1 [server/application/sdk-protocol.pg.test.ts]: PASS -- 0 code-line process.env reads of OPENBRAIN_TEST_DATABASE_URL/OPENBRAIN_LOCAL_CLONE_TEST_DATABASE_URL/OPENBRAIN_SCRATCH_ADMIN_URL, and no describe.skip, skipIf, dbDescribe
CLAUSE 2 [server/application/sdk-protocol.pg.test.ts]: PASS -- imports requireTestDatabaseUrl|requireLocalCloneTestDatabaseUrl|requireScratchAdminUrl from a require-test-database module
CLAUSE 3 [server/application/sdk-protocol.pg.test.ts]: PASS -- exited 1 and printed test_database_required

```

</details>

PR #967, opened 2026-08-29
chore(l5): src runtime closure census, 864-moved-out-of-src done-means, and the L5 cut plan (issue 864)

rodaddy:
## Summary

- L5 of `_plans/server-hardening-ladder.md` (issue 864) needs measurement before any move lane runs. This lands the three tools and no moves.
- `scripts/src-runtime-closure.ts` walks runtime imports from `server/main.ts` over tracked non-test `.ts` files and prints every `src/` file it reaches. Type-only edges (`import type`, `export type`) are excluded because they vanish at build time; dynamic `import()` edges are followed because they run. The repo root comes from `git rev-parse --show-toplevel`, so it is correct from any directory inside a checkout. `--count` prints the number, `--parents` prints the breadth-first parent and the full importer list per file. Baseline recorded in the header comment: 67 at 42944000.
- `scripts/done-means/864-moved-out-of-src.sh <old src path> [more...]` judges each named path on three clauses: A the path is no longer tracked, B it is absent from the closure output, C no TypeScript importer still names the old module by a quoted specifier that does not route through `server/`. Exit 0 all clauses pass for all paths, 1 any clause fails with `FAIL <clause> <path>` lines and every offending line printed, 3 harness error.
- All three clauses accept a SHIM at the old path, per the head's revised rule M2: a move lane repoints only `server/` and `scripts/done-means` importers and leaves the old `src/` path holding one header comment plus an `export * from "../server/<dir>/<f>.ts"` line, because the pre-commit gate lints staged files whole and editing one import line in a 1236-line unconverted file drags its pre-existing findings into the gate. `is_shim` strips comment and blank lines and requires every surviving line to be an export naming a `server/` path. A shim is tracked and stays runtime-reachable, so clauses A and B admit it explicitly, and clause C is skipped for one — the importers M2 deliberately leaves alone are exactly what the shim exists to keep correct. Shims are `src/` files and retire with `src/` at L6.
- `_plans/864-l5-cut-plan.md` carries the measurement, the seven clusters with their home directories, the four-wave order, the expected closure after each wave, and the L6 preconditions read from the ladder at lines 291-360.

## Verification

That line is run verbatim, with no arguments, and it is meaningful that way. `scripts/validate-pr-body.ts` resolves the whole `Done-means` line as one repo-relative path, so a check that required arguments could never be driven from a PR body — run bare on the previous head it exited 3 as a harness error and the verify-lane posted no receipt. With no arguments the script now discovers its own subjects: every tracked `src/*.ts` that is an L5 shim by the same test clause A already applies, judged by the unchanged clauses A, B and C, with the discovered count printed as `judged=<n> shims` before the PASS/FAIL lines. An empty tree is an explicit `judged=0 shims` and exit 0, never a silent pass. Explicit arguments keep today's behavior exactly, and exit 3 is now reserved for `bun` or `rg` missing and for not being in a git tree. Three receipts, all taken on this branch: RED `bash scripts/done-means/864-moved-out-of-src.sh src/embedding.ts` -> `FAIL A` and `FAIL B`, exit 1, because that file is still a tracked implementation. GREEN `bash scripts/done-means/864-moved-out-of-src.sh` -> `judged=0 shims` and `PASS`, exit 0, because no `src/*.ts` on this branch is a shim yet. PROBE an untracked `src/l5-shim-probe.ts` re-exporting `../server/config`, `git add`ed so discovery can see it, then the same bare run -> `judged=1 shims` and `PASS`, exit 0, proving discovery finds a real shim and clause A accepts it; the probe was then unstaged with `git restore --staged` and moved out of the tree, leaving `git status` clean apart from the committed script. A fourth probe — a shim whose own path is still in the runtime closure, to make clause B fire — is not constructible: clause B is satisfied by shim-ness by design, so no shim can fail it, and that probe is recorded as skipped rather than faked.

`bunx tsc --noEmit` exit 0. `oxlint --deny-warnings scripts/src-runtime-closure.ts` exit 0 on the committed tree. The pre-push hook ran the full isolated suite and passed. RED proof: `bash scripts/done-means/864-moved-out-of-src.sh src/embedding.ts` exits 1 with `FAIL A` and `FAIL B` because that file is still in `src/`. GREEN proof: the same script on `src/shared-namespace-retired.ts` exits 0. Shim proof: a two-line shim written at `src/l5-shim-probe.ts` was accepted with `PASS` and exit 0, then archived out of the tree; `git status` is clean apart from the committed files. On the third commit: `bash -n scripts/done-means/864-moved-out-of-src.sh` exit 0, `bun scripts/src-runtime-closure.ts --count` -> 67 exit 0, `bunx tsc --noEmit` exit 0, and the pre-push hook ran the suite again and passed.

## Critical Self-Review

- Highest-risk behavior: the closure count is the number every downstream move lane checks itself against, so a wrong regex silently mis-sizes the whole L5 program. The regex excludes `import type` and `export type` by testing the matched text, and follows dynamic `import()`; a specifier form it does not match would under-count, which is the failure direction that lets a lane claim a move it did not make.
- Assumptions that could be wrong: that every runtime edge in this repo is a relative specifier resolvable to a tracked `.ts` file. Bare-specifier and generated-path imports are ignored by design; if any `src/` file is reached only through a path alias, it is missing from the 67.
- Missing/weak tests: neither script has a unit test. The done-means script carries its own red-and-green receipts instead, and the closure script's output is checkable by hand against the tree.
- Security/permission risk: none. Both scripts are read-only over the working tree; neither touches the database, a token, or a network endpoint.
- Migration/deploy risk: none. No migration, no runtime code, no schema, no deployed surface. Nothing under `src/` or `server/` changed.
- Downstream client/runtime risk: none. No MCP tool, transport, schema, or Python client path is touched, so `docs/downstream-rollout.md` does not apply.
- Rollback/cleanup concern: the three files are additive and independent; reverting the commit removes them and leaves the tree exactly as it was.
- Fixes made before PR: the first draft nested the breadth-first walk four levels deep and oxlint refused it, so the inner loop was hoisted into `visitFile`. Prettier then refused the commit on formatting and the file was reformatted and re-staged. Both checks were re-run against the committed tree, not the working copy. Then the head revised rules M2 and M7 mid-lane to introduce the shim, which invalidated all three clauses as first written; the second commit rewrites them and re-takes every receipt, including a new shim-acceptance proof. The third commit fixes the defect the head's verify-lane found: the `Done-means` line as the validator reads it carries no arguments, and the script exited 3 on it, so no receipt could ever be posted. Discovery from the tree is the fix, and the count is printed so an empty result cannot be mistaken for a pass.
- Known residual risk: the cut plan's C3 embedding home is a head decision and is marked UNVERIFIED in the document against `_plans/463-server-rewrite-charter.md`, which names no embedding directory. The C5 and C6 drifted twins are flagged as DANGEROUS and explicitly deferred to a behavior diff before any retarget.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this adds measurement tooling and a plan document, moves no code, and changes no behavior, so no new review pattern surfaced.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or database surface is touched.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, schema, or fixture is touched by this change.

## Downstream Rollout

Notes/evidence:

- Not applicable throughout: this change adds two repo scripts and one plan document. No MCP tool, schema, transport, or client-facing surface changes, so no downstream consumer sees anything different.

### DIFF
--- a/_plans/864-l5-cut-plan.md
+++ b/_plans/864-l5-cut-plan.md
@@ -1,0 +1,83 @@
+# L5 cut plan — move every src/ file into server/ (issue 864)
+
+**Status: WRITTEN 2026-08-29, no lane merged yet.**
+
+Authority: `_plans/server-hardening-ladder.md` L5 (lines 291-330) and L6 (lines
+332-360). L5's own done-means is "zero imports from `src/` in non-test
+`server/` code"; Rico's 2026-08-29 ruling widens it — everything moves out of
+`src/`, and `src/` retires at L6.
+
+## Measurement (42944000)
+
+| Quantity | Value |
+|---|---|
+| src/ files runtime-reachable from `server/main.ts` | 67 |
+| lines in those 67 | 30675 |
+| src/ files NOT runtime-reachable | 91 |
+| — of those, imported only by `scripts/` entrypoints | 12 |
+| — retired entrypoint trio (`src/index.ts`, `src/transport.ts`, `src/server.ts`) | retires at L6 |
+| direct modules / import sites, including type-only | 25 / 38 |
+| direct modules / import sites, runtime only | 23 / 29 |
+| oxlint findings across the 67 | 191 |
+
+Re-measure with `bun scripts/src-runtime-closure.ts --count`. The number falls
+by exactly the files a lane moves and never rises (rule M3).
+
+## Clusters and home directories
+
+| # | Files | Home | Note |
+|---|---|---|---|
+| C1a | `src/chunk-write.ts` | `server/capture/` | leaf |
+| C1a | `src/decomposition.ts` | `server/domain/` | leaf |
+| C1a | `src/secret-patterns.ts` | `server/security/` | leaf |
+| C1a | `src/nats-subjects.ts` | `server/config/` | leaf |
+| C1a | `src/tools/table-constants.ts` | `server/db/` | leaf |
+| C1b | `src/promotion-nomination.ts`, `src/promotion-service.ts`, `src/sharing.ts`, `src/tiering.ts` | `server/domain/` | promotion |
+| C3 | `src/embedding.ts`, `src/chunking.ts`, `src/embedding-canonical.ts`, `src/embedding-targets.ts` | `server/embedding/` | head decision; the charter names no embedding home. UNVERIFIED against `_plans/463-server-rewrite-charter.md` |
+| C2 | `src/audit-log.ts`, `src/logger.ts`, `src/rotating-file.ts` | `server/logging/` | MERGE with the existing logger; judgment lane |
+| C4 | `src/operator-doctor.ts` | `server/application/` | doctor |
+| C4 | `src/contract.ts`, `src/contract-schemas.ts` | `server/contracts/` | |
+| C4 | `src/nats-runtime.ts`, `src/qmd-path.ts`, `src/db/pool.ts`, `src/observability/*` | per-file | pool and observability are twins — diff before moving |
+| C5 | `src/rest-api.ts`, `src/rest-promotion.ts` | `server/transport/` | REST |
+| C5 | `src/namespace-policy.ts`, `src/permissions.ts`, `src/shared-namespace.ts`, `src/read-policy.ts`, `src/table-projections.ts`, `src/extraction.ts` | `server/auth/` or `server/domain/` | three have `server/auth` twins: RETARGET after a diff read |
+| C5 | `src/tools/search-brain.ts`, `fts-config`, `source-refs` | see note | reached ONLY by `src/rest-api.ts`, while `server/tools/search-brain.ts` is the registered MCP tool. DANGEROUS drifted twin: retarget REST search onto the server/ implementation after a behavior diff |
+| C6 | `src/nats-bridge.ts`, `src/auth.ts`, `src/background-tracing.ts` | `server/application/` | NATS bridge |
+| C6 | `src/tools/agent-context-pack.ts` + its seven `agent-context-pack-*.ts`, `src/realtime/recovery-wal.ts`, `src/realtime/working-set.ts`, `src/prior-context-suppression.ts` | see note | reached only through `nats-bridge`, while `server/tools/agent-context-pack.ts` is the registered tool. Same DANGEROUS twin shape as C5 |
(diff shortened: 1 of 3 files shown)


rodaddy:
harvested: #967 A done-means check written against a rule that is still being decided will judge the wrong thing, and it fails in the direction that looks like success. This lane's three clauses — old path untracked, absent from the runtime closure, no importer naming the old basename — were correct for the move shape the brief described, and were made wrong an hour later when the head introduced the shim: a move lane now leaves a re-export at the old `src/` path so that unconverted 1000-line importers stay out of the staged set the pre-commit gate lints whole. A shim is tracked, stays runtime-reachable, and keeps its importers pointing at the old name, so every clause failed a lane that had done exactly what the rule asked. The lesson for the next checker author is that the RED-and-GREEN pair is not sufficient evidence on its own: both receipts kept passing across the rule change, because they test the extremes (a file plainly still in place, a path that never existed) and the shim is the middle case the rule actually creates. Take a third receipt against the artifact the rule produces, not only against its two endpoints — here, writing a two-line shim and observing exit 0 is what proved the revised clauses, and nothing about the original pair would have caught their absence.

rodaddy:
harvested: #967 A done-means check is driven by `scripts/validate-pr-body.ts`, which resolves the entire `Done-means` line as one repo-relative path — it cannot carry arguments, and a check that requires them exits as a harness error when the verify-lane runs the line verbatim, so no receipt is posted and the pull request cannot merge no matter how correct the check itself is. The failure is invisible to the author, because running the script by hand with arguments passes every time. A done-means script must therefore be meaningful with zero arguments: it discovers its own subjects from the tree using the same predicate its clauses already apply, and it prints the discovered count (`judged=<n> shims`) before the verdict so that an empty discovery is an explicit, readable zero rather than a silent PASS that looks identical to a real one. Explicit arguments stay supported for hand-driven iteration, and the harness exit is narrowed to genuine harness faults — a missing tool or a non-checkout — so it can never again be reached by the ordinary invocation the merge gate uses.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=a82a3ceeea5f7c1147a86bf26b5bcf2410fe39de at=2026-08-29T20:09:24.431Z

```
judged=0 shims
PASS
```

</details>

PR #968, opened 2026-08-29
refactor(l5): move chunking, embedding-canonical, embedding-targets from src/ into server/embedding (issue 864)

rodaddy:
## Summary

- Cluster C3 of the L5 move program (issue 864): the embedding provider's helpers move out of `src/` into a new `server/embedding/` directory. The charter names no embedding home; the provider is an outbound adapter, not domain, so it gets its own directory rather than being folded into an existing one.
- Moved old -> new: `src/chunking.ts` -> `server/embedding/chunking.ts`, `src/embedding-canonical.ts` -> `server/embedding/embedding-canonical.ts`, `src/embedding-targets.ts` -> `server/embedding/embedding-targets.ts`, each with its own test file alongside it.
- **`src/embedding.ts` did NOT move.** It reads nine env keys directly and `server/config.ts` parses only two of them (`EMBEDDING_BASE_URL`, `EMBEDDING_API_KEY`). The other seven are `EMBEDDING_TIMEOUT_MS`, `EMBEDDING_DIMENSIONS`, `EMBEDDING_MODEL`, `EMBEDDING_WATCHDOG_FAILURE_THRESHOLD`, `EMBEDDING_WATCHDOG_COOLDOWN_MS`, and `EMBEDDING_WATCHDOG_RESTART_SCRIPT`. `EMBEDDING_MODEL` and `EMBEDDING_DIMENSIONS` are module-scope exported constants consumed as plain values by roughly fifty files, so passing them as parameters would have to edit every one of those callers — far past the one-import-line allowance the move rules grant. The rule for that case is to stop, leave the file where it is, and report the keys, which is what this PR does.
- Each old `src/` path is left as an L5 shim: one header comment and one `export *` line pointing at the new `server/` path. Importers outside the moved set keep their import lines untouched, which keeps their pre-existing lint findings out of the staged gate. The shims are `src/` files and retire with `src/` at L6.
- Two edits were forced by the `server/` lint standard, both behavior-preserving: `chunkText`'s inline break-point search is hoisted into a named `findBreakPoint` helper (same arithmetic, same order of preference — sentence boundary, then last word boundary, then the proposed end), and the moved tests swap non-null assertions for the existing `expectDefined` helper and split three oversized `describe` blocks into siblings. No test body, name, or assertion changed otherwise.

## Verification

`bunx tsc --noEmit` exit 0 on the committed tree. `oxlint --deny-warnings` exit 0 on every file in `server/embedding/` and on all three shims. `bun run test:isolated` on the three moved test files plus `server/main.tool-logging.test.ts`: 65 pass, 0 fail across 4 files. Closure figures on this branch: `bun scripts/src-runtime-closure.ts --count` -> 67 raw, and 64 with the three L5 shims excluded, which is the number that reflects what actually left `src/`.
`bunx tsc --noEmit` exit 0 on the committed tree. `oxlint --deny-warnings` exit 0 on every file in `server/embedding/` and on all three shims. `bun run test:isolated` on the three moved test files plus `server/main.tool-logging.test.ts` and `server/config*.test.ts`: 154 pass, 0 fail. Full isolated suite on this branch: 4010 pass, 0 fail across 289 files — identical to the same suite at `origin/main` (4010 pass, 0 fail), so the move changes no test outcome anywhere.
Done-means receipt: `bash scripts/done-means/864-moved-out-of-src.sh` with no arguments -> `judged=3 shims`, `PASS`, exit 0. The no-argument form discovers the three shims this lane created and judges them, which is also the form the PR-body validator can resolve, since it reads the whole Done-means field value as a single repo path. Red/green on the explicit paths: at `origin/main`, `bash scripts/done-means/864-moved-out-of-src.sh src/chunking.ts src/embedding-canonical.ts src/embedding-targets.ts` -> exit 1 with `FAIL A`/`FAIL B` for all three; the same command on this branch -> exit 0, `PASS`.
Done-means receipt: `bash scripts/done-means/864-moved-out-of-src.sh src/chunking.ts src/embedding-canonical.ts src/embedding-targets.ts` -> exit 0, `PASS`. (The Done-means field carries the bare script path because `scripts/validate-pr-body.ts` resolves that whole field value as a single repo path — arguments on the line make it unresolvable — so the argv this lane runs it with is recorded here instead, matching the convention in #877.) The Done-means line names only the three paths this lane actually moved; `src/embedding.ts` is deliberately excluded, and the script correctly reports `FAIL A`/`FAIL B` for it, which is the honest signal that it stayed put.
Rebased onto `main` at `8ca5a6f` after pull request #967 merged, so `scripts/done-means/864-moved-out-of-src.sh` and `scripts/src-runtime-closure.ts` now come from `main` rather than from this branch. The two tooling commits this branch previously carried are gone as duplicates; its diff is the single commit and the ten files listed above and nothing else.
**This branch is based on `chore/864-l5-tooling` (pull request #967), not on `main`,** because that branch is where `scripts/done-means/864-moved-out-of-src.sh` lives and the PR-body gate refuses a Done-means line naming a script it cannot find. Merge #967 first; this PR's own diff is the ten files listed above and nothing else.

## Critical Self-Review

- Highest-risk behavior: the `findBreakPoint` hoist inside `chunkText`. Chunk boundaries feed `contentHash`, so a shifted boundary would change stored hashes and make the embedding-repair registry read healthy rows as drifted. The extracted helper computes `zoneStart` once and returns the same three-way preference the inline code had; the moved chunking tests, including the sentence-boundary and overlap-seam cases, pass unmodified.
- Assumptions that could be wrong: that no importer relies on the old `src/` paths for anything beyond their exported names. `export *` forwards named exports only, and neither moved module has a default export (checked), but a consumer doing `import * as m` and reflecting over the namespace object would see the same keys, so this holds.
- Missing/weak tests: nothing new is tested here because nothing new exists — this is a move plus mechanical lint conformance. The `describe` splits are the one place a mistake would hide, since a stray brace could silently orphan tests; the test count is unchanged before and after, which is what rules that out.
- Security/permission risk: none. No auth, namespace predicate, SQL, or token handling is touched. `embedding-targets.ts` carries namespace-isolation metadata and moved byte-identical apart from prettier reflow and its own import paths.
- Migration/deploy risk: none. No migration, no schema change, no config change. The shims mean no runtime import path breaks even if something outside this repo's search reached the old paths.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python client surface changes; `docs/downstream-rollout.md` classifies this as not applicable.
- Rollback/cleanup concern: reverting the single commit restores the previous layout wholesale. The cleanup that remains is deliberate and scheduled — the three shims are removed at L6 when `src/` retires, and until then they hold all three files in the `src/` runtime closure.
- Fixes made before PR: reworked the lane from whole-tree import repointing to the shim approach after the move rules changed mid-lane, which removed 25 files of incidental churn and kept roughly a hundred pre-existing lint findings in `src/embedding-repair.ts` and the chunk-write done-means driver out of the staged gate.
- Known residual risk: the raw closure count stays at 67 rather than dropping to 64. That is inherent to the shim strategy, not a defect in this lane — the shims are themselves `src/` files, so the three-file reduction is realized at L6 when they are deleted. A reviewer expecting the count to fall this PR should know it falls later instead.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a file move with mechanical lint conformance and no behavior change, so it surfaced no reviewable defect pattern; the one durable lesson (a move lane's closure reduction is deferred when it leaves shims) belongs to the L5 program notes, not the review knowledge base.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or tool surface changes; the moved modules are pure text/registry helpers with no service behavior of their own.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing schema is touched — the pre-push contract-parity gate reported the client/contract paths unchanged and skipped its subset.

## Downstream Rollout

Notes/evidence:

- Internal file move within this repo. No MCP tool, schema, protocol, or client-facing surface changes, so every downstream step is not applicable by the classification in `docs/downstream-rollout.md`.


rodaddy:
harvested: #968 A move lane that leaves re-export shims at the old paths does NOT lower the `src/` runtime closure, and the count staying flat is the correct result rather than a lane failure. This lane moved three files and the closure held at 67 both before and after, because each shim is itself a `src/` file that `server/main.ts` still reaches through its unchanged importers; the three-file reduction is realized only when L6 deletes the shims. The shim strategy is still right — repointing every importer instead drags whole pre-existing-violation files like `src/embedding-repair.ts` (11 findings) and `scripts/done-means/605-chunk-write-routing.driver.ts` (6) into a pre-commit gate that lints staged files whole — but it means "closure went down" cannot be the acceptance signal for a shimmed move lane, and a controller expecting 67 -> 63 will read a correct lane as broken. Two smaller traps in the same shape: the PR-body validator resolves the entire `Done-means:` field value as ONE repo path, so a script with arguments never resolves and the argv belongs in the prose below the line (convention already set in #877); and a lane whose Done-means script lives on an unmerged sibling branch must rebase onto that branch, because the pr-body-gate hook refuses the create outright rather than warning.

rodaddy:
harvested: #968 A stacked branch whose parent PR merges with a fix-up commit does not need a conflict resolution so much as a subtraction: two of this branch's three commits were the parent's tooling in its pre-fix-up form, so every conflict hunk was the branch re-proposing a file that `main` already owned in a better version. Resolving each hunk by hand would have reconstructed the superseded text; taking `git restore --source=origin/main -- <path>` for all three conflicted paths emptied both commits, and `git rebase --skip` then dropped them cleanly, leaving exactly the one commit that carried the lane's own work. The generalisation for the next stacked lane: before resolving, run `git show --stat <conflicted-commit>` and ask whether the commit still contributes anything the base lacks — when it does not, an empty staged diff against HEAD is the signal to skip rather than continue, and the check is `git diff --cached --stat HEAD` printing nothing.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=71e2f288504a20e1cd72b8ff51bb64c848871cec at=2026-08-29T20:19:50.428Z

```
judged=3 shims
PASS
```

</details>

PR #969, opened 2026-08-29
refactor(l5): move promotion-nomination and sharing from src/ into server/domain (issue 864)

rodaddy:
## Summary

- Moves two modules out of `src/` into the `server/domain` layer they already belong to, part of the L5 step of issue #864: `src/promotion-nomination.ts` -> `server/domain/promotion-nomination.ts` and `src/sharing.ts` -> `server/domain/sharing.ts` (with `src/sharing.test.ts` -> `server/domain/sharing.test.ts`).
- Each old path keeps a re-export shim so callers outside this change resolve unchanged; the shims are marked for retirement together with `src/` at L6. `server/tools/promotion.ts`, `server/tools/scan-namespace.ts`, and `server/tools/ingest-conversation-facts.ts` import the new paths directly instead of going through a shim.
- `sharing.ts` did not pass the server-side lint gate as written: `classifyShareCandidate` (complexity 24) and `shareRejectionDetail` (complexity 19) both exceeded the maximum of 10. Inline work was hoisted into named helpers — `detectSecret`, `detectPrivate`, `buildShareRejectionDetail`, `shareDecisionLogContext`, `detectorLogFields`, and `noiseRejectionReason`. No behavior change; the moved test asserts the same cases.
- Two candidates were considered and deliberately left in `src/`. `tiering.ts` needs two max-params fixes that pull `scripts/tier-lane-durable.ts` and `scripts/done-means/605-chunk-write-routing.driver.ts` (28 pre-existing findings between them) into the gate, so those two scripts get brought to standard first in their own change. `promotion-service.ts` reads `OPENBRAIN_PROMOTION_KILL_SWITCH` from `process.env` directly, which the no-`process.env`-under-`server/` rule forbids; routing that through config is its own change rather than a side effect of a move.
- Closure count moves 67 -> 65; the two shims are excluded.

## Verification

Evidence:

- `bunx tsc --noEmit` -> exit 0, no diagnostics, re-run after the rebase onto `origin/main`.
- `./node_modules/.bin/oxlint --deny-warnings` over all eight touched paths -> exit 0. The same lint against `origin/main:src/sharing.ts` is exit 1 with the two complexity errors above, which is the red baseline this change turns green.
- `bun run test:isolated server/domain/sharing.test.ts src/tools/__tests__/ingest-conversation-facts.test.ts src/tools/__tests__/scan-namespace.test.ts src/tools/__tests__/ingest-conversation-facts.pg.test.ts src/tools/__tests__/ingest-conversation-facts-isolation.pg.test.ts` -> 73 pass, 0 fail, 262 expect() calls across 5 files, tests exited 0.
- `bash scripts/done-means/864-moved-out-of-src.sh` (no arguments, judging every shim on the tree) -> exit 0, `judged=2 shims`, PASS.
- Pre-commit and pre-push gates both passed without `--no-verify`: gitleaks clean, oxlint clean on staged content, prettier clean, `tsc` clean over the staged snapshot, Bun tests green.

## Critical Self-Review

- Highest-risk behavior: the shared-worthiness classifier in `sharing.ts`. `classifyShareCandidate` and `containsSecret` decide whether a nominated entry reaches the `shared-kb` namespace, and a secret leaking into shared truth is the worst failure mode in this module. The complexity hoists rewrote the control flow of exactly those functions, so a mistake there would be a silent policy change rather than a crash.
- Assumptions that could be wrong: that the hoisted helpers preserve short-circuit ORDER, not just the final verdict. Rejection precedence is observable through the reason string a caller receives, so a reordering that still rejects the same set would change reported reasons. The moved test asserts specific reasons, which is what makes this checkable rather than assumed.
- Missing/weak tests: no test pins the shim files themselves — nothing fails if `src/sharing.ts` stops re-exporting a symbol that only an out-of-tree caller uses. `tsc` catches every in-repo consumer, so the gap is limited to callers outside this repository.
- Security/permission risk: the secret-detection surface is the security control here and it is unchanged in substance; `detectSecret` is an extraction of the existing pattern loop. No namespace predicate, auth check, or token handling is touched by this change.
- Migration/deploy risk: none. No schema, no migration, no config, no startup path. The moves are import-graph-only and the shims mean no deployed caller needs to change in the same release.
- Downstream client/runtime risk: none identified against `docs/downstream-rollout.md`. No MCP tool, schema, protocol, or client-facing contract changes; the Python package and contract fixtures are untouched, which is why the pre-push gates skipped both.
- Rollback/cleanup concern: revert is a single commit with no state to undo. The remaining cleanup is deliberate and named above — the two shims retire with `src/` at L6, and `tiering.ts` plus `promotion-service.ts` are deferred with their blockers recorded so the next lane does not rediscover them.
- Fixes made before PR: the first commit attempt was blocked by the pre-commit gate because the complexity hoists were in the working tree but never staged, so the hook linted the older staged snapshot and failed on both functions. Staging the file was the fix; the hook was not bypassed.
- Known residual risk: `sharing.ts` behavior rests on the moved test rather than on a differential run of both implementations against shared input. The test covers the classifier's documented cases and passes, but it is a test-shaped guarantee, not a proof that no input classifies differently than before.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this change surfaced no new review finding — the lint failures it fixed are the pre-existing complexity debt the move exposed, and the gate already catches that class mechanically without a knowledge-base entry.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: this is an internal module move with no runtime, transport, or tool-surface change, so a live service smoke would exercise nothing this change touches.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, schema, or client-facing shape changes — the move is internal to the TypeScript server's import graph and no fixture describes it.

## Downstream Rollout

Notes/evidence:

- The rollout gate classifies this as not applicable: no MCP tool, schema, protocol, or client-facing change. The pre-push gates confirmed it independently, skipping the Python package and contract-parity subsets because neither path was touched.


rodaddy:
harvested: #969 The pre-commit gate lints STAGED CONTENT, not the working tree, and the two are routinely different during a move-plus-refactor lane. This lane's first commit attempt failed on both complexity errors in `server/domain/sharing.ts` even though the working-tree file already had the hoists that fix them: the file was at `AM` status, staged at its pre-hoist content, and the hook faithfully judged what was staged. The tell is in the hook output itself — the reported line numbers (184, 277) did not match the working-tree definitions (225, 329), which is the cheap signal that the snapshot under review is not the file you just edited. Reading the failure as "my fix did not work" leads to re-editing code that was already correct; reading the line-number mismatch leads straight to `git add`. A lane that edits a path after staging it should re-stage before committing, and `git status --short` showing `AM` rather than `A` or `M` is that condition made visible before the hook has to catch it.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=f046fdad51af646e9e1ffcff485dd386bb444097 at=2026-08-29T20:14:03.237Z

```
judged=2 shims
PASS
```

</details>

PR #975, opened 2026-08-29
feat(config): parse the env keys the remaining src/ modules read so they can move under server/ (issue 864)

rodaddy:
## Summary

- L2 of `_plans/server-hardening-ladder.md` made `server/config.ts` the single place the environment is parsed, and `.oxlintrc.json` permits `process.env` only in `server/config.ts` and `server/main.ts`. Four L5 move lanes stopped there: the module each was moving reads a key the door does not parse, so moving it would have put a `process.env` read behind a rule that forbids one. **This lane widens the door. It moves nothing**, edits no `src/` file, and changes no existing parsed key or default.
- New fields live in `server/config/src-readers.ts`, following the split `server/config/nats.ts` and `server/config/maintenance.ts` already use: `server/config.ts` is under the 500-code-line rule and lands at 450 as it is, so a new section went in its own module rather than onto the end.
- Keys added, each with the same default and coercion its current `src/` reader applies:

| Key | `src/` reader | Default | Coercion |
|---|---|---|---|
| `EMBEDDING_TIMEOUT_MS` | `src/embedding.ts:6-7` | `8000` | `parseInt` base 10; `NaN` falls back, a negative value passes through (the guard is `isNaN` only) |
| `EMBEDDING_DIMENSIONS` | `src/embedding.ts:8-10` | `768` | `parseInt`; `NaN` or `<= 0` falls back |
| `EMBEDDING_MODEL` | `src/embedding.ts:36` | `gemini-embedding-001` | raw string, `??` on absent |
| `EMBEDDING_WATCHDOG_FAILURE_THRESHOLD` | `src/embedding.ts:99-104` | `2` | `parseInt`; `NaN` or `<= 0` falls back |
| `EMBEDDING_WATCHDOG_COOLDOWN_MS` | `src/embedding.ts:107-112` | `300000` | `parseInt`; `NaN` or `< 0` falls back, so `0` is legal |
| `EMBEDDING_WATCHDOG_RESTART_SCRIPT` | `src/embedding.ts:140-141` | absent | raw string, absent when falsy (`:141` returns early) |
| `OPENBRAIN_PROMOTION_KILL_SWITCH` | `src/promotion-service.ts:235` | `false` | exact `=== "1"`; `true`/`yes` do NOT enable |
| `OPENBRAIN_MCP_AUDIT_ENABLED` | `src/audit-log.ts:138` | `true` | exact `!== "0"` — audit is on unless explicitly disabled |
| `OPENBRAIN_MCP_AUDIT_RETENTION_DAYS` | `src/audit-log.ts:139-144` | `30` | `readBoundedInt` 1..366 |
| `OPENBRAIN_MCP_AUDIT_CLEANUP_INTERVAL_MS` | `src/audit-log.ts:145-150` | `3600000` | `readBoundedInt` 60000..86400000 |
| `OPENBRAIN_MCP_AUDIT_WRITE_TIMEOUT_MS` | `src/audit-log.ts:151-156` | `1000` | `readBoundedInt` 50..5000 |
| `QMD_INDEX_PATH` | `src/operator-doctor.ts:433-434` | absent | raw string, branch is `!== undefined`, untrimmed |
| `NODE_ENV` | `src/operator-doctor.ts:649-652` | `unknown` | one of `production`/`development`/`test`, else `unknown`; it was NOT already exposed by `server/config.ts` |
| `LOG_MAX_BYTES`, `LOG_MAX_FILES` | `src/operator-doctor.ts:685-686` | `false` | `Boolean(a) \|\| Boolean(b)` as `doctor.rotationConfigured` |

- `QMD_PATH` (`src/qmd-path.ts:17`) and `LOG_FILE` (`src/operator-doctor.ts:637`) were checked and are **already parsed** by `server/config.ts` (`QMD_PATH` via `server/config/env-groups.ts`, `LOG_FILE` in the base schema), so they are not duplicated. `EMBEDDING_BASE_URL` and `EMBEDDING_API_KEY` were likewise already parsed and are untouched.
- Exposed as `config.embedding` (with a nested `watchdog`), `config.promotion`, `config.mcpAudit`, `config.doctor` — typed, in the sections where they belong.
- **Which parse function matters, and the parsers mirror rather than tidy.** `src/embedding.ts` uses bare `parseInt`, so `9000ms` is 9000 and `1e3` is 1; `src/audit-log.ts` `readBoundedInt` tests `/^[0-9]+$/` *before* parsing, so both of those fall back there. They are reproduced as two separate helpers. Every reader here falls back on an unusable value instead of failing, so no new field rejects: tightening one would turn a currently-booting deployment into `server_configuration_invalid`, a behavior change smuggled in under a typing change.
- **One behavior cannot be reproduced in a pure parse, and is called out here rather than papered over.** `src/embedding.ts` reads its two watchdog integers *lazily*, inside `watchdogFailureThreshold` (`:99`) and `watchdogCooldownMs` (`:107`), so a test that flips `process.env` mid-process changes the answer there. A parse at the composition root is necessarily the read-once-at-startup shape. The default and the coercion are identical either way, so the key is parsed with the same default and the consumer lane decides how to thread it. (`EMBEDDING_MODEL` and the two module constants at `:6-10` are already read once at import, so they have no such gap.)
- `server/config-equivalence.test.ts` was read before touching anything: it drives its assertions from the real readers — importing exported ones (`resolveFtsConfig`, `sharedNamespaceConfig`, `parseAllowedOrigins`, `readMcpTracingConfig`) and writing out module-private expressions beside their file and line — over a fixed `DIVERGENCE_INPUTS` set, so the start-equivalence claim is falsifiable. Every reader it covers lives under `server/`. The readers this lane mirrors live under `src/`, which `server/` must not import, so it needed no change; the equivalent coverage is written out beside its coordinates in the new test file instead.

## Verification

RED, at `origin/main`, with only `server/config.ts` reverted to main's copy: `bun test server/config-src-readers.test.ts` -> `TypeError: undefined is not an object (evaluating 'embedding.timeoutMs')`, the suite failing because the parsed object has no `embedding` section.

GREEN, on the committed tree: `bun run test:isolated server/config.test.ts server/config-equivalence.test.ts server/config-extended.test.ts server/config-src-readers.test.ts server/main.pg.test.ts` -> 113 pass, 0 fail, 355 expect() calls across 5 files. `bunx tsc --noEmit` -> exit 0. `oxlint --deny-warnings --config .oxlintrc.json` on all three staged paths -> exit 0. `bash scripts/done-means/864-moved-out-of-src.sh` -> `judged=14 shims` / `PASS`. Python package untouched; no migration, no schema, no runtime surface changed.

## Critical Self-Review

- Highest-risk behavior: a parser here that silently disagrees with its `src/` reader. When the consumer lane rewires a module to take the injected value, a divergence becomes a live behavior change at the moment of the move rather than at the moment of this PR — so it would land far from its cause. Mitigated by writing each parser against the reader's own parse function and asserting the divergence inputs (`9000ms`, `1e3`, `1.5`, `1000ms`, `-1`, `0`, `""`) rather than only the happy path.
- Assumptions that could be wrong: that `parseInt` versus `/^[0-9]+$/` is the whole of the difference between the two integer families. If a reader has a third behavior I did not read — an implicit trim, a unit suffix handled elsewhere — my mirror is wrong for that input. I read each of the five reader sites in full rather than grepping for the key name, which is the check available without moving anything.
- Missing/weak tests: no test proves the parsed value equals what the live `src/` reader answers on the same input, because `server/` cannot import `src/`. The equivalence is asserted against expressions written out beside their file and line instead, which fails on an edit to this side but not on an edit to the `src/` side. That gap closes when the module moves and the two become one function.
- Security/permission risk: none introduced. No namespace predicate, auth token, or role check is touched. `OPENBRAIN_MCP_AUDIT_ENABLED` was deliberately given the `!== "0"` default-on semantics its reader has, because the failure mode of getting that backwards is silently disabling the record of who called what.
- Migration/deploy risk: none. No field rejects, so no environment that starts the server today stops starting it. Verified by construction — every new parser has a fallback branch and none returns a Zod issue — and by the malformed-value tests, which assert the fallback rather than a throw.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, protocol, or Python client surface changes; `docs/downstream-rollout.md` classifies this as not applicable. The parsed object gains four properties and loses none, so no existing consumer is affected.
- Rollback/cleanup concern: revert of one commit, three files, two of them new. Nothing was deleted and no `src/` reader was edited, so the old path is intact and still the one actually running.
- Fixes made before PR: prettier's reflow of the `NODE_ENV` loop narrowed the inferred element type to `string` and broke the `toBe` overload, caught by re-running `tsc` on the formatted tree rather than the pre-format one; fixed with `as const`. A stray doc comment left over from an earlier draft named a constant that no longer existed and was corrected to describe `KNOWN_NODE_ENVIRONMENTS`.
- Known residual risk: the lazy-read gap named in the Summary. `EMBEDDING_WATCHDOG_FAILURE_THRESHOLD` and `EMBEDDING_WATCHDOG_COOLDOWN_MS` are re-read per call today and would be read once at startup through this config; anything relying on flipping them mid-process would need the consumer lane to keep a lazy accessor. Nothing consumes the new fields yet, so the risk is entirely in the future move, and it is recorded here for the lane that makes it.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: nothing new was learned that a future reviewer of a different PR would need — the one reusable point (mirror the reader's own parse function, `parseInt` and `Number` are not interchangeable) is already the standing rule in `server/config/env-groups.ts`'s header, and this lane followed it rather than discovering it.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: nothing consumes the new fields yet, so no running behavior changes and there is nothing to smoke.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing schema is touched; this is internal config parsing behind the server boundary.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool, schema, transport, protocol, or Python client surface changes. The change is confined to `server/config.ts`, a new `server/config/src-readers.ts`, and a new test file.

### DIFF
--- a/server/config-src-readers.test.ts
+++ b/server/config-src-readers.test.ts
@@ -1,0 +1,196 @@
+/**
+ * The env keys the remaining non-`server/` modules read, parsed at the door.
+ *
+ * Split into its own file rather than grown onto `server/config.test.ts` for
+ * the same reason #868 split `config-equivalence.test.ts` out: each file stays
+ * under the max-lines rule, so the shared REQUIRED baseline is repeated.
+ *
+ * Every expectation names the reader it mirrors by file and line. The readers
+ * live under `src/`, which `server/` must not import, so a module-private
+ * expression is written out beside its coordinates instead of being called —
+ * an edit to either side then fails a test rather than drifting apart quietly.
+ */
+import { describe, expect, it } from "bun:test";
+import { parseServerConfig } from "./config.ts";
+
+const REQUIRED = {
+  DB_HOST: "db.internal",
+  DB_NAME: "open_brain_test",
+  DB_USER: "open_brain",
+  LOG_FILE: "logs/open-brain.log",
+};
+
+function configFrom(overrides: Record<string, string | undefined> = {}) {
+  const result = parseServerConfig({ ...REQUIRED, ...overrides });
+  if (!result.ok) {
+    throw new Error(`expected valid configuration: ${JSON.stringify(result.issues)}`);
+  }
+  return result.config;
+}
+
+describe("embedding keys — src/embedding.ts", () => {
+  it("defaults to the module constants when every key is unset", () => {
+    const embedding = configFrom().embedding;
+    // `:6-9`, `:36`, `:99-112`.
+    expect(embedding.timeoutMs).toBe(8_000);
+    expect(embedding.dimensions).toBe(768);
+    expect(embedding.model).toBe("gemini-embedding-001");
+    expect(embedding.watchdog.failureThreshold).toBe(2);
+    expect(embedding.watchdog.cooldownMs).toBe(300_000);
+    expect("restartScript" in embedding.watchdog).toBe(false);
+  });
+
+  it("carries a set value through", () => {
+    const embedding = configFrom({
+      EMBEDDING_TIMEOUT_MS: "1200",
+      EMBEDDING_DIMENSIONS: "1536",
+      EMBEDDING_MODEL: "embed-gemma-dense",
+      EMBEDDING_WATCHDOG_FAILURE_THRESHOLD: "5",
+      EMBEDDING_WATCHDOG_COOLDOWN_MS: "0",
+      EMBEDDING_WATCHDOG_RESTART_SCRIPT: "/opt/bounce.sh",
+    }).embedding;
+    expect(embedding.timeoutMs).toBe(1_200);
+    expect(embedding.dimensions).toBe(1_536);
+    expect(embedding.model).toBe("embed-gemma-dense");
+    expect(embedding.watchdog.failureThreshold).toBe(5);
+    // Zero is a legal cooldown: the reader guards `< 0`, not `<= 0` (`:111`).
+    expect(embedding.watchdog.cooldownMs).toBe(0);
+    expect(embedding.watchdog.restartScript).toBe("/opt/bounce.sh");
+  });
+
+  it("falls back on a malformed value instead of rejecting it", () => {
+    // Every reader here answers its default on `NaN` rather than throwing, so
+    // an environment that boots today must still boot.
+    const embedding = configFrom({
+      EMBEDDING_TIMEOUT_MS: "abc",
+      EMBEDDING_DIMENSIONS: "0",
+      EMBEDDING_WATCHDOG_FAILURE_THRESHOLD: "-1",
+      EMBEDDING_WATCHDOG_COOLDOWN_MS: "-1",
+    }).embedding;
+    expect(embedding.timeoutMs).toBe(8_000);
+    expect(embedding.dimensions).toBe(768);
+    expect(embedding.watchdog.failureThreshold).toBe(2);
+    expect(embedding.watchdog.cooldownMs).toBe(300_000);
+  });
+
+  it("keeps `parseInt` semantics rather than `Number` ones", () => {
+    // `parseInt("8000ms", 10)` is 8000 and `parseInt("1e3", 10)` is 1. A
+    // `z.coerce.number()` mirror would answer NaN and 1000 — the second is a
+    // thousand-fold divergence from a live deployment.
+    const embedding = configFrom({
+      EMBEDDING_TIMEOUT_MS: "9000ms",
+      EMBEDDING_DIMENSIONS: "1e3",
+    }).embedding;
+    expect(embedding.timeoutMs).toBe(9_000);
+    expect(embedding.dimensions).toBe(1);
+  });
+
+  it("treats a blank restart script as no restart configured", () => {
+    // `:141` returns early on any falsy value.
+    const embedding = configFrom({
+      EMBEDDING_WATCHDOG_RESTART_SCRIPT: "",
+    }).embedding;
+    expect("restartScript" in embedding.watchdog).toBe(false);
+  });
+});
+
+describe("promotion kill switch — src/promotion-service.ts:235", () => {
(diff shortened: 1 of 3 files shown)


rodaddy:
harvested: #975 A pure-parse config boundary cannot reproduce a *lazily* read env var, and the difference is invisible until the consumer moves. `src/embedding.ts` reads `EMBEDDING_WATCHDOG_FAILURE_THRESHOLD` and `EMBEDDING_WATCHDOG_COOLDOWN_MS` inside the functions that use them (`:99`, `:107`), so a test flipping `process.env` mid-process changes the answer; the same file's other keys (`:6-10`, `:36`) are read once at import and have no such gap. Parsing at the composition root is necessarily the read-once shape, so the two are equivalent on value and default but not on re-read behavior. The rule for a door-widening lane: read each reader's *call site*, not just its expression, and when a key is lazy, say so in the PR body and parse it anyway with the same default rather than inventing a lazy accessor the consumer may not want — the lane that actually moves the module is the one holding enough context to decide. Silently making it eager would land a behavior change far from its cause.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=2109c5e9d4e8e6a5d5053296b40c0b3410478166 at=2026-08-29T20:40:38.928Z

```
judged=14 shims
PASS
```

</details>

PR #976, opened 2026-08-29
refactor(l5): move distill-window and distiller from src/ into server/domain (issue 864)

rodaddy:
## Summary

- L5 of `_plans/server-hardening-ladder.md` (issue 864): two source modules move out of `src/` into `server/domain/`.
- `src/distill-window.ts` -> `server/domain/distill-window.ts`
- `src/distiller.ts` -> `server/domain/distiller.ts`
- Both old `src/` paths become shims (one header comment plus one `export *` naming the `server/` path), so every legacy importer — `src/distill-handler.ts`, `src/distill-exchange.ts`, `src/maintenance-sweep.ts`, `src/grading-*`, `src/dream-rem.ts`, `scripts/distill-run.ts`, and the rest — keeps its import line untouched and stays unstaged.
- The moved files now meet the `server/` standard whole, because the pre-commit gate lints staged files entire. `claimDistillBatch` was 117 code lines at complexity 11; its bind-parameter and SQL-fragment assembly is hoisted into module-scope `buildClaimBindings`, and its row decoding plus due-set scan into `toDistillTurn` / `readClaimRows`. Placeholder numbering and the fragments that name those placeholders are now produced in one function, which is the pairing a hand-written `$n` beside a `params.push` gets wrong.
- Non-null assertions on indexed reads are replaced by reading the element and taking the same branch that the old code took: `buildUnits` skips a missing current turn and stops the backward context walk, `nearestSpeechContext` skips, and `readClaimRows` iterates rows by value rather than indexing two arrays in parallel.
- Imports of other `src/` modules inside the moved files are rewritten to `../../src/` per M3; those files belong to later lanes and are not pulled along.
- Tests left in `src/` (retire or move at L6): `src/distill-window.test.ts` and `src/distiller.test.ts` stay byte-identical at their `src/` paths and exercise the moved code through the shims. Neither is staged, so the gate never sees them.
- Closure of `src/` under `server/main.ts`: 50 -> 48, exactly the two files moved.
- No behavior change beyond what the size and assertion rules forced.

## Verification

`bunx tsc --noEmit` exit 0. `./node_modules/.bin/oxlint --deny-warnings --config .oxlintrc.json` on all four staged paths exit 0. `bun run test:isolated src/distill-window.test.ts src/distiller.test.ts` — 69 pass, 0 fail, 182 expect() calls, which is the proof the shim route resolves. `bash scripts/done-means/864-moved-out-of-src.sh src/distill-window.ts src/distiller.ts` -> PASS, exit 0. The pre-push hook ran the full isolated suite and passed. Neither module has a `*.pg.test.ts`, so `scripts/assert-db-tests-ran.ts` is untouched.

## Critical Self-Review

- Highest-risk behavior: the `undefined` guards that replaced non-null assertions in `buildUnits`. The inner walk uses `break` and the outer uses `continue`, matching what the assertion would have done at that spot — a `continue` in the inner loop would silently skip a session boundary check and pull an unrelated session's turn into context. The loop bounds mean neither guard can fire today; they exist because the rule forbids the assertion.
- Assumptions that could be wrong: that `buildClaimBindings` reproduces the exact placeholder numbering. The push order is unchanged and the fragments are built from `params.length` at the same points, and the 69 tests plus tsc cover the shape, but the SQL is exercised against a live table only by the distill handler's own tests.
- Missing/weak tests: no test was added. The two existing suites move nothing and assert nothing new; they are the regression net and they pass unchanged. `claimDistillBatch`'s SQL path is covered indirectly rather than by a dedicated query test.
- Security/permission risk: none. The namespace and lane predicates are built from the same bind parameters in the same order; no predicate became optional, and nothing was interpolated that was previously bound.
- Migration/deploy risk: none. No migration, no schema touch, no config touch. The shims keep the `src/` module specifiers resolvable, so nothing that imports by the old path changes at runtime.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python client path is touched; `docs/downstream-rollout.md` classifies this as not applicable.
- Rollback/cleanup concern: the two shims are `src/` files and are removed with `src/` at L6, not by this lane. The two test files are the other L6 item and are named in the summary so they are not lost.
- Fixes made before PR: hoisted `claimDistillBatch` to clear the 117-line and complexity-11 findings; removed six non-null assertions across the two files; rewrote the moved files' `src/` imports to `../../src/`.
- Known residual risk: `readClaimRows` types its input as `Record<string, unknown>` and the field casts are carried over verbatim from the old `rows.map`, so a column rename would still fail at runtime rather than at compile time. That is the pre-existing shape, not something this change introduced.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical L5 move with no new defect class — the hoist-and-guard pattern is already the established program transformation across the earlier move lanes.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or tool surface changes; the change is a file location and internal function shape.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing schema is touched by a module move plus shims.

## Downstream Rollout

Notes/evidence:

- Not applicable on every downstream row: no MCP tool, schema, protocol, transport, or Python client path changed. The only externally visible artifact is the location of two TypeScript modules within this repo, and the old paths still resolve through the shims.


rodaddy:
harvested: #976 A move lane's real cost is not the `git mv` but the whole-file lint the pre-commit gate applies to the new path, and the two findings that fire together on an older `src/` module are size and indexed-access assertions. Both have a mechanical answer that costs no behavior: a long query function splits along its own seam — parameter/fragment assembly is one function, row decoding is another, and the split is worth making because placeholder numbering and the SQL fragments that name those placeholders must agree, which is exactly what a hand-written `$n` beside a `params.push` in a long body gets wrong. The assertions convert by reading the element and taking the branch the assertion implied at that spot, and the branch is NOT uniform: in `buildUnits` the outer loop wants `continue` and the inner context walk wants `break`, because a `continue` there would skip the session-boundary check and pull an unrelated session's turn into a unit's context. Reaching for `continue` in both places is the one way this transformation changes behavior, and nothing in the suite would have caught it.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=27d4e1d89ad167cf0ed216d5c99a0a0458285471 at=2026-08-29T20:48:49.304Z

```
judged=19 shims
PASS
```

</details>

PR #979, opened 2026-08-29
refactor(l5): move dream-light and dream-rem from src/ into server/maintenance (issue 864)

rodaddy:
## Summary

- L5 of `_plans/server-hardening-ladder.md` (issue 864): moves the two DREAM maintenance stages out of `src/` into `server/maintenance/`. Neither module reads `process.env`, so both move as-is.
- `src/dream-light.ts` -> `server/maintenance/dream-light.ts`
- `src/dream-rem.ts` -> `server/maintenance/dream-rem.ts`
- Split by responsibility so each moved file meets the `server/` function rules, with no behavior change: `server/maintenance/dream-light-occurrences.ts` holds `SweepRow` plus the per-turn `content_occurrences` upsert lifted out of `runLightSweep` (which also removes the four-deep nesting the sweep loop carried); `server/maintenance/dream-rem-grading.ts` holds the candidate row shape, the row -> `RemCandidate` widening, the traced grade call, and the `machine_grade` UPDATE lifted out of `runRemGrading`.
- Also hoisted the `runRemPass` argument assembly out of the `dream.rem` handler body into `buildRemPassDeps`, and replaced the single non-null assertion in `readLightQueueDepth` with an explicit guard on a query of four scalar sub-selects that always returns one row.
- Shims per M2: the two old `src/` paths now re-export from `server/maintenance/`, so every legacy `src/` and `scripts/` importer keeps its import line untouched. The only importer repointed is the one already at the lint standard, `server/domain/distill-window.ts`.
- Imports of already-moved modules (`background-tracing`, `candidate-dedupe`) name their `server/` paths directly; imports of still-in-`src/` modules (`maintenance-queue`, `tools/ingest-raw-turn`) use `../../src/` per M3.
- DreamEngine dry-run behavior is an invariant and is untouched: nothing in either module gained an archive, promote, demote, or tier mutation, and no default changed.
- Runtime closure of `src/` under `server/main.ts`: 46 -> 44, exactly the two files moved.

Tests left in src/ (retire or move at L6), all green through the shim: `src/dream-light.test.ts`, `src/dream-rem.test.ts`, `src/rem-terra-grader.test.ts`, `src/distill-window.test.ts`, `src/maintenance-bootstrap.test.ts`.

## Verification

`bash scripts/done-means/864-moved-out-of-src.sh src/dream-light.ts src/dream-rem.ts` -> exit 0, `PASS`.
`bunx tsc --noEmit` -> exit 0.
`./node_modules/.bin/oxlint --deny-warnings --config .oxlintrc.json` over all seven touched paths -> exit 0.
`bun run test:isolated src/dream-light.test.ts src/dream-rem.test.ts src/rem-terra-grader.test.ts src/distill-window.test.ts src/maintenance-bootstrap.test.ts` -> exit 0, 117 pass / 0 fail.
Pre-push hook ran the full isolated suite -> passed.
Python package is untouched and `rg -n 'dream-light|dream-rem' python/openbrain-memory/tests` returns no hits, so the M10 source-path guards need no edit.

## Critical Self-Review

- Highest-risk behavior: the `content_occurrences` upsert now returns an outcome word instead of mutating the summary inline. The old code incremented `updated` and then `corroborated` separately; the new `"corroborated"` branch increments both, which preserves the original counts. `src/dream-light.test.ts` covers created / updated / corroborated and passes unchanged.
- Assumptions that could be wrong: that `readLightQueueDepth`'s query cannot return zero rows. It is four scalar sub-selects with no FROM, so Postgres always yields one row; the guard throws rather than returning zeros, which is stricter than the assertion it replaced but unreachable in practice.
- Missing/weak tests: the two new sibling modules have no direct tests of their own; they are exercised only through `runLightSweep` and `runRemGrading`. That matches how the extracted code was tested before the split.
- Security/permission risk: none. No namespace predicate, auth check, or SQL parameterization changed; the moved SQL is byte-identical apart from indentation.
- Migration/deploy risk: none. No schema change and no migration.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python client surface changed; every legacy import path still resolves through the shims.
- Rollback/cleanup concern: the two shims are `src/` files removed at L6 with the rest of `src/`, not by this lane.
- Fixes made before PR: repointed `server/domain/distill-window.ts` at the new path; removed the now-unused `isHarnessNoise` import from `dream-light.ts`; corrected the extracted helper's parameter type to the exported `RemDeps`.
- Known residual risk: the five test files still live in `src/` and reach the moved code through the shims, so a shim removal before L6 would break them.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical move plus a responsibility split with no new failure pattern; the L5 move rules already live in the lane brief.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or tool surface changed.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing shape is touched by a file move.

## Downstream Rollout

Notes/evidence:

- Internal module relocation only. No MCP tool, schema, protocol, or client-facing behavior changed, so no downstream rollout step applies.


rodaddy:
harvested: #979 A move lane that has to split an over-rule function should extract the loop BODY, not carve the function in half: lifting the per-turn `content_occurrences` upsert out of `runLightSweep` into a named sibling cleared max-lines-per-function, complexity, AND max-depth in one edit, because the nesting the depth rule flagged was the loop's inner branch, and it left the sweep reading as claim -> record -> stamp with the counts still assembled in one place. The same shape worked on `runRemGrading`. The one thing to watch is that a summary counter incremented in two places inline (`updated` then `corroborated`) has to become a single outcome word whose branch increments both, or the corroboration count silently drops.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=3f10eb50e75a78ba27f39756e964ec6f52e6ce11 at=2026-08-29T21:04:34.193Z

```
judged=23 shims
PASS
```

</details>

PR #978, opened 2026-08-29
refactor(l5): move embedding-repair from src/ into server/embedding (issue 864)

rodaddy:
## Summary

- L5 move of `embedding-repair` out of `src/` into `server/embedding/`, split by responsibility. Old -> new: `src/embedding-repair.ts` -> `server/embedding/embedding-repair.ts` (orchestration), `server/embedding/embedding-repair-selection.ts` (row selection), `server/embedding/embedding-repair-model.ts` (types and constants).
- `src/embedding-repair.ts` remains as a re-export shim so every existing caller keeps working; it is retired with `src/` at L6.
- tests left in src/: `src/embedding-repair.test.ts` stays at its `src/` path, byte-identical to `main` and unstaged, per the M1 exception -- 963 lines with 54 standards findings does not fit this lane, and it exercises the moved code through the shim. `src/embedding-repair.pg.test.ts`, `src/embedding-repair-formula.pg.test.ts`, and `src/embedding-repair-test-helpers.ts` stay in `src/` for the same reason.
- No behavior change: the move is mechanical plus the split, with no edits to logic.

## Verification

- Done-means: scripts/done-means/864-moved-out-of-src.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

`bunx tsc --noEmit` -> exit 0. `oxlint --config .oxlintrc.json --deny-warnings` on all four touched files -> exit 0. `bun run test:isolated src/embedding-repair.test.ts src/embedding-repair.pg.test.ts src/embedding-repair-formula.pg.test.ts` -> exit 0, 69 pass 0 fail. The five `server/domain/graph-derivation-*.test.ts` split tests -> exit 0, 24 pass 0 fail. `scripts/done-means/864-moved-out-of-src.sh` -> judged=18 shims PASS at `origin/main`, judged=19 shims PASS with this branch applied. `bun scripts/src-runtime-closure.ts` -> 66 entries both before and after: the shim deliberately keeps `src/embedding-repair.ts` inside the runtime closure at L5, so the count does not drop until the shim retires at L6.

## Critical Self-Review

- Highest-risk behavior: the repair orchestration path that selects rows and rewrites embeddings. A wrong split boundary would silently change which rows are picked. Mitigated by the pre-existing 69-test suite running unmodified against the moved code through the shim.
[...]

PR #980, opened 2026-08-29
chore(l5): done-means clause A accepts an L5 adapter beside a shim (issue 864)

rodaddy:
## Summary

- Rule M9 lets a legacy `src/` path become an ADAPTER rather than a shim when the caller uses a call form the `server/` version cannot keep. An adapter holds its own code, so the shim re-export test rejects it; clause A of `scripts/done-means/864-moved-out-of-src.sh` needed a second way to accept one.
- `is_adapter`: the file's first line matches `^// L5 (shim|adapter) \(issue 864\)` AND every relative specifier it names -- `from "./..."`, `from "../..."`, `import("../...")` -- resolves under a `server/` path. Node and npm specifiers are not relative and are not judged.
- `adapter_bad_specifiers` prints the offending lines, so a `FAIL A` on a declared adapter names the specifier that broke it instead of just the path.
- `declares_l5_header` drives the no-argument discovery, so a file that claims the M9 header while naming a non-`server/` relative specifier stays in the judged set and fails out loud, rather than dropping out of the set unseen. That was a real defect found by the second probe, not a theoretical one.
- Clauses B and C skip an adapter exactly as they already skip a shim: rule M9 leaves those importers' lines untouched on purpose, and the adapter is what makes them correct.
- The count line stays `judged=<n> shims` on a shim-only tree and becomes `judged=<n> shims, <m> adapters` once any adapter is present. Nothing exported by the script is renamed.
- `_plans/864-l5-cut-plan.md`: new `## L5 adapters` section carrying rule M9 verbatim plus the three candidates (`src/operator-doctor.ts`, `src/promotion-service.ts`, `src/audit-log.ts`), and the stale `no lane merged yet` status replaced with the merged pull requests #967-#976 and #978.
- No `src/` or `server/` file is touched. The 19 shims already on `main` are unaffected -- none carries the M9 header and all 19 stay accepted by the existing re-export test.

## Verification

Receipts:

```
bash scripts/done-means/864-moved-out-of-src.sh   (at origin/main, before the change)
  -> judged=19 shims / PASS, exit 0

bash scripts/done-means/864-moved-out-of-src.sh   (committed tree, after the change)
  -> judged=19 shims / PASS, exit 0   -- same set, unchanged

PROBE (accepted): untracked src/l5-adapter-probe.ts, git add-ed so discovery sees it
  // L5 adapter (issue 864): legacy call form over server/config.ts; retired with src/ at L6.
  import { x } from "../server/config.ts";
  import { y } from "../server/logging/logger.ts";
  export function legacy() {}
  -> judged=19 shims, 1 adapters / PASS, exit 0

PROBE (rejected): the same file with its second import rewritten to "./logger.ts"
  -> judged=19 shims, 1 adapters
     FAIL A src/l5-adapter-probe.ts
     3:import { z } from "./logger.ts";
     exit 1

shellcheck scripts/done-means/864-moved-out-of-src.sh -> exit 0
bunx tsc --noEmit -> exit 0
prettier --write _plans/864-l5-cut-plan.md -> exit 0
pre-push hook (full isolated suite) -> passed
```

Both probes were unstaged with `git restore --staged` and moved out of the tree before staging; they live at `{temp_workspace}/open-brain/_scratch/session14/lane-3-probes/`.

## Critical Self-Review

- Highest-risk behavior: widening clause A could accept a `src/` file that has NOT really moved -- the whole point of the check. The accept test is dependency direction, not the header alone: a file must both declare the M9 header and name only `server/` relative specifiers. The rejecting probe proves the second half is what decides.
- Assumptions that could be wrong: `adapter_bad_specifiers` matches `from "..."` and `import("...")` with a leading `.`; a relative specifier reached some other way (a `require()`, a template-literal dynamic import, a re-export written across two physical lines) would not be seen. That is the same shape of assumption the pre-existing shim test already makes, and rule M9 describes adapters as small files under 60 code lines where this does not arise.
- Missing/weak tests: there is no committed automated test for the script -- it is itself the done-means check, and its receipts are the two probes run by hand and recorded in its header. A committed fixture pair would be stronger and is worth doing when a real adapter first lands.
- Security/permission risk: none. The script reads tracked files and runs `rg`, `git ls-files`, and one `bun` script; it writes nothing and takes no credentials.
- Migration/deploy risk: none. No runtime code, no schema, no config. The script runs in CI and by hand only.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python client surface is touched, so `docs/downstream-rollout.md` classifies this as not applicable.
- Rollback/cleanup concern: reverting the commit restores the shim-only clause A exactly; the 19 shims judged today are judged identically before and after, so a revert cannot strand a passing lane. The probe files are outside the repo.
- Fixes made before PR: the first probe run hung, because an adapter left `IS_SHIM=no` and so fell into clause C's whole-tree `rg --type ts` scan. Fixed by skipping clauses B and C for an adapter, matching how a shim is handled. The second probe then passed when it should have failed -- discovery had dropped the malformed file silently. Fixed by discovering on the declared header rather than on a successful adapter test, which is what makes the failure loud.
- Known residual risk: an adapter is accepted on dependency direction, not on being small or on genuinely preserving the legacy call form. Rule M9's under-60-code-lines and preserves-every-legacy-export conditions are reviewer-enforced, not machine-checked, and this change does not add that.
- SME review-memory update: [x] `docs/sme/` updated or [ ] not applicable because: a discovery filter that drops a malformed candidate instead of failing it turns a check silently green -- the second probe passed for the wrong reason until discovery keyed on the declared header rather than on the successful test, and the general form is that a self-discovering checker must judge everything that CLAIMS to be in scope, never only what already satisfies the rule.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, schema, or service surface changes -- a shell check script and a plan document.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing surface is touched.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool, schema, transport, event vocabulary, or Python client surface changes. The diff is `scripts/done-means/864-moved-out-of-src.sh` and `_plans/864-l5-cut-plan.md` only.

### DIFF
--- a/_plans/864-l5-cut-plan.md
+++ b/_plans/864-l5-cut-plan.md
@@ -1,6 +1,8 @@
 # L5 cut plan — move every src/ file into server/ (issue 864)
 
-**Status: WRITTEN 2026-08-29, no lane merged yet.**
+**Status: WRITTEN 2026-08-29. Merged lanes: pull requests #967, #968, #969,
+#970, #971, #972, #973, #974, #975, #976, #978.** (#977, contract and
+contract-schemas, was still open when this line was written.)
 
 Authority: `_plans/server-hardening-ladder.md` L5 (lines 291-330) and L6 (lines
 332-360). L5's own done-means is "zero imports from `src/` in non-test
@@ -67,6 +69,41 @@ with `src/` at L6, not by a move lane. `864-moved-out-of-src.sh` accepts a shim
 in clauses A and B and skips clause C for one, because the implementation has
 moved even though the path is still tracked.
 
+
+## L5 adapters
+
+Some legacy callers use a call form the `server/` version cannot keep. Rule M9
+(head decision 2026-08-29 after lanes 1 and 10) covers them, verbatim:
+
+> M9 ADAPTER (head decision 2026-08-29 after lanes 1 and 10): when a legacy
+> src/ or scripts/ caller uses a call form the server/ version cannot keep (a
+> zero-argument form that reads process.env inside, positional legacy
+> arguments, an options object widened with required env fields), the old src/
+> path becomes an ADAPTER instead of a shim: a file under 60 code lines whose
+> header line is `// L5 adapter (issue 864): legacy call form over
+> server/<dir>/<f>.ts; retired with src/ at L6.`, whose every relative import
+> names a server/ path (node and npm imports allowed), which preserves every
+> legacy export and call form, and which may read process.env itself (it is a
+> src/ file, lint-exempt, retired at L6). The server/ version takes its env
+> values as fields of one options parameter, filled by server/main.ts from
+> config. Done-means clause A accepts an adapter by the same test as a shim:
+> every relative import specifier resolves under server/. closure-count.sh
+> excludes shims and adapters alike (header regex `^// L5 (shim|adapter)`).
+
+Three candidates carry a legacy call form that rule M9 covers:
+
+| Path | Why an adapter and not a shim |
+|---|---|
+| `src/operator-doctor.ts` | zero-argument entrypoint reading env inside |
+| `src/promotion-service.ts` | positional legacy arguments its src/ callers pass |
+| `src/audit-log.ts` | options object the server/ version widens with required env fields |
+
+`scripts/done-means/864-moved-out-of-src.sh` accepts an adapter in clauses A
+and B and skips clause C for one, exactly as it does for a shim. Its
+no-argument discovery reports `judged=<n> shims, <m> adapters` once any adapter
+is on the tree, and a file that declares the M9 header while naming a relative
+specifier outside `server/` is judged and fails clause A with the offending
+line printed, rather than dropping out of the set unseen.
 ## Expected closure after each wave
 
 | After | Closure |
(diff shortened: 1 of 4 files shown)


rodaddy:
harvested: #980 A self-discovering checker must judge everything that CLAIMS to be in scope, not only what already satisfies the rule. The first implementation of the adapter clause built its target set with `is_adapter` — the same predicate clause A accepts with — so the deliberately-broken probe (M9 header present, one `from "./logger.ts"` specifier) failed the predicate, was dropped from the set, and the run printed `judged=19 shims / PASS, exit 0`. A malformed candidate made the check greener rather than redder, which is the worst direction for a done-means to fail in. The fix separates the two roles: discovery keys on the DECLARATION (`declares_l5_header`, the first line claiming to be a shim or adapter) while the accept test keys on the SUBSTANCE (every relative specifier resolves under `server/`), so a file that claims the header and breaks the rule is judged and prints `FAIL A` with the offending line. Generalised: whenever a check builds its own target set from the tree, the discovery filter and the accept test must not be the same predicate, or nothing can ever fail; and the probe that SHOULD fail is the one to write first, because a silent PASS on a broken fixture is indistinguishable from a clean run. Captured as docs/sme/entries/2026-08-29-a-self-discovering-checker-must-judge-everything-that-claims-to-be-in-scope.md (lane: correctness).

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=67ceb0cf100937d55eed070ccf099a673a7c657b at=2026-08-29T21:09:42.064Z

```
judged=19 shims
PASS
```

</details>

PR #982, opened 2026-08-29
refactor(l5): move maintenance-queue and maintenance-sweep from src/ into server/maintenance (issue 864)

rodaddy:
## Summary

- L5 of the server-hardening ladder (issue 864): `src/maintenance-queue.ts` (1018 lines) and `src/maintenance-sweep.ts` (497 lines) move into `server/maintenance/`, so `server/main.ts` reaches back into `src/` for two fewer modules.
- `maintenance-queue.ts` was over the 500-line file rule, so it is split by responsibility into three siblings: `server/maintenance/maintenance-job.ts` (job shape, error category, terminal-error marker, the queue port, row mapping, enqueue validation, backoff), `server/maintenance/maintenance-queue.ts` (the `MaintenanceQueue` class and its SQL), and `server/maintenance/maintenance-queue-runner.ts` (`MaintenanceQueueRunner`, its logger and handler types, lease heartbeat).
- Both old `src/` paths are now L5 shims re-exporting the `server/` modules, so every `src/` and `scripts/` importer keeps its import line and stays out of the pre-commit lint set.
- `server/` importers repointed: `server/maintenance/index.ts`, `server/application/index.ts`, `server/domain/candidate-dedupe.ts`, `server/domain/candidate-dedupe.test.ts`, `server/maintenance/maintenance.pg.test.ts`.
- Lint work the `server/` standard forced, with no behavior change: `validateEnqueue` split into `validateJobIdentity` / `validateRetryPolicy` / `validateScope` (complexity 22 to under the rule, same checks, same error strings, same order); the distill lane-batch SQL hoisted to a module constant so `selectDistillLaneBatches` fits the function rule (query text byte identical); the one non-null assertion on a claim lease token replaced with a typed guard that throws on the impossible null; `Promise.all` no longer spreads an already-iterable `Set`; the self-referencing active promise is a `const`.
- Runtime closure of `src/` under `server/main.ts`: 46 to 44.

Tests left in src/ (retire or move at L6), all green through the shims: `src/maintenance-queue.test.ts`, `src/maintenance-queue.pg.test.ts`, `src/maintenance-sweep.test.ts`, `src/maintenance-sweep.live.test.ts`, `src/embedding-repair-formula.pg.test.ts`, `src/db/migrations/026_maintenance_queue-retry.test.ts` and its test helpers.

## Verification

`bunx tsc --noEmit` exit 0. `oxlint --deny-warnings --config .oxlintrc.json` over every touched path exit 0. `bun run test:isolated` over the moved modules' tests plus `server/maintenance/maintenance.pg.test.ts` and `server/domain/candidate-dedupe.test.ts`: 71 pass, 0 fail, 290 expect() calls across 7 files. `bash scripts/done-means/864-moved-out-of-src.sh src/maintenance-queue.ts src/maintenance-sweep.ts` printed PASS. The pre-push hook ran the full isolated suite and passed. Python is not applicable: `rg` over `python/openbrain-memory/tests` found no reference to either module name, so no path guard needed repointing.

## Critical Self-Review

- Highest-risk behavior: the maintenance queue's lease/retry/dead-letter semantics, which the runner and the SQL both encode. The move preserves the SQL text and every method body verbatim; the only source edits are the three extracted validation helpers, the hoisted SQL constant, and the lease-token guard.
- Assumptions that could be wrong: that a claimed job always carries a lease token. The old code asserted this with `!`; the new code throws a named error instead, so if the assumption is ever false the failure is a clear message rather than a `null` reaching the SQL. That is the one place where a previously-silent path now throws, and it is the path that was already undefined behavior.
- Missing/weak tests: nothing new is tested because nothing new is intended to behave differently. The split's real risk is an import that resolves to the wrong sibling, and `bunx tsc --noEmit` plus the full isolated suite are the checks that catch that.
- Security/permission risk: none identified. No auth, namespace predicate, or token handling is touched; `validateScope` keeps the same namespace token regex and the same rejection.
- Migration/deploy risk: none. No schema change, no migration, no config key. The shims mean nothing outside `server/` had to change, so a partial deploy cannot see a missing module.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python client surface changes; these are internal server modules with no client-facing export.
- Rollback/cleanup concern: the two `src/` shims are deliberate debt, removed with `src/` at L6. They are marked with the L5 shim header the done-means script recognizes, so they are findable rather than forgotten.
- Fixes made before PR: the four lint findings the move surfaced were fixed rather than suppressed (no disable comments were added), and the extracted validation helpers were checked to preserve check order so an input that failed on the kind regex still fails there and not on the retry policy.
- Known residual risk: the moved modules now import three still-in-`src/` siblings (`graph-derivation-handler`, `distill-handler`, `distill-window`) by relative `../../src/` path, so `server/maintenance/maintenance-sweep.ts` still depends on `src/`. That is by design for this lane and is what later L5 lanes remove.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical module move with no new defect class; the split heuristic it uses is already the issue 864 program's own rule, not a new lesson.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or client-facing surface changes; the modules are internal and their behavior is unchanged.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, schema, or fixture is touched by a file move plus re-export shims.

## Downstream Rollout

Notes/evidence:

- Rollout does not apply: no MCP tool, schema, protocol, or client-facing change. The only exported surface that moved is re-exported from its old path, so no consumer sees a different module.


rodaddy:
harvested: #982 A three-way split of a large module is decided by the file's own seams, not by line count: `maintenance-queue.ts` broke cleanly at the two points where the module already changed subject (types and validation, then the class and its SQL, then the runner), and the usage matrix that proved the split -- one `rg -c` per identifier against each candidate section -- took under a minute and named exactly which module-private helpers had to become exports (`toJob`, `JOB_COLUMNS`, `validateEnqueue`, `assertPositiveInteger`, `enqueueSemanticsDiverge`, `jobColumns`, `MaintenanceJobRow`) and which consts belonged with the runner rather than the queue. Doing that matrix BEFORE writing any import block is what made the two `tsc` rounds trivially short instead of a guess-and-check loop. The second reusable point: a `git stash -u` to take a baseline measurement will not cleanly `stash pop` when the change set includes new files at paths that the stash also removed -- here the two `src/` shims sat at the paths of the moved originals, so pop restored the originals over the shims and refused the untracked half. Measure the baseline from `git show origin/main:<path>` or a separate read-only checkout instead, or verify every shim's first line after the pop before trusting the tree.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=fa726f1755e255ae7941a77fc56df442f79a63e1 at=2026-08-29T21:14:41.427Z

```
judged=23 shims
PASS
```

</details>

PR #983, opened 2026-08-29
refactor(l5): move namespace-policy and permissions from src/ into server/security (issue 864)

rodaddy:
## Summary

- L5 of `_plans/server-hardening-ladder.md` (issue 864): move two security modules out of `src/` into `server/security/`, next to the `secret-patterns.ts` that already lives there.
- `src/namespace-policy.ts` -> `server/security/namespace-policy.ts` (173 lines); `src/permissions.ts` -> `server/security/permissions.ts` (73 lines); `src/permissions.test.ts` -> `server/security/permissions.test.ts` (127 lines, passes oxlint whole under the server overrides with only its import specifiers repointed, so it moved with its module).
- Runtime closure of `src/` under `server/main.ts`: 46 -> 44, exactly the two source modules moved.
- **Namespace isolation is a security boundary, and every predicate is byte-identical apart from formatting.** No reason string, no comparison, no role list, and no evaluation order changed anywhere in `canWriteNamespace`, `isPromoterIdentity`, `writableNamespaces`, or `appendWriteNamespacePredicate`.
- The single structural edit is forced by the server standard, not chosen: `canWriteNamespace` scored complexity 14 against the server maximum of 10. The four namespace-shaped rejections that already ran ahead of any role decision (legacy shared write, frozen snapshot, `X-Namespace` header mismatch, shared-namespace promoter identity) are hoisted verbatim into a `namespaceWriteRejection` helper returning the same `NamespaceCheck` or `undefined`; `canWriteNamespace` calls it first and returns its result unchanged, so the guards still run in their original order ahead of the untouched role ladder.
- **No retarget.** `server/auth/namespace-policy.ts` and `server/auth/permissions.ts` are pre-existing twins, and every `server/` importer already resolves them. They are left exactly as they are and no `server/` import line is edited in this change.
- The old `src/` paths become L5 shims per M2 (one header comment plus one `export *` naming the `server/` path), so the ~60 `src/` and `scripts/` importers keep their import lines untouched and stay out of the staged lint set.
- Neither moved module reads `process.env`, so no caller had to start threading config.

tests left in src/ (retire or move at L6):

- `src/namespace-policy.test.ts` (298 lines) stays at its `src/` path per the M1 exception and keeps testing the moved module through the shim. It is not staged, so the gate never lints it.

## Verification

RED at `origin/main` (before the move): `bash scripts/done-means/864-moved-out-of-src.sh src/namespace-policy.ts src/permissions.ts` -> exit 1, `FAIL A` and `FAIL B` for both paths.
GREEN on this branch: the same command -> exit 0, `PASS`.

Other receipts: `bunx tsc --noEmit` -> exit 0; `oxlint --deny-warnings --config .oxlintrc.json` over all five touched paths -> exit 0, zero findings; `bun run test:isolated src/namespace-policy.test.ts server/security/permissions.test.ts server/auth/auth.test.ts src/ob-admin-role.test.ts` -> exit 0, 115 pass 0 fail across 4 files; the pre-push hook's full isolated suite -> passed. Python guards: `rg` over `python/openbrain-memory/tests` for both basenames -> no hits, so no guard repoint was needed.

## Critical Self-Review

- Highest-risk behavior: the write-authorization decision in `canWriteNamespace`. If the hoist reordered or dropped a guard, a client could gain a write it should not have — for example a frozen-snapshot or shared-namespace write slipping past into the role ladder, where `admin` returns allowed unconditionally. Mitigated by hoisting a contiguous prefix of the function whole rather than redistributing conditions, and by keeping the helper's return the exact object the original returned.
- Assumptions that could be wrong: that the four hoisted guards were genuinely a contiguous prefix with no state shared with the role ladder below them. Checked against the source — the only value they compute that the rest of the function used is `physicalTargetNamespace`, which is still computed in `canWriteNamespace` and passed in as a parameter, and `config`, which the role ladder does not read.
- Missing/weak tests: no new test was added, because the deliverable is a move and the standing rule is that behavior is unchanged. The existing coverage is what proves it: `src/namespace-policy.test.ts`, `server/auth/auth.test.ts`, and `src/ob-admin-role.test.ts` exercise the promoter identity, frozen namespaces, the header path, and the per-role ladder, and all pass unmodified against the moved code. A genuinely weak spot is that `namespaceWriteRejection` has no direct unit test of its own; it is covered only through `canWriteNamespace`, which is how it was covered before the hoist.
- Security/permission risk: this is the namespace isolation boundary, so the risk is real rather than theoretical. The mitigating fact is that the shim re-exports rather than reimplements, so there is exactly one implementation of each predicate in the tree after this change and no possibility of the two paths drifting apart.
- Migration/deploy risk: none. No schema, no migration, no configuration, and no runtime entry point changed; `server/main.ts` was not touched.
- Downstream client/runtime risk: none identified. No MCP tool, schema, transport, or wire contract changed, and the Python client is untouched — the pre-push gate skipped the Python and contract-parity lanes on that basis. Import paths that other repos could hold are unaffected because the old `src/` specifiers still resolve through the shims.
- Rollback/cleanup concern: a revert is a clean single-commit revert. The cleanup debt this deliberately creates is the two shims and the one test left in `src/`, all three of which are retired with `src/` at L6 and are named above so they are not lost.
- Fixes made before PR: the complexity finding on `canWriteNamespace` was caught by oxlint before the first commit and fixed by the hoist described above rather than by a disable comment. The moved files' relative imports of `src/types.ts` and `src/shared-namespace.ts` were repointed to `../../src/` per M3 instead of dragging those modules into this lane.
- Known residual risk: the hoist is the only place a reviewer needs to look. It is a mechanical extraction, but it is an extraction inside an authorization function, so it deserves a read against the pre-move file rather than a skim of the diff.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this lane produced no new MEDIUM-or-higher finding — the one issue encountered (a complexity score over the server maximum on a moved function) is the already-documented M4 standard behaving as designed, and the response it calls for is already written into the issue 864 lane rules.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or tool surface changed; this is a file move plus re-export within one repo.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, fixture, or client-facing schema is touched, and the pre-push gate skipped the contract parity subset on that basis.

## Downstream Rollout

Notes/evidence:

- Rollout does not apply. `docs/downstream-rollout.md` scopes the gate to MCP tool, schema, protocol, and client-facing changes; this change moves two internal modules and adds re-exports at the old paths, so no downstream consumer observes a difference. The old `src/` import specifiers still resolve, which is what keeps the blast radius inside this repo.


rodaddy:
harvested: #983 A move lane into `server/` should check for a same-basename twin under `server/` before it plans any import repointing, because the twin decides the whole shape of the lane. In this one, `server/auth/namespace-policy.ts` and `server/auth/permissions.ts` already existed and every single `server/` importer already resolved them, so the "repoint server/ importers" half of the brief had zero work in it and the lane reduced to a pure move plus shims. One `rg` for the import specifiers answered that in seconds and would have been wasted effort to discover incrementally; running it before the first `git mv` is the cheap ordering. The second reusable point is narrower: when the server complexity rule fires on an authorization function, hoist a contiguous prefix of guards whole into a helper that returns the same result type or `undefined`, rather than redistributing individual conditions across helpers. The prefix hoist keeps evaluation order and every returned object byte-identical, which is the property a security-boundary reviewer actually needs to confirm, and it makes the diff readable as "these four guards moved" instead of "this predicate was rewritten."

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=4d1ebdddb21d4b8092287d57d104cb7859199c33 at=2026-08-29T21:18:13.785Z

```
judged=23 shims
PASS
```

</details>

PR #987, opened 2026-08-29
refactor(l5): move promotion-service from src/ into server/domain behind an L5 adapter (issue 864)

rodaddy:
## Summary

- `src/promotion-service.ts` -> `server/domain/promotion-service.ts` (issue 864, L5). No basename collision in `server/domain/`.
- The `src/` closure under `server/main.ts` goes 39 -> 38.
- The server/ version takes ONE options object instead of the legacy positional tail, so it holds no environment read of its own. `promoteEntry(pool, table, id, options)` is four parameters; `targetNamespace`, `reason`, `auth`, `dryRun`, `now` and `killSwitch` are all fields. `killSwitch` is the value the old code read from `OPENBRAIN_PROMOTION_KILL_SWITCH` inline at line 235.
- `src/promotion-service.ts` remains as the M9 L5 adapter (54 lines): it keeps the old positional call form for the five legacy callers (`src/rest-promotion.ts`, `src/tools/promote-entry.ts`, `src/tools/promote-shared.ts`, `scripts/promote-lane-shared.ts`, `scripts/promote-legacy-shared.ts`), reads `process.env.OPENBRAIN_PROMOTION_KILL_SWITCH` inside the function at call time, and delegates. Every relative import in it names a `server/` path.

### The two server/tools importers

Both were repointed and both pass the kill switch from a new optional `promotionKillSwitch` field on `MemoryToolDependencies`, wired once in `server/main.ts` from `config.promotion.killSwitch` beside the `ftsCorpusConfig` and `searchEmbeddingTimeoutMs` fields that already follow exactly this pattern. `server/config/src-readers.ts` already parsed the variable and already exposed `PromotionConfigGroup.killSwitch`, so NO new reader was added and `server/config.ts` is untouched.

`server/tools/promote-entry.ts`:

```diff
-import { promoteEntry } from "../../src/promotion-service.ts";
+import { promoteEntry } from "../domain/promotion-service.ts";
@@
       dependencies.pool,
       args.table as ResourceTable,
       args.id,
-      target,
-      args.reason,
-      promotionAuth(identity),
-      { dryRun },
+      {
+        targetNamespace: target,
+        reason: args.reason,
+        auth: promotionAuth(identity),
+        dryRun,
+        killSwitch: dependencies.promotionKillSwitch ?? false,
+      },
     );
```

`server/tools/promotion.ts`:

```diff
-import { promoteEntry } from "../../src/promotion-service.ts";
+import { promoteEntry } from "../domain/promotion-service.ts";
@@
-    const result = await promoteEntry(
-      dependencies.pool,
-      table,
-      id,
-      target,
+    const result = await promoteEntry(dependencies.pool, table, id, {
+      targetNamespace: target,
       reason,
-      promotionAuth(identity),
-      { dryRun },
-    );
+      auth: promotionAuth(identity),
+      dryRun,
+      killSwitch: dependencies.promotionKillSwitch ?? false,
+    });
```

`server/main.ts` gains one line: `promotionKillSwitch: config.promotion.killSwitch,`.

### M4 seams

Linting the copy at the target path first surfaced three, all fixed without a disable comment:

- Four `source: any` parameters became a named `PromotionSourceRow` shape with an index signature, since the row is a `SELECT *` across five tables with genuinely different columns.
- `buildProvenance` scored complexity 35 on the repeated `extracted_metadata?.x ?? metadata?.x ?? null` chains; one `metadataValue(row, key)` helper collapses them and preserves the same three-way fallback.
- The 148-line `promoteEntry` body split into `readSourceRow`, `assertTargetWritable` and `buildProvenance`, each named for what it does.

The moved file imports `AuthInfo`/`Table` from `server/types.ts` (the L5 copy that exists so no server module imports from src/), and keeps `read-policy`/`namespace-policy`/`shared-namespace` at `../../src/` per M3 — those belong to later lanes.

### Tests left in src/ (retire or move at L6)

Per M1 EXCEPTION, unchanged and passing through the adapter:

- `src/rest-promotion.test.ts` (sets `OPENBRAIN_PROMOTION_KILL_SWITCH` per case, so it cannot live under `server/`)
- `src/tools/__tests__/promote-entry.pg.test.ts` (same)
- `src/tools/__tests__/promote-shared.test.ts`

## Verification

Receipts:

- RED, working tree restored to `origin/main` content for the two paths: `bash scripts/done-means/864-moved-out-of-src.sh src/promotion-service.ts` -> exit 1, `FAIL A` and `FAIL B`.
- GREEN, at the commit: same command -> exit 0, `PASS`.
- `./node_modules/.bin/oxlint --deny-warnings --config .oxlintrc.json` over all six touched paths -> exit 0.
- `bun run test:isolated server/tools/promotion.pg.test.ts src/rest-promotion.test.ts src/tools/__tests__/promote-entry.pg.test.ts src/tools/__tests__/promote-shared.test.ts` -> 44 pass, 0 fail.
- The pre-push hook ran the full isolated suite and passed.
- M10: `rg -n '"src/|src/[a-z-]+\.ts' python/openbrain-memory/tests` matches nothing for `promotion-service`, so no Python guard needed repointing.
- Closure count 39 -> 38, one file, matching the one file moved.

## Critical Self-Review

- Highest-risk behavior: the kill switch. It used to be read inside the service on every apply-mode call; now the server path reads it once at composition and the legacy path reads it per call in the adapter. A long-lived server process that had the variable flipped after start would previously have honoured the new value and now will not until restart. That is a real behavior difference on the server path, and it is the intended direction of the L2b rung (config is parsed once, at the root) rather than an accident.
- Assumptions that could be wrong: that `server/types.ts` `AuthInfo` is interchangeable with `src/types.ts` `AuthInfo` where the moved file passes it into the still-in-src policy helpers. I read both declarations and they are field-for-field identical, and `tsc` accepts it, but they are two declarations that can drift until src/ retires.
- Missing/weak tests: nothing covers the new `promotionKillSwitch` deps field directly — the server-side kill switch is exercised only through the existing tests, which reach the service through the adapter and so still test the env read rather than the field. A direct test of `promoteEntry(..., { killSwitch: true })` would be worth adding.
- Security/permission risk: none introduced. `assertTargetWritable` is the same read and write check in the same order, still executed before any insert, and it is still the service's own re-check rather than the caller's.
- Migration/deploy risk: none; no schema or SQL text changed. The INSERT/SELECT projection and the stripped-metadata expression are byte-identical.
- Downstream client/runtime risk: no MCP tool schema, transport, or client-facing contract changed. `promote_entry` and `promote_shared` keep their inputs, outputs and annotations.
- Rollback/cleanup concern: the adapter is the only lingering artifact and it is scheduled for removal with all of `src/` at L6, which is what the header comment records.
- Fixes made before PR: the first attempt passed the kill switch via `loadServerConfig()` inside the two tools. That threw under the test environment and broke three tests in `server/tools/promotion.pg.test.ts`; it was also the wrong seam, since the repo's established pattern is an optional deps field fed by the composition root. Replaced with `dependencies.promotionKillSwitch`. Separately, the pre-commit gate lints the src/ adapter too and there is no src/ exemption for `max-params`, so the adapter's seven-argument legacy signature became one typed tuple rest-parameter — same call form at every call site, still position-checked, no disable comment.
- Known residual risk: the adapter reads the environment on every call rather than once, which is a small per-call cost on the legacy path and is deliberate so the existing tests that set the variable per case keep behaving as they did.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical L5 move whose only new lesson (the src/ adapter is linted by the pre-commit gate, so a legacy over-arity signature needs a typed tuple rather than a disable comment) is being harvested as a PR comment on this pull request for the lane contract, not as a reviewer-facing SME entry.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no tool schema, transport, or hosted behavior changed; this is an internal module move plus a call-shape change behind an adapter.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract or client path was touched, and the pre-push gate reported "Client/contract paths unchanged; skipping contract parity subset".

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool schema, protocol, transport, or Python client surface changed. The two MCP tools keep identical inputs and outputs; the change is which module they import and how they pass one already-existing configuration value.

### DIFF
--- a/server/domain/promotion-service.ts
+++ b/server/domain/promotion-service.ts
@@ -1,0 +1,395 @@
+import type pg from "pg";
+import type { AuthInfo, Table } from "../types.ts";
+import {
+  appendReadNamespacePredicate,
+  canReadNamespace,
+} from "../../src/read-policy.ts";
+import { canWriteNamespace } from "../../src/namespace-policy.ts";
+import {
+  canonicalNamespace,
+  isSharedNamespace,
+  physicalNamespace,
+} from "../../src/shared-namespace.ts";
+
+export const PROMOTION_CONTENT_COLUMNS: Record<Table, string> = {
+  thoughts:
+    "content, tags, source, created_by, embedding, content_hash, embedded_at, embedding_model, tier, usefulness_score, extracted_metadata",
+  decisions:
+    "title, rationale, alternatives, context, tags, created_by, embedding, content_hash, embedded_at, embedding_model, tier, usefulness_score, extracted_metadata",
+  relationships:
+    "person_name, context, relationship_type, warmth, last_contact, email, phone, notes, tags, metadata, created_by, embedding, content_hash, embedded_at, embedding_model, tier, usefulness_score",
+  projects:
+    "name, status, description, metadata, tags, created_by, embedding, content_hash, embedded_at, embedding_model, tier, usefulness_score",
+  sessions:
+    "session_id, project, summary, tags, blockers, next_steps, key_decisions, created_by, embedding, content_hash, embedded_at, embedding_model, tier",
+};
+
+/**
+ * Nomination/audit keys that must NOT survive into a promoted copy. A promoted
+ * row that still carries `share_candidate` would re-nominate itself on the next
+ * promoter sweep (the sweep is namespace-wide), causing an endless re-scan and
+ * re-promotion attempt. Stripped from the metadata column in the SELECT.
+ */
+const STRIPPED_PROMOTION_METADATA_KEYS = [
+  "share_candidate",
+  "share_rejected_sync",
+  "share_promoted_at",
+  "memory_lifecycle_action",
+  "candidate_type",
+  "candidate_reason",
+  "candidate_confidence",
+  "candidate_scope",
+  "candidate_staleness_policy",
+  "evidence_refs",
+] as const;
+
+/**
+ * Build the SELECT projection for a promotion copy: identical to the INSERT
+ * column list, except the metadata jsonb column (`extracted_metadata` for
+ * thoughts/decisions, `metadata` for others) has the nomination keys stripped
+ * via `- key` so the promoted copy is not itself a standing nomination.
+ */
+function promotionSelectExpression(columns: string): string {
+  const metaCol = columns.includes("extracted_metadata")
+    ? "extracted_metadata"
+    : columns.includes(" metadata") || columns.startsWith("metadata")
+      ? "metadata"
+      : null;
+  if (!metaCol) return columns;
+  const stripped = STRIPPED_PROMOTION_METADATA_KEYS.map((k) => `- '${k}'`).join(" ");
+  // Replace the bare column token with the stripped expression aliased back to
+  // the column name so positional INSERT/SELECT alignment is preserved.
+  return columns
+    .split(", ")
+    .map((c) => (c === metaCol ? `(${metaCol} ${stripped}) AS ${metaCol}` : c))
+    .join(", ");
+}
+
+export interface PromotionResult {
+  status: "promoted" | "duplicate" | "dry_run";
+  new_id?: string;
+  existing_id?: string;
+  existing_archived?: boolean;
+  source_id: string;
+  source_namespace?: string;
+  target_namespace: string;
+  provenance?: Record<string, unknown>;
+  dry_run?: boolean;
+  would_insert?: boolean;
+  backoff?: {
+    retryable: boolean;
+    suggested_delay_ms: number;
+  };
+}
+
+/** Loose jsonb blob: either `extracted_metadata` or `metadata`, both optional. */
+type MetadataBlob = Record<string, unknown> | null | undefined;
+
+/**
+ * The subset of a source row this service reads.
+ *
+ * The row comes back from `SELECT *` across five different tables, so the
+ * columns genuinely differ per table; this shape names the ones the promotion
+ * path touches and leaves the rest reachable through the index signature.
+ */
+interface PromotionSourceRow {
+  namespace: string;
+  content_hash?: string | null;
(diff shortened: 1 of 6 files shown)


rodaddy:
harvested: #987 An M9 L5 adapter at an old `src/` path is linted by the pre-commit gate like any other staged file, and `.oxlintrc.json` has no `src/` exemption for `max-params` — only `no-console` and `node/no-process-env` are scoped by directory. So an adapter that must preserve a legacy call form with more than four positional arguments fails the gate on its own signature, and neither a disable comment nor an `.oxlintrc.json` edit is available to a move lane. The fix that stays inside the rules is to declare the legacy arguments as one named tuple type and take them as a single rest parameter (`export async function promoteEntry(...args: LegacyPromoteEntryArguments)`, destructured on the first line): every call site is unchanged and still type-checked position by position, the arity rule sees one parameter, and the tuple's labelled elements keep the signature self-documenting. Worth reaching for before assuming an over-arity legacy form blocks a move. The second lesson from the same lane: when a server/ module needs an environment value, the seam is an optional field on the tool dependencies interface fed once from config at the composition root (`server/main.ts`, next to `ftsCorpusConfig` and `searchEmbeddingTimeoutMs`), NOT a `loadServerConfig()` call inside the tool — the direct call throws under the test environment and broke three tests in `server/tools/promotion.pg.test.ts` before it was replaced.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=4963d64f06b4bcf9f65b8308b337af10d6b32c66 at=2026-08-29T21:29:33.457Z

```
judged=28 shims, 1 adapters
PASS
```

</details>

PR #988, opened 2026-08-29
refactor(l5): move extraction from src/ into server/domain behind an L5 adapter (issue 864)

rodaddy:
## Summary

- L5 of `_plans/server-hardening-ladder.md` (issue 864): `src/extraction.ts` -> `server/domain/extraction.ts`, moved with `git mv` so history follows.
- The moved module meets the server/ standard whole. The one change it needed was `backgroundExtract`, which took six positional parameters and failed the four-parameter signature rule. The four trailing per-row parameters are now one named object:

```ts
export interface BackgroundExtractTarget {
  entryId: string;
  namespace: string;
  text: string;
  existingTags?: string[];
}

export function backgroundExtract(
  pool: pg.Pool,
  table: string,
  target: BackgroundExtractTarget,
): void
```

  The split follows the six call sites: each passes a pool and a table name it chose, then four fields that all describe the same single row. Every other export keeps its signature untouched, and there is no logic change.

- `src/extraction.ts` is now the M9 L5 adapter, 29 code lines, importing only from `server/`:

```ts
// L5 adapter (issue 864): legacy call form over server/domain/extraction.ts; retired with src/ at L6.
export * from "../server/domain/extraction.ts";

type LegacyArgs = [
  pool: pg.Pool, table: string, entryId: string,
  namespace: string, text: string, existingTags: string[],
];

export function backgroundExtract(...args: LegacyArgs): void {
  const [pool, table, entryId, namespace, text, existingTags] = args;
  extractForTarget(pool, table, { entryId, namespace, text, existingTags });
}
```

  The rest tuple preserves the exact six-positional call form and its type checking for every legacy caller, while keeping the adapter itself inside the signature rule the gate applies to every staged file. Untouched callers: `src/rest-api.ts:224`, `src/rest-api.ts:311`, `src/tools/log-thought.ts:129`, `src/tools/log-decision.ts:124`, `src/drop-folder-collector.ts:869`, `scripts/bulk-import.ts`.

- `server/maintenance/graph-derivation-handler.ts` was the one `server/` importer and is repointed to `../domain/extraction.ts` per M2.
- Tests left in src/ (retire or move at L6), per M1 EXCEPTION: `src/__tests__/extraction.test.ts`, `src/db/migrations/027_source_registry.test.ts`. Both are unmodified and pass through the adapter.
- Runtime closure of `src/` under `server/main.ts`: 35 -> 34.

## Verification

`bash scripts/done-means/864-moved-out-of-src.sh src/extraction.ts` -> PASS.
`bunx tsc --noEmit` -> exit 0.
`./node_modules/.bin/oxlint --deny-warnings --config .oxlintrc.json` on all three staged paths -> exit 0.
`bun run test:isolated src/__tests__/extraction.test.ts src/db/migrations/027_source_registry.test.ts` -> 31 pass, 0 fail, 122 expect() calls.
Pre-commit gate (gitleaks, oxlint, prettier, whole-project tsc over the staged snapshot) and the pre-push full isolated suite both passed on the committed tree.
M10: `rg -n '"src/|src/[a-z-]+\.ts' python/openbrain-memory/tests` has no hit naming extraction, so the Python path guards need no repoint and the package is unchanged.

## Critical Self-Review

- Highest-risk behavior: the fire-and-forget enrichment UPDATE in `backgroundExtract`, whose namespace + `archived_at IS NULL` predicate is a namespace isolation boundary. The SQL, its parameter order, and the table allowlist are byte-identical to the pre-move version; only the destructuring of the incoming fields changed.
- Assumptions that could be wrong: that an explicit local `export function backgroundExtract` correctly shadows the same name arriving via `export *` from the same module. Checked rather than assumed — `bunx tsc --noEmit` is clean and both test files, which import `backgroundExtract` through the adapter and assert on the exact SQL and params, pass.
- Missing/weak tests: no test exercises the new three-parameter form directly; every existing assertion still reaches it through the adapter. That is deliberate under M1/M5 (test bodies unchanged), and it does mean the object form is covered only transitively until a `server/` test file lands at L6.
- Security/permission risk: none introduced. The namespace predicate, the archived-row guard, and the `EXTRACTION_TABLES` allowlist that stops a caller-supplied table name reaching SQL interpolation are all unchanged, and the SQL-injection rejection test (`"secrets; DROP TABLE thoughts"`) passes unchanged.
- Migration/deploy risk: none. No schema change, no migration, no config change. `server/main.ts` is untouched.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python client surface is touched; this is an internal module relocation behind an adapter that preserves every export and call form.
- Rollback/cleanup concern: single commit, revertable on its own. The adapter is cleanup that L6 owns when `src/` retires; it is not new debt beyond the M9 pattern already on main.
- Fixes made before PR: the adapter was first written with the plain six-parameter signature and the pre-commit gate's `max-params` rule refused it — src/ has no exemption from that rule for staged files. Rewritten as a rest tuple, which keeps the call form and satisfies the gate. `server/maintenance/graph-derivation-handler.ts` was also found importing the old path after the move and repointed in the same commit.
- Known residual risk: the two test files stay in `src/`, so the module's coverage still runs through the adapter rather than against the `server/` module directly. That is the M1 EXCEPTION as briefed and it resolves at L6.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: no new lesson — the max-params refusal on a src/ adapter is the already-recorded M9 pattern meeting the already-recorded staged-whole gate, and no new failure mode appeared.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no live service surface, schema, or config changes in this move.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract, client, or wire-facing surface is touched; the move is internal to the TypeScript server.

## Downstream Rollout

Notes/evidence:

- No MCP tool, schema, or protocol change: the only exported signature that changed is an internal one, and the legacy form is preserved by the adapter, so no downstream consumer sees a different surface.


rodaddy:
harvested: #988 An M9 L5 adapter that must preserve a legacy positional call form wider than four arguments cannot be written as a plain positional function: the pre-commit gate lints every staged file whole against `.oxlintrc.json`, and `max-params` has no `src/` exemption (unlike `node/no-process-env`, which is scoped to `server/`), so the adapter itself is judged by the server standard even though the code it wraps is retiring. The move that works is a named rest tuple — `type LegacyArgs = [pool: pg.Pool, table: string, ...]` with `export function backgroundExtract(...args: LegacyArgs)` destructured in the body — which keeps the call sites and their per-argument type checking byte-identical for every legacy caller while presenting one parameter to the linter. Labelled tuple elements preserve the argument names in editor hovers, so nothing is lost at the call site. Worth reaching for before considering a disable comment or widening the config; the next lane whose legacy form is a long positional list will hit the same refusal.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=117cc9216b2edc6006992703341a71b32839a5d3 at=2026-08-29T21:32:31.283Z

```
judged=32 shims, 1 adapters
PASS
```

</details>

PR #989, opened 2026-08-29
refactor(l5): move source-registry and repo-facts from src/ into server/domain (issue 864)

rodaddy:
## Summary

- L5 of issue 864: two more modules leave `src/` for `server/domain`, each split by responsibility so every sibling meets the `server/` standard whole.
- `src/source-registry.ts` (685) -> `server/domain/source-registry.ts` (register/list/update/remove/eligibility operations) + `server/domain/source-registry-model.ts` (vocabulary, input schemas, row mapping, namespace/approval helpers, content hashing).
- `src/tools/repo-facts.ts` (785) -> `server/domain/repo-facts.ts` (MCP upsert and list registration) + `server/domain/repo-facts-model.ts` (vocabulary, metadata schema and published contracts, pure identity/screening helpers) + `server/domain/repo-facts-source-urls.ts` (source-URL trust: the SSRF allowlist and the repo/path/commit match).
- Both old paths are L5 shims re-exporting the `server/` modules, so every `src/` and `scripts/` importer keeps its import line untouched and stays out of the staged-lint set.
- Server importers repointed: `server/tools/source-registry.ts`, `server/tools/ingest-conversation-facts.ts`, `server/tools/repo-facts.ts`, `server/contracts/schemas-repo-facts.ts`.
- `src/` runtime closure under `server/main.ts`: 42 -> 40.

Behavior is unchanged. Every edit is a hoist or a table lookup over the same conditions, forced by the `server/` lint standard: the `registerSource` INSERT and duplicate-identity recovery; the `updateSource` SET-clause builder and zero-row diagnosis (the eight plain updatable columns now come from one spec table); `isPrivateOrLocalHost` split into literal-prefix and parsed-octet checks; the two `sourceUrlMatchesSource` host path shapes; the `repoFactDedupeKey` metadata derivation; and both MCP tool handlers hoisted to module scope with the shared error envelope, upsert authorization, fact-text screens, embedding call, and shared-namespace read selection as named functions.

One non-mechanical consequence: `insertSourceRow` returns the inserted row and throws rather than fabricating a record when `RETURNING` yields none, replacing the index access the stricter `server/` typing rejects.

Tests left in `src/` (retire or move at L6), unmodified and green through the shims: `src/__tests__/source-registry.test.ts`, `src/db/migrations/027_source_registry.test.ts`, `src/tools/__tests__/source-registry.test.ts`, `src/tools/__tests__/repo-facts.test.ts`, `src/tools/__tests__/agent-context-pack-repo-facts.test.ts`, `src/tools/__tests__/ingest-conversation-facts*.test.ts`, `scripts/__tests__/promote-qmd-repo-facts.test.ts`.

## Verification

`bash scripts/done-means/864-moved-out-of-src.sh src/source-registry.ts src/tools/repo-facts.ts` -> exit 0, PASS on the committed tree. `bunx tsc --noEmit` -> exit 0. `oxlint --deny-warnings` over all 11 committed paths -> exit 0. `bun run test:isolated` over the seven registry/repo-fact suites -> 90 pass, 0 fail; over the three ingest-conversation-facts suites -> 25 pass, 0 fail. The pre-push hook ran the full isolated suite and passed. Python package: no `python/openbrain-memory/tests` guard pins either moved module, so the M10 repoint is not applicable.

## Critical Self-Review

- Highest-risk behavior: the source-registry approval and namespace authorization path. `registerSource`/`updateSource` decide who may approve what, and the refactor moved that logic across a file boundary and hoisted three of its branches. A dropped condition here is a privilege bug, not a crash. Mitigation: `effectiveWriteNamespace`, `canApprove`, and every `canWriteNamespace` call are unchanged in body and call order; the existing isolation and approval tests exercise them and pass unmodified.
- Assumptions that could be wrong: that `export * from "./source-registry-model.ts"` in the ops module reproduces the single-file export surface exactly. If a caller depended on a name that is now only in the model file and the re-export missed it, that is a compile error, not a silent gap -- `tsc --noEmit` at exit 0 over the whole project is what makes this checkable rather than assumed. Second assumption: that the `UPDATE_COLUMN_SPECS` table reproduces the eight `if (input.x !== undefined)` branches with identical jsonb encoding and casts. I read them off the original one at a time; the update tests cover scope/config/lifecycle/sync/content_hash/last_synced_at.
- Missing/weak tests: no test was added for the new `insertSourceRow` no-row throw, because reaching it requires a `RETURNING` insert that neither errors nor returns a row -- I could not construct that against real Postgres without faking the pool, which would test the fake. It is a defensive refusal on a branch the driver should never produce. Also thin: `readNamespaceSetWithShared` and `readSharedNamespace` are exercised only through the existing list tests, which do not cover every legacy-fallback combination; that coverage gap predates this change.
- Security/permission risk: the namespace predicate on every read and mutation had to survive the split intact. `namespaceClause` moved to the model file and is called from the same place with the same params array; the SQL it emits is byte-identical. The SSRF allowlist moved wholesale into `repo-facts-source-urls.ts` -- `isTrustedSourceUrl` keeps the https-only, no-credentials, no-private-host, two-host allowlist checks in the same order, and splitting `isPrivateOrLocalHost` preserved every prefix and both parsed ranges (172.16-31, 169.254). All SQL stays parameterized; the only interpolated fragments are the table-name constant, the boolean-chosen `NOW()`/`NULL`, and column-name literals from a closed spec table.
- Migration/deploy risk: none. No schema change, no migration, no config change. `027_source_registry.sql` is untouched and its test passes.
- Downstream client/runtime risk: none expected. No MCP tool name, input schema, output shape, or error text changed -- the handlers moved to module scope and are still registered under `upsert_repo_fact` and `list_repo_facts` with the same `inputSchema` and annotations. `server/contracts/schemas-repo-facts.ts` re-exports the same contract objects from the new path, so contract parity fixtures see no diff.
- Rollback/cleanup concern: revert is a single-commit revert; the shims mean no other file's imports need touching to undo it. The shims themselves are cleanup owed at L6, not now -- they are what keeps ~20 `src/` importers out of the staged-lint set.
- Fixes made before PR: dropped the duplicate `export *` a sed pass inserted twice into the ops module; moved a `sourceUrlMatchesSource` doc comment back onto its function after the hoist orphaned it above an interface; fixed two mis-placed import insertions (one landed inside an import block); and corrected `insertSourceRow` to return the row rather than the array after `noUncheckedIndexedAccess` rejected the index access.
- Known residual risk: `server/tools/repo-facts.ts` and `server/tools/source-registry.ts` carry ~30 lines of prettier reflow beyond my one-line import edits. Those two files were already failing `prettier --check` on `main` -- I verified this by stashing my edits and re-running the check on the pristine versions -- and the pre-commit hook blocks the commit unless staged files are prettier-clean, so the reflow is the hook's requirement rather than optional churn. It is unrelated formatting inside files I otherwise barely touched, and it will show up in review as noise.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical move/split lane under the issue-864 program with no new defect class; the lane lesson is harvested into the PR comment per the lane contract.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, schema, or deployed-service behavior changed; this is an in-repo module move with unchanged tool contracts.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract content changed. `server/contracts/schemas-repo-facts.ts` re-exports the same objects from the new module path, so the emitted contract is identical and no fixture moves.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool name, schema, protocol, or client-facing shape changed. The move is internal to this repo's file layout, both old paths remain importable as shims, and `server/contracts/schemas-repo-facts.ts` emits the same contract from the new path.


rodaddy:
harvested: #989 A move lane's real cost is not the move, it is the lint findings the target directory's stricter rules expose, and those are only visible before `git mv` if you look. Copying each file to its target path and running `oxlint --config .oxlintrc.json` on the copy turned a two-file move into a measured plan up front: 7 complexity findings, 3 oversized functions, and 2 files over the size rule, which is what determined the split seams rather than a guess about them. Two things this lane adds to that lesson. First, prettier: `--write` on a file that was ALREADY failing `prettier --check` on `main` reformats lines that have nothing to do with your change, and the pre-commit hook then requires exactly that reformat, so the churn is mandatory rather than optional -- verify the pre-existing state by stashing your edits and re-checking the pristine file, so the PR can say which. Second, `sed -i` for import surgery on a multi-line import block is a bad tool that fails loudly-but-late: it inserted a statement inside a `import { ... } from` block and duplicated a re-export line, both of which only surfaced at `tsc`. For anything touching an import block's interior, the Edit tool's exact-string match is the cheaper path, and `sed` is fine only for whole-line specifier swaps.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=1730ab2df7aed8fe45fb2babb5b096028958ac55 at=2026-08-29T21:35:00.888Z

```
judged=27 shims
PASS
```

</details>

PR #991, opened 2026-08-29
refactor(l5): move realtime working-set and recovery-wal from src/ into server/realtime (issue 864)

rodaddy:
## Summary

- L5 of issue 864: `src/realtime/working-set.ts` -> `server/realtime/working-set.ts` and `src/realtime/recovery-wal.ts` -> `server/realtime/recovery-wal.ts`.
- The server/ versions already existed on main as the hardened forks (injected logger instead of a module import, doc comments, and recovery-wal already split into five siblings: `recovery-wal-types`, `-journal`, `-records`, `-results`, `-store-parts`). The src/ copies were a stale duplicate of the same code, so this lane retires the duplication instead of a `git mv` onto files already present.
- Both old paths become M9 adapters rather than plain shims, because the server/ versions do not keep the legacy call forms their remaining src/ callers use: `WorkingSetStore` moved from a positional constructor argument to an options object, and `RecoveryWalStore.mark` moved from five positional arguments to `mark(scope, id, decision)`. The recovery-wal adapter carries the trailing three arguments as a named tuple so the method stays within the parameter-count rule.
- Every relative import in both adapters names a `server/` path. Importers left untouched by design (M2), reaching the server/ code through the adapters: `src/tools/agent-context-pack.ts`, `src/tools/index.ts`, `scripts/done-means/563-bounded-recall.driver.ts`, `src/nats-bridge.test.ts`, `src/tools/__tests__/*`.
- Tests left in src/ (retire or move at L6), per the M1 exception, unmodified and green through the adapters: `src/realtime/working-set.test.ts`, `src/realtime/recovery-wal.test.ts`.
- Closure of src/ under `server/main.ts`: 31 -> 29, exactly the two files this lane moved.
- `scripts/assert-db-tests-ran.ts` untouched: no describe name or count changed.

## Verification

Receipts, all from the committed tree:

- `bash scripts/done-means/864-moved-out-of-src.sh src/realtime/working-set.ts src/realtime/recovery-wal.ts` -> exit 0, `PASS`.
- `bunx tsc --noEmit` -> exit 0, no output.
- `./node_modules/.bin/oxlint --deny-warnings --config .oxlintrc.json src/realtime/working-set.ts src/realtime/recovery-wal.ts` -> exit 0.
- `bun run test:isolated src/realtime/ src/tools/__tests__/get-contract.test.ts src/contract.test.ts` -> exit 0, 44 pass / 0 fail.
- `bash _scratch/session14/closure-count.sh` -> 31 at `origin/main`, 29 after the change.
- M10 python guard scan: `rg -n 'src/realtime' python/openbrain-memory/tests` -> exit 1, no hits, so no guard repoint was needed and the python package is untouched.
- Pre-push full suite refused once with 4 failures, all in `scripts/__tests__/backup-restore-live*.test.ts` (the M11 drill flake, no realtime test among them); the identical `git push` was re-run once and reported `pre-push: all checks passed`. Never `--no-verify`.

## Critical Self-Review

- Highest-risk behavior: the adapters are the only thing keeping the legacy src/ call forms working. `RecoveryWalStore.mark` is the sharp one — the server form takes a decision object and the adapter re-assembles it from the legacy positional arguments, so a mis-ordered spread there would silently swap `action` and `status`, both of which are plain strings. Covered by `src/realtime/recovery-wal.test.ts`, which exercises `mark(SCOPE, id, "review", "reviewed")` and the purge variant and passes unchanged.
- Assumptions that could be wrong: that the server/ forks are behaviorally equivalent to the src/ copies they replace. I did not diff them line by line; the evidence is that the unmodified src/ tests, which were written against the src/ implementations, pass against the server/ ones through the adapters. A behavior difference the existing tests do not cover would survive this change.
- Missing/weak tests: the working-set adapter's positional-constructor path is exercised, but no test asserts the adapter and the server class produce identical results for the same inputs. The recovery-wal adapter's `compactWal` delegation is not directly covered by the src/ tests.
- Security/permission risk: none identified. No namespace predicate, auth path, SQL, or token handling is touched; both modules are process-memory stores plus a local write-ahead file whose path is caller-supplied and unchanged.
- Migration/deploy risk: none. No schema change, no migration, no config change. `server/main.ts` already imported the server/ paths before this PR and is untouched.
- Downstream client/runtime risk: none identified. No MCP tool, schema, transport, or Python client surface changes — the public export names and types of both old paths are preserved exactly, and `server/contracts/contract-payload-sections.ts` still declares the `src/realtime/...` implementation strings, which `src/contract.test.ts` asserts and which passes.
- Rollback/cleanup concern: reverting the single commit restores both src/ files verbatim; nothing else moved. The adapters are themselves scheduled cleanup — they and the two tests left in src/ retire with src/ at L6.
- Fixes made before PR: the first recovery-wal adapter extended the server class, which tsc rejected (TS2416, a subclass cannot widen `mark`'s arity), so it became composition; the resulting five-parameter `mark` then failed oxlint `max-params`, fixed by carrying the trailing arguments as a named tuple rather than a disable comment.
- Known residual risk: the two tests left in src/ still name the old paths, so any future edit to them tests the server/ code through an indirection that is not obvious from the file. The M1 exception accepts this and the PR records it.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical L5 relocation with no new failure pattern — the one novel detail (a subclass cannot widen a method's arity, so a legacy-call-form adapter is composition rather than inheritance) is recorded in the PR comment for harvest rather than as a new lane entry.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime, transport, or tool surface changes; the two modules are already the ones `server/main.ts` constructs.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract payload or fixture changes. `server/contracts/contract-payload-sections.ts` and `server/contracts/contract-types.ts` still declare `src/realtime/working-set.ts` and `src/realtime/recovery-wal.ts` as the implementation paths, those files still exist as adapters, and `src/contract.test.ts` passes unchanged.

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: this is an internal file relocation. No MCP tool name, schema, protocol, transport, or Python client surface changes, so there is nothing for a downstream client to pick up.


rodaddy:
harvested: #991 An M9 adapter that preserves a legacy positional call form must be written as composition, not inheritance: this lane's first attempt made `src/realtime/recovery-wal.ts` extend the server `RecoveryWalStore` and override `mark` with the old five-argument form, which TypeScript rejects outright (TS2416, "target signature provides too few arguments") because a subclass method may not widen its base's arity — a subclass has to remain substitutable for its base, and a five-argument override is not. Delegating to a private instance instead works, and it also makes the adapter's real surface explicit: the delegating class must list every public member the legacy callers use, which for this file was a small set (`budget`, `walPath`, `append`, `mark`, `buildContextPackFragment`, `getCounters`, `compactWal`) but is a checklist, not something inheritance supplies for free. The second trap follows immediately: the preserved legacy signature is itself over the repo's four-parameter rule, and since the whole point of an adapter is that the old form cannot change, the fix is to carry the trailing arguments as a named tuple type spread through a rest parameter — the call form at every existing call site is byte-identical and the method reads as three parameters. A disable comment is not available and was not needed. Worth noting for the next move lane: the premise "move this file with `git mv`" was wrong here because a hardened, already-split `server/realtime/` fork of both modules existed on main, so the src/ copies were duplicates rather than originals; checking whether the target path already exists before planning the move is a cheap first step that changes the whole shape of the lane.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=1e80fa5df489a48c86f47ee3be2c20a7cd383498 at=2026-08-29T21:41:35.532Z

```
judged=36 shims, 2 adapters
PASS
```

</details>

PR #986, opened 2026-08-29
refactor(l5): move operator-doctor from src/ into server/application behind an L5 adapter (issue 864)

rodaddy:
## Summary

- `src/operator-doctor.ts` (770 lines, twelve inline `process.env` reads) moves to `server/application/operator-doctor.ts`, split into three siblings by responsibility: `operator-doctor-types.ts` (the frozen `OperatorDoctorStatus` shape, `DOCTOR_CONTRACT_VERSION`, the distillation-lag ratio constants, `canReadDoctor`, `classifyDistillationLag`, the widened `OperatorDoctorOptions`), `operator-doctor-probes.ts` (the individual diagnostic questions: service version, migrations, audit storage, distillation lag, embedding availability, qmd binary and index probe worker, plus `withTimeout` and `probeUrl`), and `operator-doctor.ts` (assembly of the payload plus the single-flight short-TTL cache). Each satisfies the 500-line file rule and the 100-line function rule, and none reads `process.env`.
- The twelve env reads become fields of ONE options object. The existing 4th parameter type `OperatorDoctorCacheOptions` widened into `OperatorDoctorOptions` with the env-derived fields required there, so the parameter count stays 4 and `getOperatorDoctorStatus(pool, natsRuntimeBoundary, natsBridgeHealth, options)` keeps its signature.
- `server/main.ts:403` fills the options from the parsed `ServerConfig` it already holds: `serviceVersionFallback` ← `config.doctor.serviceVersionFallback`, `nodeEnvironment` ← `config.doctor.nodeEnvironment`, `fileLogConfigured` ← `config.logging.file`, `rotationConfigured` ← `config.doctor.rotationConfigured`, `rawTurnTtlSeconds` ← `config.doctor.rawTurnTtlSeconds`, `qmdIndexPath` ← `config.doctor.qmdIndexPath` (spread only when present, preserving the `!== undefined` branch the old resolver used).

```diff
     operatorDoctor: () =>
-      getOperatorDoctorStatus(database.pool, nats.boundary, nats.health),
+      getOperatorDoctorStatus(database.pool, nats.boundary, nats.health, {
+        serviceVersionFallback: config.doctor.serviceVersionFallback,
+        nodeEnvironment: config.doctor.nodeEnvironment,
+        fileLogConfigured: Boolean(config.logging.file?.trim()),
+        rotationConfigured: config.doctor.rotationConfigured,
+        rawTurnTtlSeconds: config.doctor.rawTurnTtlSeconds,
[...]

PR #998, opened 2026-08-29
refactor(l5): move drop-folder-collector from src/ into server/capture behind an L5 adapter (issue 864)

rodaddy:
## Summary

- L5 of issue 864: `src/drop-folder-collector.ts` (1122 lines) moves with `git mv` to `server/capture/` and splits by responsibility into seven siblings, each well under the 500-code-line file rule the pre-commit gate applies to staged files.

  | new path | responsibility |
  |---|---|
  | `server/capture/drop-folder-collector.ts` | entry: gate, discover, fold, stamp |
  | `server/capture/drop-folder-contract.ts` | source kind, input schema, receipt/result shapes, scan bounds |
  | `server/capture/drop-folder-eligibility.ts` | authority gate + durable root resolution |
  | `server/capture/drop-folder-discovery.ts` | bounded streaming walk + supported-extension allowlist |
  | `server/capture/drop-folder-read.ts` | no-follow, identity-checked, bounded per-file read |
  | `server/capture/drop-folder-durable.ts` | durable upsert with the true-no-op probe |
  | `server/capture/drop-folder-ingest.ts` | per-file disposition and the running totals |

- **The two `process.env` reads are gone from `server/`.** They were `boundedInt` over `DROP_COLLECTOR_MAX_FILES` / `_MAX_FILE_BYTES` / `_MAX_TOTAL_BYTES` / `_MAX_DEPTH`, and the `DROP_COLLECTOR_MAX_SCAN_ENTRIES` override inside `maxEntriesInspected`. They are now the fields of one options parameter on the entry function:

  ```ts
  export interface DropCollectorBounds {
    files: number;       // was DROP_COLLECTOR_MAX_FILES        (default 256)
    fileBytes: number;   // was DROP_COLLECTOR_MAX_FILE_BYTES   (default 1 MiB)
    totalBytes: number;  // was DROP_COLLECTOR_MAX_TOTAL_BYTES  (default 16 MiB)
    depth: number;       // was DROP_COLLECTOR_MAX_DEPTH        (default 8)
    scanEntries?: number; // was DROP_COLLECTOR_MAX_SCAN_ENTRIES (absent = derive from files)
  }

  export interface CollectDropFolderOptions {
    bounds?: DropCollectorBounds;
  }
  ```

  Absent options means `DEFAULT_DROP_COLLECTOR_BOUNDS`, which holds the exact values an unconfigured deployment ran with, so an environment that sets none of these keys behaves identically.

- **The importer.** `server/tools/source-registry.ts` is repointed at the `server/` path and passes the values from the deps it holds:

  ```diff
  -} from "../../src/drop-folder-collector.ts";
  +} from "../capture/drop-folder-collector.ts";
  ```
  ```diff
           const result = await collectDropFolder(
             { pool: dependencies.pool, embedFn: dependencies.embedFn },
             registryAuth(identity),
             args,
  +          dependencies.dropCollectorBounds
  +            ? { bounds: dependencies.dropCollectorBounds }
  +            : {},
           );
  ```

  `MemoryToolDependencies` gains an optional `readonly dropCollectorBounds?: DropCollectorBounds`, wired the way PR #987 wired `promotionKillSwitch`. **The composition-root fill LANDED IN THE SECOND COMMIT on this branch** (`9959833`). It was held out of the first commit on an M8 reading that turned out to be wrong — M8 does not forbid a config reader, and leaving the field unfilled would have silently dropped the five `DROP_COLLECTOR_*` overrides on the `collect_drop_folder` path, which is a behavior change. The second commit adds `positiveParsedInteger` and the five fields to `server/config/src-readers.ts` (mirroring `boundedInt` at `src/drop-folder-collector.ts:19-24` exactly), a `dropCollectorGroup` returning `DropCollectorBounds`, `ServerConfig.dropCollector` in `server/config.ts`, and the one line `dropCollectorBounds: config.dropCollector` in `server/main.ts` beside `promotionKillSwitch` — the same shape #986 and #987 used. The operator keys reach the tool path again.

- **The adapter.** `src/drop-folder-collector.ts` becomes the M9 adapter, header `// L5 adapter (issue 864): legacy call form over server/capture/drop-folder-collector.ts; retired with src/ at L6.` It reads the `DROP_COLLECTOR_*` keys **inside its functions at call time** — which is what the src/ tests depend on, since they flip those keys per test — and preserves every legacy call form:

  ```ts
  export function collectDropFolder(
    ...args: Parameters<typeof collectWithBounds> extends [infer D, infer A, infer I, ...unknown[]]
      ? [deps: D, auth: A, input: I]
      : never
  ): ReturnType<typeof collectWithBounds> {
    return collectWithBounds(args[0], args[1], args[2], { bounds: envBounds() });
  }
  ```

  plus the two-argument `discoverFiles` and the zero-argument `maxFiles` / `maxFileBytes` / `maxTotalBytes` / `maxDepth`. Every relative import in it names a `server/` path, **type imports included** (M9 NOTE): the types are taken from the server module via `Parameters<>` / `ReturnType<>` rather than imported from `src/`.

- **Restructuring forced by the server/ lint standard, no behavior change.** `writeDurableFile`'s six positional arguments become four with a `DurableFile` payload, and its two arms are hoisted into `mergeExistingTags` / `insertNewContent`; `readConfinedFile`'s open, identity check, and bounded read are hoisted into `openNoFollow` / `descriptorMatches` / `readBounded`; the discovery walk's eight closure variables become fields on a `DiscoveryWalk` whose traversal splits into `walkDirectory` / `inspectEntry` / `confine` / `retainFile`; the entry function's per-file loop body moves to `ingestOneFile` and its manifest stamp to `stampManifest`. Traversal order, every bound, the TOCTOU descriptor-identity checks, and every receipt code are unchanged — the src/ test file exercises all of it and passes unmodified.

- **Tests left in src/ (retire or move at L6)**, per M1 EXCEPTION — they set `process.env.DROP_COLLECTOR_*` at call time, which `no-process-env` forbids under `server/`, and they pass unchanged through the adapter:
  - `src/drop-folder-collector.test.ts`
  - `src/tools/__tests__/drop-folder-collector.test.ts`

- Closure count: **25 -> 24** (minus exactly the one file moved).

## Verification

Receipts, all run in this lane's clone on the committed tree:

```
bunx tsc --noEmit                                              -> exit 0
oxlint --deny-warnings --config .oxlintrc.json <10 staged>     -> exit 0
bun run test:isolated src/drop-folder-collector.test.ts \
  src/tools/__tests__/drop-folder-collector.test.ts            -> 40 pass, 0 fail
bun run test:isolated server/tools/source-registry.pg.test.ts  ->  8 pass, 0 fail
bash scripts/done-means/864-moved-out-of-src.sh (no arguments) -> PASS (38 shims, 4 adapters)
closure-count.sh at origin/main -> 25 ; on this branch -> 24
```

The RED line and its GREEN pair, both measured:

```
# RED: the file as it stands at origin/main, linted under the server/ overrides
git show origin/main:src/drop-folder-collector.ts > server/capture/_red-probe.ts
oxlint --deny-warnings --config .oxlintrc.json server/capture/_red-probe.ts
  -> exit 1, 16 errors (file over the 500-line rule, four functions over the
     complexity rule, two over the function-length rule, one non-null
     assertion, five max-depth, one max-params)

# GREEN: the same code, split, on the committed tree
oxlint --deny-warnings --config .oxlintrc.json server/capture/drop-folder-*.ts
  -> exit 0, no findings
```

Python package is not applicable: the M10 guard survey (`rg -n '"src/|/ "src"|src/[a-z-]+\.ts' python/openbrain-memory/tests`) returns no hit naming this module, so no guard needed repointing. M12 text-reader survey (`rg -n "drop-folder-collector" scripts server src python`) found only importers — no `readFileSync`/glob consumer reads this module by path.

## Critical Self-Review

- Highest-risk behavior: the discovery walk. It is the TOCTOU-hardened traversal, and turning eight closure variables into fields on a `DiscoveryWalk` touches the exact control flow that enforces the entry-scan bound and the candidate sentinel. The one place the shape genuinely changed is the sentinel: the original returned from `walk` the instant `consider` filled the buffer, whereas `retainFile` now sets `stopped` and the loop's next `if (this.stopped) return` exits. Both leave the walk with the same buffer and the same `stopped` flag, and the src/ tests that assert `truncated` and bounded `entriesInspected` on hostile trees pass unchanged.
- Assumptions that could be wrong: that the config reader parses each key the way the adapter's `boundedInt` does. That assumption is now tested rather than assumed — `server/config-src-readers.test.ts` pins the base-10 `parseInt` reading (`"512files"` is 512, `"1.5"` is 1), the fallback on blank/zero/negative/non-numeric, and the omission of `scanEntries` at or below zero, each citing the adapter line it mirrors. The earlier risk here — that a deployment setting one of these keys would silently get the defaults on the tool path — is closed by the second commit, not deferred.
- Missing/weak tests: no test covers the new `bounds` options parameter at the `server/` boundary — the src/ tests reach it only through the adapter's env read, so a caller passing `bounds` explicitly is exercised by nothing. `stampManifest`, `mergeExistingTags`, and `insertNewContent` are covered only indirectly through `collectDropFolder`.
- Security/permission risk: unchanged and re-verified. `resolveEligibleDropSource` still gates on `resolveIngestionEligibility` plus `canWriteNamespace` against the exact target namespace before any read, `configuredRoot` still takes the root only from the durable record, and `readConfinedFile` still opens `O_NOFOLLOW` and fstat-compares dev/ino against the discovery identity. The confinement predicate in `confine` is character-for-character the original.
- Migration/deploy risk: none. No schema, no migration, no new dependency.
- Downstream client/runtime risk: none. No MCP tool name, input schema, or result shape changed; `collect_drop_folder` registers identically. The `DropCollectorBounds` field on `MemoryToolDependencies` is optional and internal to the server composition.
- Rollback/cleanup concern: reverting the commit restores the single src/ file cleanly; nothing outside the ten staged paths was touched. The adapter and both src/ test files are removed at L6 with the rest of `src/`, not by this lane.
- Fixes made before PR: the first cut of the durable sibling still called `backgroundExtract` with the legacy six positional arguments, which typechecked against the src/ adapter but not the server module — corrected to the object form. The first splice of the entry file left a dangling comment fragment and an unused `DropFileReceipt` import, both caught by tsc/oxlint and removed.
- Known residual risk: the config parse is a once-at-startup snapshot, where the `src/` adapter reads the environment per call. That is the shape every other group in `server/config.ts` already has and it matches how the server runtime is started, but a process that mutated `DROP_COLLECTOR_*` after startup would have been followed by the old collector and is not by the new one. No known caller does; the `src/` adapter keeps the per-call reading for the legacy callers and their tests.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical L5 move whose only novel finding — that a module reading process.env at call time inside its own tests forces the M9 adapter rather than an M2 shim — is already the head's recorded M9/M1-EXCEPTION rule for the 864 program, so a new entry would duplicate the lane contract rather than add to the KB.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no MCP tool contract, transport, or auth surface changed; the tool registers with the same name, schema, and result shape.

## Contract Parity

- Contract parity: [x] runtime-specific because: this is an internal TypeScript module move with no client-facing contract change; the pre-push gate reported "Client/contract paths unchanged; skipping contract parity subset".

## Downstream Rollout

Notes/evidence:

- Not applicable across the board: no MCP tool name, input schema, result shape, transport behavior, or Python client surface changed. `collect_drop_folder` is registered from the same handler with the same `COLLECT_DROP_FOLDER_INPUT` and the same `publicCollection` projection; the only edit at that call site is the added options argument. The pre-push gate independently reported the Python package and contract paths as unchanged.


rodaddy:
harvested: #998 A module whose OWN TESTS flip `process.env` per test forces the M9 adapter over an M2 shim, and the tell is in the test file rather than in the module — `src/drop-folder-collector.test.ts` sets `process.env.DROP_COLLECTOR_MAX_FILE_BYTES` and `_MAX_FILES` inside individual `it` bodies and then calls the three-argument `collectDropFolder`, so the env read has to happen at CALL time, not at import time. A shim re-exporting the server function would have compiled, typechecked, and silently broken those assertions, because the server version takes its bounds from an options parameter resolved by the caller. The generalizable check for a move lane: before choosing shim vs adapter, `rg -n "process\.env" <module> <module>.test.ts` and treat a hit in the TEST file as decisive on its own — a module can be env-free itself while its tests still depend on env reaching it through the call. The second half of the lesson is that the composition-root fill may not be finishable inside a move lane: M8 forbids editing `server/config.ts`, which is exactly where a new bounds group has to live, so the honest outcome is an optional dependency field wired at the tool boundary, defaults preserved, and the residual named in the PR rather than a silent behavior change to the operator knobs.

rodaddy:
**Fix-up commit `9959833`: the composition-root fill.**

The first commit declared `MemoryToolDependencies.dropCollectorBounds` (`server/tools/types.ts:100`) and read it in the handler (`server/tools/source-registry.ts:492`) but left it unfilled at the composition root, so the server runtime would have run on `DEFAULT_DROP_COLLECTOR_BOUNDS` regardless of what the deployment set — silently dropping all five `DROP_COLLECTOR_*` overrides on the `collect_drop_folder` path. That is a behavior change. It is closed here the same way #986 (operator-doctor) and #987 (`promotionKillSwitch`) closed theirs.

- **Reader** — `server/config/src-readers.ts`: `positiveParsedInteger(fallback)` mirrors `boundedInt` at `src/drop-folder-collector.ts:19-24` line for line (`!raw` takes the fallback; base-10 `parseInt`; only an integer `> 0` wins; nothing rejects). Five fields join `srcReaderEnvironmentFields`, defaults from `DEFAULT_DROP_COLLECTOR_BOUNDS` (`server/capture/drop-folder-contract.ts:51-56`). `DROP_COLLECTOR_MAX_SCAN_ENTRIES` is **omitted** rather than zeroed when unconfigured, because the collector branches on the key being absent to derive the entry count from `files` — the adapter reaches the same shape through its `> 0` test at `:38`.
- **Group** — `server/config.ts`: `dropCollectorGroup(parsed)` returns a `DropCollectorBounds`; `ServerConfig.dropCollector` holds it, filled beside `doctor` in the same single parse.
- **Fill** — `server/main.ts:214`: `dropCollectorBounds: config.dropCollector`, one line beside `promotionKillSwitch: config.promotion.killSwitch`. Nothing else in that file changed.

**RED / GREEN**

| receipt | result |
|---|---|
| `bun test server/capture/drop-folder-bounds.test.ts` (before the config change, field absent) | **0 pass / 4 fail** |
| `bun test server/config-src-readers.test.ts` (same tree) | **16 pass / 5 fail** |
| both, after the change | **25 pass / 0 fail** |
| `bunx tsc --noEmit` | exit 0 |
| `oxlint --deny-warnings --config .oxlintrc.json` on all 5 staged paths | exit 0 |
| `bun run test:isolated server/config-src-readers.test.ts server/tools/source-registry.pg.test.ts src/drop-folder-collector.test.ts server/main.pg.test.ts server/capture/drop-folder-bounds.test.ts` | **77 pass / 0 fail** |
| `bash scripts/done-means/864-moved-out-of-src.sh` (no arguments) | **PASS** (41 shims, 10 adapters) |
| `git push --force-with-lease` | pre-push full isolated suite passed first attempt, no M11 flake |

The new `server/capture/drop-folder-bounds.test.ts` is not a shape assertion: its last case builds a real three-level tree and shows `discoverFiles` returning strictly fewer files under a configured depth of 1 than under the default depth of 8, with both values arriving through `parseServerConfig` and the dependency field. A regression at any link in that chain fails there rather than in production.

One correction found while writing the tests, recorded rather than tidied: `parseInt("1.5", 10)` is `1`, a positive integer, so the adapter accepts it — the reader now does too, and the test says so with the reason. Start-equivalence with the adapter is the bar, not a nicer parse.

The PR body's deviation paragraph, the "assumptions that could be wrong" line, and the residual-risk line are all updated to say the fill landed here.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=99598332e3682f58e5ec6450a3f0bb508e20d58f at=2026-08-29T22:20:57.433Z

```
judged=41 shims, 10 adapters
PASS
```

</details>

PR #1000, opened 2026-08-29
refactor(l5): move logger from src/ into server/logging/legacy-logger behind a settings reader and an L5 adapter (issue 864)

rodaddy:
## Summary

L5 of `_plans/server-hardening-ladder.md`, issue 864: cut `server/` free of `src/`.

`src/logger.ts` (592 lines) moved by `git mv` to `server/logging/legacy-logger.ts`
and split at the seams the file already had, so each sibling meets the whole-file
`server/` standard:

| file | responsibility |
|---|---|
| `server/logging/legacy-logger.ts` | level state, context reader, extra sinks, envelope, emit, sink-throw reporting |
| `server/logging/legacy-logger-identity.ts` | host name, service name, rotating file sink, `deriveWorkerLogPath` |
| `server/logging/legacy-logger-redaction.ts` | `redactForLog` and the marker-density guard |
| `server/logging/legacy-logger-settings.ts` | the settings reader contract |

**Collision note.** `server/logging/logger.ts` already existed and is a
DIFFERENT API — `createLogger`, `withOperation`, the pino-shaped logger
`server/main.ts` builds. The two are not merged and do not import one another.
The moved module is named `legacy-logger` precisely to keep that distinction
visible at every call site.

### Environment reads

Eight reads, every one at MODULE SCOPE in the original — the initial level, the
host label, the service label and the file sink were all resolved once at import:

| line (old `src/logger.ts`) | key | scope | now |
|---|---|---|---|
| 12 | `LOG_LEVEL` | module | reader, first use |
| 116 | `HOSTNAME` | module | reader, first use |
| 116 | `HOST` | module | reader, first use |
| 176 | `LOG_FILE` | module | reader, first use |
| 180 | `OPEN_BRAIN_WORKER_NAME` | module | reader, first use |
| 184 | `LOG_MAX_BYTES` | module | reader, first use |
| 185 | `LOG_MAX_FILES` | module | reader, first use |
| 202 | `SERVICE_NAME` | module | reader, first use |
| 203 | `OPEN_BRAIN_WORKER_NAME` | module | reader, first use |

**How first-use is preserved.** Each value is memoized on its first read instead
of at module load: `getEffectiveLevel()`, `getHostName()`, `getServiceName()` and
`getFileSink()` each resolve exactly once and cache. The value is still read once
per process — just at the moment the first line is emitted rather than at import.
That deferral is the whole mechanism: it is what lets the adapter and
`server/main.ts` register a reader during import and still be the source of the
value. `setLogLevel` called before any emission still wins, because it assigns the
resolved slot directly.

### Reader contract

Rule M13a — a reader, not a snapshot, because `src/logger-file.test.ts` spawns a
child with a constructed environment after import:

```ts
setLegacyLoggerSettingsReader(read: () => LegacyLoggerSettings)
  : () => LegacyLoggerSettings   // returns the previous reader
```

The default reader, installed before any registration, returns the
environment-absent defaults: level `info`, service `open-brain`, no file sink. A
script or test importing a `server/` module directly therefore logs exactly as it
did with none of these variables set — the unwired case behaves identically
rather than silently differently.

### Adapter

`src/logger.ts` is now an M9 adapter, 23 code lines, importing only `server/`
paths:

```ts
// L5 adapter (issue 864): legacy call form over server/logging/legacy-logger.ts; retired with src/ at L6.
import { setLegacyLoggerSettingsReader } from "../server/logging/legacy-logger-settings.ts";

setLegacyLoggerSettingsReader(() => ({
  logLevel: process.env.LOG_LEVEL,
  hostName: process.env.HOSTNAME?.trim() || process.env.HOST?.trim(),
  logFile: process.env.LOG_FILE,
  logMaxBytes: process.env.LOG_MAX_BYTES,
  logMaxFiles: process.env.LOG_MAX_FILES,
  serviceName: process.env.SERVICE_NAME,
  workerName: process.env.OPEN_BRAIN_WORKER_NAME,
}));

export * from "../server/logging/legacy-logger.ts";
```

Same keys, same precedence, same defaults as the original. Every export keeps its
name and arity: `setLogLevel`, `getLogLevel`, `deriveWorkerLogPath`,
`setLogContextReader`, `addLogSink`, `redactForLog`, `logger`.

### server/main.ts

One import plus one call in `startServer`, immediately after `const config = ...`
and before the first line can be emitted:

```ts
  const config = options.config ?? loadServerConfig();
+ // The legacy `src/logger.ts` singleton still emits from every tool this
+ // process loads, and under issue 864 it reads no environment of its own.
+ // Registering here — before the first line can be emitted — makes this
+ // process's own configuration the source of its envelope, rather than a
+ // second parse of the same variables. Fields config does not carry stay
+ // undefined, which is exactly the value the module defaults on.
+ setLegacyLoggerSettingsReader(() => ({
+   logLevel: config.logging.level,
+   hostName: config.transport.hostname,
+   logFile: config.logging.file,
+   logMaxBytes: undefined,
+   logMaxFiles: undefined,
+   serviceName: config.logging.service,
+   workerName: config.logging.workerName,
+ }));
  const logger = options.logger ?? createLogger(config.logging);
```

`server/config.ts` is UNTOUCHED: it already carries a `logging` group with
`level`, `file`, `service` and `workerName`. The rotate threshold and
retained-file count are not in config and stay `undefined`, which is the value
the module already defaults on, so no new config group and no
`config-src-readers` change was needed.

### Importers

Per rule M3 the 24 `server/` importers of `../../src/logger.ts` and every `src/`
and `scripts/` importer keep their current import line — they hold no
configuration, so the repoint belongs to L6.

### Other changes

- `.oxlintrc.json`: `server/logging/legacy-logger.ts` joins the EXISTING
  `no-console` override beside `src/logger.ts` and `server/logging/logger.ts`.
  That override follows the file rather than being a new exemption — this module
  IS the thing that wraps the console sink, which is why the entry exists.
- Two functions hoisted to meet the complexity rule, branches/order/outputs
  unchanged: `tameValue()` out of the `JSON.stringify` replacer, and
  `describeSinkThrow()` out of `reportSinkThrow()`. The sink snapshot spread
  became `Array.from` with identical semantics.

### Tests left in src/ (retire or move at L6)

- `src/logger-file.test.ts` — stays at its `src/` path untouched (rule M1
  EXCEPTION) and exercises the moved code through the adapter. It spawns a child
  with a constructed environment, which is exactly the case the reader exists to
  keep working.

Closure count: 18 -> 17.

## Verification

Receipts, each run from the lane clone:

- `bash scripts/done-means/864-moved-out-of-src.sh` (no arguments) -> exit 0,
  `judged=40 shims, 9 adapters` / `PASS`
- `bunx tsc --noEmit` -> exit 0, no diagnostics
- `./node_modules/.bin/oxlint --deny-warnings --config .oxlintrc.json` over all
  six touched source paths -> exit 0, no findings
- `bun run test:isolated src/logger-file.test.ts server/logging/ server/observability/ server/main.pg.test.ts`
  at `origin/main` BEFORE the change -> exit 0, 118 pass / 0 fail across 8 files
- the identical command AFTER the change -> exit 0, 118 pass / 0 fail across 8
  files — same count, same files
- closure count before -> after: 18 -> 17 (minus exactly the one file moved)
- `git push` pre-push hook ran the FULL isolated suite -> passed on the first
  attempt, no re-run needed
- M10: `rg` over `python/openbrain-memory/tests` for a `src/` path naming this
  module -> no hits, so no Python guard to repoint
- M12: `rg` for `readFileSync`/`Bun.file`/`read(` naming `logger.ts` across
  `scripts server src` -> no hits. The two path-as-text references
  (`src/logger-file.test.ts:24`, `src/observability/observability.test.ts:406`)
  build their specifier from `import.meta.dir` and resolve to the `src/` adapter,
  which stays in place and still works.

## Critical Self-Review

- Highest-risk behavior: the deferral of four module-load reads to first use. If
  anything emitted a line during module import — before the adapter or
  `server/main.ts` registered a reader — it would get the environment-absent
  defaults instead of the configured ones. Checked: the module emits only from
  `log()`, which is reachable only through the exported `logger` object and
  `setLogLevel`, never at import; and the adapter registers its reader in the
  statement above its own re-export, so any importer of `src/logger.ts` has a
  reader installed before it can call anything.
- Assumptions that could be wrong: that `config.transport.hostname` is an
  acceptable source for the `host` envelope label in the server runtime. It is
  the identity `/health` reports, which is the same question the label answers,
  but it is resolved by `server/transport/server-identity.ts` rather than read
  from `HOSTNAME`, so a deployment setting `HOSTNAME` to something other than its
  resolved identity would see the label change under `server/main.ts`. The `src/`
  runtime is unaffected — the adapter reads `HOSTNAME` exactly as before.
- Missing/weak tests: there is no test that asserts the default reader's values,
  and none that asserts `server/main.ts` registration wins over the adapter's in
  the server runtime. Both are covered indirectly — the 8-file suite exercises the
  module through both entry paths — but neither is pinned by a named assertion.
- Security/permission risk: none identified. Redaction moved file-to-file with no
  edit to `SECRET_PATTERNS`, `isSensitiveKey`, `redactForLog`, or the replacer's
  key check; `tameValue` preserves the branch order, so the sensitive-key test
  still runs before any value-shape branch.
- Migration/deploy risk: none. No schema, no migration, no config key added or
  removed. Rollback is reverting the commit; `src/logger.ts` still exists and
  still exports the same surface either way.
  Python-client surface is touched; `docs/downstream-rollout.md` classifies this
  as not applicable.
- Rollback/cleanup concern: the four `server/logging/legacy-logger*` files and the
  `src/logger.ts` adapter are one unit — reverting the commit reverts all of them
  together. The `.oxlintrc.json` override entry would be dead but harmless if the
  moved file were removed without the config change.
- Fixes made before PR: dropped an import left unused by the split; replaced the
  sink-snapshot spread with `Array.from` for the `no-useless-spread` rule; hoisted
  `tameValue` and `describeSinkThrow` for the complexity rule; corrected a comment
  that named a removed module-scope constant.
- Known residual risk: the `hostName` source divergence named under Assumptions —
  under `server/main.ts` the label now comes from resolved transport identity
  rather than from `HOSTNAME`. Behaviour is unchanged for every `src/` and
  `scripts/` caller, and unchanged for `server/` when the two agree.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: this is a mechanical move under the issue-864 ladder with no new review finding — the reader shape it uses is rule M13a, already recorded in the lane contract, and no MEDIUM+ finding surfaced in this lane.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no MCP tool, schema, transport or client surface changes; the moved module is an internal logging singleton and the isolated suite plus the pre-push full suite cover it.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract path is touched — the pre-push gate reported "Client/contract paths unchanged; skipping contract parity subset".

## Downstream Rollout

Notes/evidence:

- Not applicable on every row: this changes no MCP tool, no schema, no transport
  surface, and no Python client behaviour. The pre-push gate confirmed it —
  "Python package unchanged; skipping Python package gate" and "Client/contract
  paths unchanged; skipping contract parity subset".


rodaddy:
harvested: #1000 M13a says "the module reads env INSIDE function bodies (per call)" is what selects a reader over a snapshot, but this lane's module read env at MODULE SCOPE in all nine places and still needed the reader — so the trigger is not where the read sits, it is whether anything can register before the value is consumed. A module-scope read runs during import, which is strictly earlier than any importer's chance to register, so a snapshot filled at registration time is always too late for it; the fix is to keep the read a single read and defer it to first USE (memoize in an accessor) rather than to make it per-call. `getEffectiveLevel()`, `getHostName()`, `getServiceName()` and `getFileSink()` each resolve once and cache, which preserves the original's read-once semantics exactly while moving the moment from import to first emission — and that one move is what lets both the `src/` adapter and `server/main.ts` be the source of the value in their own runtimes. The second half of the lesson is the wiring risk the code-brain design names: a reader that is never registered must not fail loudly OR behave differently, so the default reader returns the env-absent defaults, which makes a script importing a `server/` module directly log identically to the unconfigured original instead of silently picking up someone else's settings.

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=1fb6a79537995eb881e545f1fab455a2bac88cb6 at=2026-08-29T22:19:22.282Z

```
judged=40 shims, 10 adapters
PASS
```

</details>

PR #1002, opened 2026-08-29
refactor(l5): move embedding from src/ into server/embedding behind a settings reader and an L5 adapter (issue 864)

rodaddy:
## Summary

- Moved `src/embedding.ts` -> `server/embedding/provider.ts` (issue 864, L5 of `_plans/server-hardening-ladder.md`). The moved module reads no `process.env`.
- **The reader contract.** The module holds `let readSettings: () => EmbeddingProviderSettings` and exports `setEmbeddingSettingsReader(read)`, which installs the reader, **returns the one it replaced**, and logs the replacement at debug through the logger the module already imported. There is no idempotence guard and no snapshot — last writer wins. Every place that used to read an `EMBEDDING_*` key now calls `readSettings().<field>` **at the same moment it used to read env**, so per-call laziness is unchanged: the watchdog threshold and cooldown inside their own functions, the restart script inside the trigger, base URL and API key inside the request, the timeout on each attempt.
- **Defaults before any reader is set** are exactly what the original computed with every key absent: `timeoutMs` 8000, `dimensions` 768, `model` `gemini-embedding-001`, `watchdog.failureThreshold` 2, `watchdog.cooldownMs` 300000, no `restartScript`, no `baseUrl`, no `apiKey`.
- **Types reused, not redeclared.** `EmbeddingProviderSettings extends EmbeddingConfigGroup` from `server/config/src-readers.ts` unchanged (`timeoutMs`, `dimensions`, `model`, and the nested `watchdog` group of `failureThreshold`/`cooldownMs`/optional `restartScript`) and adds `baseUrl?`/`apiKey?`, which mirror `config.transport.embeddingBaseUrl` and `config.transport.embeddingApiKey` (`server/config.ts:251-252`) and are the same `readonly string | undefined` there.
- `EMBEDDING_MODEL` and `EMBEDDING_DIMENSIONS` could not stay constants under a reader, so `server/` exports them as `embeddingModel()` and `embeddingDimensions()`.
- **The adapter.** `src/embedding.ts` is now an L5 adapter: `readEmbeddingEnv()` parses `process.env` with the original defaults and the original bare-`parseInt` shape, `setEmbeddingSettingsReader(readEmbeddingEnv)` runs at import, the two legacy constants are computed from env **at import time exactly as the original did**, and everything else is re-exported from `server/`. Every relative import names a `server/` path. `src/embedding.test.ts:760,786` re-import it cache-busted and still observe dimensions 3.
- **`server/embedding/provider-attempt.ts`** is a new sibling holding the per-attempt decisions. See "inherited lint debt" below.

### The `server/main.ts` diff

```ts
+import { setEmbeddingSettingsReader } from "./embedding/provider.ts";

   const config = options.config ?? loadServerConfig();
   const logger = options.logger ?? createLogger(config.logging);
+
+  const previousEmbeddingSettingsReader = setEmbeddingSettingsReader(() => ({
+    ...config.embedding,
+    ...(config.transport.embeddingBaseUrl === undefined
+      ? {} : { baseUrl: config.transport.embeddingBaseUrl }),
+    ...(config.transport.embeddingApiKey === undefined
+      ? {} : { apiKey: config.transport.embeddingApiKey }),
+  }));
```

Every value is one `main` already held; no new env read, no `server/config.ts` edit.

**Beyond the one import and one call, `shutdown` restores the reader** (and so does the failed-start path). This is not decoration — it is a defect the tests caught. The provider holds ONE module-level slot per process, so a started-and-stopped server that left its own reader installed kept answering from a dead server's config for everything that ran afterwards, which in a test process is the next test file. Measured: the combined run of `server/main.pg.test.ts` with `src/operator-doctor.test.ts` is **114 pass / 0 fail at origin/main**, went to **112 pass / 2 fail** without the restore (`src/operator-doctor.test.ts` sets `EMBEDDING_BASE_URL` at test time and expects it seen), and is **151 pass / 0 fail** with it.

### Inherited lint debt (M14)

The file failed the whole-file `server/` lint the moment it moved: `max-lines` 524, `embedOnce` at 192 lines and complexity 35, two `max-depth`, `generateEmbeddingWithMetadata` complexity 15, `isTransient` complexity 14, `no-new-array`. Split mechanically in this commit, behavior unchanged:

- `provider-attempt.ts` — `classifyHttpFailure`, `classifyThrownFailure`, `decodeEmbeddingResponse`, `isTransient`, `classifyError`, `readUsageDetails`. Same order, same comparisons, same log events, fields, and messages.
- `provider.ts` — `prepareRequest`, `runOneAttempt`, `embedSegmented` hoisted out of the loop; the loop keeps the state that is its own (`lastError`, `lastStatus`).
- One dead ternary arm collapsed: `status === 401 || status === 403 ? "client_error" : "client_error"` had two identical arms and is now `"client_error"`.
- `new Array(n).fill(0)` -> `Array.from({ length: n }, () => 0)`; the two `summed[i]!` assertions became `(summed[i] ?? 0)`, exact because the array is pre-filled over the whole range.

Closure count: **18 -> 17**.

Tests left in `src/` (retire or move at L6): `src/embedding.test.ts`, `src/embedding-repair*.test.ts` — untouched per M1 EXCEPTION, testing the moved code through the adapter.

## Verification

Receipts:

- probe (delete `EMBEDDING_WATCHDOG_RESTART_SCRIPT`, import, set it after, drive two failures) at `origin/main` and after the change -> **identical** `spawnCount=1` / `restart_configured=true`
- `oxlint --deny-warnings --config .oxlintrc.json` on all four staged paths -> exit 0
- `bun run test:isolated src/embedding.test.ts` -> **37 pass 0 fail**
- `bun run test:isolated server/config-src-readers.test.ts server/main.pg.test.ts src/operator-doctor.test.ts src/embedding.test.ts server/embedding/embedding-canonical.test.ts server/embedding/embedding-targets.test.ts server/embedding/chunking.test.ts` -> **151 pass 0 fail**
- `bash scripts/done-means/864-moved-out-of-src.sh` (no arguments) -> **PASS** (judged=40 shims, 10 adapters)
- `bash closure-count.sh` -> 18 before, 17 after
- pre-commit (gitleaks, oxlint, prettier, tsc) and pre-push (full isolated suite) both passed; nothing bypassed
- M10: `rg -n '"src/|src/[a-z-]+\.ts' python/openbrain-memory/tests` -> no hit names embedding; Python guards need no repoint
- M12 text readers: `scripts/done-means/724-backlog-recallable.sh:245` and `scripts/__tests__/bulk-import.test.ts:32,65` read `src/embedding.ts` by path; the adapter keeps that path exporting the same names, so both keep working unchanged

## Critical Self-Review

- Highest-risk behavior: which reader answers, and when. Two readers exist (the adapter's env parse, main's config reader) over one module-level slot, so ordering decides the answer. Mitigated by making `setEmbeddingSettingsReader` return the previous reader and having `main` restore it on shutdown and on failed start.
- Assumptions that could be wrong: that no caller depends on `EMBEDDING_MODEL`/`EMBEDDING_DIMENSIONS` being live rather than import-time. The original computed both once at import, and the adapter does the same, so this preserves the original behavior rather than assuming about it.
- Missing/weak tests: no test asserts the reader-restore directly — it is proven only by the combined-run count going 112/2 -> 151/0. A dedicated test in `server/embedding/` would be a better guard; it is not in this lane's moved set.
- Security/permission risk: none new. `apiKey` moves from an env read to a settings field and is still never logged; gitleaks scanned the staged changes and found nothing.
- Migration/deploy risk: none. No schema, no migration. In the server runtime the values come from `config`, which is the same env parse `server/config/src-readers.ts` already performed and `server/config-src-readers.test.ts` already proves agrees with the src/ parse.
- Downstream client/runtime risk: none. No MCP tool, schema, transport, or Python client surface changed; `src/embedding.ts` keeps every export name and arity.
- Rollback/cleanup concern: revert is one commit. The adapter and both `server/embedding/` files are retired together at L6.
- Fixes made before PR: the reader-restore in `server/main.ts` (found by the combined test run, not by review); the dead ternary arm; the two non-null assertions in `combineEmbeddings`.
- Known residual risk: a future caller that registers a reader and never restores it would repeat the cross-file leak this PR fixes. The returned-previous-reader contract makes the fix available but does not force it.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no live service surface changed; the dogfood service is unaffected by a pure module move behind an adapter that preserves every export.

## Contract Parity

- Contract parity: [x] runtime-specific because: no client-facing contract, fixture, or wire shape changed; the move is internal to the TypeScript server and the pre-push contract-parity gate reported the client/contract paths unchanged.

## Downstream Rollout

Notes/evidence:

- No MCP tool, schema, protocol, or client-facing surface changed, so every downstream item is not applicable. The pre-push gate confirmed it independently: "Python package unchanged; skipping Python package gate", "Client/contract paths unchanged; skipping contract parity subset".

### DIFF
--- a/docs/sme/correctness.md
+++ b/docs/sme/correctness.md
@@ -2482,3 +2482,45 @@ passes, or fails the rule and is not judged.
   broken fixture is the defect, not a clean run.
 - Ask of any discovery loop: what does a malformed candidate do here — fail, or
   vanish? Vanishing is the bug.
+
+
+# A process-wide settings reader must be restored by whoever replaced it
+
+When a module's configuration moves out of `process.env` and into an injected
+reader, the reader lives in ONE module-level slot for the whole process. Whoever
+installs a reader owns putting the previous one back.
+
+`server/embedding/provider.ts` holds `let readSettings` and
+`setEmbeddingSettingsReader(read)`. Two callers register: the `src/` adapter
+registers an env parse at import, and `server/main.ts` registers a config-backed
+reader when a server starts. Last writer wins, which is correct at runtime — a
+started server should answer from its own config.
+
+It is wrong the moment a server STOPS. `startServer` originally registered and
+never restored, so a server that started and shut down left its own config
+answering for everything that ran afterwards. In a test process, "afterwards" is
+the next test file.
+
+The receipt: `server/main.pg.test.ts` and `src/operator-doctor.test.ts` in one
+run were 114 pass / 0 fail before the move, 112 pass / 2 fail after it, and 151
+pass / 0 fail once `shutdown` restored the reader. The two failures were in
+`src/operator-doctor.test.ts`, which sets `EMBEDDING_BASE_URL` at test time and
+expects the provider to see it — it never did again, because a dead server's
+reader was still installed.
+
+**What to check in review.** For any injected singleton (a reader, a clock, a
+spawner, a transport):
+
+- Does the setter RETURN the value it replaced? If not, restoration is
+  impossible and the design is already wrong.
+- Does every caller that replaces it restore it on every exit path — normal
+  shutdown AND failed start? A `try/finally` around the shutdown body is the
+  cheap shape.
+- Neither file's own tests catch this. Each passes alone; only the combined run
+  shows it. When a change introduces a process-wide slot, run the new module's
+  tests TOGETHER with the tests of anything that starts a server, in one
+  process, and compare the count to the same combined run at `origin/main`. A
+  per-file green is not evidence here.
+
+This is the same class as the `mock.module` leak entry: shared mutable
+process-wide state whose contamination is invisible to any single-file run.
--- a/docs/sme/entries/2026-08-29-a-process-wide-settings-reader-must-be-restored-by-whoever-replaced-it.md
+++ b/docs/sme/entries/2026-08-29-a-process-wide-settings-reader-must-be-restored-by-whoever-replaced-it.md
@@ -1,0 +1,48 @@
+---
+lane: correctness
+severity: MEDIUM
+status: active
+order: 105
+provenance: "issue 864 L5 embedding move; found by a combined test run, not by review"
+---
+
+# A process-wide settings reader must be restored by whoever replaced it
+
+When a module's configuration moves out of `process.env` and into an injected
+reader, the reader lives in ONE module-level slot for the whole process. Whoever
+installs a reader owns putting the previous one back.
+
+`server/embedding/provider.ts` holds `let readSettings` and
+`setEmbeddingSettingsReader(read)`. Two callers register: the `src/` adapter
+registers an env parse at import, and `server/main.ts` registers a config-backed
+reader when a server starts. Last writer wins, which is correct at runtime — a
+started server should answer from its own config.
+
+It is wrong the moment a server STOPS. `startServer` originally registered and
+never restored, so a server that started and shut down left its own config
+answering for everything that ran afterwards. In a test process, "afterwards" is
+the next test file.
+
+The receipt: `server/main.pg.test.ts` and `src/operator-doctor.test.ts` in one
+run were 114 pass / 0 fail before the move, 112 pass / 2 fail after it, and 151
+pass / 0 fail once `shutdown` restored the reader. The two failures were in
+`src/operator-doctor.test.ts`, which sets `EMBEDDING_BASE_URL` at test time and
+expects the provider to see it — it never did again, because a dead server's
+reader was still installed.
+
+**What to check in review.** For any injected singleton (a reader, a clock, a
+spawner, a transport):
+
+- Does the setter RETURN the value it replaced? If not, restoration is
+  impossible and the design is already wrong.
+- Does every caller that replaces it restore it on every exit path — normal
+  shutdown AND failed start? A `try/finally` around the shutdown body is the
+  cheap shape.
+- Neither file's own tests catch this. Each passes alone; only the combined run
+  shows it. When a change introduces a process-wide slot, run the new module's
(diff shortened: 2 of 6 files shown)


rodaddy:
harvested: #1002 Moving a module's configuration out of `process.env` and into an injected reader creates ONE module-level slot for the whole process, and the lane that installs a reader owns restoring the previous one — on normal shutdown AND on failed start. `server/main.ts` originally registered its config-backed reader and never put back the `src/` adapter's env reader, so a server that started and stopped left a dead server's config answering for everything afterwards; in a test process, "afterwards" is the next test file. Neither file's own tests catch this: `server/main.pg.test.ts` and `src/operator-doctor.test.ts` each pass alone, and only the combined run moved 114 pass / 0 fail at `origin/main` to 112 pass / 2 fail, then back to 151 pass / 0 fail once `shutdown` restored the reader. The design lever that makes the fix possible is having the setter RETURN the reader it replaced; without that, restoration cannot be written at all. The reviewable rule: for any injected process-wide singleton (reader, clock, spawner, transport), check that the setter returns the previous value, that every replacing caller restores it on every exit path, and that the evidence is a COMBINED run against the same combined run at `origin/main` — a per-file green proves nothing about a shared slot. Captured as `docs/sme/entries/2026-08-29-a-process-wide-settings-reader-must-be-restored-by-whoever-replaced-it.md` (lane: correctness).

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=f094259f040ee4cdaf7c8f693db31a0dd2f39100 at=2026-08-29T22:22:58.171Z

```
judged=40 shims, 10 adapters
PASS
```

</details>

PR #1007, opened 2026-08-29
test(bulk-import): stop replacing src/logger.ts process-wide with mock.module (issue 864)

rodaddy:
## Summary

- `scripts/__tests__/bulk-import.test.ts` replaced `src/logger.ts` with a stub object through `mock.module`. That mock is process-wide and permanent in Bun -- keyed by resolved specifier, never scoped to one file, and not undone by `mock.restore()` -- so once this file had loaded, every later suite in the run received the stub instead of the real logger.
- The visible failure: a suite that registers an `addLogSink` observer and asserts on the lines the logger emits captures nothing, because the stub has no sink machinery and pushes into this file's own arrays instead. Such a suite passes alone and fails in the full run, with a stack pointing nowhere near this file.
- `src/observability/observability.test.ts:19` already states that no suite may `mock.module` the logger, and `scripts/backfill.test.ts:10` is the precedent for the alternative. This file now observes through `addLogSink`, which composes with other suites instead of overwriting them. The two assertions it makes on `logCalls.warn` are unchanged.
- Bringing the file to the whole-file lint standard came with it (M14 inherited lint debt), since the pre-commit gate lints staged files whole and this one carried 38 findings at `main`: 35 non-null assertions removed (`params![n]` loses an assertion the surrounding `as [string, unknown[]]` already makes redundant; the `files[0]!` and `warnCall![1]!` sites use `expectDefined` from `scripts/test-support/expect-defined.ts`), the 816-code-line file split by responsibility into `scripts/__tests__/bulk-import.test.ts` (parsing, reading, small/medium/large import paths) and `scripts/__tests__/bulk-import-routing.test.ts` (table routing, tag handling, dry run, argument parsing, edge cases) over a shared `scripts/__tests__/bulk-import-fixture.ts`, and two over-long describe callbacks split at a responsibility seam into `table routing` / `table routing: sessions` and `edge cases` / `edge cases: error logging and hashing`.
- No `it` name, body, order, or assertion changed. Neither file is a `*.pg.test.ts`, so `scripts/assert-db-tests-ran.ts` is untouched. 59 tests before, 59 after.

## Verification

- Done-means: scripts/done-means/864-logger-never-mocked-process-wide.sh
- [x] Relevant Open Brain tests/typecheck/migrations passed
- [x] Python package checks passed or are not applicable
- [x] Live Open Brain smoke passed or is not applicable

[...]

PR #1009, opened 2026-08-29
refactor(l5): agent-context-pack, durable-lane and durable-memory become L5 adapters over their server/tools twins (issue 864)

rodaddy:
## Summary

- Three `src/tools` context-pack modules already had `server/tools` twins, found by exported symbol across `server/` rather than by filename, so per M16 they become adapters over those twins instead of being moved: `src/tools/agent-context-pack.ts` -> `server/tools/agent-context-pack.ts` (M9 adapter, mapping the legacy `ToolDeps` call form to the twin's `MemoryToolDependencies` shape inside the adapter), `src/tools/agent-context-pack-durable-lane.ts` -> `server/tools/context-pack-durable-lane.ts` (M9 adapter, same deps mapping), and `src/tools/agent-context-pack-durable-memory.ts` -> `server/tools/context-pack-durable-memory.ts` (M2 shim, since the legacy positional wrapper had no caller in the repo).
- The environment-derived values the twins take as configuration now live in `server/tools/legacy-context-pack/legacy-deps.ts`, which the adapters fill; `server/tools/legacy-context-pack/agent-context-pack-pointers-candidates.ts`, `server/application/nats-runtime.ts`, `server/tools/realtime-append.ts` and `src/realtime/recovery-wal.ts` follow that relocation.
- `scripts/done-means/744-recall-serves-durable.sh` was rewritten per M16: its two probes no longer compare a `src` implementation against a `server` one, because at L5 both entrypoints callers reach resolve to the same twin. The comment records that the comparison collapsed at L5. The driver still reports `RESULT: FAIL`, which is the pre-existing issue-744 behaviour it was written to detect (a bare recall does not consult durable memory), not a regression from this change.
- M5 test transformations were applied to the moved suites: every non-null assertion became `expectDefined(value, label)` from `scripts/test-support/expect-defined.ts`, and the durable-memory failure-path structural guard was repointed to the twin's source path and to `logDurableFailure`, the shared two-line failure shape the twin emits.
- M14 test split: `src/tools/__tests__/agent-context-pack-durable-memory.test.ts` failed the whole-file pre-commit lint at 1035 code lines with a 987-line describe callback, so it splits mechanically by its own `it()` groups into six suites plus a shared helpers module, with assertions and `it` names unchanged. The original file is renamed with `git mv` to the first group so history follows. 26 tests before, 26 after.

**This branch carries the fix commit that pull request #1007 lands first.** `0ddd8955` is the head of `fix/864-bulk-import-logger-mock` and sits under the adapter commit here. It is required for this branch's own suite to be green: the M14 split above produced `src/tools/__tests__/agent-context-pack-durable-memory-recall-failure.test.ts`, and that suite captured zero durable records under the full run because `scripts/__tests__/bulk-import.test.ts` had replaced `src/logger.ts` process-wide with a `mock.module` stub, so the sink the test registered never received anything. It passed alone and failed in the full suite. Merge #1007 first; this branch rebases onto it and the duplicate commit drops out. Rebased 2026-08-29: this branch now carries BOTH of #1007's commits -- `0ddd8955` (the logger fix) and `1bf16fa2` (`scripts/done-means/864-logger-never-mocked-process-wide.sh`, #1007's Done-means) -- under the adapter commit, and both `scripts/done-means/864-moved-out-of-src.sh` and that new script exit 0 here.

## Verification

`bash scripts/done-means/864-moved-out-of-src.sh` with no arguments -> exit 0, `judged=48 shims, 19 adapters`, `PASS`. `bunx tsc --noEmit` -> exit 0. `bun run test:isolated` (whole suite) -> 4035 pass, 0 fail across 301 files, which is the run that failed on this branch before the fix commit was carried. The closure count of `src/` under `server/main.ts` reads 0 from this tree. `./node_modules/.bin/oxlint --deny-warnings --config .oxlintrc.json` -> exit 0 on every staged path, and the pre-push hook ran the isolated suite and passed.

## Critical Self-Review

- Highest-risk behavior: the `ToolDeps` -> `MemoryToolDependencies` mapping inside the two M9 adapters. A field dropped or misnamed there would silently change what a legacy caller passes the twin. The moved suites exercise both adapters through their legacy call form and pass unchanged, which is the evidence the mapping is faithful.
- Assumptions that could be wrong: that the twins found by exported symbol are genuinely the same behavior and not a divergent second implementation. The suites that previously tested the `src` side now pass against the twin through the adapter with their assertions unchanged, which is what would break if the two had drifted.
- Missing/weak tests: `scripts/done-means/744-recall-serves-durable.sh` no longer compares two implementations, because after this change there is only one. Its remaining value is the issue-744 probe it still reports; it is not a parity check any more and the comment says so.
- Security/permission risk: none introduced. No namespace predicate, auth check, or token-derived role changes; the adapters forward the same deps the callers already supplied.
- Migration/deploy risk: none. No schema, no migration, no config default changes. `server/main.ts` is untouched beyond import paths per M8.
- Downstream client/runtime risk: none. No MCP tool name, schema, or transport shape changes; every legacy `src/` import path still resolves and still exports the same surface, which is the point of the adapter form.
- Rollback/cleanup concern: the `src/` adapters and the one shim are removed with `src/` at L6, not by this lane. Reverting this commit restores the duplicated implementations.
- Fixes made before PR: the recall-failure suite's full-run failure was traced to the `bulk-import` logger stub rather than worked around here; it is fixed at its owning boundary in #1007 and carried under this commit so this branch is green on its own.
- Known residual risk: `scripts/done-means/744-recall-serves-durable.sh` reports `RESULT: FAIL`. That is pre-existing issue-744 behaviour, verified by reading the driver's own probes, and a reader who sees only the exit status could mistake it for a regression from this lane.
- SME review-memory update: [ ] `docs/sme/` updated or [x] not applicable because: the reusable lesson is the process-wide reach of `mock.module`, which is already recorded in-tree at `src/observability/observability.test.ts:19` and `scripts/backfill.test.ts:10` and is carried in #1007 where the fix lands.

## Review Gate

- Live Open Brain checks: [ ] linked below or [x] not applicable because: no runtime surface changes; every legacy import path resolves to the same exported surface.

## Contract Parity

- Contract parity: [x] runtime-specific because: no contract value changes. The contract string literal naming a `src/` path at `server/contracts/contract-types.ts:206` is a contract VALUE and stays until L6 rules on it, per M16.

## Downstream Rollout

Notes/evidence:

- Internal refactor. Every legacy `src/tools` import path still resolves and still exports the same surface, so no client, skill cache, or agent runtime sees a changed shape.

Refs issue 864.


rodaddy:
harvested: #1009 A lane that splits a large test file can turn a previously vacuous assertion into a real one, and the failure then lands somewhere else entirely. The M14 split here produced `src/tools/__tests__/agent-context-pack-durable-memory-recall-failure.test.ts`, which began actually observing durable-failure records through a sink; under the full suite it captured zero, while passing alone and passing across all of `src/tools/__tests__`. The cause was in a file the lane never touched: `scripts/__tests__/bulk-import.test.ts` replaced `src/logger.ts` with a stub through `mock.module`, which in Bun is keyed by resolved specifier, process-wide, and not undone by `mock.restore()`, so every later suite in the run received the stub. The reusable rule is that a pass-alone/fail-in-suite split is a shared-mutable-state question, not a question about the failing file, and the first search is for `mock.module` over a module the failing test observes -- not for a bug in the assertion. The second rule is where the fix goes: at the owning boundary (the file installing the process-wide mock) and as its own pull request from `main` (#1007), so the checker fix is reviewable independently of the refactor that exposed it, rather than being buried in the lane diff. The repo already carried both the prohibition (`src/observability/observability.test.ts:19`) and the alternative (`scripts/backfill.test.ts:10`, observe with `addLogSink` instead of replacing); the cost was paid because the offending file predated them and nothing swept for stragglers.

rodaddy:
### software-factory: errored
Reviewed at: `fd45ce6fc99fd8bdd05747d049cb0ece2320d0dc`
Run: job 8705, attempt 12153

#### Receipt
Verdict: **satisfied**

Verbatim check outcomes:
- `error` **Target checkout and SHA availability** — No \.git directory was present under /work, and Git could not resolve fd45ce6fc99fd8bdd05747d049cb0ece2320d0dc from /work/job\. Code and tests were not reviewed\.

Findings:
- [high] Review could not run because the repository checkout is absent (/work/job)

rodaddy:
verify-lane receipt: check=scripts/done-means/864-moved-out-of-src.sh exit=0 sha=fd45ce6fc99fd8bdd05747d049cb0ece2320d0dc at=2026-08-30T01:18:13.052Z

```
judged=48 shims, 19 adapters
PASS
```

</details>

